PR #88, opened 2025-03-04
DIM-111: Fully working ROSCommandQueue orchestration between nav2 and webrtc action primitives

spomichter:
Current working implementation in run_go2_ros.py

Likely should do more testing for edge cases

### DIFF
--- a/.style.yapf
+++ b/.style.yapf
@@ -1,0 +1,3 @@
+   [style]
+   based_on_style = google
+   column_limit = 80
--- a/default.env
+++ b/default.env
@@ -1,4 +1,5 @@
 OPENAI_API_KEY=
+DIMOS_MAX_WORKERS=
 
 # TODO: Cleanup and standardize robot environment variables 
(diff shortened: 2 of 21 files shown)


dimos/robot/ros_command_queue.py
```diff
@@ -0,0 +1,412 @@
...
+                    self._logger.debug(f"Robot busy state changed: {self._last_busy_state} -> {is_busy}")
+                    self._last_busy_state = is_busy
+                    
+                    # If the robot has transitioned to busy, record the time
+                    if is_busy:
+                        self._stuck_in_busy_since = current_time
+                    else:
+                        self._stuck_in_busy_since = None
+                
+                # Check if we've been waiting too long for the robot to be ready
```
lukasapaukstys, line 289: It appears that forcing queue continuation after a prolonged busy state might be a workaround for an underlying race condition or scheduling issue. We should investigate why the robot remains ‘busy’ for extended periods and address the root cause. Additionally, revisiting the connection/disconnection states and potentially integrating a connection manager might help prevent these issues. We can revisit this at a later point.

PR #105, opened 2025-03-11
Init changes in robot, skills, unitree_go2, and unitree_skills for more robust initialization of skills

spomichter:
Also easier demonstration of how Agents communicate with Robot()

Now, users can do both 
1. 
```python 
robot = Robot(... , skills=MyUnitreeSkills(),...)
agent  = Agent(... , skills=robot.get_skills())
``` 
AND
2. 
```python
robot = Robot(...)
skills_instance = MyUnitreeSkills(robot=robot)
agent = Agent(... , skills=skills_instance)
```

More importantly, as developers initialize skills in test files like so: 
```python
class Reverse(AbstractSkill): 
      def __call__(self):
                return self._robot.ros_control.reverse(distance)
```
The backend robot references will be handled out of the box and will be less likely to error out.

### DIFF
--- a/tests/test_unitree_agent_queries_fastapi.py
+++ b/tests/test_unitree_agent_queries_fastapi.py
@@ -49,9 +49,9 @@ def main():
         streams = {"unitree_video": video_stream}
         web_interface = RobotWebInterface(port=5555, **streams)
 
-        logger.info("Starting perception agent")
+        logger.info("Starting action primitive execution agent")
         agent = OpenAIAgent(
-            dev_name="UnitreeQueryPerceptionAgent",
+            dev_name="UnitreeQueryExecutionAgent",
             input_query_stream=web_interface.query_stream,
             output_dir=output_dir,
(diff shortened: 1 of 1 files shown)


dimos/robot/unitree/unitree_go2.py
```diff
@@ -111,102 +115,17 @@ def __init__(
...
-            "Denote the number you see in the image. Only provide the number, without any other text in your response. If the number is above 500, but lower than 1000, then rotate the robot at 0.5 rad/s for 1 second.",
-            image_detail="high",
-            skills=skills_instance,
-            pool_scheduler=self.thread_pool_scheduler,
-            frame_processor=frame_processor,
-        )
-
     def do(self, *args, **kwargs):
         pass
+
```
lukasapaukstys: Seems redundant? Not sure its best to have dangling unused methods if the functionality is covered internally in this class

PR #112, opened 2025-03-12
WIP Dimos-unitree README / documentation

spomichter:


README.md
```diff
@@ -1 +1,250 @@
...
+```python
+class Jump(AbstractSkill):
+    def __call__(self):
+      return self.robot.jump()
+
+class Flip(AbstractSkill):
+    def __call__(self):
+      return self.robot.flip()
+
+class JumpAndFlip(AbstractSkill):
```
spomichter, line 215: Need feedback here specifically @lukasapaukstys on best way to allow for this. 

Given how you currently have skill creation structured this shouldn't add a ton of additional logic, and would be a great feature.

README.md
```diff
@@ -1 +1,250 @@
...
+# Call a ROS2 action primitive
+robot.move(distance=1.0, speed=0.5)
+```
+
+### Creating Custom Skills (non-unitree specific)
+
+#### Create basic custom skills by inheriting from ```AbstractSkill``` and implementing the ```__call__``` method.
+
+```python
+class Flip(AbstractSkill):
```
spomichter, line 199: I know I'm missing some stuff here like the robot parameter and the __init__, but would be very nice if we could conform to something like this structure for ease of use. @lukasapaukstys

PR #128, opened 2025-03-26
Feature: direct movement velocity controls added back to ROSControl

spomichter:


dimos/robot/unitree/unitree_skills.py
```diff
@@ -248,6 +248,21 @@ def __call__(self):
             super().__call__()
             return self._robot.spin(degrees=-self.degrees)  # Spinning right is negative degrees
 
+    class MoveVel(AbstractRobotSkill):
```
spomichter, line 251: If needed can split this into MoveVel and ReverseVel but depends on model performance @alexlin2

PR #129, opened 2025-03-26
Feature: Audio (TTS, STT) Support with Agent integration

leshy:
next steps:

* Lots to be done with generic nodes.
audio normalizer should just be normalizer, if we take care of our types the same node can scale video, audio data, lidar

* How are we handling singlechannel vs multichannel audio

* Test multiple subscribers to the same observable

* Remote audio capture from ros / unitree robot

* Silence detection, keyword detection

* Generic reactivex node tests
   Run a pure node, check for mem leaks etc, easy way to specify node/io for tests

* Integration tests
   use the same test above to run full pipelines,
   pipe stuff in, expect stuff out, check memory and cleanup

### DIFF
--- a/dimos/stream/audio/node_key_recorder.py
+++ b/dimos/stream/audio/node_key_recorder.py
@@ -24,14 +24,19 @@ class KeyRecorder(AbstractAudioTransform):
     def __init__(
         self,
         max_recording_time: float = 120.0,
+        always_subscribe: bool = False,
     ):
         """
         Initialize KeyRecorder.
 
         Args:
             max_recording_time: Maximum recording time in seconds
+            always_subscribe: If True, subscribe to audio source continuously,
+                              If False, only subscribe when recording (more efficient
+                              but some audio devices may need time to initialize)
         """
         self.max_recording_time = max_recording_time
+        self.always_subscribe = always_subscribe
 
         self._audio_buffer = []
         self._is_recording = False
@@ -53,7 +58,9 @@ class KeyRecorder(AbstractAudioTransform):
 
     def consume_audio(self, audio_observable: Observable) -> "KeyRecorder":
         """
-        Subscribe to an audio observable and record on key press.
+        Set the audio observable to use when recording.
+        If always_subscribe is True, subscribes immediately.
+        Otherwise, subscribes only when recording starts.
 
         Args:
             audio_observable: Observable emitting AudioEvent objects
@@ -63,12 +70,14 @@ class KeyRecorder(AbstractAudioTransform):
         """
         self._audio_observable = audio_observable
 
-        # Subscribe to the observable
-        self._subscription = audio_observable.subscribe(
-            on_next=self._process_audio_event,
-            on_error=self._handle_error,
-            on_completed=self._handle_completion,
-        )
+        # If configured to always subscribe, do it now
+        if self.always_subscribe and not self._subscription:
+            self._subscription = audio_observable.subscribe(
+                on_next=self._process_audio_event,
+                on_error=self._handle_error,
+                on_completed=self._handle_completion,
+            )
+            logger.debug("Subscribed to audio source (always_subscribe=True)")
 
         return self
 
@@ -94,6 +103,11 @@ class KeyRecorder(AbstractAudioTransform):
         """Stop recording and clean up resources."""
         logger.info("Stopping audio recorder")
 
+        # If recording is in progress, stop it first
+        if self._is_recording:
+            self._stop_recording()
+
+        # Always clean up subscription on full stop
         if self._subscription:
             self._subscription.dispose()
             self._subscription = None
@@ -105,11 +119,13 @@ class KeyRecorder(AbstractAudioTransform):
 
     def _input_monitor(self):
         """Monitor for key presses to toggle recording."""
-        print("Press Enter to start/stop recording...")
+        logger.info("Press Enter to start/stop recording...")
 
         while self._running:
             # Check if there's input available
             if select.select([sys.stdin], [], [], 0.1)[0]:
+                sys.stdin.readline()
+
                 if self._is_recording:
                     self._stop_recording()
                 else:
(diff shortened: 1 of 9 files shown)


dimos/speech/audio/abstract.py
```diff
@@ -0,0 +1,103 @@
+from abc import ABC, abstractmethod
```
spomichter: Change file name to base.py from abstract.py

dimos/speech/text/abstract.py
```diff
@@ -0,0 +1,46 @@
+from abc import ABC, abstractmethod
```
spomichter: Right now we have all text observables as general DataProvider via ```AbstractDataProvider``` where you can define subclasses and type cast the output observable i.e.```--> Observable[str]```.

So this abstract class is likely not needed and you can inherit from there

Feel free to double check best RxPY practices with @lukasapaukstys

dimos/stream/audio/sound_processing/node_key_recorder.py
```diff
@@ -0,0 +1,271 @@
...
+import logging
+import threading
+import sys
+import select
+from reactivex import Observable, create, disposable
+from reactivex.subject import Subject, ReplaySubject
+
+from dimos.stream.audio.sound_processing.abstract import AbstractAudioTransform, AudioEvent
+
+# Set up logging
```
spomichter: logging is handled by the following syntax:

```

# Initialize global logger for the agent module 
logger = setup_logger("dimos.agents") # example for agent.py
```

So in your case you can do ```logger = setup_logger("dimos.stream.audio") # or whatever the file is called```

And just use the global reference ```logger.info("logs")``` rather than ```self.logger.info("")```

dimos/stream/audio/sound_processing/node_normalizer.py
```diff
@@ -0,0 +1,205 @@
...
+import numpy as np
+from reactivex import Observable, create, disposable
+from typing import Optional, Dict, Any, Callable
+import logging
+import time
+from dimos.stream.audio.sound_processing.volume import calculate_rms_volume, calculate_peak_volume
+from dimos.stream.audio.sound_processing.abstract import AbstractAudioTransform, AudioEvent
+
+# Set up logging
+logging.basicConfig(level=logging.INFO)
```
spomichter: Change

dimos/stream/audio/sound_processing/node_microphone.py
```diff
@@ -0,0 +1,113 @@
...
+from dimos.stream.audio.sound_processing.abstract import AbstractAudioEmitter, AudioEvent
+import numpy as np
+from typing import Optional, List, Dict, Any
+from reactivex import Observable, create, disposable
+import time
+import logging
+import sounddevice as sd
+
+# Set up logging
+logging.basicConfig(level=logging.INFO)
```
spomichter: Change

dimos/stream/audio/sound_processing/node_output.py
```diff
@@ -0,0 +1,171 @@
...
+import numpy as np
+import time
+import logging
+import sounddevice as sd
+from reactivex import Observable, create, disposable
+
+from dimos.stream.audio.sound_processing.abstract import AbstractAudioTransform, AudioEvent
+
+# Set up logging
+logging.basicConfig(level=logging.INFO)
```
spomichter: Change

dimos/stream/audio/sound_processing/node_simulated.py
```diff
@@ -0,0 +1,222 @@
+from dimos.stream.audio.sound_processing.abstract import AbstractAudioEmitter, AudioEvent
+import numpy as np
+from reactivex import Observable, create, disposable
+from typing import Dict, List, Optional, Union, Any
+import threading
+import logging
+import time
+
+# Set up logging
+logging.basicConfig(level=logging.INFO)
```
spomichter: Change

dimos/stream/audio/sound_processing/node_key_recorder.py
```diff
@@ -0,0 +1,271 @@
...
+        self._channels = None  # Will be set from first event
+        
+        self._audio_observable = None
+        self._subscription = None
+        self._output_subject = Subject()  # For real-time passthrough
+        self._recording_subject = ReplaySubject(1)  # For completed recordings
+        
+        # Start a thread to monitor for input
+        self._running = True
+        self._input_thread = threading.Thread(target=self._input_monitor, daemon=True)
```
spomichter: All threading should technically be handled internal to RxPY to prevent process locks from halting the entire system. This comes more into play as we have more agents and multiple inputs/outputs. We have some threading utilities under dimos/utils which is where we can handle this later. 

For now we can leave as is and come back if it causes an issue.

dimos/stream/audio/sound_processing/node_volume_monitor.py
```diff
@@ -0,0 +1,168 @@
...
+
+from dimos.stream.audio.sound_processing.abstract import AudioEvent, AbstractAudioConsumer, AbstractAudioEmitter
+from dimos.stream.audio.text.abstract import AbstractTextEmitter
+from dimos.stream.audio.text.node_stdout import TextPrinterNode
+
+from dimos.stream.audio.sound_processing.volume import calculate_peak_volume
+
+
+# Set up logging
+logging.basicConfig(level=logging.INFO)
```
spomichter: Change

dimos/stream/audio/utils.py
```diff
@@ -0,0 +1,10 @@
+import time
```
spomichter, line 1: This is something we need as a general utility very helpful; will put on the to do list

dimos/stream/audio/sound_processing/node_key_recorder.py
leshy: yup, agreed, this should use "record trigger" nodes, that do keyword, silence or keyboard detections etc, if done properly but didn't want to go too deep into abstractions on the first pass

dimos/stream/audio/utils.py
leshy: yup, text nodes also should be extracted from this feature, but it's early, want to see how we end up dealing with this down the line

PR #147, opened 2025-04-02
Enable Local & Remote Hugging Face and GGUF Inference Agents (GPU-Ready)

lukasapaukstys:
## Introduces Two New Agent Classes:

- **`HuggingFaceLocalAgent`**: Provides local inference capabilities using Hugging Face models. Supports GPU acceleration and is optimized for execution within Docker environments. 
**Fully tested with:** ```./run.sh hf-local```
- **`HuggingFaceRemoteAgent`**: Enables remote inference via the Hugging Face API. Functionality is endpoint-dependent.
**Fully tested with:** ```./run.sh hf-remote```
- **`CTransformersGGUFAgent`**: Enables local inference via the CTransformers.
**Fully tested with:** ```./run.sh gguf```

## Running the Agents Locally

To run the agents in a Docker container with GPU (CUDA) support:

1. Comment out all lines in `dimos/robot/__init__.py` to disable default initialization.
2. (HuggingFace Local) From the project root, run:

    ```bash
    ./run.sh hf-local
    ```
    
    (GGUF Local) From the project root, run:

    ```bash
    ./run.sh gguf
    ```

## Other Changes

- Added licensing headers to newly created files and any existing files that were missing them.
- Added a sample video to the assets folder: `assets/trimmed_video_office.mov`
- Added a convenience `run.sh` shell script to the root directory.


dimos/agents/agent_huggingface_remote.py
spomichter, line 1: @lukasapaukstys I think also totally fine to put all hugging face agents in a huggingface.py

spomichter: pretty clean if in /agents each model vendor gets their own file openai.py, claude.py, deepseek.py, etc

docker/models/hugging_face_local/docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+    env_file:
+      - ../../../.env
+    mem_limit: 8048m
+    volumes:
+      - ../../../assets:/app/assets
+      - ../../../assets/model-cache:/root/.cache/huggingface/hub
+    ports:
+      - "5555:5555"
+    runtime: nvidia
+    environment:
```
spomichter: This is good any env variables not sourced in .env can be defined here. But as i mentioned apparently bad docker compose practice to have the same env variables in both .env and environment in the docker compose as it creates circular issues

spomichter: unrelated should probably systematically go through and clean out this docker sourcing issue i fixed for ivan last well.

docker/unitree/ros_agents/docker-compose.yml
```diff
@@ -0,0 +1,67 @@
...
+  dimos-unitree-ros-agents:
+    image: dimos-unitree-ros-agents:latest
+    build: 
+      context: ../../../
+      dockerfile: docker/unitree/ros_agents/Dockerfile
+    env_file:
+      - ../../../.env
+    environment:
+      PYTHONUNBUFFERED: 1
+      ROBOT_IP: ${ROBOT_IP:-192.168.9.140}
```
spomichter: Yeah so this here is problematic lol due to my previous comment. I know you proposed commenting out 'environment' last week but that's a bit messy idk.

tests/test_claude_agent_query.py
```diff
@@ -1,4 +1,16 @@
-#!/usr/bin/env python3
+# Copyright 2025 Dimensional Inc.
```
spomichter, line 1: This is good i'm totally forgetting to add these headers. Although im pretty sure the convention i set in dimos-unitree was only license headers in /dimos, not /tests as tests aren't proprietary. So should prob remove these headers on test files unless that convention is incorrect

lukasapaukstys: We should keep the license headers in the test files too. Even though the tests aren't the core code, they're still our work and should match the license of the main codebase. Big projects like Google do this to avoid confusion if files ever get copied around. Plus it just makes everything consistent - no need to think about which files need headers and which don't

docker/models/hugging_face_local/docker-compose.yml
lukasapaukstys: Looking this up, it's completely valid to define both an .env file and an environment block in the docker-compose files. Docker has a clear precedence order for how these values are loaded:

[Docker Envvars Precedence](https://docs.docker.com/compose/how-tos/environment-variables/envvars-precedence/)

This pattern (${ROBOT_IP:-192.168.9.140}) is actually recommended in Docker as it provides sensible defaults while allowing overrides through .env files or host environment variables. There's no circular issue - the environment section takes precedence over .env files, and the syntax explicitly handles the fallback case.

docker/unitree/ros_agents/docker-compose.yml
lukasapaukstys: See above comment on docker

dimos/agents/agent_huggingface_remote.py
lukasapaukstys: That's a good idea, having multiple agents under each vendor. Will likely look at how langchain / hugginggface both do this on their stacks for file organization.

tests/test_claude_agent_query.py
spomichter: Good with this

docker/models/hugging_face_local/docker-compose.yml
spomichter: Yes as this says though 'environment' section in docker compose takes precedence over .env file. So cannot have our hardcoded IP address as it overwrites .env and other developers running the docker will not connect to their robots and will be hard to notice why

docker/unitree/ros_agents/docker-compose.yml
lukasapaukstys: Went ahead and removed this default just now

dimos/agents/agent_huggingface_local.py
```diff
@@ -0,0 +1,227 @@
...
+                 device: str = "auto",
+                 query: str = "How many r's are in the word 'strawberry'?",
+                 input_query_stream: Optional[Observable] = None,
+                 input_video_stream: Optional[Observable] = None,
+                 output_dir: str = os.path.join(os.getcwd(), "assets",
+                                                "agent"),
+                 agent_memory: Optional[AbstractAgentSemanticMemory] = None,
+                 system_query: Optional[str] = None,
+                 max_output_tokens_per_request: int = None,
+                 max_input_tokens_per_request: int = None,
```
spomichter, line 60: Added to improve "tool calling"

dimos/robot/__init__.py
```diff
@@ -1,3 +0,0 @@
-from .ros_control import ROSControl, RobotMode
```
spomichter, line 1: No longer needed

dimos/robot/unitree/unitree_go2.py
```diff
@@ -65,7 +65,7 @@ def __init__(
             use_webrtc: bool = False,
             disable_video_stream: bool = False,
             mock_connection: bool = False,
-            enable_visual_servoing: bool = True,
+            enable_visual_servoing: bool = False,
```
spomichter, line 68: Running ./run.sh 0 the video feed from unitree on the interface is halted due to visual servoing, so need to set to False here by default to prevent this. fyi @alexlin2 TO DO

docker/models/ctransformers_gguf/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+
+# Create symlink for python
+RUN ln -sf /usr/bin/python3.10 /usr/bin/python
+
+# Change working directory to /app for proper relative pathing
+WORKDIR /app
+
+COPY requirements.txt ./
+
+COPY ./dimos/perception/external ./dimos/perception/external
```
spomichter, line 35: This is bad practice just putting here now to get working

docker/models/huggingface_local/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+    env_file:
+      - ../../../.env
+    mem_limit: 8048m
+    volumes:
+      - ../../../assets:/app/assets
+      - ../../../assets/model-cache:/root/.cache/huggingface/hub
+    ports:
+      - "5555:5555"
+    runtime: nvidia
+    environment:
```
spomichter, line 17: Even running source .env does not fix the ROBOT_IP not set error so removed ROBOT_IP from this environment section. leaving here so its easy to hard code env variables from docker compose but we should force devs to use .env for ROBOT_IP or just passing directly ```ROBOT_IP=... docker compose -f ...```

docker/unitree/agents_interface/Dockerfile
```diff
@@ -99,6 +99,7 @@ COPY requirements.txt /app/
 
 WORKDIR /app
 
+COPY ./dimos/perception/external ./dimos/perception/external
```
spomichter, line 102: See discord, was required to prevent errors

docker/unitree/agents_interface/docker-compose.yml
```diff
@@ -8,13 +8,8 @@ services:
     env_file:
       - ../../../.env
     environment:
-      PYTHONUNBUFFERED: 1
-      ROBOT_IP: ${ROBOT_IP}
```
spomichter, line 12: See above comment

docker/unitree/ros_agents/docker-compose.yml
```diff
@@ -0,0 +1,67 @@
...
+  dimos-unitree-ros-agents:
+    image: dimos-unitree-ros-agents:latest
+    build: 
+      context: ../../../
+      dockerfile: docker/unitree/ros_agents/Dockerfile
+    env_file:
+      - ../../../.env
+    environment:
+      PYTHONUNBUFFERED: 1
+      ROBOT_IP: ${ROBOT_IP}
```
spomichter, line 12: TODO: remove. ros_agents is deprecated and replaced with agents_interface as the master test file

tests/test_agent_ctransformers_gguf.py
```diff
@@ -0,0 +1,44 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tests.test_header
+
+from dimos.agents.agent_ctransformers_gguf import CTransformersGGUFAgent
+
+system_query = "You are a robot with the following functions. Move(), Reverse(), Left(), Right(), Stop(). Given the following user comands return the correct function."
```
spomichter, line 19: I changed this demo to one that was a bit more relevant. Also removed the query stream since GGUF may be super slow depending on the model and machine so multiple queries will crash it. @lukasapaukstys

tests/test_agent_huggingface_local.py
```diff
@@ -0,0 +1,72 @@
...
+)
+
+# Start the query stream.
+# Queries will be pushed every 1 second, in a count from 100 to 5000.
+# This will cause listening agents to consume the queries and respond
+# to them via skill execution and provide 1-shot responses.
+query_provider.start_query_stream(
+    query_template=
+    "{query}; User: travel forward by 10 meters",
+    frequency=10,
```
spomichter, line 64: Increased frequency for easy of reading logs to see if the agent is working

tests/test_huggingface_llm_agent.py
```diff
@@ -0,0 +1,114 @@
...
+            model_name= "Qwen/Qwen2.5-3B",
+            agent_type="HF-LLM",
+            input_query_stream=query_provider.data_stream,
+            process_all_inputs=False,
+            # output_dir=self.output_dir,
+            # skills=skills_instance,
+            # frame_processor=frame_processor,
+        )
+
+        # TESTING REMOTE AGENT
```
spomichter, line 71: TODO: cleanup

tests/test_standalone_hugging_face.py
```diff
@@ -0,0 +1,153 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import tests.test_header
+
+# from transformers import AutoModelForCausalLM, AutoTokenizer
```
spomichter, line 17: TODO: definitely cleanup

PR #167, opened 2025-04-10
Removed Vector Perception as external dependency, added Navigate to Object feature

alexlin2:


dimos/robot/unitree/unitree_go2.py
```diff
@@ -136,36 +139,121 @@ def __init__(
...
             self.video_stream = None
 
-        self.enable_visual_servoing = enable_visual_servoing
         # Initialize visual servoing if enabled
-        if enable_visual_servoing and self.video_stream is not None:
-            video_stream = self.get_ros_video_stream(fps=10)
-            person_tracker = PersonTrackingStream(
+        if self.video_stream is not None:
+            self.video_stream_ros = self.get_ros_video_stream(fps=8)
+            self.person_tracker = PersonTrackingStream(
```
spomichter, line 145: It seems like since we removed ```enable_visual_servoing``` a PersonTrackingStream will start with every UnitreeGo2 initialization. Probably not ideal

dimos/robot/unitree/unitree_go2.py
```diff
@@ -136,36 +139,121 @@ def __init__(
...
                 camera_intrinsics=self.camera_intrinsics,
                 camera_pitch=self.camera_pitch,
                 camera_height=self.camera_height
             )
-            person_tracking_stream = person_tracker.create_stream(video_stream)
-            self.visual_servoing = VisualServoing(tracking_stream=person_tracking_stream)
+            self.object_tracker = ObjectTrackingStream(
```
spomichter, line 150: Same here with the ObjectTrackingStream

dimos/robot/unitree/unitree_go2.py
alexlin2: Yeah, I can move the person_tracking stream into the follow_human call and only instantiate when it get's called, but that would be make it not possible to visualize it in the web server for debugging purposes unless we start the person following before starting the web server, which doesn't make sense.  Also I remember Lucas saying that only instantiating the stream won't actually run the code in the pipeline. ReactiveX Observables are designed as "lazy" constructs—they only execute their internal logic when they have an active subscriber. So I think it should be fine to just initialize the stream but not subscribe to it. The subscription only happens when you call person follow.

dimos/robot/unitree/unitree_go2.py
alexlin2: See comment above

dimos/robot/unitree/unitree_go2.py
spomichter: Thanks yes this is correct. Resolving for now. Eventually as robots start to have many streams we will want to move this to some other constructor like RobotPerception or something, as ideally the Robot class is relating to hardware commands not perception.

tests/test_object_tracking_with_qwen.py
```diff
@@ -0,0 +1,181 @@
...
+import cv2
+import numpy as np
+import queue
+import threading
+import json
+from reactivex import Subject, operators as RxOps
+from openai import OpenAI
+
+# Add the parent directory to the Python path
+sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```
spomichter: also reminder to use import tests.test_header. i will fix

dimos/models/qwen/video_query.py
```diff
@@ -116,7 +116,7 @@ def get_bbox_from_qwen(
             
             # Extract and validate bbox
             if 'bbox' in result and len(result['bbox']) == 4:
-                return result['bbox']
+                return result['bbox'], result['size']
```
spomichter, line 119: This breaks all tests that treat output from qwen as a direct BBOX not a tuple with bbox, size @alexlin2

dimos/robot/unitree/unitree_go2.py
```diff
@@ -136,36 +139,120 @@ def __init__(
...
+                camera_pitch=self.camera_pitch,
+                camera_height=self.camera_height
+            )
+            person_tracking_stream = self.person_tracker.create_stream(self.video_stream_ros)
+            object_tracking_stream = self.object_tracker.create_stream(self.video_stream_ros)
+
             self.person_tracking_stream = person_tracking_stream
+            self.object_tracking_stream = object_tracking_stream
 
     def follow_human(self, distance: int = 1.5, timeout: float = 20.0, point: Tuple[int, int] = None):
```
spomichter, line 161: TODO: Move this under /skills

dimos/robot/unitree/unitree_go2.py
```diff
@@ -136,36 +139,120 @@ def __init__(
...
+            x_vel = output.get("linear_vel")
+            z_vel = output.get("angular_vel")
+            logger.debug(f"Following human: x_vel: {x_vel}, z_vel: {z_vel}")
+            self.ros_control.move_vel_control(x=x_vel, y=0, yaw=z_vel)
+            time.sleep(0.05)
+        person_visual_servoing.stop_tracking()
+        del person_visual_servoing
+        return success
+        
+    def navigate_to(self, object_name: str, distance: float = 1.0, timeout: float = 40.0, max_retries: int = 3):
```
spomichter, line 179: And this

PR #198, opened 2025-04-22
feat/websocket visualization

leshy:
Implements websocket based visualizer for common types like costmap, path and vector.

- Non invasive to the codebase, only updates these types with serialize() methods
- Any submodule can return an observable of these types and they will be automatically visualized 

```py
websocket_vis.connect(planner.vis_stream())
```

in planner (at any point):

```py
        self.vis("global_costmap", smudge)
        self.vis("pos", pos)
```

(multiple subsystems can draw in the same context, so local/global planners, semantic map can share the vis)

initial UI is scrapped quickly - main part is that we can pass real time internal state to web, so can use this in the dimos console or other UIs

run

```sh
python tests/test_websocketvis.py
```
for demo, or
```sh
python tests/test_websocketvis.py --live
```

for live robot connection/mouse control

### DIFF
--- a/dimos/robot/global_planner/planner.py
+++ b/dimos/robot/global_planner/planner.py
@@ -21,65 +21,17 @@ from dimos.types.path import Path
 from dimos.types.costmap import Costmap
 from dimos.robot.global_planner.algo import astar
 from dimos.utils.logging_config import setup_logger
-from dimos.web.websocket_vis.types import Drawable, Visualizable
-from reactivex.subject import Subject
-from reactivex import Observable
+from dimos.web.websocket_vis.types import Visualizable
 
 logger = setup_logger("dimos.robot.unitree.global_planner")
 
 
 @dataclass
 class Planner(Visualizable):
-    local_nav: Callable[[VectorLike], bool]
+    local_nav: Callable[[Path], bool]
 
     @abstractmethod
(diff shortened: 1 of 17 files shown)


dimos/robot/global_planner/planner.py
```diff
@@ -15,28 +15,52 @@
...
 from dimos.types.path import Path
 from dimos.types.costmap import Costmap
 from dimos.robot.global_planner.algo import astar
 from dimos.utils.logging_config import setup_logger
-from nav_msgs import msg
+from dimos.web.websocket_vis.types import Drawable, Visualizable
+from reactivex.subject import Subject
+from reactivex import Observable
 
 logger = setup_logger("dimos.robot.unitree.global_planner", level=logging.DEBUG)
```
spomichter: TODO: in production no hardcoding of logging levels. DIMOS_LOG_LEVEL env variable is used to manage verbose logging

dimos/web/websocket_vis/deno.lock
```diff
@@ -0,0 +1,157 @@
+{
```
spomichter, line 1: This should be added to gitignore right.

leshy: no, this pins deps to specific package versions

PR #209, opened 2025-04-25
Major Refactor: Moved person following skill and navigate to object skill under skills folder

alexlin2:
Move both skills to visual_navigation_skills.py and added stop function to both skills. Also made some parameter changes to local planner and slowdown logic when turning. Tested with claude thinking agent.

### DIFF
--- a/dimos/robot/unitree/unitree_go2.py
+++ b/dimos/robot/unitree/unitree_go2.py
@@ -183,82 +183,10 @@ class UnitreeGo2(Robot):
 
         # Create the visualization stream at 5Hz
         self.local_planner_viz_stream = self.local_planner.create_stream(frequency_hz=5.0)
-        
-        # Add visual navigation skills if the skill library exists
-        if self.skill_library is not None:
-            self.skill_library.add(FollowHuman)
-            self.skill_library.add(NavigateToObject)
-            
-            # Create instances of the skills
-            self.skill_library.create_instance("FollowHuman", robot=self)
-            self.skill_library.create_instance("NavigateToObject", robot=self)
-
-    def follow_human(self, distance: int = 1.5, timeout: float = 20.0, point: Tuple[int, int] = None):
-        """
-        Follow a human using visual servoing.
-        
-        Args:
-            distance: Desired distance to maintain from person in meters
-            timeout: Maximum time to follow in seconds
-            point: Optional starting point for tracking (x,y pixel coordinates)
-            
-        Returns:
-            bool: True if successful, False otherwise
-        """
-        if self.skill_library is None:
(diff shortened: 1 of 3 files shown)


dimos/robot/unitree/unitree_go2.py
```diff
@@ -182,23 +183,45 @@ def __init__(
 
         # Create the visualization stream at 5Hz
         self.local_planner_viz_stream = self.local_planner.create_stream(frequency_hz=5.0)
+        
+        # Add visual navigation skills if the skill library exists
+        if self.skill_library is not None:
```
spomichter: This can be removed as we can should add skills to unitree go2 in unitree_skills. 

Or as shown in test_claude_agent_skills_query we can add it right in the test file

dimos/robot/unitree/unitree_go2.py
```diff
@@ -182,23 +183,45 @@ def __init__(
...
+            self.skill_library.add(FollowHuman)
+            self.skill_library.add(NavigateToObject)
+            
+            # Create instances of the skills
+            self.skill_library.create_instance("FollowHuman", robot=self)
+            self.skill_library.create_instance("NavigateToObject", robot=self)
 
     def follow_human(self, distance: int = 1.5, timeout: float = 20.0, point: Tuple[int, int] = None):
```
spomichter: Having this method still here defeats the purpose of the refactor to skills. 

Skills will be called by the agent that then run the .call() method themselves outside of any Robot class or subclass

dimos/robot/unitree/unitree_go2.py
```diff
@@ -182,23 +183,45 @@ def __init__(
...
+                                            timeout=timeout,
+                                            point=point)
+            
+            # For backward compatibility, return a boolean success indicator
+            return "Failed" not in result if isinstance(result, str) else False
+        except Exception as e:
+            logger.error(f"Error calling FollowHuman skill: {e}")
+            return False
 
     def navigate_to(self, object_name: str, distance: float = 1.0, timeout: float = 40.0, max_retries: int = 3):
```
spomichter: Same with this method. this should be in the call method of the skill in question

dimos/robot/unitree/unitree_skills.py
```diff
@@ -293,23 +293,4 @@ class Wait(AbstractRobotSkill):
 
         def __call__(self):
             super().__call__()
-            return time.sleep(self.seconds)
-
-    class FollowHuman(AbstractRobotSkill):
```
spomichter, line 298: You're right to remove these from unitree skills as theyre not unitree related, they are general

PR #212, opened 2025-04-26
Pytest implementation

spomichter:


pytest.ini
leshy, line 1: can leave this in pyproject.toml, no need for a separate file

run_tests.sh
leshy, line 1: We have a /bin dir where we've been placing helper scripts.

Also probably it's best to not use a script but configure pytest so that just "pytest" command does a correct thing. (excludes hardware, integration tests) 

because proxies for tooling usually restrict how you can use them, for example first thing I want is to run is `pytest -s -k some_filter` - within my actual work dir to test my components and downstream.

PR #237, opened 2025-04-30
Local planner finetunes, major TTS update, Speak Skill updates, new master testfile for demos

spomichter:
Changes:
- Cosmetic changes to global planner printing
- Local planner disabled for testing
- Local planner parameter finetunes @alexlin2 
- Major TTS/Speak Skill changes: Added speech queue to prevent pyaudio crashes and the robot talking over itself, fixes to audio event handling to prevent invalid, corrupted, or short audio events. Also fixes buffer size issues
- Added TTS custom voices and speed parameter

### DIFF
--- a/dimos/skills/navigation.py
+++ b/dimos/skills/navigation.py
@@ -377,6 +377,9 @@ class Navigate(AbstractRobotSkill):
             # The scheduler approach isn't working, switch to direct threading
             # Define a navigation function that will run on a separate thread
             def run_navigation():
+                skill_library = self._robot.get_skills()
+                self.register_as_running("Navigate", skill_library)
+                
                 try:
                     logger.info(f"Starting navigation to ({pos_x:.2f}, {pos_y:.2f}) with rotation {theta:.2f}")
                     # Pass our stop_event to allow cancellation
@@ -394,6 +397,8 @@ class Navigate(AbstractRobotSkill):
                 except Exception as e:
                     logger.error(f"Unexpected error in navigation thread: {e}")
                     return False
+                finally:
+                    self.stop()
             
             # Cancel any existing navigation before starting a new one
             # Signal stop to any running navigation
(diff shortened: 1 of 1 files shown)


dimos/stream/audio/node_key_recorder.py
leshy, line 1: this whole file probably can be reviewed and reduced by at least  50% of lines, but we don't care :D

dimos/skills/speak.py
```diff
@@ -1,8 +1,65 @@
...
+            
+        except queue.Empty:
+            # No tasks in queue, just continue waiting
+            continue
+        except Exception as e:
+            logger.error(f"Error in audio queue processor: {e}")
+            # Continue processing other tasks
+
+
+def start_audio_queue_processor():
```
leshy, line 46: can we extract to a proper speech module?

robot.speech = new Speech(audio_in, audio_out) 
robot.speech.say('bla')
robot.speech.listen().subscribe(..)

idk something like this

dimos/skills/speak.py
```diff
@@ -12,14 +69,87 @@ class Speak(AbstractSkill):
...
+            logger.error("No TTS node provided to Speak skill")
+            return "Error: No TTS node available"
+        
+        # Create a result queue to get the result back from the audio thread
+        result_queue = queue.Queue(1)
         
-        # Connect the Subject to the TTS node
-        self._tts_node.consume_text(text_subject)
+        # Define the speech task to run in the audio queue
+        def speak_task():
```
leshy, line 85: I assume this will be super short with a speech module

dimos/skills/speak.py
spomichter: Yes TODO

dimos/stream/audio/node_key_recorder.py
spomichter: I assume so, not sure as I did not write it. Most likely.

dimos/skills/speak.py
spomichter: Yes we need some speech queue to prevent overlapping outputs from TTS.

PR #239, opened 2025-05-01
Feat: Yolo Query Stream for that can be attached to each prompt

alexlin2:
Run test file: `python tests/test_yolo_query_stream.py --mode robot` to test on robot or --mode webcam to run on webcam

### DIFF
--- a/dimos/agents/claude_agent.py
+++ b/dimos/agents/claude_agent.py
@@ -49,7 +49,7 @@ from dimos.utils.threadpool import get_scheduler
 load_dotenv()
 
 # Initialize logger for the Claude agent
-logger = setup_logger("dimos.agents.claude", level=logging.DEBUG)
+logger = setup_logger("dimos.agents.claude")
 
 # Response object compatible with LLMAgent
 class ResponseMessage:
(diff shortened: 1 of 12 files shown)


dimos/perception/yolo_query_stream.py
```diff
@@ -0,0 +1,247 @@
...
+from dimos.models.depth.metric3d import Metric3D
+from dimos.perception.detection2d.utils import (
+    calculate_depth_from_bbox,
+    calculate_object_size_from_bbox,
+    calculate_position_rotation_from_bbox
+)
+from dimos.types.vector import Vector
+
+
+class YoloQueryStream:
```
spomichter, line 16: This is a good interface for the YOLO master node. Think also worth adding some simple utilities / getter methods that allow me in ClaudeAgent to very easily to run yolo_query_stream.get_label() and i get the most up to date labels, no need to do any sunscription or run the ops.just() operator

spomichter, line 16: Although perhaps we can make this .get_stream() utility work for any of these query stream types.

spomichter, line 16: Also is there any parent class that this should inherit from? IT seems likely that we will have many other types of streams and thus maybe already have this ```Base``` type maybe the ```DataProvider``` or ```AbstractDataProvider``` class

leshy: ya writing all this in the next pass for universal streams, for now supported by topics

PR #246, opened 2025-05-03
Feature - Network Video Provider via FFMPEG

lukasapaukstys:
Adds network video ingestion via an FFmpeg shim that implements our `VideoProvider` interface. This replaces the unreliable RTSP ingestion in OpenCV and removes the complexity of its numerous import options by leveraging FFmpeg’s superior performance and stability.

## To test locally

1. Add `TEST_RTSP_URL` to your `.env` file.  
2. Run:
   ```bash
   python3 -m tests.test_rtsp_video_provider
 The live RTSP stream will be saved as JPEG frames (in the assets folder).

 The stream will also be served by the built-in web server.

 Note: Firefox may freeze when viewing the stream due to missing back-pressure handling; Chrome does not exhibit this issue.

### DIFF
--- a/dimos/stream/rtsp_video_provider.py
+++ b/dimos/stream/rtsp_video_provider.py
@@ -28,13 +28,14 @@ from reactivex.disposable import Disposable
 from reactivex.observable import Observable
 from reactivex.scheduler import ThreadPoolScheduler
 
+from dimos.utils.logging_config import setup_logger
 from dimos.stream.frame_processor import FrameProcessor
 from dimos.stream.video_operators import VideoOperators as vops
 # Assuming AbstractVideoProvider and exceptions are in the sibling file
 from .video_provider import (AbstractVideoProvider, VideoFrameError,
                              VideoSourceError, get_scheduler)
 
-logger = logging.getLogger(__name__)
+logger = setup_logger("dimos.stream.rtsp_video_provider")
 
 
 class RtspVideoProvider(AbstractVideoProvider):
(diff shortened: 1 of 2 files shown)


dimos/stream/rtsp_video_provider.py
```diff
@@ -0,0 +1,347 @@
...
+from reactivex.observable import Observable
+from reactivex.scheduler import ThreadPoolScheduler
+
+from dimos.stream.frame_processor import FrameProcessor
+from dimos.stream.video_operators import VideoOperators as vops
+# Assuming AbstractVideoProvider and exceptions are in the sibling file
+from .video_provider import (AbstractVideoProvider, VideoFrameError,
+                             VideoSourceError, get_scheduler)
+
+logger = logging.getLogger(__name__)
```
spomichter: Need to use our dimOS logger

tests/test_rtsp_video_provider.py
```diff
@@ -0,0 +1,144 @@
...
+
+import numpy as np
+import reactivex as rx
+from reactivex import operators as ops
+
+from dimos.stream.frame_processor import FrameProcessor
+from dimos.stream.video_operators import VideoOperators as vops
+from dimos.stream.video_provider import get_scheduler
+
+logger = logging.getLogger(__name__)
```
spomichter: same here

dimos/stream/rtsp_video_provider.py
spomichter: Fixed

PR #275, opened 2025-05-14
Point Cloud Filtering and Segmentation, Full 6DOF Object pose estimation, Grasp generation, ZED driver support

alexlin2:
Implementation of pointcloud filtering for object mask from object detection stream. ZED driver support and integration with dimos stack. 

Current issues:
-ZED has some threading issue, it has to be ran from the main thread and is blocking. Using observe_on with reactivex doesn't work. Results in choppy and slow frames

### DIFF
--- a/dimos/perception/manip_aio_processer.py
+++ b/dimos/perception/manip_aio_processer.py
@@ -178,28 +178,30 @@ class ManipulationProcessor:
 
             # Get full point cloud
             full_pcd = self.pointcloud_filter.get_full_point_cloud()
-            
+
             # Extract misc/background points and create voxel grid
             misc_start = time.time()
             all_filtered_objects = segmentation_filtered_objects + detected_objects
             misc_clusters, misc_voxel_grid = extract_and_cluster_misc_points(
-                full_pcd, 
+                full_pcd,
                 all_filtered_objects,
                 eps=0.03,
                 min_points=100,
                 enable_filtering=True,
-                voxel_size=0.02
+                voxel_size=0.02,
             )
             misc_time = time.time() - misc_start
-            
+
             # Store results
-            results.update({
-                'detected_objects': detected_objects,
-                'all_objects': all_objects,
-                'full_pointcloud': full_pcd,
-                'misc_clusters': misc_clusters,
-                'misc_voxel_grid': misc_voxel_grid
-            })
+            results.update(
+                {
+                    "detected_objects": detected_objects,
+                    "all_objects": all_objects,
+                    "full_pointcloud": full_pcd,
+                    "misc_clusters": misc_clusters,
+                    "misc_voxel_grid": misc_voxel_grid,
+                }
+            )
 
             # Create point cloud visualizations
             base_image = colorize_depth(depth_image, max_depth=10.0)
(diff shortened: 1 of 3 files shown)


dimos/stream/stereo_camera_streams/zed.py
```diff
@@ -0,0 +1,164 @@
...
+try:
+    import pyzed.sl as sl
+except ImportError:
+    sl = None
+    logging.warning("ZED SDK not found. Please install pyzed to use ZED camera functionality.")
+
+logger = logging.getLogger(__name__)
+
+
+class ZEDCameraStream:
```
spomichter, line 33: Inherit from AbstractVideoProvider or VideoProvider (may be harder)

dimos/stream/stereo_camera_streams/zed.py
```diff
@@ -0,0 +1,164 @@
...
+            try:
+                # Initialize camera
+                if not self.is_opened:
+                    self._initialize_camera()
+
+                while True:
+                    # Capture frame directly
+                    left_img, right_img, depth_img = self.zed_camera.capture_frame()
+
+                    if left_img is not None and depth_img is not None:
```
spomichter, line 100: Maybe want to have seperate observables for left right and depth but probably doesnt really matter

dimos/stream/stereo_camera_streams/zed.py
```diff
@@ -0,0 +1,164 @@
...
+            finally:
+                # Clean up resources
+                self._cleanup_camera()
+                observer.on_completed()
+
+        return create(emit_frames).pipe(
+            ops.share(),  # Share the stream among multiple subscribers
+        )
+
+    def get_camera_info(self) -> Dict[str, float]:
```
spomichter, line 122: what is this method for if you can just run self.zed_camera.get_camera_info() and access intrinsics that way? Ideally anything relating to hardware should be accessed from the hardware interface in /hardware and not via /streams

tests/test_pointcloud_filtering.py
```diff
@@ -0,0 +1,248 @@
...
+            camera_intrinsics=camera_intrinsics,
+            min_confidence=min_confidence,
+            class_filter=None,
+            detector=detector,
+            video_stream=video_stream,
+            disable_depth=True,
+        )
+
+        # Initialize point cloud filtering
+        pointcloud_filter = PointcloudFiltering(
```
spomichter: May be a better API interface to have PointCloudFiltering object take in the CameraInterface / HardwareInterface as a init, so then internally it can extract the necessary intrinsics from camera.intrinsics

tests/test_pointcloud_filtering.py
```diff
@@ -0,0 +1,248 @@
...
+                    filtered_objects = pointcloud_filter.process_images(
+                        rgb, depth, result["objects"]
+                    )
+
+                    if filtered_objects:
+                        # Create base image (colorized depth)
+                        base_image = colorize_depth(depth, max_depth=10.0)
+
+                        # Create point cloud overlay visualization
+                        overlay_viz = create_point_cloud_overlay_visualization(
```
spomichter: This stream operation is a great example of where we can begin to define PerceptionStream classes that define pre-built stream 'recipes'. in this case for pointcloud overlay. So we can remove this logic from the runfile to some class like that.

.gitignore
```diff
@@ -17,6 +17,7 @@ __pycache__
 
 # Ignore default runtime output folder
 /assets/output/
+/assets/rgbd_data/
```
spomichter, line 20: All of /assets should be in gitignore

PR #278, opened 2025-05-16
Development container & CI

leshy:
# DEV env & CI sketch

## Goal
- Provide a unified development environment that uses the same tooling across different OSes and IDEs and CI
- Performant and easy way to enter dimos shell cross platform and cross ide
- If running builds, do it as fast as possible since this hurts the dev cycle

## Implementation
- We build a chain of docker images (ros -> python -> dev env) and run tests on top of the dev env. (these are supposed to be run for each branch only if needed (ros not rebuilt if there are no ros level changes in the branch)

- Example run (skipped ros build) https://github.com/dimensionalOS/dimos/actions/runs/15163702331

- Dev env image conforms to devcontainer spec and can be used by IDEs locally. (this is Recomended and supported by IDEs like Cursor/VS Code https://code.visualstudio.com/docs/devcontainers/tutorial)

- We run CI tests on top of the newly built dev container

once builds finish you can run them from github
'docker pull ghcr.io/dimensionalos/dev:env_devcontainer2'
image is tagged with your branch name

## Todo
- validate dev container in cursor/vscode - write readme for devs
- implement CI/dev container linting and mypy checks

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,7 @@ opencv-python
 python-dotenv
 openai
 anthropic>=0.19.0
+numpy>=1.26.4,<2.0.0
 colorlog==6.9.0
 yapf==0.40.2
 typeguard
@@ -22,7 +23,7 @@ git+https://github.com/dimensionalOS/rxpy-backpressure.git
 pytest-asyncio==0.26.0
 asyncio==3.4.3
 -e git+https://github.com/legion1581/go2_webrtc_connect.git@fe64abb5987594e8c048427a98445799f6f6a9cc#egg=go2_webrtc_connect
-# -e git+https://github.com/legion1581/aioice.git@ff5755a1e37127411b5fc797c105804db8437445#egg=aioice
+#-e git+https://github.com/legion1581/aioice.git@ff5755a1e37127411b5fc797c105804db8437445#egg=aioice
 
 # Web Extensions
 fastapi>=0.115.6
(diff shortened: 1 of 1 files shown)


docker/depricated/agent/Dockerfile
spomichter, line 1: since you're adding an extra directory 'deprecated' to the tree every docker-compose will break as it uses relative paths to find the project root + volume mounts

docker/depricated/jetson/jetson_requirements.txt
spomichter, line 1: jetson should likely not be deprecated as its standalone

requirements.txt
```diff
@@ -89,3 +88,7 @@ lvis
 nltk
 git+https://github.com/openai/CLIP.git
 git+https://github.com/facebookresearch/detectron2.git@v0.6
+
+# Mapping
+open3d
+numpy==1.26.4
```
spomichter: May need range here as mentioned between 2.0 and 1.24 or whatever it was

docker/depricated/agent/Dockerfile
leshy: True, after discussion on discord we will review these builds on a per-need basis

docker/depricated/jetson/jetson_requirements.txt
leshy: True, after discussion on discord we will review these builds on a per-need basis

requirements.txt
leshy: fixed, rebuilding to confirm

PR #294, opened 2025-05-22
env/devcontainer ruff formatting/typing

leshy:
💞 Implements ruff formatting/typing on top of devcontainers and newly auto-built dev images 💞
cursor/vscode will detect a dev container, just allow it to use it, the rest should be automatic

### DIFF
--- a/docker/dev/dev-requirements.txt
+++ b/docker/dev/dev-requirements.txt
@@ -1,3 +1,2 @@
-ruff                                     0.11.10
-mypy                                     1.15.0
-mypy_extensions                          1.1.0
+ruff==0.11.10
+mypy==1.15.0


.devcontainer/devcontainer.json
```diff
@@ -1,13 +1,26 @@
 {
   "name": "dimos-dev",
-  "image": "ghcr.io/dimensionalos/dimos-dev:latest",
-  "postCreateCommand": "pre-commit install",
+  "image": "ghcr.io/dimensionalos/dev:dev",
```
spomichter, line 3: There is an issue jacob mentioned as this image isn't public, so he can't use. TODO on my end.

leshy: aha didn't know people are actually using our dev images. yeah I have no permission access

PR #295, opened 2025-05-22
Global code reformat with ruff

leshy:
Code formatting
`ruff format *`

With small fixes (like unused imports)
`ruff check --fix *`

Depends on (uses tooling from) https://github.com/dimensionalOS/dimos/pull/294

Everything is default ruff formatting, just as requested I changed line lengths to
`line-length = 120`
might want 100?  idk

Also it went through the whole dimos/ dir, touching stuff like `models/Detic` - not sure if it should, lmk if some dirs need an ignore

if you want to play with this locally, 
```sh
git fetch
git checkout env/devcontainer-ruff
```

edit ruff settings in `pyproject.toml` and just  rerun on the whole codebase, see if you like it

`ruff format dimos/*`

`ruff check dimos/*`

### DIFF
--- a/docker/dev/dev-requirements.txt
+++ b/docker/dev/dev-requirements.txt
@@ -1,3 +1,2 @@
-ruff                                     0.11.10
-mypy                                     1.15.0
-mypy_extensions                          1.1.0
+ruff==0.11.10
+mypy==1.15.0


dimos/agents/memory/chroma_impl.py
```diff
@@ -119,53 +113,51 @@ def create(self):
...
+
         self.model_name = model_name
         super().__init__(collection_name=collection_name)
 
     def create(self):
         """Create local embedding model and initialize the ChromaDB client."""
         # Load the sentence transformer model
         # Use CUDA if available, otherwise fall back to CPU
-        device = 'cuda' if torch.cuda.is_available() else 'cpu'
+        device = "cuda" if torch.cuda.is_available() else "cpu"
```
spomichter, line 138: what is the parentheses convention that ruff enforces

PR #296, opened 2025-05-22
moved robot_location type to spatial memory file

leshy:
robot_location type is very specific to spatial memory, so moved it there from core types which likely will be rewritten to be LCM types soon


dimos/perception/spatial_perception.py
```diff
@@ -83,24 +179,25 @@ def __init__(
...
         self.visual_memory_path = visual_memory_path
         self.output_dir = output_dir
-        
+
         # Setup ChromaDB client if not provided
         self._chroma_client = chroma_client
         if chroma_client is None and db_path is not None:
             # Create db directory if needed
             os.makedirs(db_path, exist_ok=True)
-
```
spomichter, line 97: Need to re-apply latest Ruff reformat to this PR due to formatting revertions

dimos/perception/spatial_perception.py
```diff
@@ -16,43 +16,139 @@
...
 from dimos.agents.memory.visual_memory import VisualMemory
 from dimos.types.vector import Vector
-from dimos.types.robot_location import RobotLocation
+from dataclasses import dataclass, field
 
 logger = setup_logger("dimos.perception.spatial_memory")
 
+
+@dataclass
+class RobotLocation:
```
spomichter, line 40: Need a test for this before moving. Will implement full robot testing pipeline in sim soon.

PR #298, opened 2025-05-26
Auto-delivery of binary data files for testing, rewrite of dev script

leshy:
- creates a pre-push system that compresses and pushes data dirs for testing into github LFS
- creates easy testing facilities for auto-downloading & decompression of test data
- ./bin/dev now uses standard devcontainer interface (auto downloads required image & devcontainer CLI)
- uses a pre-push hook to compress new test data (coming in a follow up PR)

### DIFF
--- a/dimos/utils/test_testing.py
+++ b/dimos/utils/test_testing.py
@@ -3,6 +3,7 @@ import os
 import subprocess
 from dimos.utils import testing
 
+
 def test_pull_file():
     repo_root = testing._get_repo_root()
     test_file_name = "cafe.jpg"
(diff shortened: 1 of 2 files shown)


bin/dev
```diff
@@ -38,47 +71,81 @@ remove_image() {
...
+    exec $devcontainer_bin $1 --workspace-folder="$REPO_ROOT" "${@:2}"
+}
 
-# Parse flags and commands
-while [[ $# -gt 0 ]]; do
+if [[ $# -eq 0 ]]; then
+    run_devcontainer exec bash
+else
     case "$1" in
-        -c|--cuda)
```
spomichter, line 46: Where was cuda flag moved to?

leshy: this is intended, we are not managing cuda via docker-compose. devcontainer took this over via
https://github.com/dimensionalOS/dimos/blob/d7350cb18080ce8465b7eb4606888f2d23f47f85/bin/dev#L120

PR #304, opened 2025-05-27
Add Cerebras Agent

joshuajerin:


dimos/agents/cerebras_agent.py
```diff
@@ -0,0 +1,421 @@
...
+             "Colors in Optical Flow are determined by the movement of light, and can be used to track the movement of objects."
+             ),
+            ("id4",
+             "Json is a data interchange format that is easy for humans to read and write, and easy for machines to parse and generate."
+             ),
+        ]
+        for doc_id, text in context_data:
+            self.agent_memory.add_vector(doc_id, text)
+
+    def _build_prompt(self, messages: list, base64_image: Optional[Union[str, List[str]]] = None,
```
spomichter: The reason build_prompt was overridden in ClaudeAgent was due to the different messages format and syntax. Looks like cerebras uses OpenAI syntax so likely can use the build_prompt method in LLMAgent

PR #305, opened 2025-05-28
test/mapper

leshy:
uses LCM data delivery to test the map component using a real sensor stream

https://github.com/dimensionalOS/dimos/blob/792582c647d61408bddf31f2e3bae2d4df242d38/dimos/robot/unitree_webrtc/type/test_map.py#L44

### DIFF
--- a/dimos/robot/unitree_webrtc/type/test_map.py
+++ b/dimos/robot/unitree_webrtc/type/test_map.py
@@ -1,9 +1,10 @@
 import pytest
+
+from dimos.robot.unitree_webrtc.testing.helpers import show3d, show3d_stream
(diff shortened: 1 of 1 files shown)


dimos/robot/unitree_webrtc/type/vector.py
```diff
@@ -116,7 +116,6 @@ def __add__(self: T, other: Union["Vector", Iterable[float]]) -> T:
 
     def __sub__(self: T, other: Union["Vector", Iterable[float]]) -> T:
         if isinstance(other, Vector):
-            print(self, other)
```
spomichter, line 119: Print lines not ideal here as they clog up the logging.

PR #309, opened 2025-05-29
base typing cleanup, sensor reply tests+docs

leshy:
- Costmap moved from unitree_webrtc to base types
- Position inherits from Vector now (adds .rot property for rotation vector)
- Odometry inherits from Position
- Adds sensor reply tests for mapping & odometry
- Removex .nix files entirely (we are using pure ubuntu dev containers for now)

### DIFF
--- a/docs/testing_stream_reply.md
+++ b/docs/testing_stream_reply.md
@@ -54,8 +54,8 @@ total_rad = (
     .stream()
     .pipe(
         ops.map(lambda odom: odom.rot.z),
-        ops.pairwise(),            # → [(v₀,v₁), (v₁,v₂), …]
-        ops.starmap(sub),          # → [v₁−v₀, v₂−v₁, …]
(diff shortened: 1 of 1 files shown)


dimos/types/vector.py
```diff
@@ -107,10 +107,6 @@ def serialize(self) -> Tuple:
         """Serialize the vector to a tuple."""
         return {"type": "vector", "c": self._data.tolist()}
 
-    def __len__(self) -> int:
```
spomichter, line 110: This file still imports Vector3 from ROS

flake.nix
spomichter, line 1: should flake.nix and flake.lock be added to gitignore i assume @leshy

leshy: Deleted from this PR, was included in the repo by mistake

dimos/types/vector.py
leshy: yes, ros cleanup in the follow up PR

spomichter: Kk

PR #333, opened 2025-05-31
auto prepend licence to files

leshy:
not very important, was easy to add, merge if interesting. expecting CI to add licences to all missing files in the follow up auto-commit

### DIFF
--- a/dimos/perception/common/cuboid_fit.py
+++ b/dimos/perception/common/cuboid_fit.py
@@ -1,3 +1,17 @@
+# Copyright 2025 Dimensional Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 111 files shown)


.pre-commit-config.yaml
```diff
@@ -1,15 +1,34 @@
...
 repos:
+
+  - repo: https://github.com/Lucas-C/pre-commit-hooks
+    rev: v1.5.5
+    hooks:
+      - id: forbid-crlf
+      - id: remove-crlf
+      - id: insert-license
+        files: \.py$
+        exclude: __init__\.py$
```
spomichter, line 12: Need to also exclude the main LICENCE file at project root since the wording is different there than in the license headers

PR #341, opened 2025-06-05
noros builds

leshy:
- added polyfill for ros libs if ros is unavailable, could be that we can use a similar polyfil to switch over to LCM? not sure
- non ros build workflow running parallel with ros builds
- running tests on ros and non ros build in parallel
- docs
- reintroduced flake.nix and flake.lock since we can run dimos now without docker as well, docs on this to follow later

we have 5 docker images now: 
```
┌──────┐
│ubuntu│
└┬────┬┘
┌▽──┐┌▽───────┐
│ros││python  │
└┬──┘└───────┬┘
┌▽─────────┐┌▽──┐
│ros-python││dev│
└┬─────────┘└───┘
┌▽──────┐
│ros-dev│
└───────┘
```

### DIFF
--- a/docker/dev/Dockerfile
+++ b/docker/dev/Dockerfile
@@ -31,8 +31,8 @@ COPY /docker/dev/tmux.conf /root/.tmux.conf
 
 # Install nodejs (for random devtooling like copilot etc)
 RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.1/install.sh | bash
-RUN /root/.nvm/nvm.sh
-RUN nvm install node # v24 atm
+ENV NVM_DIR=/root/.nvm
+RUN bash -c "source $NVM_DIR/nvm.sh && nvm install 24"
 
 # This doesn't work atm
 RUN echo "  v_${GIT_BRANCH}:${GIT_COMMIT} | $(date)" >> /etc/motd


flake.lock
```diff
@@ -0,0 +1,61 @@
+{
```
spomichter, line 1: @leshy Wasn't flake.lock added to gitignore

docker/python/Dockerfile
```diff
@@ -0,0 +1,43 @@
+ARG FROM_IMAGE
+FROM ${FROM_IMAGE}
```
spomichter, line 2: Check these default value errors @leshy

leshy: from_image is dynamic but can set the default value to something reasonable yeah

flake.lock
leshy: flake.* is either totally out or totally in, reintroducing here since we can run on any linux without docker using this. I'll tweak & work on it & document in the following weeks.

dimos/robot/local_planner/local_planner.py
```diff
@@ -30,7 +30,7 @@
 
 from dimos.types.vector import VectorLike, Vector, to_tuple
 from dimos.types.path import Path
-from nav_msgs.msg import OccupancyGrid
+from dimos.types.costmap import Costmap
 
 logger = setup_logger("dimos.robot.unitree.local_planner", level=logging.DEBUG)
```
spomichter, line 35: Should not hard code logging level so you can get logs by running ```DIMOS_LOG_LEBEL=DEBUG python run.py```

PR #345, opened 2025-06-11
ONNX Support for YOLO, SAM2 + Unit tests for CLIP, YOLO, SAM2

spomichter:
Model unit tests for @mdaiter 

Not merged as I need to test on my Mac. Ran so far on my Ubuntu 22.04 machine with CUDA 11.7/Torch 2.0.1. Also need to push clean python install in setup.py so you can run ```pip install -e ".[mac]"``` or ```pip install -e ".[arm,gpu]"```

To run:
```
pytest -s dimos/agents/memory dimos/perception
```

### DIFF
--- a/dimos/perception/detection2d/utils.py
+++ b/dimos/perception/detection2d/utils.py
@@ -18,17 +18,6 @@ import cv2
 from dimos.utils.ros_utils import distance_angle_to_goal_xy
 
 
-def is_cuda_available():
-    try:
-        import pycuda.driver as cuda
-        import pycuda.autoinit  # implicitly initializes the CUDA driver
-
-        cuda.init()
-        return cuda.Device.count() > 0
-    except Exception:
-        return False
-
-
 def filter_detections(
     bboxes,
     track_ids,
--- a/dimos/perception/detection2d/yolo_2d_det.py
+++ b/dimos/perception/detection2d/yolo_2d_det.py
@@ -15,7 +15,6 @@
 import cv2
 from ultralytics import YOLO
 from dimos.perception.detection2d.utils import (
-    is_cuda_available,
     extract_detection_results,
     plot_results,
     filter_detections,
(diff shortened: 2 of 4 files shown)


dimos/agents/memory/test_image_embedding.py
```diff
@@ -0,0 +1,224 @@
...
+
+class TestImageEmbedding:
+    """Test class for CLIP image embedding functionality."""
+
+    @pytest.fixture(scope="class")
+    def video_path(self):
+        """Return the path to the test video."""
+        # Use a video file from assets directory
+        base_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), "../../../assets"))
+        video_file = "trimmed_video_office.mov"  # Use the same test video as YOLO test
```
leshy: should use LFS store for these files.
merge this, https://github.com/dimensionalOS/dimos/pull/342 then read:
https://github.com/dimensionalOS/dimos/blob/dev/docs/testing_stream_reply.md

would be something like
```python
dir = testData("my_video_dir")
```
the rest the same, or you can upload individual videos separately
```python
video_path = testData("some_specific_video.mp4")
```

can do in the follow up PR if annoying but plz let's do it :D

dimos/agents/memory/test_image_embedding.py
```diff
@@ -0,0 +1,224 @@
...
+import time
+import numpy as np
+import pytest
+import reactivex as rx
+from reactivex import operators as ops
+from dimos.stream.video_provider import VideoProvider
+from dimos.agents.memory.image_embedding import ImageEmbeddingProvider
+
+
+class TestImageEmbedding:
```
leshy, line 29: all of this works as a non-class, not sure why we are embedding tests in class methods? not very important, might care about it for dimos-wide uniformity

spomichter: Will fix

dimos/agents/memory/test_image_embedding.py
spomichter: Nice sounds good

spomichter: So LFS seems to only be setup for tests/data dir right @leshy ? Is it worth adding to /assets or automatically any directories that include "*data" so that @mdaiter can store any onnx yolo binaries in ```dimos/models/yolo/data/yolo.onnx``` or whatever

spomichter: @leshy Will do this after merge since your testing stream stuff in dev

PR #368, opened 2025-06-27
LFS moved to utils from testing

leshy:
lfs data dir is now in /data (not in tests/data)

new generic function for data fetching is at 

`from dimos.utils.data import get_data`

stream sim stuff stayed in dimos.utils.testing


dimos/perception/segmentation/sam_2d_seg.py
```diff
@@ -12,25 +12,27 @@
...
-import cv2
+import os
 import time
+from collections import deque
+from concurrent.futures import ThreadPoolExecutor
+
+import cv2
+import onnxruntime
 from ultralytics import FastSAM
+
```
spomichter, line 23: Why all of these import changes for no reason

dimos/perception/test_spatial_memory.py
```diff
@@ -13,19 +13,19 @@
...
-import time
+import shutil
 import tempfile
-import pytest
-import numpy as np
+import time
+
 import cv2
-import shutil
+import numpy as np
```
spomichter, line 21: Import changes pretty confusing to read @leshy . Idk why diff does this

PR #377, opened 2025-07-01
Dask lcm multiprocess

leshy:
pretty big but separate

artisinal hand crafted multiprocess + pluggable transport between nodes

- builds a generic structure for services and generic Pubsub on top, with LCM, redis and in-memory implementations (for grid testing of my spec) - allows us to be independant from transport protocols (can implement other types of services in the future like KV stores, RPC etc)

- builds a bunch of ROS/LCM native types - this is a bit time consuming and awkward - likely will pivot into fixing autogen vs rewriting so much

- builds a multiprocessing architecture with (per stream) configurable transport strategy - so we can use zenoh, dask, lcm, even within the same deployment

core works well, but API will get syntax sugar on top - I have some basics in mind, pls also give suggestions for follow up PRs 

Dask manages multiprocesing but we are not tied to dask for this, can switch this out without rewriting anything built on top.

Supports convinient way to delegate actual classes to separate processes and talk to them, but nothing stops you from creating ros-style separate python scripts either.

API example (paste from `dimos/core/test_core.py`)

```python

def test_deployment(dimos):
    robot = dimos.deploy(RobotClient)
    target_stream = RemoteOut[Vector](Vector, "target")

    print("\n")
    print("lidar stream", robot.lidar)
    print("target stream", target_stream)
    print("odom stream", robot.odometry)

    nav = dimos.deploy(Navigation)

    # transport override - this stream encodes proper LCM messages
    robot.lidar.transport = LCMTransport("/lidar", LidarMessage)

    # transport override - odometry is transfered using a pickle over LCM
    robot.odometry.transport = pLCMTransport("/odom")

    # move messages we are sending using the (default) dask transport so we are not specifying mov transport
    # nav.mov.transport = pLCMTransport("/mov")

    nav.lidar.connect(robot.lidar)
    nav.odometry.connect(robot.odometry)
    robot.mov.connect(nav.mov)

    print("\n" + robot.io().result() + "\n")
    print("\n" + nav.io().result() + "\n")
    robot.start().result()
    nav.start().result()

    time.sleep(1)
    robot.stop().result()

    print("robot.mov_msg_count", robot.mov_msg_count)
    print("nav.odom_msg_count", nav.odom_msg_count)
    print("nav.lidar_msg_count", nav.lidar_msg_count)

    assert robot.mov_msg_count >= 9
    assert nav.odom_msg_count >= 9
    assert nav.lidar_msg_count >= 9

```
output:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -98,6 +98,9 @@ open3d
 # Inference (CPU)
 onnxruntime
 
+# Terminal colors
+rich==14.0.0
+
 # multiprocess 
 dask[complete]==2025.5.1
 git+https://github.com/dimensionalOS/python_lcm_msgs@main#egg=lcm_msgs
--- a/dimos/core/test_core.py
+++ b/dimos/core/test_core.py
@@ -119,6 +119,7 @@ class Navigation(Module):
         self.lidar.subscribe(_lidar)
 
 
+@pytest.mark.tool
 def test_deployment(dimos):
     robot = dimos.deploy(RobotClient)
     target_stream = RemoteOut[Vector](Vector, "target")
@@ -131,10 +132,10 @@ def test_deployment(dimos):
     nav = dimos.deploy(Navigation)
 
     # this one encodes proper LCM messages
-    # robot.lidar.transport = LCMTransport("/lidar", LidarMessage)
+    robot.lidar.transport = LCMTransport("/lidar", LidarMessage)
     # odometry & mov using just a pickle over LCM
-    # robot.odometry.transport = pLCMTransport("/odom")
-    # nav.mov.transport = pLCMTransport("/mov")
+    robot.odometry.transport = pLCMTransport("/odom")
+    nav.mov.transport = pLCMTransport("/mov")
 
     nav.lidar.connect(robot.lidar)
     nav.odometry.connect(robot.odometry)


dimos/msgs/sensor_msgs/test_image.py
```diff
@@ -0,0 +1,63 @@
...
+import numpy as np
+import pytest
+
+from dimos.msgs.sensor_msgs.Image import Image, ImageFormat
+from dimos.utils.data import get_data
+
+
+@pytest.fixture
+def img():
+    image_file_path = get_data("cafe.jpg")
```
spomichter, line 24: Probably cafe not be in the system data root maybe as data/cafe.jpg maybe data/test where we put all of the test assets like videos, image, pcd, lidar, etc.

dimos/protocol/pubsub/lcmpubsub.py
```diff
@@ -0,0 +1,152 @@
...
+        if self.lcm_type is None:
+            return self.topic
+        return f"{self.topic}#{self.lcm_type.name}"
+
+
+class LCMbase(PubSub[Topic, Any], Service[LCMConfig]):
+    default_config = LCMConfig
+    lc: lcm.LCM
+    _stop_event: threading.Event
+    _thread: Optional[threading.Thread]
```
spomichter, line 67: Isn't threading handled via Dask why do we have threading events set via python threading? @leshy

dimos/robot/unitree_webrtc/type/lidar.py
```diff
@@ -48,29 +49,51 @@ class RawLidarMsg(TypedDict):
...
 
-@dataclass
-class LidarMessage(Timestamped):
-    ts: datetime
-    origin: Vector
-    resolution: float
-    pointcloud: o3d.geometry.PointCloud
-    raw_msg: RawLidarMsg = field(repr=False, default=None)
-    _costmap: Costmap = field(init=False, repr=False, default=None)
+class LidarMessage(PointCloud2):
```
spomichter, line 52: This type does not look unitree specific?

pyproject.toml
```diff
@@ -149,6 +149,15 @@ exclude = [
     "src"
 ]
 
+[tool.mypy]
+# mypy doesn't understand plum @dispatch decorator
+# so we gave up on this check globally
+disable_error_code = ["no-redef", "import-untyped", "import-not-found"]
```
spomichter, line 155: No better way to do this?

requirements.txt
```diff
@@ -96,4 +96,11 @@ git+https://github.com/facebookresearch/detectron2.git@v0.6
...
-onnxruntime
\ No newline at end of file
+onnxruntime
+
+# Terminal colors
+rich==14.0.0
+
+# multiprocess 
+dask[complete]==2025.5.1
+git+https://github.com/dimensionalOS/python_lcm_msgs@main#egg=lcm_msgs
```
spomichter, line 106: I will add to pyproject build

dimos/msgs/sensor_msgs/test_image.py
leshy: issue is that we don't support dir trees for data/ (there was a good reason why but need to think, obviously would be nice to have)

dimos/robot/unitree_webrtc/type/lidar.py
leshy: it is, for now, I've moved out some types that started unitree specific to dimos types/ once they feel ready and need to be shared so that's my policy, once we are implementing a new lidar, we can see if we can make this type universal

pyproject.toml
leshy: I've spent a bit of time on it and decided to do this. mypy is a typecheck that only I use for now, so I've been keeping it's config loose, it's also possible that this is the best way unfortunately, not sure.

requirements.txt
leshy: thanks

dimos/protocol/pubsub/lcmpubsub.py
leshy: dask runs you in a separate process, once there you can do whatever you want. in this case I need some separate loop to run message broadcast loop, two things though:

- I don't like "threading" too much, and prefer asyncio, in this case cursor vibecodeded a thread and I said "ok fine"

- In the case above given we don't do anything apart from the message broadcast loop - I might not need a "thread" - will test

PR #400, opened 2025-07-10
UnitreeSpeak skill over webrtc, Voice Interface added on localhost, Voice interface on mobile device on network

spomichter:
Changes:
- New UnitreeSpeak skill to send audio/speech commands to the robot over webRTC 
- Fix hardcoded 0.0.0.0 and CORS issues over HTTP so DimOS terminal can be accessed and send audio/text commands to robot by a phone/tablet/device on the same network. 
- Fix ObjectDetectionStream bug due to get_pose type updating 
- Agent and ClaudeAgent additional error catching
- Skills error catching 
- Cerebras only runfile with fewer skills implemented and higher token cutoff to maximize our Cerebras token limits
- Added voice input buttons to the Dimensional Web Interface and Terminal interface via svelte
- Added audio streaming to server.py and additional voice stream added to stream.ts
- Minimal Prompt.txt changes
- Prompt updates to Navigation skills, new flag to go straight to SpatialMemory query circumventing VLM
- Fix Observe skill agent recursion bug 
- Commented out unitree_skills that don't work on 1.17


dimos/skills/unitree/unitree_speak.py
```diff
@@ -0,0 +1,280 @@
...
+    "UPLOAD_AUDIO_FILE": 2001,
+    "ENTER_MEGAPHONE": 4001,
+    "EXIT_MEGAPHONE": 4002,
+    "UPLOAD_MEGAPHONE": 4003,
+}
+
+PLAY_MODES = {"NO_CYCLE": "no_cycle", "SINGLE_CYCLE": "single_cycle", "LIST_LOOP": "list_loop"}
+
+
+class UnitreeSpeak(AbstractRobotSkill):
```
leshy, line 47: This is a bit weird, openai bindings etc,

I'd expect to implement audio_input stream on unitree connection.py
then I'd use some generic speaker that speaks into an arbitrary audio_input stream?

don't have big stake in this if it's temporary and we are on the same page on the above? if this was easier to implement then my suggestion then we need to map the arch issues

spomichter: Yes openAI client is used for whisper STT/TTS. should have a local option.

PR #411, opened 2025-07-13
WIP: foxglove bridge stub

leshy:


pyproject.toml
```diff
@@ -91,7 +91,11 @@ dependencies = [
 
     # Multiprocess 
     "dask[complete]==2025.5.1",
-    "lcm_msgs @ git+https://github.com/dimensionalOS/python_lcm_msgs.git@main#egg=lcm_msgs"
+    "lcm_msgs @ git+https://github.com/dimensionalOS/python_lcm_msgs.git@main#egg=lcm_msgs",
```
leshy: hey we need one or the other no? lcm_msgs is just a copy of a dir from dimos_utils

spomichter: Oh yes youre right this is now redundant. Will remove

leshy: from lcm_msgs.geometry_msgs import Vector3 feels better then
from dimos_utils.lcm_msgs... etc

I think dimos_utils repo is only lcm_msgs at this stage, proly should rename?

PR #471, opened 2025-07-17
Stream fixes, Twist, Pose, Quaternion updates

leshy:


dimos/robot/unitree_webrtc/connection.py
```diff
@@ -81,55 +81,24 @@ def start_background_loop():
         self.thread.start()
         self.connection_ready.wait()
 
-    def move(self, velocity: Vector, duration: float = 0.0) -> bool:
```
spomichter: Move has not been tested

PR #481, opened 2025-07-21
SpatialMemory converted to Dask module, input LCM odom and video streams

spomichter:


dimos/robot/unitree_webrtc/multiprocess/test_unitree_go2_cpu_module.py
```diff
@@ -0,0 +1,216 @@
...
+
+            # Wait for initialization
+            await asyncio.sleep(3)
+
+            # Test get methods
+            odom = connection.get_odom()
+            assert odom is not None, "Should get odometry"
+            logger.info(f"Got odometry: position={odom.position}")
+
+            pos = connection.get_pos()
```
spomichter, line 156: @leshy Lmk best and most robust way to test that module is working properly here. For now just do some basic checks that we're getting odom, lidar, etc packets. And then MovementControl tests movecmd stream.

.github/workflows/docker.yml
```diff
@@ -129,10 +129,7 @@ jobs:
     uses: ./.github/workflows/_docker-build-template.yml
     with:
       should-run: ${{
-        needs.check-changes.result == 'success' &&
-        ((needs.ros-python.result == 'success') ||
-        (needs.ros-python.result == 'skipped' &&
-        needs.check-changes.outputs.dev == 'true'))
+        needs.check-changes.result == 'success' && ((needs.ros-python.result == 'success') || (needs.ros-python.result == 'skipped')) && (needs.check-changes.outputs.dev == 'true')
```
leshy, line 132: can't remember what we decided here, this is not changed by accident?

.github/workflows/tests.yml
```diff
@@ -47,10 +47,18 @@ jobs:
           sudo chown -R $USER:$USER ${{ github.workspace }} || true
           
       - uses: actions/checkout@v4
+        with:
+          lfs: true
+          
+      - name: Configure Git LFS
```
leshy, line 53: same as below

.github/workflows/docker.yml
```diff
@@ -181,9 +178,58 @@ jobs:
...
+    steps:
+      - name: Fix permissions
+        run: |
+          sudo chown -R $USER:$USER ${{ github.workspace }} || true
+          
+      - uses: actions/checkout@v4
+        with:
+          lfs: true
+          
+      - name: Configure Git LFS
```
leshy, line 217: not neccessary, if tests use utils.data, lfs autodownload will happen, I think this pre-downloads whole lfs store?

dimos/core/core.py
```diff
@@ -202,6 +202,19 @@ class RemoteOut(RemoteStream[T]):
     def connect(self, other: RemoteIn[T]):
         return other.connect(self)
 
+    def observable(self):
```
leshy, line 205: tofix - observable needs to unsubscribe from the stream once observing stops, probably disposable needs to be returned

dimos/perception/person_tracker.py
```diff
@@ -72,83 +249,10 @@ def create_stream(self, video_stream: Observable) -> Observable:
...
-            return result
-
-        return video_stream.pipe(ops.map(process_frame))
+        return video_stream.pipe(ops.map(self._process_tracking))
 
+    @rpc
     def cleanup(self):
         """Clean up resources."""
-        pass  # No specific cleanup needed for now
+        # CUDA cleanup is now handled by WorkerPlugin in dimos.core
```
leshy, line 257: is it? can't see where, not sure if you need to adress this

dimos/protocol/service/lcmservice.py
```diff
@@ -65,71 +69,122 @@ def check_multicast() -> list[str]:
...
+            int(result.stdout.split("=")[1].strip()) if result.returncode == 0 else None
+        )
+        if not current_default or current_default < 2097152:
             commands_needed.append(f"{sudo}sysctl -w net.core.rmem_default=2097152")
-    except Exception:
+    except:
         commands_needed.append(f"{sudo}sysctl -w net.core.rmem_default=2097152")
 
-    return commands_needed
+    return commands_needed, current_max
```
leshy, line 102: as far as I can tell we are doing all this complex processing just to return current_max for the error message? might not be needed?

dimos/protocol/service/lcmservice.py
```diff
@@ -65,71 +69,122 @@ def check_multicast() -> list[str]:
...
+        for cmd in multicast_commands:
+            logger.error(f"  {cmd}")
+        logger.error("\nThen restart your application.")
         sys.exit(1)
 
+    # Buffer configuration is just for performance
+    elif buffer_commands:
+        if current_buffer_size:
+            logger.warning(
+                f"UDP buffer size limited to {current_buffer_size} bytes ({current_buffer_size // 1024}KB). Large LCM packets may fail."
```
leshy, line 132: is this the only reason for parsing buffer sizes?

dimos/robot/unitree_webrtc/multiprocess/test_unitree_go2_cpu_module.py
```diff
@@ -0,0 +1,216 @@
...
+        cmd = Vector3(x, y, yaw)
+        self.movecmd.publish(cmd)
+        self.commands_sent.append(cmd)
+        logger.info(f"Sent move command: x={x}, y={y}, yaw={yaw}")
+
+    @rpc
+    def send_explore_sequence(self):
+        """Send a sequence of exploration commands."""
+
+        def send_commands():
```
leshy, line 62: why send movement commands if it's a sensor replay?

dimos/robot/unitree_webrtc/multiprocess/test_unitree_go2_cpu_module.py
leshy: awesome that we have this now.

can write a spec for which functionality is needed for proper module testing - fast limited sensor reply basically, mem leak detection, maybe a bit of pytest stuff, I expect we will replace this file soon so no need for fixes?

would be really good to measure & store exec speed per module (how do we do this across machines with diff cpus?) - influxdb? so we can detect if some feature slows down or speeds up the system, attach a report to the pr thread

.github/workflows/docker.yml
spomichter: this is what worked

.github/workflows/docker.yml
spomichter: No this is needed in self hosted runner unfort

dimos/core/core.py
spomichter: yes

dimos/perception/person_tracker.py
spomichter: oh yeah i added this to workerplugin and then deleted since it didnt work

dimos/protocol/service/lcmservice.py
spomichter: No we need to parse buffer size that was your stuff from before. TO check if we need to run the autoconf commands

PR #487, opened 2025-07-22
Full Unitree (Local planner, Explore, SpatialMemory) FakeRTC/WebRTC LCM modules working in self-hosted devcontainer 

spomichter:


dimos/protocol/service/lcmservice.py
```diff
@@ -38,6 +38,17 @@ def check_root() -> bool:
         return False
 
 
+@cache
+def is_dev_container() -> bool:
+    """Return True if we're running in a dev container or similar restricted environment."""
+    # Check if we can access the network sysctls (common limitation in containers)
+    try:
+        subprocess.run(["sysctl", "net.core.rmem_max"], capture_output=True, text=True, check=True)
```
leshy: instead of is_dev_container() should change the code to fail only on multicast / route add commands and not all commands.

allow sysctl to fail - just emit a warning about "not being able to auto-configure udp sizes (likely docker image)"

(logger.warning so it's suppressable)

spomichter: Changing

PR #492, opened 2025-07-25
Added modules documentation

spomichter:


docs/modules.md
```diff
@@ -0,0 +1,112 @@
...
+# Auto-configure LCM system buffers (required in containers)
+from dimos.protocol import pubsub
+pubsub.lcm.autoconf()
+```
+
+This architecture enables building complex robotic systems as composable, distributed modules that communicate efficiently via streams and RPC, scaling from single machines to clusters.
+
+## Multiprocess Application Design  
+
+1. @lesh can you add considerations here for best practices, issues, etc. One for example I will add is handling CUDA properly in multiprocess which is WIP.
```
spomichter: @leshy

docs/modules.md
```diff
@@ -0,0 +1,112 @@
...
+from dimos.core import Module, In, Out, rpc
+from dimos.msgs.geometry_msgs import Vector3
+
+class MyModule(Module):
+    # Declare inputs/outputs as class attributes initialized to None
+    data_in: In[Vector3] = None  
+    data_out: Out[Vector3] = None
+    
+    # Call parent Module init
+    super().__init__()
```
leshy, line 27: oops, needs to be in __init__ and doesn't need to exist if MyModule.__init__ is not defined

docs/modules.md
```diff
@@ -0,0 +1,112 @@
...
+Modules are deployed across Dask workers using the `dimos.deploy()` method:
+
+```python
+from dimos import core
+
+# Start Dask cluster with N workers
+dimos = core.start(4)
+
+# Deploy modules with initialization parameters
+my_module = dimos.deploy(MyModule, param1="value1", param2=123)
```
leshy, line 51: doesn't take params in init so confusing what this does

docs/modules.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+### 3. Stream Connections
+Modules communicate via reactive streams using LCM transport:
+
+```python
+# Configure LCM transport for outputs
+module1.data_out.transport = core.LCMTransport("/topic_name", MessageType)
```
leshy, line 60: the context of this is unclear, do I run this within a module?

docs/modules.md
```diff
@@ -0,0 +1,112 @@
...
+### 4. Module Lifecycle
+```python
+# Start modules to begin processing
+module.start()  # Calls the @rpc start() method if defined
+
+# Inspect module I/O configuration  
+print(module.io().result())  # Shows inputs, outputs, and RPC methods
+
+# Clean shutdown
+dimos.close()
```
leshy, line 81: no need for .close?

docs/modules.md
spomichter: Oh i thought it does? We do this in ConnectionModule. We pass in the ip parameter like this no?

docs/modules.md
spomichter: ah right

docs/modules.md
spomichter: This is walking folks through piece by piece each separate component of how modules work, as is done in any documentation. So yes this code block doesn't standalone but that is intentional

spomichter: How this all fits together is answered by your simple full examples which we should add to the end of this doc @leshy

dimos/models/qwen/video_query.py
```diff
@@ -85,15 +85,15 @@ def query_single_frame_observable(
 
 
 def query_single_frame(
-    image: "PIL.Image",
+    image: np.ndarray,
```
leshy, line 88: Ideally we are using our types, if diff storage format is needed we can support this on Image type. I want to be able to feed any module output into this

docs/modules.md
```diff
@@ -0,0 +1,183 @@
...
+
+```python
+from dimos.core import Module, In, Out, rpc
+from dimos.msgs.geometry_msgs import Vector3
+
+class MyModule(Module):
+    # Declare inputs/outputs as class attributes initialized to None
+    data_in: In[Vector3] = None  
+    data_out: Out[Vector3] = None
+<<<<<<< HEAD
```
leshy, line 18: some sort of merge issue here

PR #539, opened 2025-08-07
MuJoCo Simulation

paul-nechifor:
Add MuJoCo simulation for the Unitree Go1 robot with odom, video, and lidar from the simulation.

### DIFF
--- a/dimos/robot/unitree_webrtc/unitree_go2.py
+++ b/dimos/robot/unitree_webrtc/unitree_go2.py
@@ -164,6 +164,7 @@ class ConnectionModule(Module):
             raise ImportError("'mujoco' is not installed. Use `pip install -e .[sim]`")
 
         from dimos.robot.unitree_webrtc.mujoco_connection import MujocoConnection
+
         connection = MujocoConnection(self.ip)
         connection.start()
         return connection


assets/robots/go1/assets/hfield.png
leshy, line 1: in the follow up PRs we need to decide on asset policy for sim, ok for now, probably will be managed by get_data function and our existing lfs on demand pull system

dimos/navigation/bt_navigator/goal_validator.py
```diff
@@ -40,6 +41,7 @@ def find_safe_goal(
         min_clearance: Minimum clearance from obstacles in meters (default: 0.3m)
         max_search_distance: Maximum distance to search from original goal in meters (default: 5.0m)
         connectivity_check_radius: Radius in cells to check for connectivity (default: 3)
+        treat_unknown_as_safe: Whether to treat UNKNOWN cost values as safe (default: False)
```
leshy, line 44: I think good intervention to get sim up & running, we need a policy for exploration and mapping wrt sensors that don't generate floor hits

paul-nechifor: I created https://github.com/dimensionalOS/dimos/issues/545

assets/robots/go1/assets/hfield.png
paul-nechifor: Yeah, this won't last much. I'm guessing the next steps will be a to use a realistic office environment, so these particular assets will be deleted. I'll look at get_data later.

PR #551, opened 2025-08-13
Correct way to build package directly in non-editable mode, no manife…

spomichter:
…st needed

### DIFF
--- a/docker/dev/Dockerfile
+++ b/docker/dev/Dockerfile
@@ -29,10 +29,7 @@ ENV UV_SYSTEM_PYTHON=1
 RUN curl -LsSf https://astral.sh/uv/install.sh | sh
 ENV PATH="/root/.local/bin:$PATH"
 
-# Copy entire project first to ensure proper package installation
-COPY . /app/
-
-# Install dependencies with UV (10-100x faster than pip)
+# Install dependencies with UV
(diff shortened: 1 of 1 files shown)


docker/dev/Dockerfile
```diff
@@ -24,20 +24,17 @@ RUN git config --global --add safe.directory '*'
...
-RUN mkdir -p /app/dimos
-RUN touch /app/dimos/__init__.py
-
 # Install UV for fast Python package management
 ENV UV_SYSTEM_PYTHON=1
 RUN curl -LsSf https://astral.sh/uv/install.sh | sh
 ENV PATH="/root/.local/bin:$PATH"
 
+# Copy entire project first to ensure proper package installation
+COPY . /app/
```
leshy: I think no need to copy /app twice, dev image depends on python image, and /app is already there in py image

spomichter: I dont think it brings over the dimos package files from the base images

spomichter: When I exec in couldnt find it anywhere

spomichter: Double checking

spomichter: Yes youre right fixed

PR #554, opened 2025-08-18
Office environment mujoco

paul-nechifor:
* Add a office scene for the Mujoco simulation with collision detection.
* Make the Lidar look more realistic.
* Made the robot move faster.
* Add a command to trigger exploration inside the simulator.
* Use a seed for all LLM tests so they are reproducible (for example one test failed because it expected "3+3", but the result from the LLM was "Three plus three" sometimes).

### DIFF
--- a/dimos/agents/test_agent_image_message.py
+++ b/dimos/agents/test_agent_image_message.py
@@ -219,7 +219,7 @@ def test_agent_mixed_content():
         model="openai::gpt-4o-mini",
         system_prompt="You are a helpful assistant that can see images when provided.",
         temperature=0.0,
-        seed=42,
+        seed=100,
     )
 
     # Text-only query
--- a/dimos/agents/test_conversation_history.py
+++ b/dimos/agents/test_conversation_history.py
@@ -235,7 +235,7 @@ def test_conversation_history_with_tools():
         system_prompt="You are a helpful assistant with access to a calculator.",
         skills=TestSkillLibrary(),
         temperature=0.0,
-        seed=42,
+        seed=100,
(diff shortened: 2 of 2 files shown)


bin/explore-cmd
leshy, line 1: ok for now for convinience, we need to make a small controller for robots, even for movement mesages, you can check dimos/utils/cli for cli tooling and follow the example sometime

dimos/agents/test_agent_image_message.py
```diff
@@ -216,6 +219,7 @@ def test_agent_mixed_content():
         model="openai::gpt-4o-mini",
         system_prompt="You are a helpful assistant that can see images when provided.",
         temperature=0.0,
+        seed=100,
```
leshy, line 222: oh wait holy shit seed will always make the llm reply with the same thing? awesome

dimos/navigation/frontier_exploration/wavefront_frontier_goal_selector.py
```diff
@@ -93,6 +93,8 @@ class WavefrontFrontierExplorer(Module):
     costmap: In[OccupancyGrid] = None
     odometry: In[PoseStamped] = None
     goal_reached: In[Bool] = None
+    explore_cmd: In[Bool] = None
```
leshy, line 96: this stuff can be an rpc also, probably more appropriate (you decorate a method with @rpc and it's remote callable) but just letting you know, not asking for the change I don't own the structure here. unclear to me why this takes odometry and costmap as streams.

PR #555, opened 2025-08-20
WIP: Matt shared mem

leshy:


dimos/protocol/pubsub/shared_memory_transport.py
```diff
@@ -0,0 +1,140 @@
...
+import threading
+from dataclasses import dataclass
+from typing import Any, Callable, Optional
+
+from dimos.protocol.pubsub.spec import PubSub, PubSubEncoderMixin, PickleEncoderMixin
+from dimos.protocol.service.shmservice import SharedMemoryConfig, SharedMemoryService
+
+
+@dataclass
+class SHMTopic:
```
leshy: why not use just str for topic? seems like this is just a string

dimos/protocol/pubsub/shared_memory_transport.py
```diff
@@ -0,0 +1,140 @@
...
+            try:
+                unsub()
+            except Exception:
+                pass
+
+
+# ---- Encoders + concrete classes (parallel to LCM / PickleLCM) ------------
+
+
+class SharedMemoryBytesEncoderMixin(PubSubEncoderMixin[str | SHMTopic, bytes]):
```
leshy: doesn't to do anything looks like? if you want to throw you can do it in publish() of your pubsub but also no need, it's typed, type system should catch

dimos/protocol/service/shmservice.py
```diff
@@ -0,0 +1,244 @@
+# Copyright 2025 Dimensional Inc.
```
leshy: this seems like entirely a pubsub implementation, so would put it in a pubsub/shared_memory stuff,
redis, lcm service can do other (non pubsub) things that's why implementation is in service

dimos/protocol/pubsub/shared_memory_transport.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def __str__(self) -> str:
+        return self.topic
+
+
+def _as_str_topic(t: str | SHMTopic) -> str:
+    return t if isinstance(t, str) else str(t)
+
+
+class SharedMemoryPubSubBase(PubSub[str | SHMTopic, Any], SharedMemoryService):
```
leshy: can just name it SharedMemory since this class is directly usable without subclassing

dimos/protocol/pubsub/shared_memory_transport.py
mdaiter: done

PR #556, opened 2025-08-21
Add g1

paul-nechifor:
Run Unitree Go2 with fake data:

    python dimos/robot/unitree_webrtc/unitree_robot.py 
    # or
    CONNECTION_TYPE=fake python dimos/robot/unitree_webrtc/unitree_go2.py 

Run Unitree Go2 in webrtc with a real robot:

    ROBOT_IP=... python dimos/robot/unitree_webrtc/unitree_go2.py 

Run Mujoco with Unitree Go1:

    CONNECTION_TYPE=mujoco ROBOT_MODEL=unitree_go1 python dimos/robot/unitree_webrtc/unitree_go2.py 

Run Mujoco with Unitree G1:

    CONNECTION_TYPE=mujoco ROBOT_MODEL=unitree_g1 python dimos/robot/unitree_webrtc/unitree_go2.py 

You can select a different scene with SCENE=empty (only other one so far).

### DIFF
--- a/dimos/simulation/mujoco/model.py
+++ b/dimos/simulation/mujoco/model.py
@@ -49,8 +49,13 @@ def load_model(input_device: InputController, robot: str, scene: str):
 
     mujoco.mj_resetDataKeyframe(model, data, 0)
 
+    match robot:
+        case "unitree_g1":
+            sim_dt = 0.002
+        case _:
+            sim_dt = 0.005
+
     ctrl_dt = 0.02
-    sim_dt = 0.002
     n_substeps = int(round(ctrl_dt / sim_dt))
     model.opt.timestep = sim_dt
 


dimos/robot/unitree_webrtc/robot_config.py
```diff
@@ -0,0 +1,58 @@
...
+ConnectionType = Literal["webrtc", "fake", "mujoco"]
+
+
+RobotModel = Literal["unitree_go1", "unitree_go2", "unitree_g1"]
+
+
+MujocoScene = Literal["empty", "office1"]
+
+
+class RobotConfig(BaseModel):
```
spomichter, line 32: Need to decide whether we need a robot config and how to structure, maybe better as some json / yaml file it would be shorter. currently these robot confirm params we just init inside Robot instances like robot_ip and output_dir

paul-nechifor: I see this as a temporary fix to select between the two supported robots. I can refactor better into separate UnitreeGo1, UnitreeGo2, and UnitreeG1 classes once agent2 is merged. Is that all right?

leshy: yeah we don't have a unified config interface, topic prefixes (for multiples of same robot), individual module defaults overrides etc, even transport protocols should all be propagaded from the root init, we'll sort it out but can have whatever is the easiest temporary approach for now

dimos/robot/unitree_webrtc/unitree_g1.py
```diff
@@ -55,14 +56,12 @@ class G1ConnectionModule(Module):
...
-    ip: str
-    connection_type: str = "webrtc"
+    robot_config: RobotConfig
 
     _odom: PoseStamped = None
 
-    def __init__(self, ip: str = None, connection_type: str = "webrtc", *args, **kwargs):
-        self.ip = ip
-        self.connection_type = connection_type
+    def __init__(self, robot_config, *args, **kwargs):
```
spomichter, line 63: This is a breaking change for g1_run.py so I cannot merge this yet. Forcing this RobotConfig thing is tough because it adds an extra layer of complexity to init

spomichter: We need to make sure no instnaces of unitreeg1 or unitreego2 pass IP into the init, or we simply revert the robotconfig stuff

spomichter: Backward compatibility here may be handled automatically by the kwargs in Unitreeg1 and unitreego2 __init__ but even so im pretty sure you're missing some references to self.ip that would error our now that that and connection_type is gone

paul-nechifor: I've changed g1_run to also pass RobotConfig.

PR #570, opened 2025-08-25
Twist message for all move command, added keyboard teleop for easy robot control in sim

alexlin2:


bin/explore-cmd
leshy, line 1: was this removed on purpose?

dimos/web/command-center-extension/.gitignore
leshy, line 1: what is this stuff?

bin/explore-cmd
paul-nechifor: Yes, this is my commit from the Foxglove stuff. Exploration can be triggered from Foxglove so I removed it.

dimos/web/command-center-extension/.gitignore
paul-nechifor: "command-center-extension" is the Foxglove extension I wrote. This gitignore is needed to un-ignore package.json.

paul-nechifor: @leshy About the merging of the two branches. Basically since this adds Twist and the foxglove extension adds the keyboard teleop which needs twist. To do it separately, Alex would have to have a Twist PR, I would have a Foxglove PR without Twist and then a separate PR to use Twist in Foxglove once the previous two are merged. It's less work to have everything on a single PR.

Sorry that this makes reviewing a bit hard. I think it's easier to review by each commit.

PR #578, opened 2025-08-26
[DRAFT]: Pixel kalman filter for temporal + spatial depth smoothing

mdaiter:


dimos/models/depth/metric3d.py
```diff
@@ -41,6 +48,8 @@ def __init__(self, camera_intrinsics=None, gt_depth_scale=256.0):
         self.pad_info = None
         self.rgb_origin = None
 
+        self.kalman_filter = PixelKalmanFilter2D((720, 1280), dt=1.0/30.0) if kalman_filter == True else None
```
mdaiter, line 51: (yes, this is hard-coded, removing this and cleaning it up in a bit -- just put this in to get it working with unitree_go2)

PR #581, opened 2025-08-26
Stereo Zed world mapper

spomichter:
…ame to build world map

To run: `python dimos/hardware/stereo_mapper.py --replay zed`

### DIFF
--- a/dimos/hardware/zed_camera.py
+++ b/dimos/hardware/zed_camera.py
@@ -565,6 +565,7 @@ class ZEDModule(Module):
         set_floor_as_origin: bool = True,
         publish_rate: float = 30.0,
         frame_id: str = "camera_link",  # ZED outputs in ROS frame (z-up, x-forward)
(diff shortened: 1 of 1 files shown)


dimos/robot/unitree_webrtc/type/map.py
```diff
@@ -43,20 +43,30 @@ def __init__(
         global_publish_interval: Optional[float] = None,
         min_height: float = 0.15,
         max_height: float = 0.6,
+        frame_id: str = "world",
+        max_map_points: int = 100000,  # Limit accumulated map size
```
leshy, line 47: a bit weird to limit this way, I assume strange downstream things would happen if your global map suddenly changes resolution, I think just resolution setting should be enough, you can get it really coarse and efficient once correct voxelizaion works.. processing 100k points should be cheap and easy also, limit is quite low, unitree single message has 16k points

PR #584, opened 2025-08-28
Zed stereo mapper old data

paul-nechifor:


dimos/robot/unitree_webrtc/type/passthrough_map.py
```diff
@@ -0,0 +1,116 @@
...
+    @rpc
+    def start(self):
+        logger.info(f"PassthroughMap.start() - Publishing every {self.global_publish_interval}s")
+        logger.info("This map module passes through ZED's spatial map without accumulation")
+
+        # Subscribe to lidar messages
+        self.lidar.subscribe(self.process_lidar_message)
+
+        # Publish global map periodically
+        def publish(_):
```
spomichter, line 77: Im surprised this runs i would assume this would return LCM RPC call timeout error @paul-nechifor

paul-nechifor: So this class is just copying the pointcloud from /lidar into the /global_map . It doesn't do any work.

PR #589, opened 2025-08-30
CPU isolated shared mem

mdaiter:
Implements only CPU-bound shard memory.

### DIFF
--- a/dimos/robot/foxglove_bridge.py
+++ b/dimos/robot/foxglove_bridge.py
@@ -25,8 +25,9 @@ class FoxgloveBridge(Module):
     _thread: threading.Thread
     _loop: asyncio.AbstractEventLoop
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, shm_channels=None, **kwargs):
         super().__init__(*args, **kwargs)
+        self.shm_channels = shm_channels or []
         self.start()
 
     @rpc
@@ -35,7 +36,13 @@ class FoxgloveBridge(Module):
             self._loop = asyncio.new_event_loop()
             asyncio.set_event_loop(self._loop)
             try:
-                bridge = LCMFoxgloveBridge(host="0.0.0.0", port=8765, debug=False, num_threads=4)
+                bridge = LCMFoxgloveBridge(
+                    host="0.0.0.0",
+                    port=8765,
+                    debug=False,
+                    num_threads=4,
+                    shm_channels=self.shm_channels,
+                )
                 self._loop.run_until_complete(bridge.run())
             except Exception as e:
                 print(f"Foxglove bridge error: {e}")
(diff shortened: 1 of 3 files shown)


dimos/protocol/pubsub/test_spec.py
```diff
@@ -84,6 +84,28 @@ def lcm_context():
     print("LCM not available")
 
 
+try:
```
leshy: no need for try/except here

onnx/metric3d_vit_small.onnx
leshy, line 1: I assume this file is added by accident

dimos/msgs/sensor_msgs/Image.py
```diff
@@ -343,6 +343,22 @@ def lcm_decode(cls, data: bytes, **kwargs) -> "Image":
             **kwargs,
         )
 
+    def as_memoryview(self) -> memoryview:
```
leshy: shouldn't change image type, this needs to work with all types via pickle

onnx/metric3d_vit_small.onnx
mdaiter: Nope - strangely, `metric3d` was complaining about it not being there. I might have to Git LFS pull instead.

dimos/protocol/pubsub/test_spec.py
mdaiter: Oh, okay - was just following how it was done before!

dimos/msgs/sensor_msgs/Image.py
mdaiter: Don't think we need this as much anymore as well, can remove

dimos/protocol/pubsub/shmpubsub.py
```diff
@@ -0,0 +1,329 @@
...
+
+            try:
+                L = struct.unpack("<I", host[:4].tobytes())[0]
+                if L == 0 or L < 0 or L > st.capacity:
+                    continue
+
+                payload = host[4 : 4 + L].tobytes()
+
+                # Drop exactly the number of local echoes we created
+                cnt = st.suppress_counts.get(payload, 0)
```
paul-nechifor, line 279: I'm not sure it's a good idea to use the entire payload as a key. An alternative would be to give each message a sequence number, but you'd also have to prefix it with a process id. But that's probably a lot more complicated.

dimos/core/transport.py
```diff
@@ -106,6 +107,54 @@ def subscribe(self, callback: Callable[[T], None], selfstream: In[T] = None) ->
         return self.lcm.subscribe(self.topic, lambda msg, topic: callback(msg))
 
 
+class pSHMTransport(PubSubTransport[T]):
```
paul-nechifor, line 110: NIT: This is rather cryptic. How about `PickledSharedMemTransport`

onnx/metric3d_vit_small.onnx
paul-nechifor: I'm not sure why it's needed. Nothing in the PR uses it so wouldn't `dev` break without it?

The other onnx files are in ./data/

dimos/protocol/pubsub/shmpubsub.py
mdaiter: Ah, found a *kind* of workaround to this. I'm gonna feed in a memory view and let the view be the key. No memory copied, per-process pointer, and basically a way of de-duplicating copies. I think it should work? tests pass

PR #609, opened 2025-09-09
CUDA IPC additions

mdaiter:
Hey all - this is the first addition of CUDA IPC.
It *will* break `unitree_go2.py`. You need to run with `DIMOS_IPC_BACKEND=cpu` to make this work.
However - this is the formal addition of CUDA to IPC. Happy to default `DIMOS_IPC_BACKEND=cpu` to make this mergeable.

### DIFF
--- a/dimos/protocol/pubsub/shm/ipc_factory.py
+++ b/dimos/protocol/pubsub/shm/ipc_factory.py
@@ -1,4 +1,4 @@
-# Copyright 2025 Dimensional Inc.
+# Copyright 2025-2026 Dimensional Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/dimos/agents/agent.py
+++ b/dimos/agents/agent.py
@@ -1,4 +1,4 @@
-# Copyright 2025 Dimensional Inc.
+# Copyright 2025-2026 Dimensional Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/dimos/agents/agent_config.py
+++ b/dimos/agents/agent_config.py
@@ -1,4 +1,4 @@
-# Copyright 2025 Dimensional Inc.
+# Copyright 2025-2026 Dimensional Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/dimos/agents/agent_ctransformers_gguf.py
+++ b/dimos/agents/agent_ctransformers_gguf.py
@@ -1,4 +1,4 @@
-# Copyright 2025 Dimensional Inc.
+# Copyright 2025-2026 Dimensional Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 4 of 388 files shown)


dimos/protocol/pubsub/shm/ipc_factory.py
```diff
@@ -274,7 +311,254 @@ def close(self):
...
+            "ptr1": int(self._bufs[1].data.ptr),
+            "handle0": self._handles[0],
+            "handle1": self._handles[1],
+        }
+
+    @classmethod
+    def attach(cls, desc: dict) -> "CudaIpcChannel":
+        import cupy as cp
+
+        obj = object.__new__(cls)
```
paul-nechifor, line 433: I'm confused about the way you're initializing here instead of `cls(...)`. Is it to avoid calling the constructor? If so, couldn't the constructor handle both situations, maybe by passing `attack=True`?

In any case, a comment could help. :)

paul-nechifor: Also, is this the reason you always do `getattr(self, "_is_owner", False)` instead of `self._is_owner or False`?

dimos/protocol/pubsub/shm/ipc_factory.py
```diff
@@ -274,7 +311,254 @@ def close(self):
...
+            if isinstance(frame, cp.ndarray):
+                cp.copyto(dst, frame, casting="no")
+            else:
+                dst.set(frame)
+        self._h2d_stream.synchronize()
+        ts = np.int64(time.time_ns())
+        # Publish order: ts -> idx -> seq
+        self._ctrl[1] = ts
+        self._ctrl[2] = inactive
+        self._ctrl[0] += 1
```
paul-nechifor, line 389: I probably don't know what I'm talking about here, but isn't `.synchronize()` defeating the purpose of doing things in parallel? Can this be replaced with `.add_callback()`?

```python
def cb():
    ts = np.int64(time.time_ns())
    # Publish order: ts -> idx -> seq
    self._ctrl[1] = ts
    self._ctrl[2] = inactive
    self._ctrl[0] += 1

self._h2d_stream.add_callback(cb)
```

mdaiter: Ah. So I was being a bit paranoid here: I wanted to guarantee sequential frames being sent, instead of potentially parallel frames arriving before or after one another. I didn't know whether any of the neural nets would have strange results if they saw potentially out of order frames.

This can be replaced by `add_callback`, and I'll try that now. Good shout-out, and we might want to throw in a flag like `sequentially_guaranteed` to the constructor of this class.

dimos/protocol/pubsub/shm/ipc_factory.py
mdaiter: Good callout to leave comments:
1. `getattr(self, "_is_owner", False)` was used instead of `self._is_owner or False`, because occasionally accessing the field `self._is_owner` would result in fatal error. I suspected this was from referencing between processes, but I wasn't exactly sure. So I decided to be extra cautious, and using `getattr(self, "_is_owner", False)` solved those issues. Occasionally, `_is_owner` would just disappear, along with fairly finicky `weakref` issues.
2. I wanted an uninitialized class, in order to not construct CUDA buffers that'd be tossed and not initialize variables that'd be overwritten. This can also probably just be separated by a flag in the class initializer. Happy to do so.

PR #611, opened 2025-09-10
Gstreamer module (relative time)

paul-nechifor:


dimos/hardware/gstreamer_camera.py
```diff
@@ -0,0 +1,325 @@
...
+gi.require_version("GstApp", "1.0")
+from gi.repository import Gst, GLib
+
+logger = setup_logger("dimos.hardware.gstreamer_camera", level=logging.INFO)
+
+Gst.init(None)
+
+
+class GstreamerCameraModule(Module):
+    """Module that captures frames from a remote camera using GStreamer TCP with absolute timestamps.
```
spomichter, line 44: is it absolute timestamps? I thought relative?

paul-nechifor: Normally, when `buffer.pts = time.time()` is set on the sender, the protocol strips that out and replaces it a timestamp relative to the start of the stream. But the Matroska container preserves the real absolute timestamp now.

PR #612, opened 2025-09-10
Image upgrades! Impls for CUDA + numpy, along with an abstraction and full backwards compatibility

mdaiter:
@paul-nechifor is leading this integration, according to @spomichter .

Full on abstraction for `Image` away from a locked-in platform.

Added all functions @alexlin2 requested in native CUDA + OpenCV numpy fallbacks.

Also, CUDA IPC and CPU IPC are now native.

### DIFF
--- a/dimos/robot/unitree_webrtc/unitree_go2.py
+++ b/dimos/robot/unitree_webrtc/unitree_go2.py
@@ -695,7 +695,6 @@ class UnitreeGo2(UnitreeRobot):
         return self.connection.get_odom()
 
 
-
 def main():
     """Main entry point."""
     ip = os.getenv("ROBOT_IP")


dimos/msgs/sensor_msgs/image_impls/CudaImage.py
```diff
@@ -0,0 +1,674 @@
...
+        if not isinstance(tracker, _CudaTemplateTracker):
+            raise TypeError("Expected CUDA tracker instance")
+        gray = self.to_grayscale().data.astype(cp.float32)
+        x, y, w, h = tracker.update(gray)
+        return True, (int(x), int(y), int(w), int(h))
+
+    # PnP – Gauss–Newton (no distortion in batch), iterative per-instance
+    def solve_pnp(
+        self,
+        object_points: np.ndarray,
```
mdaiter, line 691: (yes, I know, there are a few object points accepted in `np.ndarray` and put on-device. gonna go after this in a bit

mdaiter: the reason being that you might not *need* these parameters on the GPU by the time you're calling this function. It's basically user-convenience. Can shift to GPU or keep them on CPU. @alexlin2, feedback please.

dimos/msgs/sensor_msgs/Image.py
```diff
@@ -12,511 +12,389 @@
...
-        image_bytes = self.data.tobytes()
-        msg.data_length = len(image_bytes)
-        msg.data = image_bytes
-
+            now = time.time()
+            msg.header.stamp.sec = int(now)
+            msg.header.stamp.nsec = int((now - int(now)) * 1e9)
+
+        arr = (
+            self.to_opencv()
```
mdaiter, line 363: nit: repeat

dimos/msgs/sensor_msgs/image_impls/CudaImage.py
```diff
@@ -0,0 +1,674 @@
...
+
+def _gray_to_rgb_cuda(gray):
+    return cp.stack([gray, gray, gray], axis=-1)  # type: ignore
+
+
+def _rgb_to_gray_cuda(rgb):
+    r = rgb[..., 0].astype(cp.float32)  # type: ignore
+    g = rgb[..., 1].astype(cp.float32)  # type: ignore
+    b = rgb[..., 2].astype(cp.float32)  # type: ignore
+    y = 0.299 * r + 0.587 * g + 0.114 * b
```
mdaiter, line 376: (it's a YUV conversion, taking the Y channel. It's luminosity. lemme know if I should drop a comment here)

dimos/msgs/sensor_msgs/Image.py
```diff
@@ -12,511 +12,389 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import base64
+from __future__ import annotations
```
paul-nechifor, line 15: We might not need this (if we support Python 3.12+ only https://discord.com/channels/1341146487186391173/1357833666751107174/1424850698667032629 ).

dimos/msgs/sensor_msgs/Image.py
```diff
@@ -12,511 +12,389 @@
...
-        """Return total number of pixels."""
-        return self.height * self.width
+        return int(self.height * self.width)
+
+
+# Re-exports for tests
+HAS_CUDA = HAS_CUDA
+ImageFormat = ImageFormat
+NVIMGCODEC_LAST_USED = NVIMGCODEC_LAST_USED
+HAS_NVIMGCODEC = HAS_NVIMGCODEC
```
paul-nechifor, line 521: Should this be `__all__ = ["HAS_CUDA", "ImageFormat", "NVIMGCODEC_LAST_USED", "HAS_NVIMGCODEC", ...]
`

dimos/msgs/sensor_msgs/image_impls/CudaImage.py
paul-nechifor: Yeah, you could mention that's the BT.601 standard YUV conversion to make the numbers less magical.

dimos/msgs/sensor_msgs/image_impls/CudaImage.py
```diff
@@ -0,0 +1,672 @@
...
+    if rgb.dtype == cp.uint8:  # type: ignore
+        y = cp.clip(y, 0, 255).astype(cp.uint8)  # type: ignore
+    return y
+
+
+def _resize_bilinear_hwc_cuda(img, out_h: int, out_w: int):
```
paul-nechifor, line 382: I think biliniar resize is implemented in cupy as `cupyx.scipy.ndimage.zoom`  with biliniar being `zoom(..., order=1)` . Cubic (the default) is `order=3`.

dimos/msgs/sensor_msgs/image_impls/CudaImage.py
```diff
@@ -0,0 +1,672 @@
...
+    theta = np.arccos(trace)
+    if np.isclose(theta, 0):
+        return np.zeros((3, 1), dtype=np.float64)
+    rx = R[2, 1] - R[1, 2]
+    ry = R[0, 2] - R[2, 0]
+    rz = R[1, 0] - R[0, 1]
+    r = np.array([rx, ry, rz])
+    r = r / (2.0 * np.sin(theta))
+    rvec = r * theta
+    return rvec.reshape(3, 1).astype(np.float64)
```
paul-nechifor: It doesn't look like this is used. Also, is this the equivalent of `cv2.Rodrigues()`

dimos/msgs/sensor_msgs/image_impls/CudaImage.py
```diff
@@ -0,0 +1,672 @@
...
+            gx = cndimage.convolve1d(gx, smooth5, axis=0, mode="reflect")  # type: ignore
+            gy = cndimage.convolve1d(gray, deriv5, axis=0, mode="reflect")  # type: ignore
+            gy = cndimage.convolve1d(gy, smooth5, axis=1, mode="reflect")  # type: ignore
+            magnitude = cp.hypot(gx, gy)  # type: ignore
+            mean_mag = float(cp.asnumpy(magnitude.mean()))  # type: ignore
+        except Exception:
+            return 0.0
+        if mean_mag <= 0:
+            return 0.0
+        return float(np.clip((np.log10(mean_mag + 1) - 1.7) / 2.0, 0.0, 1.0))
```
paul-nechifor, line 661: Is this similar to ([using cucim](https://docs.rapids.ai/api/cucim/stable/api/#cucim.skimage.filters.sobel)):

```
from cucim.skimage import filters
edges = filters.sobel(camera)
sharpness = cp.mean(edges)
```

dimos/msgs/sensor_msgs/image_impls/CudaImage.py
```diff
@@ -0,0 +1,672 @@
...
+
+    # CUDA tracker (template NCC with small scale pyramid)
+    def create_csrt_tracker(self, bbox: Tuple[int, int, int, int]):
```
paul-nechifor: I think it would be useful to alias this BBox type. There are other places in the codebase where we use a BBox to mean [x1, y1, x2, y2]. An alias would be useful so we don't mix them.

Of course, it doesn't guarantee type safety. Maybe we should just define:

```
@dataclass
class BBox:
    x: int
    y: int
    w: int
    h: int
```

dimos/msgs/sensor_msgs/image_impls/CudaImage.py
```diff
@@ -0,0 +1,672 @@
...
+                return cp.stack(
+                    [
+                        cp.stack([O, -vz, vy], axis=-1),
+                        cp.stack([vz, O, -vx], axis=-1),
+                        cp.stack([-vy, vx, O], axis=-1),
+                    ],
+                    axis=-2,
+                )
+
+            def rodrigues(r):
```
paul-nechifor: This function is also defined in `solve_pnp()`. Same for `skew`. Should they be moved outside?

Also, `_rodrigues_from_R_np` is unused. Did you mean to use it here?

dimos/msgs/sensor_msgs/image_impls/CudaImage.py
```diff
@@ -0,0 +1,672 @@
...
+                theta = cp.linalg.norm(r)
+                I = cp.eye(3, dtype=cp.float64)
+                if float(theta) < 1e-12:
+                    return I
+                k = r / theta
+                Kx = skew(k)
+                s, c = cp.sin(theta), cp.cos(theta)
+                return I + s * Kx + (1.0 - c) * (Kx @ Kx)
+
+            for _ in range(20):
```
paul-nechifor: I see that the number of iterations is different in all three versions? Should it be configurable?

tests/test_image_backend_utils.py
```diff
@@ -0,0 +1,279 @@
...
+    img = Image(data=arr, format=fmt, frame_id="cam", ts=123.456)
+    K = np.array(
+        [[100.0, 0, arr.shape[1] / 2], [0, 100.0, arr.shape[0] / 2], [0, 0, 1]], dtype=np.float64
+    )
+    D = np.zeros(5, dtype=np.float64)
+    out = rectify_image(img, K, D)
+    assert out.shape[:2] == arr.shape[:2]
+    assert out.format == fmt
+    assert out.frame_id == "cam"
+    assert abs(out.ts - 123.456) < 1e-9
```
paul-nechifor, line 63: nit: `assert out.ts == pytest.approx(123.456, rel=1e-9)`

tests/test_image_backend_utils.py
```diff
@@ -0,0 +1,279 @@
...
+    import cupy as cp  # type: ignore
+
+    depth_np = np.zeros((16, 20), dtype=np.float32)
+    depth_np[4:8, 5:15] = 2.0
+    out_cpu = colorize_depth(depth_np, max_depth=4.0, overlay_stats=False)
+    out_gpu = colorize_depth(cp.asarray(depth_np), max_depth=4.0, overlay_stats=False)
+    np.testing.assert_array_equal(cp.asnumpy(out_gpu), out_cpu)
+
+
+def test_draw_bounding_box_cpu():
```
paul-nechifor, line 209: Just a question. What's your opinion on using actually images for testing instead of zeros or noise?

That is:

1. add a small sample image
2. perform the modification on the image and save it as expected.png
3. in the test assert the result image is similar to expected.png

This way we can see what it's supposed to look like. And if we change the implementation and a test fails, we can compare what it should have been and what the new code is generating.

dimos/msgs/sensor_msgs/Image.py
mdaiter: Fair. Does it hurt to leave it in? Happy to take it out, but if it allows backwards compatibility, a comment might help here.

dimos/msgs/sensor_msgs/Image.py
mdaiter: What would `__all__` add here? I'm just not following.

dimos/msgs/sensor_msgs/image_impls/CudaImage.py
mdaiter: Not quite. This matches the OpenCV / PyTorch style, whereas cupyx matches its own coordinate system. You could do this:
```
zoom_factors = (out_h / in_h, out_w / in_w, 1)
out = cupyx.scipy.ndimage.zoom(img, zoom=zoom_factors, order=1, mode='nearest', grid_mode=True)
```
to trick the system though. It's a good idea - when AWS is back up, I can test and see if parity hits

dimos/msgs/sensor_msgs/image_impls/CudaImage.py
mdaiter: Good call - it is *not* used. I can get rid of it.

dimos/msgs/sensor_msgs/image_impls/CudaImage.py
mdaiter: Almost - this is a bit numerically different. We use a 5x5 Sobel filter here with a second order derivative, whereas the normal Sobel uses a 3x3 with a first derivative. But it's a good shout - let me check if we can use this

mdaiter: Ahhh boooo cucim pre-bakes the kernel

dimos/msgs/sensor_msgs/image_impls/CudaImage.py
mdaiter: Good shout. When I get an AWS instance back, I can run this

dimos/msgs/sensor_msgs/image_impls/CudaImage.py
mdaiter: Another good shout: on it when I have AWS back

dimos/msgs/sensor_msgs/image_impls/CudaImage.py
mdaiter: (when I have AWS back)

tests/test_image_backend_utils.py
mdaiter: (on it when I have AWS back)

tests/test_image_backend_utils.py
mdaiter: We could do that! Hit me with an image, happy to include.

dimos/msgs/sensor_msgs/Image.py
paul-nechifor: I wasn't sure why you're doing `x = x`. I thought that's meant to say that the variable should be exported. If so, then using `__all__` is a more explicit way of exporting. Otherwise, I'm not sure what this does.

dimos/msgs/sensor_msgs/Image.py
paul-nechifor: No, you can leave it. Looks like people haven't replied to that comment yet.

dimos/msgs/sensor_msgs/image_impls/CudaImage.py
```diff
@@ -0,0 +1,716 @@
...
+        raise ValueError(f"Unsupported format: {self.format}")
+
+    def resize(self, width: int, height: int, interpolation: int = cv2.INTER_LINEAR) -> "CudaImage":
+        return CudaImage(
+            _resize_bilinear_hwc_cuda(self.data, height, width), self.format, self.frame_id, self.ts
+        )
+
+    def sharpness(self) -> float:
+        if cp is None:
+            return 0.0
```
leshy, line 644: should throw an error here, not return a magic number, is there a situation in which cp is None here?

leshy: is this a diff algo then cpu image sharpness? does this return the same value? is it fair to compare in tests?

tests/test_image_backends.py
```diff
@@ -0,0 +1,517 @@
...
+    cpu = Image.from_numpy(arr, format=ImageFormat.BGR)
+    gpu = Image.from_numpy(arr, format=ImageFormat.BGR, to_cuda=True)
+    import time
+
+    t0 = time.perf_counter()
+    for _ in range(5):
+        _ = cpu.resize(320, 240)
+    cpu_t = time.perf_counter() - t0
+    t0 = time.perf_counter()
+    for _ in range(5):
```
leshy: we need to measure aloc time as well, not init once then measure calculation, and this needs to be done for all tests, since every module will need to initialize image before doing operations on it, and init will happen on every frame on ingest

dimos/msgs/sensor_msgs/image_impls/CudaImage.py
mdaiter: 1. Agreed, that makes sense. Technically, a zero sharp image is just blurry. But I understand where this comment is derivative from.
2. Check the parity tests. It's explicitly compared.

PR #616, opened 2025-09-11
Pshm to lcm

paul-nechifor:


dimos/msgs/vision_msgs/__init__.py
leshy, line 1: do relative paths work? is this a mistake?

paul-nechifor: Yeah, I modeled it after `dimos/msgs/std_msgs/__init__.py`

PR #631, opened 2025-09-22
modules 2.0

leshy:


docs/specs/modules/notes-0.md
```diff
@@ -0,0 +1,61 @@
...
+    detections3d: Out[Detection3DArray] = None  # type: ignore
+```
+
+above doesn't work, all I/O needs to be re-specified. 
+
+# configuration inheritance
+
+protocol/service/spec.py - Configurable module built, experimenting with config classes.
+
+seems like dataclasses have inheritance issues (something related to optional values)
```
paul-nechifor, line 22: Persoanlly, I'd keep it simple. I don't understand why config is hoisted up to `Configurable`. We get better type safety by composing configs (instead of inheriting) and just passing them down.


```python
@dataclass(frozen=True)
class ModuleConfig:
    tf_transport: type[TFSpec] = LCMTF

@dataclass(frozen=True)
class ConnectionModuleConfig
    module_config: Optional[ModuleConfig] = None
    ip: Optional[str] = None
    connection_type: str = "webrtc"
    ...

class ConnectionModule(Module):
    _config: ConnectionModuleConfig

    def __init__(self, config: ConnectionModuleConfig):
        super().__init__(config.module_config)
        self._config = config
```

docs/specs/modules/notes-0.md
```diff
@@ -0,0 +1,61 @@
+# module deps/inheritance
```
paul-nechifor, line 1: Is this sort of like PEP (Python Enhancement Proposal). Maybe we can call yours DEP1.

docs/specs/modules/notes-0.md
```diff
@@ -0,0 +1,61 @@
...
+should be able to run Detection3DModule alone (2d processsing is inherited)
+or as 2d + 3d as separate nodes, in some very convinienet way
+
+stream/audio/* stuff is really convininet to use, modules aren't, they should be.
+
+`player(normalizer(microphone_input))` is super fast, building this with streams, then calling dimos.deploy manually would be annoying, modules have full inspectability and they should support this type of an api or something similar.
+
+# Stream generalization
+
+skills are generators, streams between modules are something DIY, streams in modules are reactivex sometimes.
```
paul-nechifor, line 51: About skills, the inheritance is quite odd:

```
DaskModule -> ModuleBase -> SkillContainer
```

SkillContainer is the root class. I would expect SkillContainer to be a subclass of DaskModule, not an ancestor.

This is what confused me about SkillCoordinator:

```python
class SkillCoordinator(Module):
    def __init__(self) -> None:
        SkillContainer.__init__(self)
```

It skips calling `DaskModule.__init__` and `ModuleBase.__init__` and calls `SkillContainer.__init__` directly.

docs/specs/modules/notes-0.md
```diff
@@ -0,0 +1,61 @@
...
+
+class Detection3DModule(Detection2DModule[how_does_this_work?]):
+   def __init__(self, **how_does_typing_work?)_:
+   ...
+```
+
+# composibility/decomposibility
+
+should be able to run Detection3DModule alone (2d processsing is inherited)
+or as 2d + 3d as separate nodes, in some very convinienet way
```
paul-nechifor, line 43: What does Detection2DModule even have in common with Detection3DModule? They should just be separate classes.

docs/specs/modules/notes-0.md
```diff
@@ -0,0 +1,61 @@
...
+We need a good general interface for "a stream" that can be interpreted as a generator, rxpy, loop etc. from outside and inside
+
+# performance
+
+all streams should have easy way to test performance speed, we want to detect perforamnce deteriroration in CI, related to testing task below
+
+# testing
+
+All modules should have an easy way to feed "a moment in time" into them, or a full replay, some slice of all streams they care for - to check the module output. - important for testing complex higher order processing without running an actual robot or nodes below. Makes it easy to test performance of the isolated module
+
```
paul-nechifor, line 61: What I would really like from modules is to have a clean way to start then and stop them. I would add an ExitStack and a CompositeDisposable on every module to manage cleanup and implement the context manager protocol.

Also, we need a root Dimos class for the system. It should instantiate modules and shut them down. The RPC system uses the name of the module to generate the topic so we can't have multiple instances of a module running anyway.

docs/specs/modules/notes-0.md
leshy: actually true, the idea behind configurable was to apply defaults correctly, maybe read the env and overlay on the config etc, but we can have a smart configuration class do all that, and use proper inheritance

docs/specs/modules/notes-0.md
leshy: maybe yes, more the point is that process orchestration should be so easy that one doesn't care or need to change much

docs/specs/modules/notes-0.md
leshy: yeah that's a typo probably? SkillCoordinator is supposed to inherit a SkillContainer (the idea is that it behaves on the outside the same as SkillContainer, but provides access to many Skillcontainers)

but yeah unrealted to the above, I meant we have a bunch of diff ways to "stream things" and it should be the same shared interface

docs/specs/modules/notes-0.md
leshy: Audio system is for example a pleasure to work with, it's close to like,

`recorder().pipe(normalizer).pipe(stt).pipe(console)`

can I run like
`robot.pipe(agent).pipe(robot)`
robot has `camera.pipe(detection2d).pipe(detection3d).pipe(objectdb)` inside?

idk stuff like this, with easy way to run something as a process or not

maybe rxpy or similar (better) streaming interface but works with processes. we should be able to build any interface we want since each module is fully inspectable around it's I/O

docs/specs/modules/notes-0.md
leshy: dataclases have some issue with default/non-default arg and inheritance. It's very reasonable to inherit configs as you build higher order stuff, so probably we'd need pydantic. I can document if you are not familiar

PR #632, opened 2025-09-22
Alexl ros nav intergration

alexlin2:


AUTONOMY_STACK_README.md
```diff
@@ -0,0 +1,277 @@
+# Autonomy Stack API Documentation
+
+## Prerequisites
+
+- Ubuntu 24.04
+- [ROS 2 Jazzy Installation](https://docs.ros.org/en/jazzy/Installation.html)
+
```
paul-nechifor, line 7: I would add 'clone the autonomy_stack_mecanum_wheel_platform repo, and follow the instructions in the readme'.

dimos/robot/nav_bot.py
```diff
@@ -0,0 +1,404 @@
...
+
+
+# then deploy this module in any other run file.
+############################################################
+class NavigationModule(Module):
+    goal_reached: In[Bool] = None
+
+    goal_pose: Out[PoseStamped] = None
+    cancel_goal: Out[Bool] = None
+    joy: Out[Joy] = None
```
spomichter: Great in/out interface would be better if this was Twist rather that Joy though @alexlin2

alexlin2: This joy is to emulate change the mode from autonomy to joystick so that the navstack will allow movement, there is a twist subscriber

dimos/msgs/sensor_msgs/test_Joy.py
```diff
@@ -0,0 +1,243 @@
...
+
+if __name__ == "__main__":
+    test_lcm_encode_decode()
+    test_initialization_methods()
+    test_equality()
+    test_string_representation()
+    if ROS_AVAILABLE:
+        test_ros_conversion()
+    test_edge_cases()
+    print("\nAll Joy tests passed! ✓")
```
paul-nechifor: I've seen these and I don't understand the purpose. Do you not use pytest to run the tests?

dimos/robot/nav_bot.py
```diff
@@ -0,0 +1,426 @@
...
+
+    def __init__(self, sensor_to_base_link_transform=None, *args, **kwargs):
+        """Initialize NavigationModule."""
+        Module.__init__(self, *args, **kwargs)
+        if not rclpy.ok():
+            rclpy.init()
+        Node.__init__(self, "navigation_module")
+
+        self.goal_reach = None
+        self.tf = TF()
```
leshy: there is no need for this, you get a warning when running, self.tf is already available on all modules

dimos/robot/nav_bot.py
```diff
@@ -0,0 +1,426 @@
...
+            ROSPointCloud2, "/registered_scan", self._on_ros_registered_scan, 10
+        )
+
+        logger.info("NavigationModule initialized with ROS2 node")
+
+    @rpc
+    def start(self):
+        """Start the navigation module and set up bridging."""
+        # Set up ROS2 spinning in a thread
+        import threading
```
leshy: let's not import in the middle of files if no reason for it

dimos/robot/nav_bot.py
paul-nechifor: I don't think you need this. You already have it in `ModuleBase`:

```python
    @property
    def tf(self):
        if self._tf is None:
            self._tf = self.config.tf_transport()
        return self._tf

    @tf.setter
    def tf(self, value):
        import warnings

        warnings.warn(
            "tf is available on all modules. Call self.tf.start() to activate tf functionality. No need to assign it",
            UserWarning,
            stacklevel=2,
        )
```

dimos/robot/nav_bot.py
```diff
@@ -0,0 +1,426 @@
...
+                self.soft_stop_pub.publish(soft_stop_msg)
+                return self.goal_reach
+            time.sleep(0.1)
+
+        self.stop()
+        logger.warning(f"Navigation timed out after {timeout} seconds")
+        return False
+
+    @rpc
+    def stop(self) -> bool:
```
paul-nechifor: I'd rename `stop` to `cancel_navigation` and `shutdown` to `stop`.

`start` starts the module. So `stop` should stop the module, not the navigation.

dimos/robot/nav_bot.py
```diff
@@ -0,0 +1,426 @@
...
+        cancel_msg.data = True
+        self.cancel_goal_pub.publish(cancel_msg)
+
+        soft_stop_msg = ROSInt8()
+        soft_stop_msg.data = 2
+        self.soft_stop_pub.publish(soft_stop_msg)
+
+        return True
+
+    def shutdown(self):
```
paul-nechifor: This needs `@rpc` too, right?

dimos/robot/nav_bot.py
```diff
@@ -0,0 +1,426 @@
...
+        )
+        self.cmd_vel_sub = self.create_subscription(
+            ROSTwistStamped, "/cmd_vel", self._on_ros_cmd_vel, 10
+        )
+        self.goal_waypoint_sub = self.create_subscription(
+            ROSPointStamped, "/way_point", self._on_ros_goal_waypoint, 10
+        )
+        self.registered_scan_sub = self.create_subscription(
+            ROSPointCloud2, "/registered_scan", self._on_ros_registered_scan, 10
+        )
```
paul-nechifor: You have to call destroy_subscription when the module is stopped.

dimos/robot/nav_bot.py
```diff
@@ -0,0 +1,426 @@
...
+            0.0,
+            0.0,
+        ]
+        self.spin_thread = None
+
+        # ROS2 Publishers
+        self.goal_pose_pub = self.create_publisher(ROSPoseStamped, "/goal_pose", 10)
+        self.cancel_goal_pub = self.create_publisher(ROSBool, "/cancel_goal", 10)
+        self.soft_stop_pub = self.create_publisher(ROSInt8, "/soft_stop", 10)
+        self.joy_pub = self.create_publisher(ROSJoy, "/joy", 10)
```
paul-nechifor: You have to call destroy_publisher when the module is stopped.

dimos/robot/nav_bot.py
```diff
@@ -0,0 +1,426 @@
...
+from geometry_msgs.msg import PointStamped as ROSPointStamped
+from nav_msgs.msg import Odometry as ROSOdometry
+from sensor_msgs.msg import PointCloud2 as ROSPointCloud2, Joy as ROSJoy
+from std_msgs.msg import Bool as ROSBool, Header, Int8 as ROSInt8
+from tf2_msgs.msg import TFMessage as ROSTFMessage
+
+logger = setup_logger("dimos.robot.unitree_webrtc.nav_bot", level=logging.INFO)
+
+
+class NavigationModule(Module, Node):
```
paul-nechifor: I disagree with multiple inheritance.

Module has nothing in common with Node and Node is external. If they add a method like `start` it will interfere with our code and lead to bugs. The simple solution is to just use Node.

```python
self._node = Node("navigation_module")
```

https://en.wikipedia.org/wiki/Composition_over_inheritance

dimos/robot/nav_bot.py
alexlin2: Ah I see

dimos/robot/nav_bot.py
alexlin2: Will do

dimos/robot/nav_bot.py
alexlin2: Yes, will make that change

dimos/robot/nav_bot.py
alexlin2: Yes

alexlin2: destroy_node() already does that

dimos/robot/nav_bot.py
alexlin2: destroy_node() already does that too

dimos/robot/nav_bot.py
alexlin2: I need both module and node to subscribe and publish both ROS and DIMOS topics, @leshy anyway to solve this?

AUTONOMY_STACK_README.md
alexlin2: yes

dimos/robot/nav_bot.py
leshy: yeah above what paul suggested should be ok

dimos/robot/nav_bot.py
```diff
@@ -0,0 +1,425 @@
...
+    """
+    Handles navigation control and odometry remapping.
+    """
+
+    goal_req: In[PoseStamped] = None
+    cancel_goal: In[Bool] = None
+    soft_stop: In[Int8] = None
```
leshy: can we move cancel_goal and soft_stop to a single rpc as discussed? stop_navigation() ?

dimos/navigation/rosnav/nav_bot.py
```diff
@@ -0,0 +1,425 @@
...
+    goal_req: In[PoseStamped] = None
+    cancel_goal: In[Bool] = None
+    soft_stop: In[Int8] = None
+
+    pointcloud: Out[PointCloud2] = None
+    global_pointcloud: Out[PointCloud2] = None
+
+    goal_active: Out[PoseStamped] = None
+    path_active: Out[Path] = None
+    goal_reached: Out[Bool] = None
```
leshy, line 67: do we need this given navigate_to is rpc? not totally sure atm, but I'd remove and add if needed, since all navs need to mirror, it's good to keep the it light

dimos/navigation/rosnav/nav_bot.py
```diff
@@ -0,0 +1,425 @@
...
+    odom: Out[Odometry] = None
+    cmd_vel: Out[TwistStamped] = None
+    odom_pose: Out[PoseStamped] = None
```
leshy, line 70: to remove, odom_pose, we discussed, we have Odometry and transforms

dimos/robot/nav_bot.py
alexlin2: Done

dimos/navigation/rosnav/nav_bot.py
alexlin2: Stash said to use topics instead of rpc calls to pass goals and receive goal updates

dimos/navigation/rosnav/nav_bot.py
alexlin2: I need it for legacy modules such as all the working module we have now, object_tracker, spatial_memory, we can remove this once those modules are updated

leshy: cool can clean in a follow up

dimos/navigation/rosnav/nav_bot.py
spomichter: Yes because certain transports like LCM will drop messages so need to send a stream of Bools to ensure cancel goal

dimos/navigation/rosnav/nav_bot.py
```diff
@@ -0,0 +1,421 @@
...
+        self.tf.publish(sensor_to_base_link_tf, map_to_world_tf, ros_tf)
+
+    def _on_goal_pose(self, msg: PoseStamped):
+        self.navigate_to(msg)
+
+    def _on_cancel_goal(self, msg: Bool):
+        if msg.data:
+            self.stop()
+
+    def _set_autonomy_mode(self):
```
spomichter, line 224: @alexlin2 Like discussed this is a generic NavBot but set autonomy mode is specific to ROS Mechanum drive stack. Move to a wrapper child class. Called MechanumDriveNavBot or something

alexlin2: This is not the generic interface, this class is an implementation of the nav stack spec, so it will have implementation specifics

spomichter: NavigationModule I thought is the generic interface. If this isn't it, where is the generic interface.

spomichter: Its called NavigationModule thats very generic

alexlin2: https://github.com/dimensionalOS/dimos/blob/navspec/dimos/navigation/rosnav.py This is the generic interface

alexlin2: Ok, I have changed it to inherit from the navspec @leshy gave

dimos/navigation/rosnav/nav_bot.py
leshy: it's possible to overload the transport but we shouldn't build the systems above under assumptions that underlying systems don't work. now with shared memory also this is not an issue

leshy: Also this is a driver implementation, we want absolute minimum required to achieve full control, because every future driver needs to implement these functions.

If we want goal reached, cancel goal, and any other extra topic, specific navigation related skills, awesome visualization functions, we can implement all this but not on an individual driver level

we implement it in a base class, so that all future drivers get the same functionality

PR #640, opened 2025-09-27
PRODUCTION Nav skills on drone with tracking

spomichter:
### Summary
Complete drone integration via RosettaDrone MAVLink bridge with autonomous object tracking capabilities.

### Key Features
- **MAVLink drone control** - Arm, takeoff, land, fly_to GPS coordinates, velocity movement
- **Visual servoing** - Object tracking using Qwen detection + CSRT tracker + PID controller
- **Indoor/Outdoor modes** - Conservative indoor PID (1 m/s max) vs aggressive outdoor (5 m/s max)
- **Depth estimation** - Metric3D integration for 3D perception
- **Replay mode** - Full test coverage with recorded MAVLink/video data

### New Files
```
dimos/robot/drone/
├── drone.py                           # Main orchestrator
├── connection_module.py               # MAVLink + skills
├── camera_module.py                   # Video + depth
├── drone_tracking_module.py           # Visual servoing
├── drone_visual_servoing_controller.py # PID controller
├── mavlink_connection.py              # MAVLink protocol
├── dji_video_stream.py                # GStreamer capture
└── test_drone.py                      # Unit tests
```

### Other Changes
- Renamed `street_map_query` → `map_query` in OSM skill
- Added position marker to OSM map visualization
- Added Foxglove dashboard for drone telemetry

### Test Plan
- [x] Unit tests pass (`pytest dimos/robot/drone/test_drone.py`)
- [ ] Manual test with replay mode
- [ ] Manual test with real drone (outdoor mode)

### DIFF
--- a/dimos/robot/drone/README.md
+++ b/dimos/robot/drone/README.md
@@ -9,7 +9,7 @@ Complete integration for DJI drones via RosettaDrone MAVLink bridge with visual 
 # Test with replay mode (no hardware needed)
 python dimos/robot/drone/drone.py --replay
 
-# Real drone - indoor (visual odometry)
+# Real drone - indoor (IMU odometry)
 python dimos/robot/drone/drone.py
 
 # Real drone - outdoor (GPS odometry)
@@ -200,14 +200,7 @@ Parameters: `(Kp, Ki, Kd, (min_output, max_output), integral_limit, deadband_pix
 
 ### Unit Tests
 ```bash
-# Test visual servoing controller
-pytest dimos/robot/drone/test_drone_visual_servoing_controller.py
-
-# Test tracking module
-pytest dimos/robot/drone/test_drone_tracking.py
-
-# Test stream communication
-pytest dimos/robot/drone/test_tracking_stream.py
+pytest -s dimos/robot/drone/
 ```
 
 ### Replay Mode (No Hardware)


dimos/robot/drone/connection_module.py
```diff
@@ -0,0 +1,440 @@
...
+    def stop(self):
+        """Stop the module."""
+        self._running = False
+        if self.video_stream:
+            self.video_stream.stop()
+        if self.connection:
+            self.connection.disconnect()
+        logger.info("Drone connection module stopped")
+
+    @skill(output=Output.image)
```
spomichter, line 482: @leshy

dimos/mapping/osm/current_location_map.py
```diff
@@ -67,10 +68,41 @@ def _fetch_new_map(self) -> None:
...
+        radius = 20
+        draw.ellipse(
+            [x - radius, y - radius, x + radius, y + radius],
+            fill=(255, 0, 0),
+            outline=(0, 0, 0),
+            width=3,
+        )
+
+        self._map_image.image.data[:] = np.array(pil_image)
+
```
paul-nechifor, line 89: Why are you drawing a circle on the current position? `query_for_one_position_and_context` already sends the position textually: `    my_location = f"I'm currently at x={x}, y={y}."` in https://github.com/dimensionalOS/dimos/blob/dev/dimos/mapping/osm/query.py#L39

This position will be out of date anyway as `_fetch_new_map` is only called when the robot strays from the center of the map.

dimos/web/websocket_vis/websocket_vis_module.py
```diff
@@ -174,9 +186,21 @@ async def click(sid, position):
...
+            # Add to goal points list for visualization
+            self.gps_goal_points.append(goal)
+            logger.info(f"Added GPS goal to list. Total goals: {len(self.gps_goal_points)}")
+
+            # Emit updated goal points back to all connected clients
+            await self.sio.emit("gps_travel_goal_points", self.gps_goal_points)
+            logger.debug(
+                f"Emitted gps_travel_goal_points with {len(self.gps_goal_points)} points: {self.gps_goal_points}"
+            )
+
```
paul-nechifor, line 239: This is not the best way to do it. Firstly, it duplicates what `set_gps_travel_goal_points` is doing, but the main issue is that if you set the visualized goal only from foxglove, then you can't visualize the gps goal coming from other places (such as from agents).

The fix I added is in https://github.com/dimensionalOS/dimos/commit/c7f55f74e0244177631d1b1a5d832af3adc0e8d6

dimos/robot/drone/connection_module.py is in charge of updating the visualization of the goal based on what the drone is actually commanded to move to.

dimos/web/websocket_vis/websocket_vis_module.py
```diff
@@ -187,6 +211,13 @@ async def stop_explore(sid):
             logger.info("Stopping exploration")
             self.stop_explore_cmd.publish(Bool(data=True))
 
+        @self.sio.event
+        async def clear_gps_goals(sid):
+            logger.info("Clearing all GPS goal points")
+            self.gps_goal_points.clear()
+            await self.sio.emit("gps_travel_goal_points", self.gps_goal_points)
+            logger.info("GPS goal points cleared and updated clients")
+
```
paul-nechifor: It doesn't look like this is used? Did you perhaps forget to commit a change to `dimos/web/command-center-extension`?

dimos/robot/drone/test_drone.py
```diff
@@ -0,0 +1,978 @@
...
+from dimos.robot.drone.dji_video_stream import FakeDJIVideoStream
+from dimos.robot.drone.connection_module import DroneConnectionModule
+from dimos.robot.drone.drone import Drone
+from dimos.msgs.geometry_msgs import Vector3, PoseStamped, Quaternion
+from dimos.msgs.sensor_msgs import Image, ImageFormat
+import numpy as np
+import os
+
+
+class TestMavlinkProcessing(unittest.TestCase):
```
paul-nechifor, line 36: Out of curiosity, why are you using unittest tests? pytest tests are so much nicer, in my humble opinion.

Especially because of nice fixtures. In this case, I'd use:

```
@pytest.fixture
def fake_mavlink_connection():
    conn = MavlinkConnection("udp:0.0.0.0:14550")

    # Mock the mavlink connection
    conn.mavlink = MagicMock()
    conn.connected = True
    return conn
```

and avoid all the repetition.

dimos/robot/drone/drone_tracking_module.py
```diff
@@ -0,0 +1,377 @@
...
+    def _get_latest_frame(self) -> Optional[np.ndarray]:
+        """Get the latest video frame as numpy array."""
+        with self._frame_lock:
+            if self._latest_frame is None:
+                return None
+            # Convert Image to numpy array
+            return self._latest_frame.data
+
+    @rpc
+    def start(self):
```
paul-nechifor: You need a `def stop()` to close all the subscriptions opened in start. You also need to call `self._close_module()` to stop all the things started in `ModuleBase.__init__()`

dimos/robot/drone/dji_video_stream.py
```diff
@@ -0,0 +1,214 @@
...
+                if line:
+                    msg = line.decode("utf-8").strip()
+                    if "ERROR" in msg or "WARNING" in msg:
+                        logger.warning(f"GStreamer: {msg}")
+                    else:
+                        logger.debug(f"GStreamer: {msg}")
+            except:
+                pass
+
+    def stop(self):
```
paul-nechifor: You also need to close _capture_thread and _error_thread

dimos/robot/drone/dji_video_stream.py
```diff
@@ -0,0 +1,214 @@
...
+        width, height = 640, 360
+        channels = 3
+        frame_size = width * height * channels
+
+        logger.info(f"Capturing frames: {width}x{height} RGB ({frame_size} bytes per frame)")
+
+        frame_count = 0
+        total_bytes = 0
+
+        while self._running:
```
paul-nechifor: In this thread and the others, it's not safe to use `self._running`. Use events as they are threadsafe: `self._stop_event = threading.Event()`

dimos/robot/drone/dji_video_stream.py
```diff
@@ -0,0 +1,214 @@
...
+            try:
+                line = self._process.stderr.readline()
+                if line:
+                    msg = line.decode("utf-8").strip()
+                    if "ERROR" in msg or "WARNING" in msg:
+                        logger.warning(f"GStreamer: {msg}")
+                    else:
+                        logger.debug(f"GStreamer: {msg}")
+            except:
+                pass
```
paul-nechifor: This is not a good pattern.

```python
while True:
    try:
        ...
    except:
         pass
```

The issue is that if you have a failure (for example `self._process` closes for some reason) then you have an infinite loop with 100% thread utilization.

The best solution is to just not handle the exception and let it close the thread. It's not like you can recover.

An alternative is to sleep like you do in the `_capture_loop` thread. That will prevent wasting CPU. But I still think it's a bad idea.

dimos/robot/drone/dji_video_stream.py
```diff
@@ -0,0 +1,214 @@
...
+        channels = 3
+        frame_size = width * height * channels
+
+        logger.info(f"Capturing frames: {width}x{height} RGB ({frame_size} bytes per frame)")
+
+        frame_count = 0
+        total_bytes = 0
+
+        while self._running:
+            try:
```
paul-nechifor, line 115: The idea behind `while _running: try-except` is that you can resume after temporary errors.

But I don't think this applies here. If the exception is that you're trying to read from the stdout of a closed process, that will just spam errors at 10Hz. It's never going to recover.

A better way is to just remove the try-except and let the threads die. Have a separate way of checking if `DJIDroneVideoStream` is running, and if it's not, restart the whole thing (`_process` and all)

dimos/robot/drone/dji_video_stream.py
```diff
@@ -0,0 +1,214 @@
...
+            return True
+
+        except Exception as e:
+            logger.error(f"Failed to start video stream: {e}")
+            return False
+
+    def _capture_loop(self):
+        """Read frames with fixed size."""
+        # Fixed parameters
+        width, height = 640, 360
```
paul-nechifor: I think these should be instance properties and used in `start` here too: `"video/x-raw,width=640,height=360",`

dimos/robot/drone/connection_module.py
```diff
@@ -0,0 +1,440 @@
...
+    def _on_gps_goal(self, cmd: LatLon) -> None:
+        current_alt = self._latest_telemetry.get("GLOBAL_POSITION_INT", {}).get("relative_alt", 0)
+        self.connection.fly_to(cmd.lat, cmd.lon, current_alt)
+
+    def _on_tracking_status(self, status: String) -> None:
+        with self._latest_status_lock:
+            self._latest_status = json.loads(status.data)
+
+    @rpc
+    def stop(self):
```
paul-nechifor: There are a lot more things you have to stop here like `_telemetry_thread` and the subscriptions.

dimos/robot/drone/connection_module.py
```diff
@@ -0,0 +1,440 @@
...
+        self._latest_telemetry = telemetry
+
+        if "GLOBAL_POSITION_INT" in telemetry:
+            tel = telemetry["GLOBAL_POSITION_INT"]
+            self.gps_location.publish(LatLon(lat=tel["lat"], lon=tel["lon"]))
+
+    def _telemetry_loop(self):
+        """Continuously update telemetry at 30Hz."""
+        frame_count = 0
+        while self._running:
```
paul-nechifor, line 227: Same as in that other close. This should use an Event(), not a boolean as it's not threadsafe.

dimos/robot/drone/connection_module.py
```diff
@@ -0,0 +1,440 @@
...
+                            position=Vector3(0, 0, 0),
+                            orientation=Quaternion(0, 0, 0, 1),
+                            frame_id="world",
+                            ts=time.time(),
+                        )
+                        self._publish_tf(default_pose)
+                        logger.debug("Publishing default odometry")
+
+                frame_count += 1
+                time.sleep(0.033)  # ~30Hz
```
paul-nechifor, line 247: This being in sync with this below is a requirement, right?

```python
                vx, vy, vz = self.servoing_controller.compute_velocity_control(
                    target_x=current_x,
                    target_y=current_y,
                    center_x=center_x,
                    center_y=center_y,
                    dt=0.033,  # ~30Hz
                    lock_altitude=True,
                )
```

You should use a common constant for both.

Also, if you need to achieve actual 30Hz, a better way would be `time.sleep(1/30 - processing_time` where `processing_time` is the time in between the sleeps.

dimos/robot/drone/connection_module.py
```diff
@@ -0,0 +1,440 @@
...
+            tel = telemetry["GLOBAL_POSITION_INT"]
+            self.gps_location.publish(LatLon(lat=tel["lat"], lon=tel["lon"]))
+
+    def _telemetry_loop(self):
+        """Continuously update telemetry at 30Hz."""
+        frame_count = 0
+        while self._running:
+            try:
+                # Update telemetry from drone
+                self.connection.update_telemetry(timeout=0.01)
```
paul-nechifor: This thread calls  update_telemetry in connection, which makes `connection.telemetry` unsafe to use without a lock.

Ideally, connection should be in charge of updating its telemetry with its own thread. But you still need a lock in either case.

dimos/robot/drone/dji_video_stream.py
spomichter: fixex

dimos/robot/drone/dji_video_stream.py
spomichter: fixed

dimos/robot/drone/camera_module.py
```diff
@@ -48,14 +49,14 @@ class DroneCameraModule(Module):
...
     # Outputs
-    color_image: Out[Image] = None
-    depth_image: Out[Image] = None
-    depth_colorized: Out[Image] = None
-    camera_info: Out[CameraInfo] = None
-    camera_pose: Out[PoseStamped] = None
+    color_image: Out[Image]
+    depth_image: Out[Image]
+    depth_colorized: Out[Image]
+    camera_info: Out[Any]
```
paul-nechifor: I don't think this should be Any. It won't match in blueprints.

dimos/robot/drone/camera_module.py
```diff
@@ -19,8 +19,9 @@
 
 import threading
 import time
+from typing import Any
 
-from dimos_lcm.sensor_msgs import CameraInfo
+from dimos_lcm.sensor_msgs import CameraInfo  # type: ignore[import-untyped]
```
paul-nechifor, line 24: Should this perhaps be `from dimos.msgs.sensor_msgs import CameraInfo`

dimos/robot/drone/drone.py
```diff
@@ -344,7 +390,7 @@ def stop(self) -> None:
             self.foxglove_bridge.stop()
 
         if self.dimos:
-            self.dimos.shutdown()
+            self.dimos.shutdown()  # type: ignore[no-untyped-call]
```
paul-nechifor: I think you want to call `self.dimos.close_all()` here.

spomichter: yes

dimos/robot/drone/camera_module.py
```diff
@@ -0,0 +1,291 @@
...
+            self._processing_thread.join(timeout=2.0)
+
+        # Cleanup Metric3D
+        if self.metric3d:
+            self.metric3d.cleanup()
+
+        logger.info("Camera module stopped")
+
+    @rpc
+    def get_camera_intrinsics(self) -> list[float]:
```
paul-nechifor: not used

dimos/robot/drone/drone.py
```diff
@@ -0,0 +1,502 @@
...
+    agent = Agent(
+        system_prompt="""You are controlling a DJI drone with MAVLink interface.
+        You have access to drone control skills you are already flying so only run move_twist, set_mode, and fly_to.
+        When the user gives commands, use the appropriate skills to control the drone.
+        Always confirm actions and report results. Send fly_to commands only at above 200 meters altitude to be safe.
+        Here are some GPS locations to remember
+        6th and Natoma intersection: 37.78019978319006, -122.40770815020853,
+        454 Natoma (Office): 37.780967465525244, -122.40688342010769
+        5th and mission intersection: 37.782598539339695, -122.40649441875473
+        6th and mission intersection: 37.781007204789354, -122.40868447123661""",
```
paul-nechifor, line 475: These are just for testing, right? They should probably be removed.

dimos/robot/drone/drone_tracking_module.py
```diff
@@ -0,0 +1,399 @@
...
+            self._tracking_active = True
+
+            # Start tracking in thread
+            self._tracking_thread = threading.Thread(
+                target=self._visual_servoing_loop, args=(tracker, duration), daemon=True
+            )
+            self._tracking_thread.start()
+
+            # Wait for tracking to complete
+            self._tracking_thread.join(timeout=duration + 1)
```
paul-nechifor: This hangs the RPC call. The default `duration` is 120 which is also the RPC call timeout. 120+1 means all default calls to `track_objects` will result in RPC timeouts.

I think you have to change the caller to poll for the result.

There are multiple places where we need to make a long call to an RPC method. @leshy , what do you think about modifying the system to add a way to have async responses to RPC calls. I.e., the initial call ends immediately, and the callee responds back to the caller when it's ready.

dimos/robot/drone/drone.py
spomichter: yeah true

dimos/robot/drone/camera_module.py
spomichter: removed

dimos/robot/drone/drone_tracking_module.py
spomichter: Yeah good idea here. Perhaps some multiprocess / thread handler where we can spawn long processes in RPC calls cleanly

spomichter: For now removed blocking

PR #648, opened 2025-09-30
Navspec

leshy:
Nav spec proposal, not super set on naming so feel free to adjust. This would not require a ros bridge but does comms to ros internally, like other drivers communicate to external things

Is posestamped the most correct message type for nav? since current ros nav uses something else

### DIFF
--- a/dimos/navigation/test_rosnav.py
+++ b/dimos/navigation/test_rosnav.py
@@ -1,0 +1,37 @@
+# Copyright 2025 Dimensional Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Protocol
+
+from dimos.mapping.spec import Global3DMapSpec
+from dimos.navigation.rosnav import ROSNav
+from dimos.navigation.spec import NavSpec
+from dimos.perception.spec import PointcloudPerception
+
+
+class RosNavSpec(NavSpec, PointcloudPerception, Global3DMapSpec, Protocol):
+    """Combined protocol for navigation components."""
+
+    pass
+
+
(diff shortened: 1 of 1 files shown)


dimos/mapping/spec.py
```diff
@@ -0,0 +1,31 @@
...
+
+from abc import ABC
+
+from dimos.core import Out
+from dimos.msgs.nav_msgs import OccupancyGrid
+from dimos.msgs.sensor_msgs import PointCloud2
+
+
+class Global3DMapSpec(ABC):
+    global_pointcloud: Out[PointCloud2] = None
```
paul-nechifor: I've seen this a lot, but isn't the actual type `Optional[Out[PointCloud2]]` since you're setting it to `None`?

dimos/navigation/spec.py
```diff
@@ -0,0 +1,32 @@
...
+    goal_req: In[PoseStamped] = None  # type: ignore
+    goal_active: Out[PoseStamped] = None  # type: ignore
+    path_active: Out[Path] = None  # type: ignore
+    ctrl: Out[Twist] = None  # type: ignore
+
+    # identity quaternion (Quaternion(0,0,0,1)) represents "no rotation requested"
+    def goto(self, target: PoseStamped) -> None:
+        pass
+
+    def stop(self) -> None:
```
paul-nechifor: I wouldn't use `stop` because it conflicts with existing modules where `stop` means shutdown the module.

I'd use `navigate_to_target` and `stop_navigating`

leshy: true!

dimos/navigation/rosnav.py
```diff
@@ -0,0 +1,41 @@
...
+    goal_req: In[PoseStamped] = None  # type: ignore
+    goal_active: Out[PoseStamped] = None  # type: ignore
+    path_active: Out[Path] = None  # type: ignore
+
+    ctrl: Out[Twist] = None  # type: ignore
+
+    pointcloud: Out[PointCloud2] = None  # type: ignore
+    global_pointcloud: Out[PointCloud2] = None  # type: ignore
+
+    config: Config
```
paul-nechifor: `self.config` already exists in `Configurable` in `Module`

dimos/mapping/spec.py
leshy: issue is it's not optional, class needs to know this is always available on .self

These are auto-created by the module init, but the way the class self-inspects, these things need to have a value assigned (so None in this case) - this is all fixable by better self inspection

wouldn't be surprised if we rewrite Module and Stream now (without changing the API actual modules use)

PR #649, opened 2025-10-01
Fix leaks

paul-nechifor:
* Fix all thread leaks.
* Fail tests which leak threads.

### DIFF
--- a/dimos/perception/detection2d/yolo_2d_det.py
+++ b/dimos/perception/detection2d/yolo_2d_det.py
@@ -26,7 +26,6 @@ from dimos.perception.detection2d.utils import (
 from dimos.utils.data import get_data
 from dimos.utils.gpu_utils import is_cuda_available
 from dimos.utils.logging_config import setup_logger
-from dimos.utils.path_utils import get_project_root
 
 logger = setup_logger("dimos.perception.detection2d.yolo_2d_det")
 
@@ -104,7 +103,7 @@ class Yolo2DDetector:
         """
         return plot_results(image, bboxes, track_ids, class_ids, confidences, names)
 
-    def cleanup(self):
+    def stop(self):
         """
         Clean up resources used by the detector, including tracker threads.
         """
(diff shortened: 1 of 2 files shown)


dimos/perception/detection2d/yolo_2d_det.py
```diff
@@ -104,6 +104,20 @@ def visualize_results(self, image, bboxes, track_ids, class_ids, confidences, na
         """
         return plot_results(image, bboxes, track_ids, class_ids, confidences, names)
 
+    def cleanup(self):
```
leshy: should this be called `stop()` for convention reasons?

dimos/protocol/skill/skill.py
```diff
@@ -165,9 +166,22 @@ def stop(self):
             self._skill_thread_pool.shutdown(wait=True)
             self._skill_thread_pool = None
 
-        if hasattr(self, "_close_rpc"):
+        # If this container is also a Module, close the module properly
+        if hasattr(self, "_close_module"):
+            self._close_module()
+        elif hasattr(self, "_close_rpc"):
```
leshy, line 172: I'm sure there is a reason but just to understand, shouldn't we just check if super has stop()? and it's stop() responsibility to have underlying knowledge about rpc etc

dimos/perception/detection2d/yolo_2d_det.py
paul-nechifor: Yes, it should be stop, I used cleanup because i thought it was a module and I wanted to avoid conflicts.

dimos/protocol/skill/skill.py
paul-nechifor: The weird hasattr checks are because of the __init__ calls. SkillCoordinator is a Module (and there for a SKillContainer), but its __init__ only calls SkillContainer.__init__ and not Module.__init__ so the things like `self.rpc` are never initialized in Module. That's why I'm checking for everything, just to be sure.

A later fix would be to improve the inheritance and `__init__` methods.

dimos/perception/detection2d/yolo_2d_det.py
paul-nechifor: Fixed.

PR #654, opened 2025-10-03
Modules navigate object bbox

spomichter:
Changes:
- Created BBox navigator to take in Detection2D type and output goal pose
- Extracted from object tracker 2d and 3d object trackers. 2d tested. 3d not yet tested. 
- Remove navigate_to_object method that should not be in the unitree runfile 

To run: 
`python dimos/robot/unitree_webrtc/run_agents2.py`

To give agent input: 
`human-cli`

### DIFF
--- a/dimos/agents2/skills/test_navigation.py
+++ b/dimos/agents2/skills/test_navigation.py
@@ -36,7 +36,20 @@ def test_take_a_look_around(fake_robot, create_navigation_agent, mocker):
 
 
 def test_go_to_object(fake_robot, create_navigation_agent, mocker):
-    fake_robot.navigate_to_object.return_value = True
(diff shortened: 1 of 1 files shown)


dimos/navigation/bbox_navigation.py
alexlin2, line 1: This just going towards the target for self.goal_distance m. It is not using any sort of distance estimate of the target, this is a todo right?

spomichter: No we need clean modules that are composable. This only vectors toward a bounding box

spomichter: The distance estimation is handled on @leshy side since he has ground truth object locations via his localization

PR #668, opened 2025-10-08
Detection 2d bbox navigator

spomichter:


dimos/perception/detection/type/detection2d/bbox.py
```diff
@@ -314,7 +314,7 @@ def to_ros_detection2d(self) -> ROSDetection2D:
                     )
                 )
             ],
-            id=str(self.track_id),
```
spomichter, line 407: @leshy Lmk if this will break anything its needed so we can access the object names easily via the .id attribute

leshy: it's fine for this yes, but generally as mentioned in person, for inter module-comms about detections we can also use pLCM and our own types, with optional conversion to ros and emission if needed for some third party integration, given stuff like class_id, track_id, confidence, person keypoints, segmentations etc are missing from ros types and are communicated in non standard ways anyway

paul-nechifor:
Closed because it's old and it's before the rebase.

PR #670, opened 2025-10-09
Migrate from pip to uv

naveenkul:
Installation time with pip: ~4m 02s
Installation time with uv: ~2m 30s

1.6x speedup


requirements.txt
```diff
@@ -1,96 +0,0 @@
-opencv-python
-python-dotenv
-openai
-anthropic>=0.19.0
-cerebras-cloud-sdk
```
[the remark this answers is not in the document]
naveenkul: 👍

pyproject.toml
```diff
@@ -71,11 +71,16 @@ dependencies = [
...
     "wasmtime",
-    
+    "rich>=14.0.0",
+
     # Audio
     "openai-whisper",
     "soundfile",
     
+    # PyTorch (version constraints from base-requirements.txt)
+    "torch==2.0.1",
```
spomichter, line 81: No we cannot lock torch versions

README.md
```diff
@@ -59,39 +59,41 @@ We are shipping a first look at the DIMOS x Unitree Go2 integration, allowing fo
 Tested on Ubuntu 22.04/24.04
 
 ```bash
-sudo apt install python3-venv
```
spomichter, line 62: Why was this removed

README.md
```diff
@@ -59,39 +59,41 @@ We are shipping a first look at the DIMOS x Unitree Go2 integration, allowing fo
...
-pip install torch==2.0.1 torchvision torchaudio --index-url https://download.pytorch.org/whl/cu118
-```
 
-#### Install dependencies
-```bash
-# CPU only (reccomended to attempt first)
-pip install -e .[cpu,dev]
+# Install dependencies
+# CPU only (recommended to attempt first)
+uv sync --extra dev --extra cpu
```
spomichter, line 82: Idk this syntax i usually do `uv pip install -e .[cpu]` maybe its the same

README.md
naveenkul: redundant. `uv venv --python 3.10` replaces it.

README.md
naveenkul: uv sync and uv add are more native options from uv. uv sync installs only the subset of dependencies that are missing or have changed, making it faster than checking everything like pip install does. 

uv pip install will install the packages in global env if local isn't activated, uv sync needs opt-in. given how specific some dependencies are, we wouldn't want to mess with the system level dependencies accidentally

pyproject.toml
naveenkul: following the same constraint as the requirements file. https://github.com/dimensionalOS/dimos/blob/753c56a18ac6d9a13bfa8198e20e77013391fd41/base-requirements.txt#L1  what's the best course of action here? I see a lot of dependencies in `dev` branch with version locks.

PR #673, opened 2025-10-10
alibaba env var fix

leshy:


.envrc
leshy, line 1: deleted this by mistake, just re-adding

spomichter: you deleted my dotenv init in the bottom of envrc also

dimos/perception/detection2d/module2D.py
```diff
@@ -40,7 +39,6 @@
 class Config:
     max_freq: float = 5  # hz
     detector: Optional[Callable[[Any], Detector]] = lambda: Yolo2DDetector()
-    vlmodel: VlModel = QwenVlModel
```
spomichter, line 43: one sec dont merge yet

PR #684, opened 2025-10-14
Add autoconnect back2

paul-nechifor:


dimos/agents2/skills/osm.py
```diff
@@ -12,46 +12,45 @@
...
 
-class OsmSkillContainer(SkillContainer, Resource):
-    _robot: Robot
-    _disposables: CompositeDisposable
+class OsmSkill(SkillModule):
     _latest_location: Optional[LatLon]
-    _position_stream: Observable[LatLon]
     _current_location_map: CurrentLocationMap
-    _started: bool
+    _skill_started: bool
```
leshy, line 36: (just leaving low effort notes with initial thoughts)

this feels a bit weird, why do I as a user need to manage the started state in a module that's meant to provide the skeleton I need?

dimos/core/README_BLUEPRINTS.md
```diff
@@ -0,0 +1,219 @@
...
+```python
+class ModuleA(Module):
+
+    @rpc
+    def get_time(self) -> str:
+        ...
+
+class ModuleB(Module):
+    @rpc # Note that it has to be an rpc method.
+    def set_ModuleA_get_time(self, rpc_call: RpcCall) -> None:
```
leshy, line 157: hm, would it be cleaner if we just expect the whole module type at __init__ ?  I can imagine having like 5 or more of those, gets goofy? I think you had an issue with passing full modules right?

paul-nechifor: I don't know about `__init__` because we should be able to start modules and introduce them to each other after. If we do it in `__init__` that requires a specific order of initialization.

My thinking is to do something like this:

```python
class SomeModule(Module):
    foreign_function_requests = { # from FFI, but could use a better name
        "start_navigation": "NavigationModule.start_navigation",
        "add_to_spatial_memory": "SpatialMemory.add_to_memory",
    }
```

and you can later call them as:

```
self.foreign_function['start_navigation']()
```

or maybe

```
self.foreign_function.start_navigation()
```

dimos/agents2/skills/osm.py
paul-nechifor: Yeah, good, point. I initially added because I couldn't see why it didn't work only to realize I forgot to call .start(). But since modules are started automatically this should be an issue now.

dimos/core/module_coordinator.py
```diff
@@ -12,23 +12,31 @@
...
 from dimos.core import DimosCluster, Module
+from dimos.core.global_config import GlobalConfig
 from dimos.core.resource import Resource
 
 T = TypeVar("T", bound="Module")
 
 
-class Dimos(Resource):
+class ModuleCoordinator(Resource):
     _client: Optional[DimosCluster] = None
```
leshy, line 27: should ModuleCoordinator actually be == DimosCluster - the same abstract type, with dask being an implementation?

dimos/navigation/global_planner/planner.py
```diff
@@ -216,3 +218,8 @@ def plan(self, goal: Pose) -> Optional[Path]:
 
         logger.warning("No path found to the goal.")
         return None
+
+
+astar_planner = partial(create_module_blueprint, AstarPlanner)
```
leshy: we have a lot of this boilerplate, and I worry that diff between
astar_planner and AstarPlanner is not clear when importing

should we create .blueprint classmethod on Module? 
which does this partial(create_module_blueprint) ?

dimos/core/module_coordinator.py
paul-nechifor: Could be. There's a lot of code in that `__init__.py` file which should be extracted out. The monkey patching in it is a bit unnecessary.

I was thinking of refactoring all of that but for now I just wrapped the DimosCluster in `ModuleCoordinator` to keep the scope smaller. But it's definitely something that should be improved.

dimos/navigation/global_planner/planner.py
paul-nechifor: Yep, good idea.

dimos/core/module_coordinator.py
leshy: yes ok, we refactor later, I agree

dimos/core/README_BLUEPRINTS.md
leshy: Just a note for later if needed - so we are aware of this - we can create RPCClients before actual modules they point towards are deployed

PR #685, opened 2025-10-14
Add metaquest module

naveenkul:
## Add VR Teleoperation Module for Quest 3/3S Controllers

Adds native Quest 3/3S controller support for robot teleoperation via WebXR.

**What's New:**
- **MetaQuestModule** - Full dimos integration with output streams for controller data
- **90Hz streaming** - Position, rotation, buttons, and thumbstick data at high frequency  
- **WebXR interface** - Browser-based, no APK required
- **Auto SSL setup** - Generates certificates for HTTPS (WebXR requirement)
- **Standalone server** - For testing and development

**Usage:**
```python
quest = dimos.deploy(MetaQuestModule, port=8881)
quest.start()
quest.controller_right.observable().subscribe(handle_controller)
```

**How it works:** Quest browser connects to HTTPS server, streams controller data via WebSocket to dimos streams. 

**Files:** `dimos/vr/` module, test server, SSL cert generator, WebXR interface


dimos/vr/modules/metaquest.py
```diff
@@ -0,0 +1,382 @@
...
+                logger.error(f"Failed to subscribe to camera stream: {e}")
+            finally:
+                # Cleanup
+                if camera_key in self._camera_disposables:
+                    self._camera_disposables[camera_key].dispose()
+                    del self._camera_disposables[camera_key]
+
+        return generate
+
+    def _close_module(self):
```
paul-nechifor: I added ModuleBase._close_module to cleanup things before we had start/stop. It's now called in ModuleBase.stop so you don't need to call _close_module anymore.

I think you should move this code to `def stop()` above and also add `super().stop()` at the end (to clean up all the module stuff, including _close_module).

dimos/vr/run_vr_server.py
```diff
@@ -0,0 +1,175 @@
...
+import sys
+from pathlib import Path
+
+import uvicorn
+from fastapi import FastAPI, WebSocket, WebSocketDisconnect
+from fastapi.responses import HTMLResponse
+from fastapi.staticfiles import StaticFiles
+from pydantic import ValidationError
+
+sys.path.insert(0, str(Path(__file__).parent.parent.parent))
```
paul-nechifor: What is this for?

dimos/vr/modules/metaquest.py
```diff
@@ -0,0 +1,382 @@
...
+            self._running = False
+            logger.error(f"Failed to start VR server: {e}", exc_info=True)
+            raise
+
+    @rpc
+    def stop(self):
+        """Stop the VR teleoperation server."""
+        if not self._running:
+            logger.info("VR server not running")
+            return
```
paul-nechifor: If there's an exception in the thread `self._running` gets set to false. That means that `module.stop()` cannot run fully because of the `if not self._running` condition.

I think you need to replace `self._running = False` with `self.stop()` in these two `try-except`s

dimos/vr/modules/metaquest.py
```diff
@@ -0,0 +1,382 @@
...
+            "errors": self.stats["errors"],
+            "last_frame_timestamp": self.stats["last_frame"].timestamp
+            if self.stats["last_frame"]
+            else None,
+            "host": self.host,
+            "port": self.port,
+        }
+
+    @rpc
+    def generate_certificate(self, cert_dir: str = "dimos/vr/certificates"):
```
paul-nechifor: I think `dimos` is for source code. You should probably place certificates in `assets/vr/certificates`. You can get it with:

```python
from dimos.constants import DIMOS_PROJECT_ROOT
_CERTIFICATES_DIR = DIMOS_PROJECT_ROOT / "assets" / "vr" / "certificates"
_CERTIFICATES_DIR.mkdir(parents=True, exist_ok=True)
```

dimos/vr/modules/metaquest.py
naveenkul: PoseStamped added and moved camera cleanup to under stop()

dimos/vr/run_vr_server.py
naveenkul: Removed.

dimos/vr/modules/metaquest.py
naveenkul: fixed now. using self.stop()

dimos/vr/modules/metaquest.py
naveenkul: switched to using constants

PR #692, opened 2025-10-17
Ivan g1

leshy:
### detections
- bugfixes on detection stuff for precise bounding boxes, independant human skeleton tracking and static object tracking

### unitree general rewrite
- moved my stuff from unitree_webrtc/ to just unitree/, removed old unitree/ dir completely
- rewrote the structure, simplified minimal run files

### unitree go2 rewrite
- lots of nuts things, fixed, system load down from 13 to 3 now for me, runs very well

### modular run files, no robot deps
- more to do here, very quick minimal first pass

### brutalized nav skills
- I kinda beat this into submission to not depend on robot, it's not great, I'd consider it needing a full revisit and a second PR, IMO lots has changed and it's ok to rewrite

### rosnav
- throttling of lidar frames (was high frequency, high bandwidth and had issues with remote foxglove, possibly can bump up once images are jpeg if we need to
- standard configuration 
- removed boolean topics to conform to spec (can discuss)
- added some basic skills, passive stream for global position, relative movement, and absolute movement skills

### testing
```sh
cd dimos/robot/unitree 
python run.py --ip mujoco go2/go2.py
python run.py --ip mujoco g1/g1detector.py # or any other
```

### DIFF
--- a/dimos/agents/agent.py
+++ b/dimos/agents/agent.py
@@ -80,7 +80,7 @@ class Agent:
         agent_type: str = "Base",
         agent_memory: AbstractAgentSemanticMemory | None = None,
         pool_scheduler: ThreadPoolScheduler | None = None,
-    ) -> None:
+    ):
         """
         Initializes a new instance of the Agent.
 
@@ -97,7 +97,7 @@ class Agent:
         self.disposables = CompositeDisposable()
         self.pool_scheduler = pool_scheduler if pool_scheduler else get_scheduler()
 
-    def dispose_all(self) -> None:
+    def dispose_all(self):
         """Disposes of all active subscriptions managed by this agent."""
         if self.disposables:
             self.disposables.dispose()
@@ -157,7 +157,7 @@ class LLMAgent(Agent):
         input_query_stream: Observable | None = None,
         input_data_stream: Observable | None = None,
         input_video_stream: Observable | None = None,
-    ) -> None:
+    ):
         """
         Initializes a new instance of the LLMAgent.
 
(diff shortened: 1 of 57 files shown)


dimos/navigation/rosnav.py
alexlin2, line 1: I need odom and cancel goal, you removing the cancel goal topic will not stop the navigation properly on the ROS side, I don't want to reiterate this over and over again. Odometry topic is VERY important to any current existing robotics developers, can you either add it back to ROSNav or add it to another module.

leshy: you have a stop_navigation rpc call, if driver isn't doing something it should when calling `stop_navigation` that's a problem to fix

pose odometry is a subset of tf - I'm the one reiterating this over and over again. 

Actual odometry message we can discuss re-adding, I can imagine this being a type of message a navigation system uses while everyone downstream works with the best fused guess available. - actual pose, will you use covariance for anything that's not internal to navigation?

alexlin2: All of my nodes subscribes to an Odom topic, Odom does not just contain the pose of the robot, it is a state estimation that includes position, velocity, uncertainty, tf and Odom are just not the same. I need a Odom message in the system and it is non-negotiable. Ask any robotics engineer how important Odom is, it is like the single most used topic in any mobile robot. If people don't find Odom and are forced to use tf lookups to get position, then they won't take our framework seriously. There is no such thing as Odom is a subset of tf, they are wildly different concepts.

PR #693, opened 2025-10-21
Add transport which encodes images as JPEG to improve performance.

paul-nechifor:
## Usage

I've changed dimos-lcm, so you need to install it again.

```bash
pip uninstall dimos-lcm
pip install "dimos-lcm @ git+https://github.com/dimensionalOS/dimos-lcm.git@3aeb724863144a8ba6cf72c9f42761d1007deda4"
```

I've added a new blueprint (`unitree-go2-jpeglcm`) which uses the new transport `JpegLcmTransport`.

Check that the sim works first:

```bash
dimos-robot --use-simulation run unitree-go2-jpeglcm
```

If so, use the real robot:

```bash
dimos-robot --robot-ip=... run unitree-go2-jpeglcm
```

For JPEG SHM use:

```bash
dimos-robot run unitree-go2-jpegshm
```

## Troubleshooting

If the image is a red X, dimos-lcm hasn't been upgraded. Make sure to wipe it and install it again.

## Additional testing

You can compare the compressed/uncompressed versions with demo encoding.

To run the uncompressed version:

```bash
python dimos/utils/demo_image_encoding.py
```

With JPEG compression:

```bash
python dimos/utils/demo_image_encoding.py --use-jpeg
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -107,7 +107,7 @@ dependencies = [
     "dask[complete]==2025.5.1",
 
     # LCM / DimOS utilities
-    "dimos-lcm @ git+https://github.com/dimensionalOS/dimos-lcm.git@bceb618ed431562a93f3d806c79656f85695a15d",
+    "dimos-lcm @ git+https://github.com/dimensionalOS/dimos-lcm.git@3aeb724863144a8ba6cf72c9f42761d1007deda4",
 
     # CLI
     "pydantic-settings>=2.11.0,<3",


dimos/core/transport.py
```diff
@@ -88,7 +88,8 @@ class LCMTransport(PubSubTransport[T]):
 
     def __init__(self, topic: str, type: type, **kwargs):
         super().__init__(LCMTopic(topic, type))
-        self.lcm = LCM(**kwargs)
+        if not hasattr(self, "lcm"):
```
spomichter, line 92: What is this conditional check

paul-nechifor: This is so I can override `self.lcm` in the `JpegLcmTransport` subclass (to be `JpegLCM()` there).

The conditional only initializes `self.lcm` if it's not already set in a subclass.

PR #702, opened 2025-10-27
Realtime XArm driver for joint positiona and velocitt control

mustafab0:
This PR does the following:

1. Key Features

- **100Hz control loop** for real-time position/velocity control
- **LCM pub/sub** for distributed system integration
- **RPC methods** for direct hardware control
- **Position mode** (radians) and **velocity mode** (deg/s)
- **Component-based API**: motion, kinematics, system, gripper control

2. The XArm driver has the following: Topics
Subscribed:
- `/xarm/joint_position_command` - JointCommand (positions in radians)
- `/xarm/joint_velocity_command` - JointCommand (velocities in deg/s)

Published:
- `/xarm/joint_states` - JointState (100Hz)
- `/xarm/robot_state` - RobotState (10Hz)
- `/xarm/ft_ext`, `/xarm/ft_raw` - WrenchStamped (force/torque)

3. **Testing**

 With Mock Hardware (No Physical Robot)

```
bash
# Unit tests with mocked xArm hardware
python tests/test_xarm_rt_driver.p
```

### DIFF
--- a/dimos/hardware/manipulators/xarm/xarm_driver.py
+++ b/dimos/hardware/manipulators/xarm/xarm_driver.py
@@ -823,16 +823,28 @@ class XArmDriver(
                     # Check if data is nested (list of lists) or flat
                     if isinstance(data[0], (list, tuple)):
                         # Nested: [[positions], [velocities], [efforts]]
+                        # NOTE: SDK may return extra joints (e.g., gripper) - truncate to num_joints
                         position = (
-                            list(data[0]) if len(data) > 0 else [0.0] * self.config.num_joints
+                            list(data[0])[: self.config.num_joints]
+                            if len(data) > 0
+                            else [0.0] * self.config.num_joints
                         )
-                        velocity = list(data[1]) if len(data) > 1 else [0.0] * len(position)
-                        effort = list(data[2]) if len(data) > 2 else [0.0] * len(position)
+                        velocity = (
+                            list(data[1])[: self.config.num_joints]
+                            if len(data) > 1
+                            else [0.0] * self.config.num_joints
+                        )
+                        effort = (
+                            list(data[2])[: self.config.num_joints]
+                            if len(data) > 2
+                            else [0.0] * self.config.num_joints
+                        )
                     else:
                         # Flat: just positions
-                        position = list(data)
-                        velocity = [0.0] * len(position)
-                        effort = [0.0] * len(position)
+                        # NOTE: SDK may return extra joints (e.g., gripper) - truncate to num_joints
+                        position = list(data)[: self.config.num_joints]
+                        velocity = [0.0] * self.config.num_joints
+                        effort = [0.0] * self.config.num_joints
                 else:
                     # Older firmware: only get_servo_angle available
                     code, position = self.arm.get_servo_angle(is_radian=self.config.is_radian)
@@ -840,16 +852,20 @@ class XArmDriver(
                         logger.warning(f"get_servo_angle failed with code: {code}")
                         continue
 
+                    # NOTE: SDK may return extra joints (e.g., gripper) - truncate to num_joints
+                    position = position[: self.config.num_joints]
+
                     # Calculate velocity from position difference
-                    velocity = [0.0] * len(position)
+                    velocity = [0.0] * self.config.num_joints
                     if initialized:
                         dt = curr_time - prev_time
                         if dt > 0:
                             velocity = [
-                                (position[i] - prev_position[i]) / dt for i in range(len(position))
+                                (position[i] - prev_position[i]) / dt
+                                for i in range(self.config.num_joints)
                             ]
 
-                    effort = [0.0] * len(position)
+                    effort = [0.0] * self.config.num_joints
 
                 # Update joint state message
                 self._joint_state_msg.ts = curr_time


dimos/hardware/manipulators/xarm/xarm_driver.py
```diff
@@ -0,0 +1,1181 @@
...
+                # Send command if available
+                if joint_cmd is not None and len(joint_cmd) == self.config.num_joints:
+                    code = None  # Initialize code variable
+
+                    if self.config.velocity_control:
+                        # Velocity control mode (mode 4)
+                        # NOTE: velocities are in degrees/second, not radians!
+                        # Always send velocity commands (velocity changes frequently)
+                        code = self.arm.vc_set_joint_velocity(
+                            joint_cmd, False, self.config.velocity_duration
```
luqiang21, line 970: Bug here: `joint_cmd` is in radian, but this call expects degrees.

dimos/hardware/manipulators/xarm/xarm_driver.py
```diff
@@ -0,0 +1,1181 @@
...
+        self._last_cmd_time: float = 0.0  # Timestamp of last command received (for timeout)
+
+        # Thread management
+        self._running = False
+        self._state_thread: Optional[threading.Thread] = None  # Joint state publishing
+        self._control_thread: Optional[threading.Thread] = None  # Command sending
+        self._stop_event = threading.Event()
+
+        # Subscription management
+        self._disposables = CompositeDisposable()
```
luqiang21: You should call this line in `start()` not here. Creating here will not allow the driver to receive command after a stop/start cycle. 

Reason: `stop()` disposed `CompositeDisposable()`, when you call another `start()` , the new subscription will be added to an already-disposed `CompositeDisposable()`.

dimos/hardware/manipulators/xarm/sample_trajectory_generator.py
```diff
@@ -0,0 +1,565 @@
...
+                f"to {math.degrees(self._trajectory_end_positions[joint_index]):.2f}° "
+                f"over {duration}s"
+            )
+
+            return (
+                f"Started moving joint {joint_index + 1} by {delta_degrees:.1f}° over {duration}s"
+            )
+
+    @rpc
+    def move_joint_velocity(self, joint_index: int, velocity_deg_s: float, duration: float) -> str:
```
luqiang21, line 261: here `velocity_deg_s` vis in degree, and `_publish_velocity_command()` is publishing it in degree as well. But in the driver, it is accepting radians, will there be a problem?

dimos/hardware/manipulators/xarm/xarm_driver.py
```diff
@@ -0,0 +1,1181 @@
...
+                        zero_vel = [0.0] * self.config.num_joints
+                        self.arm.vc_set_joint_velocity(
+                            zero_vel, True, self.config.velocity_duration
+                        )
+                    continue
+                else:
+                    timeout_logged = False
+
+                # Send command if available
+                if joint_cmd is not None and len(joint_cmd) == self.config.num_joints:
```
luqiang21, line 962: You should check readiness of the robot before sending command, otherwise commands may fail silently or cause error accumulation.

dimos/hardware/manipulators/xarm/xarm_driver.py
```diff
@@ -0,0 +1,1181 @@
...
+            # Verify recovery
+            logger.info(
+                f"  Recovery complete: state={self.curr_state}, mode={self.curr_mode}, error={self.curr_err}"
+            )
+            logger.info(f"[{self.config.ip_address}] ✓ Error recovery successful!")
+
+        except Exception as e:
+            logger.error(f"[{self.config.ip_address}] Error recovery failed: {e}")
+
+    def _xarm_is_ready_write(self) -> bool:
```
luqiang21, line 516: Variables `_last_state`, `_last_mode`, `_last_not_ready` are used but may not be initialized before first use
They are only set when conditions are met, but if `_xarm_is_ready_write()` is called before any state changes, these variables don't exist

dimos/hardware/manipulators/xarm/xarm_driver.py
```diff
@@ -0,0 +1,1181 @@
...
+                        logger.warning(
+                            f"Command timeout: no commands received for {time_since_last_cmd:.2f}s. "
+                            f"Stopping robot."
+                        )
+                        timeout_logged = True
+                    # Send zero velocity to stop
+                    if self.config.velocity_control:
+                        zero_vel = [0.0] * self.config.num_joints
+                        self.arm.vc_set_joint_velocity(
+                            zero_vel, True, self.config.velocity_duration
```
luqiang21, line 955: When command timeout occurs, zero velocity is sent with is_radian=True (line 944)
But the control loop uses is_radian=False for velocity commands (line 959)

You should make them consistent by changing this like to False as well

dimos/hardware/manipulators/xarm/xarm_driver.py
```diff
@@ -0,0 +1,1181 @@
...
+            self.arm.clean_error()
+            self.arm.clean_warn()
+
+            # Enable motion
+            self.arm.motion_enable(enable=True)
+
+            # Enable servo mode if configured
+            if self.config.enable_on_start:
+                logger.info("Enabling servo mode (mode=1)...")
+                code = self.arm.set_mode(1)  # Servo mode
```
luqiang21, line 716: Velocity Mode Never Enabled

Problem:
  - _initialize_arm() always sets servo mode (set_mode(1)) and expects mode 1, even when self.config.velocity_control is True.
  - The control loop later sends vc_set_joint_velocity commands assuming mode 4, so the controller rejects every command (code 9).

  Impact:

  - Critical: velocity-control deployments can never leave servo mode, so no velocities are accepted.

Fix:
- Make _initialize_arm() check velocity_control like _attempt_error_recovery() does

tests/test_xarm_rt_driver.py
```diff
@@ -0,0 +1,385 @@
...
+    mock.vc_set_joint_velocity.return_value = 0
+
+    # Force/torque sensor data
+    mock.ft_ext_force = [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
+    mock.ft_raw_force = [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
+
+    return mock
+
+
+def test_basic_connection():
```
paul-nechifor, line 96: Shouldn't these be pytest tests? If they have to be executed manually, they are bound to break.

dimos/hardware/manipulators/xarm/test_xarm_driver.py
```diff
@@ -0,0 +1,562 @@
...
+    logger.info("Starting driver...")
+    driver.start()
+
+    # Wait for initialization
+    time.sleep(2.0)
+
+    # Check connection via RPC
+    logger.info("Checking connection via RPC...")
+    code, version = driver.get_version()
+    if code == 0:
```
paul-nechifor: This should be:

```python
assert code == 0, "Failed to get firmware version"
```

dimos/hardware/manipulators/xarm/test_xarm_driver.py
```diff
@@ -0,0 +1,562 @@
...
+        logger.info(f"✓ Firmware version: {version}")
+    else:
+        logger.error(f"✗ Failed to get firmware version: code={code}")
+        dimos.stop()
```
paul-nechifor: You can use a fixture like this and not have to manually clean up:

```python
@pytest.fixture
def dimos_cluster():
    dimos = core.start(1)
    try:
        yield dimos
    finally:
        dimos.stop()
```

dimos/hardware/manipulators/xarm/xarm_driver.py
paul-nechifor: self._disposables already exists on Module.

dimos/hardware/manipulators/xarm/xarm_driver.py
```diff
@@ -0,0 +1,1181 @@
...
+
+        Publishes robot state at robot_state_rate Hz (default 10Hz).
+        Uses state variables updated by _report_data_callback.
+        """
+        period = 1.0 / self.config.robot_state_rate
+        next_time = time.time()
+
+        logger.info(f"Robot state loop started at {self.config.robot_state_rate}Hz")
+
+        while self._running:
```
paul-nechifor: Do you need `self._running` if you have `self._stop_event`? Events are thread safe. This could just be

```python
while not self._stop_event.is_set()
```

dimos/hardware/manipulators/xarm/xarm_driver.py
mustafab0: @luqiang21 the xarm SDK uses radians when sending joint position commands,

But when sending velocity commands it expects inputs as deg/s

hence the difference here.

dimos/hardware/manipulators/xarm/sample_trajectory_generator.py
mustafab0: The driver accepts joint velocity in deg/s only

dimos/hardware/manipulators/xarm/xarm_driver.py
mustafab0: @luqiang21 in my testing i found that the velocity control and position control both work using the SERVO Mode (1) and they are not exclusive to their specific modes.

dimos/hardware/manipulators/xarm/test_xarm_driver.py
mustafab0: Fixed

tests/test_xarm_rt_driver.py
mustafab0: I tried building this with pytest last time and couldnt figure it out, so went with this basic thing. I have added a new pytest now.

dimos/hardware/manipulators/xarm/README.md
```diff
@@ -0,0 +1,149 @@
...
+**On boot** (Important)
+```bash
+sudo ifconfig lo multicast
+sudo route add -net 224.0.0.0 netmask 240.0.0.0 dev lo
+```
+
+**Option A: Command-line argument** (recommended)
+```bash
+python test_xarm_driver.py --ip 192.168.1.235
+python interactive_control.py --ip 192.168.1.235
```
paul-nechifor, line 18: It would be nice to have a blueprint for these commands. Picking up arguments from environment variables or CLI is supported.

dimos/hardware/manipulators/xarm/README.md
```diff
@@ -0,0 +1,149 @@
...
+### With Real Hardware
+
+**⚠️ Note:** Interactive control and hardware tests require a physical xArm connected to the network. Interactive control, and sample_trajectory_generator are part of test suite, and will be deprecated.
+
+**Using Alfred Embodiment:**
+
+To test with real hardware using the current Alfred embodiment:
+
+1. **Turn on the Flowbase** (xArm controller)
+2. **SSH into dimensional-cpu-2:**
```
paul-nechifor, line 126: This is a bit overly specific to us. We should probably have internal docs somewhere else.

dimos/hardware/manipulators/xarm/components/state_queries.py
```diff
@@ -0,0 +1,172 @@
...
+class StateQueryComponent:
+    """
+    Component providing state query RPC methods for XArmDriver.
+
+    This component assumes the parent class has:
+    - self.arm: XArmAPI instance
+    - self.config: XArmDriverConfig instance
+    - self._joint_state_lock: threading.Lock
+    - self._joint_states_: Optional[JointState]
+    - self._robot_state_: Optional[RobotState]
```
paul-nechifor, line 41: I'm a bit unsure about the architecture here. It seems like XArmDriver is a single class which had it's methods moved into five mixins: MotionControlComponent, StateQueryComponent, SystemControlComponent, KinematicsComponent, GripperControlComponent.

I don't think they are real components as they are not used independently in other places and the dependencies are described in comments like here.

Having a single class would make it a bit harder to read, but it would make it more robust as you can depend on mypy for typechecking instead of comments.

mustafab0: Thanks for taking a look!

The reason for making the five mixins was essentially to keep the main driver file simpler. The thought process behind this is as follows.

For most manipulation tasks to move the robot we are going to give it joint position commands or joint velocity commands from the main control loop thread. The mixins/components are essentially all the features that the xarm SDK exposes for the arm. Most of these will not be used in "planned tasks" but they are necessary, for example clear error etc.

I debated putting all the mixins in the main driver file but it got too long. 

But further more another motivation for breaking them into components is to strategically categories them such that it becomes a standard template for creating drivers for other arms as well. As piper arm SDK does not have all the features that xarm SDK has and vice versa, it is easy to make additional mixins or omit whatever isn't available, while the main driver's position and velocity control remains largely unchanged.

paul-nechifor: I completely agree with breaking large classes into reusable and interchangeable components, the problem is that mixins are not the best way to achieve that because they are not good at encapsulation and providing interfaces.

For example, instead of assuming that `self.arm` exists in the `MotionControlComponent` mixin, you can change it to an actual class that accepts `arm`


```python
class XArmDriver:
    def __init__(...):
        self.arm = ...
        self.motion_control = MotionControlComponent(self.arm)
```

Currently, MotionControlComponent reaches into the internals of XArmDriver, for example by setting `self._joint_cmd_`. That's why I'm saying XArmDriver is really a single class split into multiple files. For components to be reusable, they need to have an established interface. This can be verified statically so we don't end up altering one implementation without the others.

---

But if you want to keep mixins, you can change this comment to a Protocol.

Something like:

```python
class StateQueryHost(Protocol):
    arm: XArmAPI
    config: XArmDriverConfig
    _joint_state_lock: threading.Lock
    _joint_states_: JointState | None
    _robot_state_: RobotState | None
```

and then annotate the self arguments.

```python
class StateQueryComponent:
    def get_joint_state(self: StateQueryHost) -> JointState | None:
```

paul-nechifor: I'm not saying it should be done now, though. This PR is quite large and I don't want to block you. It would be nice to do it in a subsequent PR though. :)

tests/test_xarm_driver_unit.py
```diff
@@ -0,0 +1,344 @@
...
+    driver.start()
+    time.sleep(0.3)
+
+    code, position = driver.get_position()
+
+    assert code == 0, "get_position should return success code"
+    assert len(position) == 6, "Position should have 6 values [x,y,z,roll,pitch,yaw]"
+    assert position[0] == 200.0, "X position should match mock"
+
+    driver.stop()
```
paul-nechifor, line 200: Stop is already called by the driver fixture. I don't think you have to call it in every test.

tests/test_xarm_driver_unit.py
```diff
@@ -0,0 +1,344 @@
...
+        joint_states_received.append(msg)
+
+    # Try to subscribe (will fail without transport, but that's OK for unit test)
+    try:
+        driver.joint_state.subscribe(on_joint_state)
+    except (AttributeError, ValueError):
+        pass  # Expected - no transport in unit test
+
+    driver.start()
+    time.sleep(1.0)
```
paul-nechifor, line 270: Is there a way to avoid the sleep by waiting for the specific event to happen?

The worst thing about sleep is that it takes the same amount of time no matter how many cores you have. :D

dimos/hardware/manipulators/xarm/xarm_driver.py
luqiang21: > [@luqiang21](https://github.com/luqiang21) in my testing i found that the velocity control and position control both work using the SERVO Mode (1) and they are not exclusive to their specific modes.

Is it desired behavior?

tests/test_xarm_rt_driver.py
```diff
@@ -0,0 +1,385 @@
+# Copyright 2025 Dimensional Inc.
```
spomichter, line 1: This dir is deprecated so look at how unit tests implemented w pytset throughout the codebase.

PR #713, opened 2025-10-30
transport improvements

leshy:
- implements exception passing to RPC clients (if there is an exception on the server, it will be raised on the caller side)
- stress tests for all transport protocols (we expect 10k messages < sec, 100% reliability expected across all transports)
- stress tests for RPC (1k rpc calls < sec, 100% reliability expected) - SHM actually fails this - has 25% reliability on RPC tests due to how it handles message reception, this is an issue that means we can't use SHM across the board in dimos

- moves rpc call threadpool from LCMRPC to the rpc base so it's used cross-protocol
- shares a single cpp LCM instance/loop per process for all instantiated transports

- investigated kernel networking stats and LCM cpp implementation, optimized details around this to get LCM to perform much better

### DIFF
--- a/dimos/protocol/rpc/pubsubrpc.py
+++ b/dimos/protocol/rpc/pubsubrpc.py
@@ -19,7 +19,6 @@ from collections.abc import Callable
 from concurrent.futures import ThreadPoolExecutor
 import threading
 import time
-import traceback
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -63,7 +62,7 @@ class RPCRes(TypedDict, total=False):
 
 
 class PubSubRPCMixin(RPCSpec, PubSub[TopicT, MsgT], Generic[TopicT, MsgT]):
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
         # Thread pool for RPC handler execution (prevents deadlock in nested calls)
         self._call_thread_pool: ThreadPoolExecutor | None = None
@@ -200,7 +199,7 @@ class PubSubRPCMixin(RPCSpec, PubSub[TopicT, MsgT], Generic[TopicT, MsgT]):
         self.publish(topic_req, self._encodeRPCReq(req))
 
         # Return unsubscribe function that removes this callback from the dict
-        def unsubscribe_callback():
+        def unsubscribe_callback() -> None:
             with self._response_subs_lock:
                 if topic_res_key in self._response_subs:
                     _, callbacks_dict = self._response_subs[topic_res_key]
@@ -256,7 +255,7 @@ class PubSubRPCMixin(RPCSpec, PubSub[TopicT, MsgT], Generic[TopicT, MsgT]):
 
 
 class LCMRPC(PubSubRPCMixin, PickleLCM):
-    def __init__(self, **kwargs):
+    def __init__(self, **kwargs) -> None:
         # Need to ensure PickleLCM gets initialized properly
         # This is due to the diamond inheritance pattern with multiple base classes
         PickleLCM.__init__(self, **kwargs)
@@ -272,7 +271,7 @@ class LCMRPC(PubSubRPCMixin, PickleLCM):
 
 
 class ShmRPC(PubSubRPCMixin, PickleSharedMemory):
-    def __init__(self, prefer: str = "cpu", **kwargs):
+    def __init__(self, prefer: str = "cpu", **kwargs) -> None:
         # Need to ensure SharedMemory gets initialized properly
         # This is due to the diamond inheritance pattern with multiple base classes
         PickleSharedMemory.__init__(self, prefer=prefer, **kwargs)
--- a/dimos/protocol/rpc/rpc_utils.py
+++ b/dimos/protocol/rpc/rpc_utils.py
@@ -35,12 +35,12 @@ class RemoteError(Exception):
     Preserves the original exception type and full stack trace from the remote side.
     """
 
-    def __init__(self, type_name: str, type_module: str, args: tuple, traceback: str):
+    def __init__(self, type_name: str, type_module: str, args: tuple, traceback: str) -> None:
         super().__init__(*args if args else (f"Remote exception: {type_name}",))
         self.remote_type = f"{type_module}.{type_name}"
         self.remote_traceback = traceback
 
-    def __str__(self):
+    def __str__(self) -> str:
         base_msg = super().__str__()
         return (
             f"[Remote {self.remote_type}] {base_msg}\n\nRemote traceback:\n{self.remote_traceback}"
--- a/dimos/protocol/rpc/test_rpc_utils.py
+++ b/dimos/protocol/rpc/test_rpc_utils.py
@@ -14,8 +14,6 @@
 
 """Tests for RPC exception serialization utilities."""
 
-import pytest
-
 from dimos.protocol.rpc.rpc_utils import (
     RemoteError,
     deserialize_exception,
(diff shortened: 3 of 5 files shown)


dimos/protocol/pubsub/test_spec.py
```diff
@@ -263,3 +275,40 @@ async def consume_messages() -> None:
...
+            received_messages.append(message)
+            last_message_time[0] = time.time()
+
+        # Subscribe to the topic
+        x.subscribe(topic, callback)
+
+        # Publish 10000 messages
+        num_messages = 10000
+        for _ in range(num_messages):
+            x.publish(topic, values[0])
```
paul-nechifor, line 282: Just an idea: instead of sending the same vector, maybe send `Vector3(i, i, i)`, and assert that all 10000 numbers were received.

dimos/protocol/rpc/pubsubrpc.py
```diff
@@ -152,3 +266,39 @@ def _encodeRPCRes(self, res: RPCRes) -> dict:
...
+from dimos.protocol.pubsub.shmpubsub import PickleSharedMemory
+
+
+class LCMRPC(PassThroughPubSubRPC, PickleLCM):
+    def __init__(self, **kwargs):
+        # Need to ensure LCMPubSubBase gets initialized since it's not in the direct super() chain
+        # This is due to the diamond inheritance pattern with multiple base classes
+        PickleLCM.__init__(self, **kwargs)
+        # Initialize PubSubRPCMixin's thread pool
+        PubSubRPCMixin.__init__(self, **kwargs)
```
paul-nechifor: What's the issue with calling `super().__init__`? Only calling a few of the constructors is an indication that the inheritance graph could be too complex.

dimos/protocol/rpc/pubsubrpc.py
```diff
@@ -152,3 +266,39 @@ def _encodeRPCRes(self, res: RPCRes) -> dict:
...
+
+    def topicgen(self, name: str, req_or_res: bool) -> Topic:
+        from dimos.constants import LCM_MAX_CHANNEL_NAME_LENGTH
+        from dimos.utils.generic import short_id
```
paul-nechifor: Maybe move these (and other imports) to the top.

dimos/protocol/rpc/test_lcmrpc.py
```diff
@@ -1,45 +0,0 @@
...
-
-@pytest.fixture
-def lcmrpc() -> Generator[LCMRPC, None, None]:
-    ret = LCMRPC()
-    ret.start()
-    yield ret
-    ret.stop()
-
-
-def test_short_name(lcmrpc) -> None:
```
paul-nechifor, line 31: Not very important tests, but didn't hurt to leave them in. :P

dimos/protocol/service/lcmservice.py
```diff
@@ -223,57 +349,38 @@ def __str__(self) -> str:
...
-        self._call_thread_pool_lock = threading.RLock()
+        self._stopped = False
 
     def start(self) -> None:
-        # Reinitialize LCM if it's None (e.g., after unpickling)
-        if self.l is None:
-            if self.config.lcm:
-                self.l = self.config.lcm
-            else:
-                self.l = lcm.LCM(self.config.url) if self.config.url else lcm.LCM()
```
paul-nechifor, line 276: I think this breaks pickling. Example if you run `dimos-robot run unitree-go2-agentic`:

```
Pickle encoding error: cannot pickle 'LCM' object                                                                                                                                                                                                                                                                                                                            
Traceback (most recent call last):                                                                                                                                                                                                                                                                                                                                           
  File "/home/p/pro/dimensional/dimos/dimos/protocol/pubsub/spec.py", line 145, in encode                                                                                                                                                                                                                                                                                    
    return pickle.dumps(msg)                                                                                                                                                                                                                                                                                                                                                 
           ^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                                                                                                                 
TypeError: cannot pickle 'LCM' object                                                                                                                                                                                                                                                                                                                                        
Tried to pickle: {'name': 'LlmAgent/register_skills', 'args': ((<dimos.agents2.cli.human.HumanInput object at 0x7b9d9f1d0230>,), {'run_implicit_name': 'human'}), 'id': 1761861481.7190497}                                                                                                                                                                                  
2025-10-30 23:58:31,719 - /home/p/pro/dimensional/dimos/dimos/protocol/rpc/pubsubrpc.py - ERROR - Exception in RPC handler for HumanInput/set_LlmAgent_register_skills: RPC call to 'LlmAgent/register_skills' timed out after 30.0 seconds                                                                                                                                  
Traceback (most recent call last):                                                                                                                                                                                                                                                                                                                                           
  File "/home/p/pro/dimensional/dimos/dimos/protocol/rpc/pubsubrpc.py", line 233, in execute_and_respond                                                                                                                                                                                                                                                                     
    response = f(*args[0], **args[1])                                                                                                                                                                                                                                                                                                                                        
               ^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                                                                                                        
  File "/home/p/pro/dimensional/dimos/dimos/protocol/rpc/spec.py", line 94, in override_f                                                                                                                                                                                                                                                                                    
    return getattr(module, fname)(*args, **kwargs)                                                                                                                                                                                                                                                                                                                           
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                                                                                           
  File "/home/p/pro/dimensional/dimos/dimos/agents2/cli/human.py", line 52, in set_LlmAgent_register_skills                                                                                                                                                                                                                                                                  
    callable(self, run_implicit_name="human")                                                                                                                                                                                                                                                                                                                                
  File "/home/p/pro/dimensional/dimos/dimos/core/rpc_client.py", line 69, in __call__                                                                                                                                                                                                                                                                                        
    result, unsub_fn = self._rpc.call_sync(f"{self._remote_name}/{self._name}", (args, kwargs))                                                                                                                                                                                                                                                                              
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                                              
  File "/home/p/pro/dimensional/dimos/dimos/protocol/rpc/spec.py", line 57, in call_sync                                                                                                                                                                                                                                                                                     
    raise TimeoutError(f"RPC call to '{name}' timed out after {rpc_timeout} seconds")                                                                                                                                                                                                                                                                                        
TimeoutError: RPC call to 'LlmAgent/register_skills' timed out after 30.0 seconds 
```

leshy: ok I'll investigate this, ran replay and all was fine, so wasn't sure where to check more, it's weird we ever pickle transport objects, hosting objects should have custom picklers defined

dimos/protocol/rpc/pubsubrpc.py
leshy: good catch, and yes inheritance is too complex

dimos/protocol/pubsub/test_spec.py
leshy: yeah, could check a bit, issue is this is part of a grid test now, which tests universal pubsub, which means other transports might move only binary data other only json etc, so I need custom per transport generators, will see

dimos/protocol/rpc/test_lcmrpc.py
leshy: true, sorry I'll re-intro LCM specific tests

dimos/protocol/service/lcmservice.py
leshy: btw this means that something  important isn't tested since tests are happy

paul-nechifor: Ah, good point. The agentic tests I've written don't deploy the skill modules, so they're not testing the system as used in practice. I'll add a test for it.

paul-nechifor: @leshy Added this test: https://github.com/dimensionalOS/dimos/pull/716

leshy: Passed now after dev merge, are we good :D ?

PR #718, opened 2025-11-01
Remove unneeded code files

paul-nechifor:
* Remove entire files which should no longer be needed.
* In https://github.com/dimensionalOS/dimos/pull/725 I remove unused functions.

### DIFF
--- a/dimos/robot/unitree_webrtc/modular/__init__.py
+++ b/dimos/robot/unitree_webrtc/modular/__init__.py
@@ -1,2 +1,1 @@
 from dimos.robot.unitree_webrtc.modular.connection_module import deploy_connection
-from dimos.robot.unitree_webrtc.modular.navigation import deploy_navigation


dimos/agents/agent_ctransformers_gguf.py
```diff
@@ -1,215 +0,0 @@
-# Copyright 2025 Dimensional Inc.
```
spomichter, line 1: Ah shit we do need to re support the local GGUF implementations since people want to run local llms @paul-nechifor

paul-nechifor: Yep, but not on Agent1 code, right? Just agent2.

Langchain does support GGUF, but I haven't tried it. Should I add a blueprint using a local LLM?

paul-nechifor: @spomichter I've added https://github.com/dimensionalOS/dimos/pull/739 . It supports local GGUF files through ollama and it also has support for huggingface.

Can this old agent code be deleted now?

PR #720, opened 2025-11-01
Map evals and interpret map skill (WIP)

s-desh:
This PR is one of many for encoding maps for agents. (#804 )

Adds 
- Evals for point placement and map comprehension.
- Encodes costmap as RGB image and overlays robot pose.
- Uses this encoding in a skill that can pull maps and place points based on the query for navigation. (Incomplete)


dimos/msgs/nav_msgs/OccupancyGrid.py
```diff
@@ -272,6 +277,28 @@ def __repr__(self) -> str:
...
+        # convert to grayscale image:
+        # - unknown (-1) -> black (0)
+        # - free (0) -> white (255)
+        # - obstacles (1..100) -> gradient to black (100 -> 0)
+        image_arr = np.full(self.grid.shape, 0, dtype=np.uint8)
+
+        known_mask = self.grid != CostValues.UNKNOWN
+        if np.any(known_mask):
+            # map from [0..100] to [255..0]
+            image_arr[known_mask] = (255 - (self.grid[known_mask] * 255 // 100)).astype(np.uint8)
```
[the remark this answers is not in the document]
s-desh: works

dimos/robot/unitree_webrtc/unitree_skill_container.py
```diff
@@ -51,12 +56,21 @@ def __init__(self, robot: UnitreeGo2 | None = None) -> None:
...
+
         # Dynamically generate skills from UNITREE_WEBRTC_CONTROLS
         self._generate_unitree_skills()
 
     @rpc
     def start(self) -> None:
         super().start()
+        # subscribe to local costmap
+        self._local_costmap_transport = core.LCMTransport("/local_costmap", OccupancyGrid)
+        self._costmap_unsub = self._local_costmap_transport.subscribe(self._on_local_costmap)
```
s-desh: Is this the right way to subscribe to /local_costmap? 
`self._latest_local_costmap` is always None.

paul-nechifor: It should be like this.

```python
class UnitreeSkillContainer(SkillModule):
    local_costmap: In[OccupancyGrid] = None

    @rpc
    def start(self) -> None:
        self._disposables.add(self.local_costmap.subscribe(self._on_local_costmap))

    def _on_local_costmap(self, costmap: OccupancyGrid) -> None:
        self.latest_costmap_variable = costmap
```

dimos/agents2/temp/run_unitree_agents2.py
```diff
@@ -89,6 +89,7 @@ def setup_agent(self, skillcontainers, system_prompt: str) -> Agent:
             agent.register_skills(container)
 
         agent.run_implicit_skill("human")
+        agent.run_implicit_skill("get_map")
```
paul-nechifor: It's unclear which get_map should be called because you have two get_map skills.

But also, I wouldn't use this file, it's in `temp` and should be deleted.

I'll add below in another comment how to run an implicit skill now.

dimos/robot/unitree_webrtc/unitree_skill_container.py
```diff
@@ -105,6 +115,28 @@ def current_time(self):
             yield str(datetime.datetime.now())
             time.sleep(1)
 
+    @skill()
+    def get_map(self):
```
paul-nechifor: I don't think this skill has anything to do with Unitree. I'd create a separate skill class (which is also easier if you want an implicit skill, in order to register it).

```python
class AsciiMapSkill(Module):  # or a better name

    @rpc
    def start(self) -> None:
        super().start()

    @rpc
    def stop(self) -> None:
        super().stop()

    @skill(..., hide_skill=True) # hide_skill so it's never called by the agent
    def get_map(self):
        """docstring"""
        ...

    # this is how you get a reference to LlmAgent.register_skills
    # and request that `get_map` be called implicitly
    @rpc
    def set_LlmAgent_register_skills(self, callable: RpcCall) -> None:
        callable.set_rpc(self.rpc)
        callable(self, run_implicit_name="get_map")

ascii_map_skill = AsciiMapSkill.blueprint

__all__ = ["AsciiMapSkill", "ascii_map_skill"]


```

dimos/agents2/temp/run_unitree_agents2.py
s-desh: Using the one in unitree_skills, I'll it move to a skill class and use blueprints

dimos/msgs/nav_msgs/test_OccupancyGrid.py
```diff
@@ -321,6 +321,21 @@ def test_filter_below() -> None:
...
+
+    origin = Pose(1.0, 2.0, 0.0)
+    grid = OccupancyGrid(grid=data, resolution=0.05, origin=origin, frame_id="odom")
+    image = grid.grid_to_image()
+
+    assert image.data.shape == (20, 30)
+    # save image
+    # image.save("occupancy_grid_image_debug.png")
+
+    grid.agent_encode()
```
paul-nechifor: This should have an assert, no?

dimos/msgs/nav_msgs/test_OccupancyGrid.py
```diff
@@ -321,6 +321,21 @@ def test_filter_below() -> None:
...
+def test_grid_image_conversion() -> None:
+    data = np.zeros((20, 30), dtype=np.int8)
+    data[5:10, 10:20] = 100  # Add some obstacles
+    data[15:18, 5:8] = -1  # Add unknown area
+
+    origin = Pose(1.0, 2.0, 0.0)
+    grid = OccupancyGrid(grid=data, resolution=0.05, origin=origin, frame_id="odom")
+    image = grid.grid_to_image()
+
+    assert image.data.shape == (20, 30)
```
paul-nechifor: I'd assert the values too. You can make the array smaller and assert the actual values.

dimos/msgs/nav_msgs/OccupancyGrid.py
```diff
@@ -272,6 +284,93 @@ def __repr__(self) -> str:
             f"occupied={self.occupied_cells}, free={self.free_cells}, "
             f"unknown={self.unknown_cells})"
         )
+    
+    def grid_to_image(self) -> Image:
```
paul-nechifor: What is the context in which this would be used? It might be useful to use a colored one. This is what we use in the foxglove plugin (could be better):



With this grayscale image, both unknown and impassable are represented by black. We might want to let agent know the difference between the two.

dimos/msgs/nav_msgs/test_OccupancyGrid.py
s-desh: yeah, yet to update tests

dimos/msgs/nav_msgs/OccupancyGrid.py
s-desh: Later if we find images work better for larger maps / debugging, but I can get rid of it as not used right now. Kept it grayscale to keep the agent call light.

dimos/agents2/skills/interpret_map.py
```diff
@@ -0,0 +1,164 @@
...
+            "Prioritize selecting a goal position in free space (blue area) over exactly matching the description. \n"
+            "MAKE SURE there is a clear path from the robot's current position to the goal position without crossing any obstacles. \n"
+            "MAKE SURE the goal position is located in the blue area (free space) of the map and few pixels away from obstacles or objects. \n"
+            "Return ONLY a JSON object with this exact format:\n"
+            "{'point': [x, y]}\n"
+            f"where x,y are the pixel coordinates of the goal position in the image. \n"
+        )
+
+        response = query_single_frame(image, prompt)
+        x, y = parse_qwen_points_response(response)
```
paul-nechifor: query_single_frame is an agents1 function which is deprecated. The replacement is `QwenVlModel` used in agents2. You can parse the response with `extract_json_from_llm_response`. (I see `parse_qwen_points_response` is already defined in `dimos/skills/manipulation/pick_and_place.py`)

I think the query aspect can be abstracted a bit like it's done in `dimos/navigation/visual/query.py` or `dimos/mapping/osm/query.py`. It would allow us to pass a different VL model if we need to.

dimos/agents2/skills/interpret_map.py
```diff
@@ -0,0 +1,164 @@
...
+        if not costmap.is_free_space(x, y):
+            logger.warning(
+                f"Identified goal position ({x}, {y}) is not in free space, choosing nearest free space instead."
+            )
+            closest_free_point = costmap.get_closest_free_point(x, y)
+            if closest_free_point is not None:
+                x, y = closest_free_point
+
+        if x is not None and y is not None:
+            debug_image_with_identified_point(image, (x, y), filepath="./debug_goal_position.png")
```
paul-nechifor: I think you should put debugging code behind a env variable flag like `DEBUG_INTERPRET_MAP=true`

tests/mockdata/local_costmap.pkl
paul-nechifor, line 1: This is quite a big file. I think it should be in added to LFS.

You can move it to `data/mockdata_local_costmap.pkl`.

Compress it with:

```
cd data; tar -czf .lfs/mockdata_local_costmap.pkl.tar.gz mockdata_local_costmap.pkl
```

and then use it in the code with:

```
get_data('mockdata_local_costmap.pkl')
```

It automatically downloads the file from LFS and decompresses it into `data/mockdata_local_costmap.pkl`.

dimos/agents2/skills/interpret_map.py
s-desh: The results differ using QwenVLModel while my input remains same. `query_single_frame(image, prompt)` does better and identifies all 3 rooms while `vl_model.query(opencv_image, prompt)` identifies 2. 

What might be causing this? The query and frame are unchanged and I pass a numpy array for consistency in QwenVLModel.

paul-nechifor: Interesting... this might be a BGR issue. QwenVlModel.query accepts either Image or np.ndarray. But if you pass an ndarray, you have to use `image.to_bgr().to_opencv()`.

s-desh: Yeah, this works.

dimos/agents2/skills/interpret_map.py
```diff
@@ -85,7 +88,7 @@ def get_goal_position(self, description: str | None = None) -> Pose | None:
 
         image = costmap.grid_to_image(size=(1024, 1024), flip_vertical=True)
 
-        image = image.to_opencv()
+        image = image.to_bgr().to_opencv()
```
paul-nechifor: You can delete this line. `vl_model.query` accepts both Image and np.ndarray.

dimos/agents2/skills/test_interpret_map_skill.py
```diff
@@ -0,0 +1,267 @@
...
+    return prompt
+
+
+def interpretability_prompt(question: str) -> str:
+    prompt = (
+        "Look at this image carefully \n"
+        "it represents a 2D occupancy grid map where,\n"
+        " - blue area is free space, \n"
+        " - yellow area is unknown space, \n"
+        " - red (and its shades) areas are obstacles/walls, \n"
```
leshy, line 120: specific shades are part of a costmap for specific navigation algos, depend on the embodiment and can change wildly, they shouldn't be a part of agents understanding of the space

dimos/agents2/skills/test_interpret_map_skill.py
```diff
@@ -0,0 +1,267 @@
...
+    test_cases = [
+        (
+            "a clear area near the center of the map",
+            (
+                (costmap.info.width * 0.25, costmap.info.width * 0.75),
+                (costmap.info.height * 0.25, costmap.info.height * 0.75),
+            ),
+        ),
+        (
+            "an open space close to the bottom right corner",
```
leshy: can we ask more difficult questions from https://github.com/dimensionalOS/dimos/issues/783 ? 

You are giving a heavy hint of where to place a point and I have a feeling that I can have a high level of success in this test by guessing and not looking at an image, or looking at an extremely blurry image.

We really want to gauge here the abilities of different models, I assume that results of these will be success rates discovered through multiple queries per question per model we use

dimos/agents2/skills/test_interpret_map_skill.py
```diff
@@ -0,0 +1,267 @@
...
+            (
+                (costmap.info.width * 0.5, costmap.info.width * 1.0),
+                (costmap.info.height * 0.0, costmap.info.height * 0.5),
+            ),
+        ),
+    ]
+
+    for description, expected_range in test_cases:
```
leshy: we want an ability to easily specify large numbers of questions and ranges of acceptable answers so this needs an abstraction. as a user I'd like to specify something like

```python

Models: Tuple[VLModel] = ...

type Question = Tuple[Question, AcceptedBox]
type Questions = Sequence[Question]
```

And we want to be able to specify multiple models we'd like to test, could check grid/parametric testing as in https://github.com/dimensionalOS/dimos/blob/dev/dimos/protocol/pubsub/test_spec.py#L218 but not sure, you are free to approach by whatever seems best

This is our first true eval on a specific task, likely in the future we'll need a third party model eval framework

dimos/agents2/skills/test_interpret_map_skill.py
s-desh: This is an old test, will be removing it. Questions from #783 are tested below, stored in the test_map_interpretability.yaml.

dimos/agents2/skills/test_interpret_map_skill.py
```diff
@@ -0,0 +1,267 @@
...
+def load_test_cases(filepath: str):
+    import yaml
+
+    with open(filepath) as f:
+        data = yaml.safe_load(f)
+    return data
+
+
+@pytest.fixture
+def vl_model():
```
leshy, line 47: `class VlModel(ABC)` 

Is there a reason we are deferring spatial queries to a one shot VlModel?

since this interface has no memory, conversation history etc. 

In a normal operations I assume ideally we want agent_encode to be used with an agent `dimos.agents2.Agent` (and queried via agent.query()) which keeps the spatial understanding in the context of long term goals and conversation

We can implement VlModel as a corner case in which an agent controling the robot is not multimodal, or is bad at spatial stuff but good at speed let's say, but I assume in default ops it is a core Agent model itself that's interacting with the map?

Your skill can literally return the costmap object
costmap object can have agent_encode together with the textual description (I'm not 100% sure atm if agent_encode can return 2 messages, image and description though)

dimos/agents2/skills/test_interpret_map_skill.py
s-desh: Isn't the goal to test if our encoding of maps can be understood? And adding how we encode occupancy grids to images helps.

dimos/agents2/skills/interpret_map.py
```diff
@@ -0,0 +1,145 @@
...
+        self.vl_model = QwenVlModel()
+
+    @rpc
+    def stop(self) -> None:
+        super().stop()
+
+    def _on_local_costmap(self, costmap: OccupancyGrid) -> None:
+        self._latest_local_costmap = costmap
+
+    def _on_lidar(self, lidar: LidarMessage) -> None:
```
leshy: you shouldn't depend on lidar for localizing robot on the costmap (costmap can be generated by different sensors and algos)

you can get the robot pose any time by just running `self.tf.get("world","base_link")` in a module

dimos/agents2/skills/test_interpret_map_skill.py
s-desh: Getting the agent to send pixel coordinates, then reading and converting them to world frame is 2 skill calls, plus one more trigger navigation. And using VLModel, it's 2.

Also interpretability was not as good as calling VLModel. I'll try with the new encoding and see.

dimos/agents2/skills/test_interpret_map_skill.py
leshy: > Isn't the goal to test if our encoding of maps can be understood? And adding how we encode occupancy grids to images helps.

Yes I just mean should remove the gradients " - red (and its shades) areas are obstacles/walls" and rely only on walls/explored/free

This is because costs in costmaps (shades of red you mention to an agent) are specific to both goals and algos used and can change wildly from robot to robot and moment by moment, so are not stable or useful for an agent to answer "where the room on the right is"

a person follow has different cost map from a* global search, terrain height costs etc - https://chizaruu.github.io/roguebasin/dijkstra_maps_visualized

dimos/agents2/skills/test_interpret_map_skill.py
leshy: oh  see, I think you forgot to include this file

dimos/agents2/skills/test_interpret_map_skill.py
s-desh: I agree it should be parametric and it's done below in `test_point_placement` and `test_map_comprehension`. Will add support for multiple models and separate these tests from skill tests.

dimos/agents2/skills/test_interpret_map_skill.py
s-desh: it's in the lfs maps/ folder

dimos/agents2/skills/test_interpret_map_skill.py
leshy: > Getting the agent to send pixel coordinates, then reading and converting them to world frame is 2 skill calls, plus one more trigger navigation. And using VLModel, it's 2.

sry might be missing something

```
agent: give_me_map_toolcall
system: *map*
agent: tool_goto_map_pixels(x,y)
system: *navigating..."
```
- tool does the pixel->world_coordinates conversion?

also we have this concept of passive skills that append data to an agent implicitly (there is webcam_agent.py as an example with a camera image) and we experimented with permanent latest image frame available in the context for an agent running the robot. we can do the same for the map, so agent can always just say "goto_map_pixels" or whatever and robot starts walking

but! switching to an full on agent, passive skills etc might also be too complicated for this initial PR, we can keep VLM then explore further later, especially if one-shot VLM performs better etc, so your call how you want to approach this

s-desh: > Your skill can literally return the costmap object
costmap object can have agent_encode together with the textual description (I'm not 100% sure atm if agent_encode can return 2 messages, image and description though)

Yes, should send image + default description via agent_encode but also allow injecting text  from skills (in the same message, need to figure this out).

dimos/msgs/nav_msgs/OccupancyGrid.py
```diff
@@ -276,6 +292,254 @@ def __repr__(self) -> str:
...
+
+        if flip_vertical:
+            pixel_y = size[1] - pixel_y
+
+        grid_x = (pixel_x / size[0]) * self.width
+        grid_y = (pixel_y / size[1]) * self.height
+
+        return self.grid_to_world(Vector3(grid_x, grid_y, 0.0))
+
+    def augment_image_with_robot_pose(self, image_arr: np.ndarray) -> np.ndarray:  # type: ignore[type-arg]
```
s-desh, line 429: I think this class is getting crowded with encoding related functions and they should be in a separate class so other implementations can be easily added/swapped.

Also it relies on information LCMOccupancyGrid does not provide, right now its robot_pose and later maybe object detections.

What do you think? @paul-nechifor @leshy

paul-nechifor: I agree. In my branch I'm actually adding some OccupancyGrid functions too, but I'm adding them in different files.

I actually think this about all code in `dimos/msgs/`.

There are two issues at play here:

* IMHO, code should not provide transformations which are not exact and depend on a particular strategy. For example, `get_closest_free_point` depends on how you want to implement it. Should it be the closest point "as the crow flies"? Should it be the closest point which is reachable contiguously? Those strategies don't yield equal results all the time. Also, what if you want both? The simple solution here is to just make it an external function (or functions) which takes `OccupancyGrid` as a param.

* `dimos/msgs/` implies this code provides data structures. If they're just structs, outside of simple helper functions, they shouldn't be providing so many unrelated functionality.

I don't think we should be changing anything drastically at the moment, just food for thought.

But I do think the complex functions should be extracted out and grouped by the concern they handle.

leshy: Yes message files are holy and definitely should go out, for my stuff the policy has been always implement in your own codebase, upgrade if generic enough and needed elsewhere

s-desh: understood, will move it

PR #726, opened 2025-11-12
Draft: Migrate Transport to Rust

Kaweees:


README.md
```diff
@@ -20,11 +20,11 @@
 
 ## What is Dimensional?
 
-Dimensional is an open-source framework for building agentive generalist robots. DimOS allows off-the-shelf Agents to call tools/functions and read sensor/state data directly from ROS. 
+Dimensional is an open-source framework for building agentive generalist robots. DimOS allows off-the-shelf Agents to call tools/functions and read sensor/state data directly from ROS.
```
paul-nechifor, line 23: NITPICK: I see you've fixed trailing whitespace. That's good, but it should be separate PR because it obscures what's changed and can cause merge conflicts.

pyproject.toml
```diff
@@ -219,12 +222,12 @@ sim = [
...
-    "onnxruntime-gpu @ https://pypi.jetson-ai-lab.io/jp6/cu126/+f/4eb/e6a8902dc7708/onnxruntime_gpu-1.23.0-cp310-cp310-linux_aarch64.whl",
-    "xformers @ https://pypi.jetson-ai-lab.io/jp6/cu126/+f/731/15133b0ebb2b3/xformers-0.0.33+ac00641.d20250830-cp39-abi3-linux_aarch64.whl",
+    # # Jetson Jetpack 6.2 with CUDA 12.6 specific wheels
+    # # Note: Alternative torch wheel from docs: https://developer.download.nvidia.com/compute/redist/jp/v61/pytorch/torch-2.5.0a0+872d972e41.nv24.08.17622132-cp310-cp310-linux_aarch64.whl
+    # "torch @ https://pypi.jetson-ai-lab.io/jp6/cu126/+f/564/4d4458f1ba159/torch-2.8.0-cp310-cp310-linux_aarch64.whl",
+    # "torchvision @ https://pypi.jetson-ai-lab.io/jp6/cu126/+f/1c0/3de08a69e9554/torchvision-0.23.0-cp310-cp310-linux_aarch64.whl",
+    # "onnxruntime-gpu @ https://pypi.jetson-ai-lab.io/jp6/cu126/+f/4eb/e6a8902dc7708/onnxruntime_gpu-1.23.0-cp310-cp310-linux_aarch64.whl",
+    # "xformers @ https://pypi.jetson-ai-lab.io/jp6/cu126/+f/731/15133b0ebb2b3/xformers-0.0.33+ac00641.d20250830-cp39-abi3-linux_aarch64.whl",
```
paul-nechifor, line 230: Why is this commented out?

dimos_rs/uv.lock
paul-nechifor, line 1: Why is there a uv.lock file here too?

.editorconfig
```diff
@@ -0,0 +1,31 @@
...
+
+[*.md]
+indent_size = 3
+
+[*.nix]
+indent_size = 2
+
+[*.{py,ipynb}]
+indent_size = 4
+max_line_length = 120
```
paul-nechifor, line 24: I don't think these are the values we're using. For example for line-length we have 100 in pyproject.toml

README.md
Kaweees: Agreed, I fixed this via [EditorConfig](https://editorconfig.org/) and a couple of other tools. Since I made these changes earlier, I opened up a pull request [here](https://github.com/dimensionalOS/dimos/pull/728) that contains these changes.

dimos_rs/uv.lock
Kaweees: Stash said we should track it during a call we had. I can remove it if you would like.

.editorconfig
Kaweees: These are the common practice values for each of the respective file extensions, I can change them based on the project specific configurations.

pyproject.toml
Kaweees: Commenting this was a hacky way to get this project working locally, I didn't mean to include this whoops. Speaking of which, in a different pull request, I would like to group the difference packages from different registrars based on the deployment environment (Jetpack, CPU, Cuda 12 GPU) for this project in a similar way to how [Roboflow does it](https://github.com/roboflow/inference/blob/main/inference_experimental/pyproject.toml). They have a feature which prevents different torch sources from being installed.

https://github.com/roboflow/inference/blob/aa387396a4c1f9b341f1516c3c649142a2c62058/inference_experimental/pyproject.toml#L193C1-L210C1

.editorconfig
spomichter: Yeah we use ruff may be overkill to use both ruff and this editorconfig

dimos_rs/uv.lock
paul-nechifor: Ah no, I was asking if we need two of them. Should dimos_rs have its own requirements or be in the root? Just asking.

.editorconfig
paul-nechifor: Ruff doesn't fix the non-python files. Also the pre-commit editorconfig checker is quite fast.

Many people have their editors set to automatically remove trailing whitespace. If we allow it we'll end up with many conflicts from future contributors.

PR #728, opened 2025-11-12
Fix trailing whitespace

Kaweees:
This introduces [EditorConfig](https://editorconfig.org/), which automatically removes trailing whitespace and removes/resizes tabs from files that are edited (assuming the [EditorConfig plugin is installed](https://editorconfig.org/#download)).


.editorconfig
```diff
@@ -0,0 +1,31 @@
...
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+indent_size = 3
```
paul-nechifor: 3 is very odd, isn't it supposed to be 4 (or 2)?

Kaweees: Agreed, changed to 4

PR #737, opened 2025-11-13
Working Ivan g1 detection in blueprints

spomichter:


assets/foxglove_g1_detections.json
leshy, line 1: `foxglove_unitree_yolo.yaml` has a stupid name but is supposed to be a dash for this, do we have a reason to add another one? otherwise all good from me

spomichter: This .yaml doesnt pull up the detection streams like we have in the demos

leshy: let's just delete it then, we only need a single yaml for detectors

PR #747, opened 2025-11-16
Use a process for MuJoCo

paul-nechifor:


dimos/simulation/mujoco/shared_memory.py
```diff
@@ -0,0 +1,293 @@
...
+    def write_odom(self, pos: NDArray[Any], quat: NDArray[Any], timestamp: float) -> None:
+        odom_array: NDArray[Any] = np.ndarray((8,), dtype=np.float64, buffer=self.shm.odom.buf)
+        odom_array[0:3] = pos
+        odom_array[3:7] = quat
+        odom_array[7] = timestamp
+        self._increment_seq(2)
+
+    def write_lidar(self, lidar_msg: LidarMessage) -> None:
+        data = pickle.dumps(lidar_msg)
+        data_len = len(data)
```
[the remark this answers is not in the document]
leshy: it actually is pickleable since we have a custom pickler for o3d defined. all good

dimos/simulation/mujoco/constants.py
```diff
@@ -0,0 +1,35 @@
...
+
+# Lidar constants
+LIDAR_RESOLUTION = 0.05
+
+# Simulation timing constants
+STEPS_PER_FRAME = 7
+VIDEO_FPS = 20
+LIDAR_FPS = 2
+
+LAUNCHER_PATH = Path(__file__).parent / "mujoco_process.py"
```
leshy, line 35: I think ideally all this stuff is just part of the standard typed config for the mujoco module? not requesting to rewrite now, just mentioning

dimos/simulation/mujoco/shared_memory.py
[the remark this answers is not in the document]
paul-nechifor: Yeah, I saw it working. Codex, bad bot! :smile:

dimos/simulation/mujoco/constants.py
paul-nechifor: Ah, true, it should be configurable.

dimos/simulation/mujoco/shared_memory.py
```diff
@@ -0,0 +1,286 @@
+# Copyright 2025 Dimensional Inc.
```
leshy, line 1: wasn't this whole file avoidable if we import & use standard shm or LCM pubsub?

pubsubs per spec have just .publish(topic, data) and .subscribe(topic, callback) so easy to use independantly. though I'm totally ok with this one shot implementation, it doesn't matter since we won't further build on top of this

paul-nechifor: I started like this because I thought it was easier to use without the callbacks. :sweat_smile: But, yeah, shouldn't be too different.

PR #749, opened 2025-11-18
Deduplicate Unitree connections/entrypoints.

paul-nechifor:
* Removed the old `unitree_go2.py` and `unitree_g1.py` entrypoints. Now you can either use `dimos` or `run.py`.
* Remove duplicate connection classes and modules.
* Renamed `spec.Image.image` to `color_image` because it's clearer (`image` alone could also mean depth image).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,13 +53,13 @@ dependencies = [
     "uvicorn>=0.34.0",
 
     # Agents
-    "langchain>=0.3.27",
-    "langchain-chroma>=0.2.5",
-    "langchain-core>=0.3.72",
-    "langchain-openai>=0.3.28",
-    "langchain-text-splitters>=0.3.9",
-    "langchain-huggingface>=0.3.1",
-    "langchain-ollama>=0.3.10",
+    "langchain>=0.3.27,<1",
+    "langchain-chroma>=0.2.6,<1",
+    "langchain-core>=0.3.79,<1",
+    "langchain-openai>=0.3.33,<1",
+    "langchain-text-splitters>=0.3.11,<1",
+    "langchain-huggingface>=0.3.1,<1",
+    "langchain-ollama>=0.3.10,<1",
     "bitsandbytes>=0.48.2,<1.0",
     "ollama>=0.6.0",
 


dimos/robot/unitree/connection/connection.py
```diff
@@ -127,7 +127,7 @@ def stop(self) -> None:
 
         async def async_disconnect() -> None:
             try:
-                self.move(TwistStamped())
+                self.move(Twist())
                 await self.conn.disconnect()
             except Exception:
                 pass
```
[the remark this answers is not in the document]
paul-nechifor: Fixed.

dimos/robot/unitree/connection/g1.py
```diff
@@ -13,55 +13,155 @@
...
+        # Publish camera_link transform
+        camera_link = Transform(
+            translation=Vector3(0.3, 0.0, 0.0),
+            rotation=Quaternion(0.0, 0.0, 0.0, 1.0),
+            frame_id="base_link",
+            child_frame_id="camera_link",
+            ts=time.time(),
+        )
+
+        map_to_world = Transform(
```
leshy: this is navigation specific mapping, rosnav already does this automatically, but g1 shouldn't concern itslef with this

https://github.com/dimensionalOS/dimos/blob/dev/dimos/navigation/rosnav.py#L222

dimos/robot/unitree/connection/g1.py
```diff
@@ -13,55 +13,155 @@
...
         super().stop()
 
+    def _publish_tf(self, msg: PoseStamped) -> None:
+        if self.odom.transport:
+            self.odom.publish(msg)
+
+        self.tf.publish(Transform.from_pose("base_link", msg))
+
+        # Publish camera_link transform
+        camera_link = Transform(
```
leshy: I assume this is a transform for our camera that we glued on top of g1? this situation is defined and transform is provided by the Webcam module itself, this isn't g1 specific

check https://github.com/dimensionalOS/dimos/blob/dev/dimos/robot/unitree/g1/g1zed.py#L51

 - I know we need to consolidate my run files and autoconnect, will do

dimos/robot/unitree/connection/go2.py
```diff
@@ -41,6 +48,28 @@
 logger = setup_logger(__file__, level=logging.INFO)
 
 
+_PARAMS_DIR = Path(__file__).parent / "../../unitree_webrtc/params"
+
+
+def _get_lcm_camera_info(connection_type: str) -> tuple[CameraInfo, Any, Any]:
```
leshy: this seems like something a connection iself should provide. not as 
```py
    return lcm_camera_info, camera_matrix, dist_coeffs
```

but as a standard CameraInfo type.

for loading these standard yaml files I implemented

```py
    @classmethod
    def from_yaml(cls, yaml_file: str) -> CameraInfo:
```

at 

https://github.com/dimensionalOS/dimos/blob/dev/dimos/msgs/sensor_msgs/CameraInfo.py#L99


(note would be good to adapt and support dimos rectification tool that generates these yamls)

dimos/robot/unitree/connection/go2.py
```diff
@@ -134,71 +163,68 @@ def video_stream(self):
...
+    color_image: Out[Image] = None  # type: ignore
     camera_info: Out[CameraInfo] = None  # type: ignore
-    connection_type: str = "webrtc"
 
     connection: Go2ConnectionProtocol
-
-    ip: str | None
-
     camera_info_static: CameraInfo = _camera_info_static()
+    camera_matrix: Any
```
leshy: camerainfo contains all of these values, we shouldn't have them in this random format on the connection itself..

there is also camera hardware spec
https://github.com/dimensionalOS/dimos/blob/dev/dimos/hardware/camera/spec.py#L32

that connection could just conform to?

IMO in general if we do good, we don't have separate specs for hardware implementations and actual modules

`dimos/spec/*` vs hardware component specs `dimos/hardware/camera/spec.py`

but one spec location, format and you can deploy what you want, outputs are streams, which can be dimos transprots or just rxpy streams etc.. but we are not there yet, so separate specs?

dimos/robot/unitree/connection/go2.py
```diff
@@ -134,71 +163,68 @@ def video_stream(self):
...
         *args,
         **kwargs,
     ) -> None:
-        match ip:
-            case None | "fake" | "mock" | "replay":
-                self.connection = ReplayConnection()
-            case "mujoco":
-                from dimos.robot.unitree_webrtc.mujoco_connection import MujocoConnection
+        self._global_config = global_config or GlobalConfig()
+        self.rectify_image = not self._global_config.simulation
```
leshy: image rectification doesn't seem to belong to unitree at all but is a more general thing, afaik I should be the only user of this, and I don't need this here so we can just remove.

if we want it in, it should be (at most) at some Camera superclass that provides it implicitly from existing Image and CameraInfo properties

or even better on an Image class
```py
Image.rectify(CameraInfo) -> Image
```

in general we shouldn't encourage rectification if projecting a few points, it's cheaper to project those from the unrectified image through the camera info distortion, then to rectify each pixel

dimos/robot/unitree/connection/go2.py
```diff
@@ -244,31 +268,48 @@ def _odom_to_tf(cls, odom: PoseStamped) -> list[Transform]:
...
         """Send movement command to robot."""
-        self.connection.move(twist, duration)
+        return self.connection.move(twist, duration)
 
     @rpc
-    def standup(self):
+    def standup(self) -> None:
         """Make the robot stand up."""
-        return self.connection.standup()
+        self.connection.standup()
```
leshy: don't we want success returns ideally here?

dimos/robot/unitree/connection/go2.py
```diff
@@ -244,31 +268,48 @@ def _odom_to_tf(cls, odom: PoseStamped) -> list[Transform]:
             sensor,
         ]
 
-    def _publish_tf(self, msg) -> None:
+    def _publish_tf(self, msg: PoseStamped) -> None:
         self.tf.publish(*self._odom_to_tf(msg))
+        if self.odom.transport:
+            self.odom.publish(msg)
+
+    def _on_lidar(self, msg: LidarMessage) -> None:
```
leshy: these don't feel very natural, why do we have those? shouldn't connection provide streams that we pipe into outputs?

```py
self.connection.lidar_stream().subscribe(self.pointcloud.publish),
```

what are these functions for?

I think what bothers us is that unconnected transports are not publishable, I agree - we should allow unconnected noop publishing

dimos/robot/unitree/connection/g1.py
```diff
@@ -13,55 +13,155 @@
...
+        self._publish_tf(
+            PoseStamped(
+                ts=msg.ts,
+                frame_id=msg.frame_id,
+                position=msg.pose.pose.position,
+                orientation=msg.pose.orientation,
+            )
+        )
+
+    def _publish_sim_odom(self, msg: SimOdometry) -> None:
```
leshy: I think generic g1 connection shouldn't care about sim odom vs odom, it should get the same data from the connection implementation given they conform to some Protocol,  why are we doing this?

I also don't know why aren't we doing this as reactivex but individual methods, seems much more verbose then let's say
```py
connection.odom().pipe(self._publish_tf)
```

paul-nechifor: Unitree Go2 gets the odometry from the webrtc connection (so `self.connection.odom_stream()`. MujocoConnection does it the same.

But G1 gets odometry from `self.odom_in` which is coming from ROS, not the connection.

dimos/robot/unitree/connection/go2.py
paul-nechifor: Fixed.

dimos/robot/unitree/connection/go2.py
paul-nechifor: Fixed.

dimos/robot/unitree/connection/go2.py
paul-nechifor: I've removed rectification.

dimos/robot/unitree/connection/go2.py
paul-nechifor: I've removed this.

dimos/robot/unitree/connection/go2.py
paul-nechifor: I've removed camera_matrix

dimos/robot/unitree/connection/g1.py
paul-nechifor: Removed. Now it's in g1sim.py

dimos/robot/unitree/connection/g1.py
paul-nechifor: Removed. Now it's in g1sim.py

dimos/robot/unitree/connection/g1.py
paul-nechifor: Removed.

PR #764, opened 2025-11-23
docs: Add conceptual doc for Blueprints

ym-han:
Copies / adapts some of the stuff from Paul Nechifor's excellent dimos/core/README_BLUEPRINTS.md, but also adds other stuff (and doesn't include some stuff that gets discussed in other docs) and goes at a slower pace. In particular, it adds more scaffolding and background for people new to the framework.

### DIFF
--- a/docs/concepts/blueprints.md
+++ b/docs/concepts/blueprints.md
@@ -160,14 +160,16 @@ coordinator = blueprint.build()
 module_coordinator.stop()  # Clean up when finished
 ```
 
-## How the blueprint system works, in more detail
+## How the blueprint system works, in more detail (Advanced)
 
 Now that we've seen what blueprints are, at a high level, and how to build and run them,
 we are in a position to dive into details that are helpful for building more complicated systems, such as
 
-- how the blueprint system matches compatible streams
-- what happens if the same module is supplied more than once to `autoconnect`
+- how the blueprint system matches compatible streams, and what to do when stream names don't match
 - how to override the default configuration
+
+> [!TIP]
+> Feel free to skip this section on first read -- you can always come back when you need the details.
 
 ### How the blueprint system matches compatible streams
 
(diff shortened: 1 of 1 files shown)


docs/concepts/blueprints.md
```diff
@@ -0,0 +1,344 @@
...
+)
+# These streams share a transport instance
+# Data published by ProducerModule.image flows to ConsumerModule.image
+```
+
+Matching on not just the name but also the *type* prevents mistakes: an `Out[Temperature]` won't connect to `In[Pressure]` even if both are named `data`.
+
+### Last-wins duplicate elimination
+
+If you include the same module multiple times, the last one wins:
```
paul-nechifor: You might want to include the purpose for this. It's so you can extend a blueprint but override the previous params it was using for that specific module.

docs/concepts/blueprints.md
```diff
@@ -0,0 +1,344 @@
...
+class ProducerModule(Module):
+    image: Out[Image] = None
+```
+
+will be `/image`.
+
+The stream property name is used as the topic only if it's unique. If two streams share the same name but have different types, the blueprint system assigns each group a random topic id like `/SGVsbG8sIFdvcmxkI` to keep them separate.
+
+<!-- Citation: blueprints.py:233 - topic = f"/{name}" if unique -->
+<!-- Citation: blueprints.py:233 - topic = f"/{short_id()}" if not unique; blueprints.py:249-250 - _is_name_unique checks count -->
```
paul-nechifor: This should be removed. It can no longer happen but the code for it is still there.

Currently, if there are two streams with the same name but different types, it will error. The previous way (described here) led to confusion.

docs/concepts/blueprints.md
```diff
@@ -0,0 +1,344 @@
...
+blueprint = blueprint.transports({
+    ("image", Image): pSHMTransport(
+        "/go2/color_image", default_capacity=DEFAULT_CAPACITY_COLOR_IMAGE
+    ),
+    ("start_explore", Bool): pLCMTransport(),
+})
+```
+
+> [!NOTE]
+> `expanded_blueprint` does not get the transport overrides because it's created from the initial value of `blueprint`, not the second.
```
paul-nechifor: I've just noticed now that this could be a bit confusing because it's talking about two points. I think you should delete the `expanded_blueprint = autoconnect(blueprint, ...)` line and the note, and talk about this separately.

Essentially, what the note is saying should be a separate topic: the fact that blueprints are immutable and autoconnect/.transports/etc returns a new one. So `blueprint.transports` doesn't override the transports on the previous `blueprint` or on `expanded_blueprint`.

docs/concepts/blueprints.md
ym-han: Nice, thanks for catching this! (I see that there's even tests for this in dimos/core/test_blueprints.py) 
I'm changing it to
> Streams with the same name must have the same type -- this is how the blueprint system knows to wire them together. If two streams share a name but have different types, `build()` raises a `ValueError`.

docs/concepts/blueprints.md
ym-han: Thanks for catching this too -- it was something I had just adapted / copied and pasted from the blueprints README in a rush. I'm just going to delete the note for now, and maybe just note in the "Last-wins duplicate elimination" (or maybe somewhere before that) that blueprints are immutable and the various methods always return a new blueprint.

docs/concepts/blueprints.md
ym-han: [TODO] will return to this once I get further along with the tutorials

PR #731, opened 2025-11-13
Make dimos pip-installable

paul-nechifor:
# New stuff

* I've copied rxpy_backpressure in since it's so small.
* I've tested the release of dimos-lcm to test PyPI.
* Replaced our go2-webrtc-connect clone with the original.
* I'm currently cloning the `dev` repo in order to get LFS files when using `pip install dimos`. I need to switch to `main` when we release.
* Added fresh instructions in README_installation.md .

---

# Old stuff

Now you can:

```bash
pip install dimos
dimos-robot --robot-ip 192.168.x.xxx run unitree-go2
```

or with `uv`:

```bash
uv add dimos
uv run dimos-robot --robot-ip 192.168.x.xxx run unitree-go2
```

or even better, no need to install:

```bash
uvx --from dimos dimos-robot --robot-ip=192.168.x.xxx run unitree-go2
```

Or even better, rename `dimos-robot` to `dimos` (https://github.com/dimensionalOS/dimos/pull/740) and it becomes just:

```bash
uvx dimos --robot-ip=192.168.x.xxx run unitree-go2
```

PR #781, opened 2025-11-29
docs: Add quickstart

ym-han:
I made a separate quickstart doc and linked to it from the README, to make the README less cluttered -- that's how https://github.com/Eventual-Inc/Daft/blob/main/README.rst does it too. But I could also inline it into the README if that's what is preferred.

The quickstart doesn't get you to run a quick Mujoco simulation, but maybe it should. Then again, it's also important for a quickstart to be super concise.

The README needs more cleaning up -- I would personally go for a structure like Daft's -- but that probably shouldn't be part of this PR.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -1,0 +1,138 @@
+# Quickstart
+
+DimOS is a modular framework for building agentive robots. In this quickstart, you'll learn the basics of DimOS by building an LLM agent that can make greetings.
+
+## Installation
+
+### Requirements
+
+- Python 3.10+
+- OpenAI API key in environment
+
+### Install DimOS
+
+```python
+# TODO: Ideally, when this is released, this should be as simple as
+# pip install dimos
+```
+<!-- TODO: Check what the installation directions should be, post-release; hopefully it will be this easy after the library gets published -->
+
+## Define a skill
+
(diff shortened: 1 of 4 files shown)


docs/quickstart.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Requirements
+
+- Python 3.10+
+- OpenAI API key in environment
+
+### Install DimOS
+
+```python
+pip install -U dimos
```
spomichter: this wouldnt work im pretty sure since dimos isnt in PyPi public distro yet

ym-han: yes, the install instructions are aspirational / placeholders -- there's a comment to that effect. But maybe I should make it clearer it's TBD

ym-han: I've updated it to make it more explicit that the installation instructions part is just a placeholder. Hopefully after #731 lands and after release, it really will be that easy to install dimos. (Actually after that PR lands, it might be possible to do `pip install git+ssh://git@github.com/dimensionalOS/dimos.git` even without adding it to PyPi).

(I could link to the current installation instructions, but I really do think the installation needs to be made simpler for non-contributor users. The user-oriented quickstart needs to look quick.)

PR #792, opened 2025-12-02
New planner

paul-nechifor:
* If you want to see the navigation map in Foxglove use this:
```bash
DEBUG_NAVIGATION=1 dimos --robot-ip 192.168.x.xxx run unitree-go2-test-nav
```

* Added ability to generate a mujoco room from an occupancy map grid:
```bash
 dimos --mujoco-start-pos '4,2' --mujoco-camera-position "4.13,0,4.9,7.14,89,-50" --mujoco-room-from-occupancy data/occupancy_simple.npy run unitree-go2-test-nav
```

* Added ability to pre-load the global map from a previously saved pointcloud:
```bash
dimos --mujoco-start-pos '1.64,9.4' --mujoco-camera-position "4.13,0,4.9,7.14,89,-50" --mujoco-room-from-occupancy data/three_paths.png --mujoco-global-map-from-pointcloud data/three_paths.ply run unitree-go2-test-nav
```

I've fixed a bug with voxels being rounded down instead of to the nearest pixel. This produced these kinds of artifacts which were covered up by filling the empty space and inflating the occupied areas.

<table>
  <tr>
    <td>before<br>before</td>
    <td>after<br>after</td>
  </tr>
</table>

I've introduced another style for rendering occupancy maps. It renders fully occupied as magenta so it's easier to tell apart from unknown. This is the style used in tests. It doesn't have to be shown to the user.

<table>
  <tr>
    <td>turbo<br>visualize_occupancy_turbo</td>
    <td>rainbow<br>visualize_occupancy_rainbow</td>
  </tr>
</table>

I've changed the A* algorithm to prioritize minimum cost above path length.

<table>
  <tr>
    <td>before<br>astar_general</td>
    <td>after<br>astar_min_cost</td>
  </tr>
</table>

I've added a new function to resample A* paths. It produces much smoother paths (the robot is represented by a 0.2x0.4 rectangle at each pose):

<table>
  <tr>
    <td>before<br>resample_path_simple</td>
    <td>after<br>resample_path_smooth</td>
  </tr>
</table>

### DIFF
--- a/dimos/navigation/replanning_a_star/local_planner.py
+++ b/dimos/navigation/replanning_a_star/local_planner.py
@@ -214,7 +214,6 @@ class LocalPlanner(Resource):
                 cmd_vel = None
 
             if cmd_vel is not None:
-                print(cmd_vel.linear.x, cmd_vel.linear.y, cmd_vel.angular.z)
                 self.cmd_vel.on_next(cmd_vel)
 
             elapsed = time.perf_counter() - start_time
@@ -271,8 +270,6 @@ class LocalPlanner(Resource):
         first_yaw = quaternion_to_euler(first_pose.orientation).z
         robot_yaw = current_odom.orientation.euler[2]
         yaw_error = normalize_angle(first_yaw - robot_yaw)
-
-        print("yaw_error", yaw_error)
 
         if abs(yaw_error) < self._orientation_tolerance:
             with self._lock:


data/.lfs/apartment.tar.gz
leshy, line 1: do you think we should move all these into some planner/ dir or something? I assume we either need most or none

dimos/mapping/occupancy/visualize_path.py
```diff
@@ -0,0 +1,89 @@
...
+import numpy as np
+
+from dimos.mapping.occupancy.visualizations import visualize_occupancy_grid
+from dimos.msgs.nav_msgs import Path
+from dimos.msgs.nav_msgs.OccupancyGrid import OccupancyGrid
+from dimos.msgs.sensor_msgs.Image import Image
+from dimos.msgs.sensor_msgs.image_impls.AbstractImage import ImageFormat
+
+
+def visualize_path(
```
leshy, line 25: Just a note, I personally like to emit to foxglove and let it overlay stuff, idk if you thought of it or not

dimos/navigation/replanning_a_star/module.py
```diff
@@ -0,0 +1,106 @@
...
+from dimos.core.global_config import GlobalConfig
+from dimos.msgs.geometry_msgs import PoseStamped, Twist
+from dimos.msgs.nav_msgs import OccupancyGrid, Path
+from dimos.msgs.sensor_msgs import Image
+from dimos.navigation.base import NavigationInterface, NavigationState
+from dimos.navigation.replanning_a_star.global_planner import GlobalPlanner
+
+
+class ReplanningAStarPlanner(Module, NavigationInterface):
+    odom: In[PoseStamped]  # TODO: Use TF.
```
leshy, line 30: it's so easy to use tf, let's obsolete this
```py
self.tf.get('world', 'base_link') 
```

dimos/navigation/replanning_a_star/module.py
```diff
@@ -0,0 +1,106 @@
...
+from dimos.core import In, Module, Out, rpc
+from dimos.core.global_config import GlobalConfig
+from dimos.msgs.geometry_msgs import PoseStamped, Twist
+from dimos.msgs.nav_msgs import OccupancyGrid, Path
+from dimos.msgs.sensor_msgs import Image
+from dimos.navigation.base import NavigationInterface, NavigationState
+from dimos.navigation.replanning_a_star.global_planner import GlobalPlanner
+
+
+class ReplanningAStarPlanner(Module, NavigationInterface):
```
leshy, line 29: there is also dimos.spec (used by rosnav.py) 

I also see you have .target and .set_goal etc, what's the reason for it? could we try and solidify the spec?

dimos/core/global_config.py
```diff
@@ -13,18 +13,36 @@
...
+    mujoco_camera_position: str | None = None
     mujoco_room: str | None = None
+    mujoco_room_from_occupancy: str | None = None
+    mujoco_global_costmap_from_occupancy: str | None = None
+    mujoco_global_map_from_pointcloud: str | None = None
+    mujoco_start_pos: str = "-1.0, 1.0"
     robot_model: str | None = None
+    robot_width: float = 0.3
+    robot_rotation_diameter: float = 0.6
+    planner_strategy: NavigationStrategy = "simple"
```
leshy, line 43: hmmm not a fan of centralized config that's expanded like this in a single place, "core" shouldn't care about your particular "planner_strategy"

we have a standard for shared config that comes with Configurable, overlayed on top of ModuleConfig, that also applies defaults.
 
https://github.com/dimensionalOS/dimos/blob/0c646081eb5d0f41c62055346d3a897103c25310/dimos/mapping/voxels.py#L47

For global config we should be able to easily apply a deep dict, like

```python
{
"memory_limit": ..., 
"n_dask_workers": ...

# this is class name
"astar_navigator": { 
    "strategy": ...
}
}
```

then if we want convenient CLI we can play with some sort of deep matching etc, but for a start `--astar_navigator_strategy=...` is also ok?

dimos/navigation/replanning_a_star/module.py
```diff
@@ -0,0 +1,106 @@
...
+    odom: In[PoseStamped]  # TODO: Use TF.
+    global_costmap: In[OccupancyGrid]
+    goal_request: In[PoseStamped]
+    target: In[PoseStamped]
+
+    goal_reached: Out[Bool]
+    navigation_state: Out[String]  # TODO: set it
+    cmd_vel: Out[Twist]
+    path: Out[Path]
+    debug_navigation: Out[Image]
```
leshy, line 39: as long as we are aware we have a rich visual language on fg (likely mapping to similar primitives in rerun) but still prefer this for now, all good.

https://github.com/dimensionalOS/dimos/blob/0c646081eb5d0f41c62055346d3a897103c25310/dimos/perception/detection/type/detection3d/imageDetections3DPC.py#L26

dimos/core/global_config.py
paul-nechifor: Most of these are meant to serve as a sort of [feature flags](https://martinfowler.com/articles/feature-toggles.html): added as a way to test which is working better and removed once we're certain which is best.

The issue with a nested config is where to place the value. In this example, `planner_strategy` is used by `make_navigation_map` which is used by both `navigation/global_planner/planner.py` and `navigation/replanning_a_star/navigation_map.py`. So I'm not sure where to put it in this particular case.

I'm okay with using a nested config, though. Do you think it would be better if you added the code to read/pass the config? I'd do it myself, but I'm sensing a slight difference of opinion here (for example on yaml too) so it would be cool if I could see what you have in mind. :smiley:

dimos/navigation/replanning_a_star/module.py
leshy: def nvm on this since rerun is coming

dimos/core/global_config.py
leshy: aha this is important so let's make sure that I understand your perspective first also, don't want to trample over it if I don't understand it :D :D

dimos/navigation/replanning_a_star/module.py
paul-nechifor: NavigationInterface contains the methods that are used in the agent skills. The skills module binds to `NavigationInterface.set_goal` etc to communicate with whichever planner is deployed.

Yeah, I'll look at unifying.

dimos/navigation/replanning_a_star/module.py
paul-nechifor: Yep, I'll do this.

paul-nechifor: tf doesn't have subscribe. How would I subscribe to the latest position?

data/.lfs/apartment.tar.gz
paul-nechifor: Hm, I'm not sure. There are quite a few things which need to be moved in there and doing it in this PR would make it even bigger than it already is. I'll add this on my todo list.

dimos/core/global_config.py
leshy: had a talk, implementing a config system in a separate PR

PR #802, opened 2025-12-04
Fix incorrect `= None`

paul-nechifor:
We've typically defined streams as:

```python
color_image: Out[Image] = None
```

This is incorrect as `color_image` cannot be none. But it's required because Dask's actor model looks at class properties not type annotations.

To avoid `= None` on every stream definition, I've moved them to `DaskModule.__init_subclass__`. That handles the initialization so we only have to define:

```python
color_image: Out[Image]
```

### DIFF
--- a/dimos/core/_test_future_annotations_helper.py
+++ b/dimos/core/_test_future_annotations_helper.py
@@ -1,0 +1,36 @@
+# Copyright 2025 Dimensional Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


dimos/core/module.py
```diff
@@ -279,6 +280,33 @@ class DaskModule(ModuleBase):
...
+            hints = get_type_hints(cls, globalns=globalns, include_extras=True)
+        except (NameError, AttributeError, TypeError):
+            hints = {}
+
+        for name, ann in hints.items():
+            origin = get_origin(ann)
+            if origin in (In, Out):
+                # Set class-level attribute if not already set.
+                if not hasattr(cls, name) or getattr(cls, name) is None:
+                    setattr(cls, name, None)
```
leshy, line 308: tnx long time coming, did you check if we can do this by augmenting the system that assigns input/output classes to the module, it iterates through properties so we needed `= None`, should it iterate through hints as well, so that we don't have a step of assinging None here?

if hard no worries at all

PR #787, opened 2025-12-01
[Convenience PR / branch: all of the docs / all of YM's changes]

ym-han:
This is a convenience PR that collects all the changes. When actually reviewing, though, it'd probably be easier to review the constituent PRs.

To understand the dependency structure of the PRs, see https://gist.github.com/ym-han/1eefc2057ffd02428ac3405821959768

To get a quick sense for what docs have been done and what else is left, see https://github.com/dimensionalOS/dimos/issues/736#issuecomment-3577190229 (please do read this comment first).

## How to see the docs

First, make sure you've installed the docs deps:
```
uv pip install .[docs]
```

### Local Development Server

To start a local server with hot reload:

```bash
mkdocs serve
# Then open <http://127.0.0.1:8000/> in your browser.
```

### Build Static Site

To build the static documentation site:

```bash
mkdocs build
```

The output (which includes the various `llm.tx`es) will be in the `site/` directory.

See also the note in docs/development.md on marimo notebooks.

## What needs to get added to CI

This was mentioned in the comment, but just in case:
- need to set up CI for `xdoctest`
- tests / checks of the marimo notebooks, if you decide to stick with marimo (but as I told Ivan, it's easy to switch to a different option because it's easy to export marimo notebooks to markdown etc)

Depending on the docs platform you go with, you might also need to add CI checks to make sure the links work etc.

### Doc site packages

Ivan was asking, so: for reference, the non-Mintlify docs sites solutions I'm aware of include:

- Astro Starlight (https://starlight.astro.build/)
- fumadocs: https://fumadocs.dev/
- Zensical -- but this might not be stable enough rn: https://squidfunk.github.io/mkdocs-material/blog/2025/11/05/zensical/

Unfortunately, I don't know which of these would be most suitable for you. But the Zensical stuff does suggest that mkdocs isn't a good long-term solution (I went with mkdocs initially just because it's a common choice; hadn't known then that it didn't seem well-maintaned).

Issue #736, opened 2025-11-13
TODO list before release


paul-nechifor:
* Delete dead code (https://github.com/dimensionalOS/dimos/pull/718 , https://github.com/dimensionalOS/dimos/pull/725)
* Clean up the code (whitespace: https://github.com/dimensionalOS/dimos/pull/728)
* Documentation (https://github.com/dimensionalOS/dimos/pull/710)
* ??? Switch to `uv`?
* ??? Add tests for important untested code
* ??? Improve logging (https://github.com/dimensionalOS/dimos/pull/715)
* ??? Fully working Docker
* ??? Fully working `pip install dimos` (https://github.com/dimensionalOS/dimos/pull/731 )
* ??? Local LLM agents (https://github.com/dimensionalOS/dimos/pull/739)
* ??? Rename agents to agents_deprecated and agents2 to agents

PR #809, opened 2025-12-05
[Convenience PR / branch: all of the docs / all of YM's changes]

ym-han:
Re-opening new PR b/c couldn't open old one. (leshy had also left some comments on the old PR: https://github.com/dimensionalOS/dimos/pull/787)

To get a quick sense for what docs have been done and what else is left, see https://github.com/dimensionalOS/dimos/issues/736#issuecomment-3577190229 (please do read this comment first).

## How to see the docs

First, make sure you've installed the docs deps:
```
uv pip install .[docs]
```

### Local Development Server

To start a local server with hot reload:

```bash
mkdocs serve
# Then open <http://127.0.0.1:8000/> in your browser.
```

### Build Static Site

To build the static documentation site:

```bash
mkdocs build
```

The output (which includes the various `llm.tx`es) will be in the `site/` directory.

See also the note in docs/development.md on marimo notebooks.

## Notes on doc tech stack choices

- xdoctest because it's better than the stock doctest facilities at floating-point comparisons, and uses a more robust AST-based approach. PyTorch uses this.
- Marimo: like jupyter notebooks, except more git-friendly. Easy to export to markdown or python; easy to migrate away from it if you don't like it.

### Doc site packages

Ivan was asking, so: for reference, the non-Mintlify docs sites solutions I'm aware of include:

- Astro Starlight (https://starlight.astro.build/)
- fumadocs: https://fumadocs.dev/
- Zensical -- but this might not be stable enough rn: https://squidfunk.github.io/mkdocs-material/blog/2025/11/05/zensical/

Unfortunately, I don't know which of these would be most suitable for you. But the Zensical stuff does suggest that mkdocs isn't a good long-term solution (I went with mkdocs initially just because it's a common choice; hadn't known then that it didn't seem well-maintaned).


docs/concepts/skills.md
```diff
@@ -0,0 +1,213 @@
...
+
+Finally, if there's information you want to get to an agent, you need to do that with skills -- more on this shortly.
+
+## What is a skill?
+
+At a high level, skills are wrappers over lower-level robot capabilities. But at a more prosaic level, a skill is just a method on a Module decorated with `@skill` that:
+
+1. **Becomes an agent-callable tool** - The decorator generates an OpenAI-compatible function schema from the method signature and docstring
+2. **Executes in background threads** - Skills run concurrently without blocking the agent
+3. **Reports state via messages** - Each execution tracks state (pending → running → completed/error)
```
leshy: this is an internal detail of messaging protocol related to skills, there are many details to this protocol, and users need not understand it so I'm not sure why are we explaining this here?

this is also an incomplete explanation, actual states are 
```
    pending = 0
    start = 1
    stream = 2
    reduced_stream = 3
    ret = 4
    error = 5
```

docs/concepts/skills.md
```diff
@@ -0,0 +1,213 @@
...
+<!-- Citation: dimos/core/module.py:77 - ModuleBase inherits from SkillContainer -->
+
+Finally, if there's information you want to get to an agent, you need to do that with skills -- more on this shortly.
+
+## What is a skill?
+
+At a high level, skills are wrappers over lower-level robot capabilities. But at a more prosaic level, a skill is just a method on a Module decorated with `@skill` that:
+
+1. **Becomes an agent-callable tool** - The decorator generates an OpenAI-compatible function schema from the method signature and docstring
+2. **Executes in background threads** - Skills run concurrently without blocking the agent
```
leshy: it executes on a hosting module which is a separate process, I'm not sure thread is meant abstractly here, but this is an important detail to be clear about

docs/concepts/skills.md
```diff
@@ -0,0 +1,213 @@
...
+```
+<!-- Citation: dimos/core/module.py:77 - ModuleBase inherits from SkillContainer -->
+
+Finally, if there's information you want to get to an agent, you need to do that with skills -- more on this shortly.
+
+## What is a skill?
+
+At a high level, skills are wrappers over lower-level robot capabilities. But at a more prosaic level, a skill is just a method on a Module decorated with `@skill` that:
+
+1. **Becomes an agent-callable tool** - The decorator generates an OpenAI-compatible function schema from the method signature and docstring
```
leshy: tool call definition encoding can be dependant on the model we are integrating, I'm not sure if this is openai every time? langchain controls this layer, I honestly don't know, and I'm not sure why we are explaining this to a user at this point

docs/tutorials/skill_basics/greeter.py
```diff
@@ -0,0 +1,129 @@
...
+        super().start()
+
+    @rpc
+    def stop(self) -> None:
+        super().stop()
+
+    # The following dunder methods are for Dask serialization:
+    # Module instances are serialized when deployed to worker processes.
+    # We return {} in __getstate__ since this class has no custom state to preserve.
+    def __getstate__(self):
```
leshy, line 76: what is this?

docs/concepts/modules.md
```diff
@@ -0,0 +1,144 @@
...
+
+**Composability** - Modules connect in flexible topologies without enforced hierarchies. A camera module can feed multiple perception modules; an agent can coordinate several navigation modules.
+
+<!-- Citation: blueprints.py:41-45 - ModuleBlueprint has no hierarchical constraints, just module class + connections -->
+
+**Safety through isolation** - Because modules are mapped onto separate processes, every module has its own address space. Even if one module fails catastrophically (e.g. a segfault), it won't bring down the others.
+
+<!-- Citation: module.py:113-126 - _close_module() shows independent resource management (loop, rpc, tf, disposables) -->
+
+**Distributed execution** - Modules run as Dask actors across a cluster. Deploy computationally intensive perception on GPU workers while running navigation on CPU workers. The system handles network communication, serialization, and RPC automatically.
```
leshy: GPU workers vs CPU workers? there is no such thing in dimos

docs/concepts/modules.md
```diff
@@ -0,0 +1,144 @@
...
+
+```python
+class ModuleA(Module):
+    image: Out[Image] = None
+    start_explore: Out[Bool] = None
+```
+
+In particular, these declarations are done with *streams*: `In[T]` for input and `Out[T]` for output, where `T` is the type variable for the type of data the stream carries.
+
+Streams provide reactive, push-based data flow between modules, built on ReactiveX. The [blueprint system](./blueprints.md) validates that connected streams have compatible types at build time.
```
leshy, line 54: we support reactivex, we are not built on reactivex

docs/concepts/transport.md
```diff
@@ -0,0 +1,243 @@
...
+        # - pSHMTransport (local shared memory)
+        # - LCMTransport (network)
+        # - JpegLcmTransport (compressed network)
+        self.add_to_memory(img)
+```
+
+<!-- Citation: perception/spatial_perception.py:64 - Real SpatialMemory module with color_image: In[Image] declaration -->
+
+> [!NOTE]
+> The core takeaway: modules can operate on any underlying transport. The same module code deploys across local shared memory, network protocols, or distributed clusters by changing just the transport configuration.
```
leshy, line 32: a lot of this is theoretical, like ok, how do I run on a distributed cluster? We have no code for this, we don't want to claim things like this that don't exist yet

docs/concepts/skills.md
```diff
@@ -0,0 +1,213 @@
...
+
+**Write clear docstrings** - They become tool descriptions. Be specific about what the skill does and what parameters mean.
+<!-- Citation: dimos/protocol/skill/schema.py - function_to_schema() extracts docstrings -->
+
+**Handle errors gracefully** - Return contextual error messages for agent recovery, not raw exceptions.
+
+**Monitor long-running skills** - Use `skillspy` to watch skill execution in real-time. Skills are tracked in an execution database showing what's currently running and what has completed—invaluable for debugging navigation or other long operations. See the [skill basics tutorial](../tutorials/skill_basics/tutorial.md) for an example of this.
+
+> [!WARNING]
+> **Don't use both `@skill` and `@rpc` decorators on a single method** - The `@skill` wrapper can't be pickled for LCM transport. Use `@skill()` for agent tools, `@rpc` for module-to-module calls.
```
leshy, line 201: this is likely a bug? you should be able to use both decorators though @skill implies @rpc

docs/concepts/skills.md
```diff
@@ -0,0 +1,213 @@
...
+See these tutorials for examples:
+
+* [Equip an agent with skills](../tutorials/skill_with_agent/tutorial.md).
+* [Build a RoboButler multi-agent system](../tutorials/multi_agent/tutorial.md)
+
+## Updating agents with results from skills
+
+We've seen how skills can be made available to agents as tools they can call. Often, however, we don't just want agents making tool calls -- we also want to relay updates from the tool calls, from the skills, back to the agent.
+
+Some of this behavior already comes as a default: if you decorate the method with `@skill()`, the agent will be notified with the *return value* of the method when the skill finishes (because the default value of the `ret` parameter is `Return.call_agent`).
```
leshy: `ret` parameter? what is that? why are we talking about this? this is the first mention of this or `Return.call_agent`

docs/concepts/skills.md
```diff
@@ -0,0 +1,213 @@
...
+
+### Notifying the agent whenever there's updates
+
+But often we want to update the agent not just when the skill is finished, but also whenever there's progress. Think, e.g., of a 'move to certain coordinates' skill, where we might want to stream progress updates continously to the agent.
+
+This can be done by making  the method a generator and setting the `stream` parameter of `@skill` to `Stream.call_agent`:
+
+```python
+@skill(stream=Stream.call_agent, reducer=Reducer.string)
```
leshy: why are we introducing reducer=... here? makes example unclear and also it's not needed here

docs/concepts/skills.md
```diff
@@ -0,0 +1,213 @@
...
+
+**Monitor long-running skills** - Use `skillspy` to watch skill execution in real-time. Skills are tracked in an execution database showing what's currently running and what has completed—invaluable for debugging navigation or other long operations. See the [skill basics tutorial](../tutorials/skill_basics/tutorial.md) for an example of this.
+
+> [!WARNING]
+> **Don't use both `@skill` and `@rpc` decorators on a single method** - The `@skill` wrapper can't be pickled for LCM transport. Use `@skill()` for agent tools, `@rpc` for module-to-module calls.
+
+
+## See also
+
+- [The Skills API reference](../api/skills.md)
```
leshy, line 206: I feel like after this I don't understand the concepts of reducers, active/passive streams etc, idk about tooling (skillspy, agentspy, humancli) idk if this is explained elsewhere, will go through tutorials

docs/concepts/skills.md
ym-han: I think I got an error when literally trying to do something like
```
# but can't remb the exact order I had tried this in
@skill
@rpc
def method():
```

docs/concepts/skills.md
ym-han: Good catch, this needs to be refactored, yeah.

I think this was meant to be SkillStateEnum, which, as of the time of this PR, only had 
```
    pending = 0
    running = 1
    completed = 2
    error = 3
```
(I.e., this isn't factually wrong if it's meant to be SkillStateEnum.)

I don't remember this part very well now, but I think I was tempted to leave that there because it might be helpful for understanding skillspy's output. But you are right in that it either needs more explanation or should just be removed. Maybe the thing to do is to only bring up SkillStateEnum when actually explaining Skillspy -- will remove that third point, unless you disagree.

docs/concepts/skills.md
ym-han: Yes, this is misleading in a bad way. Am changing this to "**Executes asynchronously** - Skills run without blocking the agent"

docs/concepts/modules.md
ym-han: Yeah. In principle Modules should help with this, but yes, it's too misleading in this context. Will remove that sentence.

docs/concepts/skills.md
ym-han: Changing this to "**Becomes an agent-callable tool** - The decorator generates a tool schema from the method signature and docstring"

docs/concepts/skills.md
ym-han: Good catch, will remove it.

docs/concepts/skills.md
ym-han: Good catch, I've removed that parenthetical comment.

docs/tutorials/skill_basics/greeter.py
ym-han: I can't remb well now, but I'm guessing I added this to explain how to write a skill 'from scratch', without the helper SkillModule. (The SkillModule had this, and the blueprint readme also had it.) But it might be more confusing than helpful

docs/concepts/transport.md
ym-han: Yeah you're right, I had let this slide because I was thinking too much about what this in principle allows for.

docs/concepts/modules.md
ym-han: Since you are re-writing the modules doc, I'll let you handle this

Issue #804, opened 2025-12-04
Encoding maps [WIP]


s-desh:
Goal: A way to encode maps for agents to understand space.

**v0**
Requirements:
- The agent should be able to act on queries like "go to the largest room", "go to first room after the corridor" etc.

PRs:
#815 
- Adds evals for point placement and map comprehension.  
- Encodes costmap as RGB image and overlays robot pose.
- Uses this encoding in a skill that can pull maps and place points based on the query for navigation. (Incomplete)

TODOs:
Evals:
1. Add larger and realistic maps for eval (only for evals - directly test images instead of image -> costmap -> image)
2. Support to test multiple models
3. Separate encoding into OccupancyGridImage

Skill:
1. Use Agents to send map instead of `VLModel`
2. Complete interpret map skill after comparing agent / vlmodel perf.
3. Testing with agents.

**v0.1**

- Overlay object bboxes?

**v0.2**
- multifloor? (side by side images)
- history? (can use time dimension in video models)

PR #815, opened 2025-12-08
Encode maps

s-desh:
This PR is one of many for encoding maps for agents. #804 

Adds
- Evals for point placement and map comprehension.
- Encodes costmap as RGB image and overlays robot pose.
- Uses this encoding in a skill that can pull maps and place points based on the query for navigation. (Incomplete)

### DIFF
--- a/dimos/agents2/skills/interpret_map.py
+++ b/dimos/agents2/skills/interpret_map.py
@@ -14,9 +14,11 @@
 
 import json
 import os
+from typing import TYPE_CHECKING
 
 import cv2
 import numpy as np
+from numpy.typing import NDArray
 
 from dimos.core.module import Module
 from dimos.core.rpc_client import RpcCall
@@ -99,15 +101,13 @@ class InterpretMapSkill(SkillModule):
             "MAKE SURE there is a clear path from the robot's current position to the goal position without crossing any obstacles. \n"
             "MAKE SURE the goal position is located in the blue area (free space) of the map and few pixels away from obstacles or objects. \n"
             "Return ONLY a JSON object with this exact format:\n"
-            "{'point': [x, y]}\n"
+            '{"point": [x, y]}\n'
             f"where x,y are the pixel coordinates of the goal position in the image. \n"
         )
 
         response = self.vl_model.query(image, prompt)
         point = extract_json_from_llm_response(response)
-        if point is None or "point" not in point:
-            return "Failed to parse goal position from model response."
-        x, y = point["point"]
+        x, y = extract_coordinates(point)
 
         # ensure point is in free space, else choose nearest free space
         if not costmap.is_free_space(x, y):
@@ -131,8 +131,18 @@ class InterpretMapSkill(SkillModule):
         return goal_pose
 
 
+def extract_coordinates(point: dict[str, list[int]] | None) -> list[int]:
+    if point is None:
+        raise ValueError("Failed to parse goal position: response is None.")
+    if "point" not in point:
+        raise ValueError("Failed to parse goal position: missing 'point' key.")
+    if not isinstance(point["point"], list):
+        raise ValueError("Failed to parse goal position: 'point' is not a list.")
+    return point["point"]
+
+
 def debug_image_with_identified_point(
-    image_frame: np.ndarray, point: tuple[int, int], filepath: str
+    image_frame: NDArray[np.uint8], point: tuple[int, int], filepath: str
 ) -> None:
     """Utility to visualize identified points on the image for debugging."""
     debug_image = image_frame.copy()
--- a/dimos/agents2/skills/test_map_eval.py
+++ b/dimos/agents2/skills/test_map_eval.py
@@ -58,24 +58,20 @@ def build_occupancygrid_from_image(image: Image, resolution: float = 0.05) -> "O
         image_arr = image_arr[:, :, :3]
 
     # Define colors and threshold
-    RED = np.array([255, 0, 0])
-    BLUE = np.array([0, 0, 200])
+    RED = np.array([255, 0, 0], dtype=np.float32)
+    BLUE = np.array([0, 0, 200], dtype=np.float32)
     color_threshold = 20
 
-    for y in range(height):
-        for x in range(width):
-            pixel = image_arr[y, x]
+    # Convert to float32 for distance calculations
+    image_float = image_arr.astype(np.float32)
 
-            # calculate distances to target colors
-            red_dist = np.sqrt(np.sum((pixel.astype(np.float32) - RED) ** 2))
-            blue_dist = np.sqrt(np.sum((pixel.astype(np.float32) - BLUE) ** 2))
+    # Calculate distances to target colors using broadcasting
+    red_dist = np.sqrt(np.sum((image_float - RED) ** 2, axis=2))
+    blue_dist = np.sqrt(np.sum((image_float - BLUE) ** 2, axis=2))
 
-            # assign based on closest color within threshold
-            if red_dist <= color_threshold:
-                grid[y, x] = 100  # Obstacle
-            elif blue_dist <= color_threshold:
-                grid[y, x] = 0  # Free space
-            # else: remains -1 (unknown)
+    # Assign based on closest color within threshold
+    grid[red_dist <= color_threshold] = 100  # Obstacle
+    grid[blue_dist <= color_threshold] = 0  # Free space
 
     occupancy_grid = OccupancyGrid()
     occupancy_grid.info.width = width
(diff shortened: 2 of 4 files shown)


dimos/agents2/skills/interpret_map.py
```diff
@@ -0,0 +1,146 @@
...
+    @rpc
+    def stop(self) -> None:
+        super().stop()
+
+    def _on_local_costmap(self, costmap: OccupancyGrid) -> None:
+        self._latest_local_costmap = costmap
+        self._robot_pose = self.tf.get("world", "base_link")
+
+    @skill()
+    def get_goal_position(self, description: str | None = None) -> Vector3 | str:
```
paul-nechifor, line 59: I think the agents only work on strings. Vector3 is going to be turned into a string, so you might as well transform it here.

dimos/agents2/skills/interpret_map.py
```diff
@@ -0,0 +1,146 @@
...
+            "Return ONLY a JSON object with this exact format:\n"
+            "{'point': [x, y]}\n"
+            f"where x,y are the pixel coordinates of the goal position in the image. \n"
+        )
+
+        response = self.vl_model.query(image, prompt)
+        point = extract_json_from_llm_response(response)
+        if point is None or "point" not in point:
+            return "Failed to parse goal position from model response."
+        x, y = point["point"]
```
paul-nechifor: You would be surprised how imprecise LLM responses can be. For example, even if you tell it to return `{'point': [x, y]}` (btw, that's invalid JSON as single quotes are not allowed) it sometimes might return `{"point": {"x": 123, "y": 123}}`.

I'd also check if `point["point"]` is a list.

dimos/agents2/skills/test_map_eval.py
```diff
@@ -0,0 +1,218 @@
...
+            # calculate distances to target colors
+            red_dist = np.sqrt(np.sum((pixel.astype(np.float32) - RED) ** 2))
+            blue_dist = np.sqrt(np.sum((pixel.astype(np.float32) - BLUE) ** 2))
+
+            # assign based on closest color within threshold
+            if red_dist <= color_threshold:
+                grid[y, x] = 100  # Obstacle
+            elif blue_dist <= color_threshold:
+                grid[y, x] = 0  # Free space
+            # else: remains -1 (unknown)
```
paul-nechifor: Looping in Python is very slow. You can do this directly in Numpy.

```suggestion
    RED = np.array([255, 0, 0], dtype=np.float32)
    BLUE = np.array([0, 0, 200], dtype=np.float32)
    color_threshold = 20

    # Convert to float32 for distance calculations
    image_float = image_arr.astype(np.float32)

    # Calculate distances to target colors using broadcasting
    red_dist = np.sqrt(np.sum((image_float - RED) ** 2, axis=2))
    blue_dist = np.sqrt(np.sum((image_float - BLUE) ** 2, axis=2))

    # Assign based on closest color within threshold
    grid[red_dist <= color_threshold] = 100  # Obstacle
    grid[blue_dist <= color_threshold] = 0   # Free space
```

dimos/agents2/skills/test_map_eval.py
```diff
@@ -0,0 +1,218 @@
...
+
+def interpretability_prompt(question: str) -> str:
+    prompt = (
+        "Look at this image carefully \n"
+        "it represents a 2D occupancy grid map where,\n"
+        " - blue area is free space, \n"
+        " - yellow area is unknown space, \n"
+        " - red (and its shades) areas are obstacles/walls, \n"
+        " - green object represents the robot's position and points to the direction it is facing. \n"
+        f"Answer the following question based on this image: {question}\n"
```
paul-nechifor, line 133: You're using different prompts here from what the actual code uses. I think the prompt, querying, and response parsing should be extracted to a separate function and tested here as it's used.

It can be extracted as something like:

```python
def find_location_in_occupancy_map(vl_model: VlModel, description: str) -> Vector3 | None: ...
```

dimos/agents2/skills/test_map_eval.py
```diff
@@ -0,0 +1,218 @@
...
+        else:
+            debug_image_with_identified_point(
+                image.to_opencv(),
+                (x, y),
+                filepath=f"./debug_goal_placement_{test_map['map_id']}_{qna['query'].replace(' ', '_')}.png",
+            )
+
+    # assert score >= len(test_map["questions"]) * 0.7, (
+    #     f"Goal placement score too low: {score}/{len(test_map['questions'])}"
+    # )
```
paul-nechifor: It's not really a test if it doesn't assert anything. :)

Does this fail frequently? Maybe you should assert something different?

dimos/agents2/skills/test_map_eval.py
```diff
@@ -0,0 +1,218 @@
...
+@pytest.mark.parametrize(
+    "test_map",
+    [
+        test_map
+        for test_map in load_test_cases(TEST_DIR / "test_map_interpretability.yaml")[
+            "point_placement_tests"
+        ]
+    ],
+)
+def test_point_placement(test_map, vl_model):
```
paul-nechifor, line 147: These two tests take a long time because the ask a lot of questions:

```
43.82s call     dimos/agents2/skills/test_map_eval.py::test_map_comprehension[test_map0]
29.17s call     dimos/agents2/skills/test_map_eval.py::test_point_placement[test_map0]
```

but you can run them in parallel since they're requests, not local. In parallel they're much faster:

```
9.73s call     dimos/agents2/skills/test_map_eval.py::test_point_placement[test_map0]
8.56s call     dimos/agents2/skills/test_map_eval.py::test_map_comprehension[test_map0]
```

dimos/msgs/nav_msgs/OccupancyGrid.py
```diff
@@ -114,6 +128,8 @@ def __init__(
             self.info = MapMetaData(map_load_time=self._to_lcm_time())  # type: ignore[no-untyped-call]
             self.grid = np.array([], dtype=np.int8)
 
+        self.robot_pose = robot_pose  # type: ignore[assignment]
```
paul-nechifor: You don't need this ignore. The property is defined incorrectly as `Pose`. It should be `Pose | None`.

dimos/msgs/nav_msgs/OccupancyGrid.py
```diff
@@ -276,6 +292,254 @@ def __repr__(self) -> str:
...
+        """
+        pixel_x = round((grid_x / self.width) * size[0])
+        pixel_y = round((grid_y / self.height) * size[1])
+
+        if flip_vertical:
+            pixel_y = size[1] - pixel_y
+
+        return (pixel_x, pixel_y)
+
+    def pixel_to_grid(
```
paul-nechifor, line 384: It's probably not worth to do it now, but in the future I'd extract these to a class just for the image. Something like:

```python
class OccupancyGridImage:
    image: Image

    @classmethod
    def from_occupancy_grid(occupancy_grid: OccupancyGrid, size: ...) -> 'OccupancyGridImage': ...

    def grid_to_pixel(...):
    def pixel_to_grid(...):
```

dimos/msgs/nav_msgs/OccupancyGrid.py
```diff
@@ -276,6 +292,254 @@ def __repr__(self) -> str:
...
+
+        if flip_vertical:
+            pixel_y = size[1] - pixel_y
+
+        grid_x = (pixel_x / size[0]) * self.width
+        grid_y = (pixel_y / size[1]) * self.height
+
+        return self.grid_to_world(Vector3(grid_x, grid_y, 0.0))
+
+    def augment_image_with_robot_pose(self, image_arr: np.ndarray) -> np.ndarray:  # type: ignore[type-arg]
```
paul-nechifor: You shouldn't ever use `np.ndarray`. This should be:

```python
from numpy.typing import NDArray
def augment_image_with_robot_pose(self, image_arr: NDArray[np.uint8]) -> NDArray[np.uint8]:
```

dimos/msgs/nav_msgs/OccupancyGrid.py
```diff
@@ -276,6 +292,254 @@ def __repr__(self) -> str:
...
+            (rgx, rgy),
+            (rgx + arrow_dx, rgy + arrow_dy),
+            (0, 255, 0),
+            line_thickness,
+            tipLength=0.4,
+        )
+
+        return image_arr
+
+    def quat_to_yaw(self, q: Quaternion) -> np.ndarray:
```
paul-nechifor: This is unrelated to OccupancyGrid. It should probably be on Quaternion instead. But you can already do this with `q.euler[2]`

dimos/robot/unitree_webrtc/unitree_go2_blueprints.py
```diff
@@ -20,6 +20,7 @@
 from dimos.agents2.cli.human import human_input
 from dimos.agents2.cli.web import web_input
 from dimos.agents2.ollama_agent import ollama_installed
+from dimos.agents2.skills.interpret_map import interpret_map_skill
```
paul-nechifor: This is imported but not used.

dimos/agents2/skills/test_map_eval.py
s-desh: Yes, it is more an evaluation than a must pass test for now. There are only 3 questions for this test, more are needed before this assert is useful.

I'll reduce the score requirement for now.

dimos/agents2/skills/test_map_eval.py
s-desh: Can do this with pytest?

Also this is not an test for the skill, it is an eval and probably should not run on every CI trigger.

dimos/msgs/nav_msgs/OccupancyGrid.py
s-desh: Yeah, have a draft [here](https://github.com/dimensionalOS/dimos/pull/818/files#diff-cb13da6a12f1ca5a4b296d035b07d2f6e5e3b78ce954c51a98c9ddd3691ca538R39)

dimos/agents2/skills/test_map_eval.py
s-desh: Will match the prompts. These tests are evals and are not meant for the skill, as it will change soon. 

If you still think I should extract and test it, I can add a simple skill test for this func in another file.

dimos/agents2/skills/interpret_map.py
s-desh: I don't see the benefit of doing it here if agent_encode is being called in the background anyways.

dimos/agents2/skills/test_map_eval.py
```diff
@@ -95,19 +91,34 @@ def occupancy_grid_from_image(image_path) -> OccupancyGrid:
     return occupancy_grid
 
 
+def extract_coordinates(point: dict | None) -> list | str:
```
paul-nechifor, line 94: You have this function in dimos/agents2/skills/interpret_map.py as well.

dimos/agents2/skills/test_map_eval.py
paul-nechifor: You can parallelize it inside the test. Ask Claude rewrite the evals to run in parallel.

dimos/agents2/skills/interpret_map.py
paul-nechifor: I guess the argument is that it makes it more clear what the agent receives. I'm not sure how Vector3 is turned into a string, but if it's by calling `str(vector)` it ends up with an unnecessary Unicode arrow encoding.

PR #819, opened 2025-12-09
models/ refactor

leshy:
- adds ignore missing imports for ros/mujoco
 
- writes consistent type spec for models
    visual language models `models/vl/base.py`
    embedding models `models/embedding/base.py`
    local models `models/base.py`
    huggingface models `models/base.py` 

```mermaid
classDiagram
    class LocalModel {
        +device
        +dtype
        +_model
    }

    class HuggingFaceModel {
        +model_name
        +from_pretrained()
    }

    class Captioner {
        <<ABC>>
        +caption(image)
    }

    class VlModel {
        +query(image, query)
        +query_json()
        +query_detections()
    }

    class EmbeddingModel~E~ {
        <<ABC>>
        +embed(*images)
        +embed_text(*texts)
    }

    class Embedding {
        +vector
        +to_numpy()
        +to_torch()
    }

    LocalModel <|-- HuggingFaceModel

    Captioner <|-- VlModel

    Captioner <|-- Florence2Model
    HuggingFaceModel <|-- Florence2Model

    EmbeddingModel <|-- CLIPModel
    HuggingFaceModel <|-- CLIPModel

    EmbeddingModel <|-- MobileCLIPModel
    LocalModel <|-- MobileCLIPModel

    VlModel <|-- Qwen : remote API
    VlModel <|-- Moondream : local GPU
    VlModel <|-- MoondreamHosted : remote API

    EmbeddingModel ..> Embedding : produces


```

- comprehensive parametric tests for all `vl/` `embedding/` and `captioning/` models (you don't write your own tests for a new module, just add it to a list) 

- ensures models are resources, so have start() and stop() methods that clean up cuda resources, adds optional argument for warmup when initializing

- adds methods for batch querying of models (for speed)

- adds optional auto-resize of input images to models (for speed)

- adds florence2 fast captioning model (implements captioning model type, all vlms are captioners, not all captioners are vlms) 

- adds torch reid models to lfs

- adds mobileclip models to lfs

- cleans up dead code (models/segmentation, pointcloud, labels, qwen)

- adds a quick cli graphing system at dimos/utils/cli/plot.py for profiling

### DIFF
--- a/dimos/models/vl/base.py
+++ b/dimos/models/vl/base.py
@@ -187,10 +187,6 @@ class VlModel(Captioner, Resource, Configurable[VlModelConfig]):
         except Exception:
             pass
 
-    def stop(self) -> None:
-        """Stop the model (Resource interface). Override in subclasses for cleanup."""
-        pass
-
     # requery once if JSON parsing fails
     @retry(max_retries=2, on_exception=json.JSONDecodeError, delay=0.0)  # type: ignore[misc, untyped-decorator]
     def query_json(self, image: Image, query: str) -> dict:  # type: ignore[type-arg]
@@ -225,7 +221,11 @@ class VlModel(Captioner, Resource, Configurable[VlModelConfig]):
 
         for track_id, detection_tuple in enumerate(detection_tuples):
             # Scale coordinates back to original image size if resized
-            if scale != 1.0 and isinstance(detection_tuple, (list, tuple)) and len(detection_tuple) == 5:
+            if (
+                scale != 1.0
+                and isinstance(detection_tuple, (list, tuple))
+                and len(detection_tuple) == 5
+            ):
                 detection_tuple = [
                     detection_tuple[0],  # label
                     detection_tuple[1] / scale,  # x1
--- a/dimos/models/vl/moondream.py
+++ b/dimos/models/vl/moondream.py
@@ -26,16 +26,10 @@ class MoondreamConfig(HuggingFaceModelConfig):
     auto_resize: tuple[int, int] | None = MOONDREAM_DEFAULT_AUTO_RESIZE
 
 
-class MoondreamVlModel(VlModel, HuggingFaceModel):
+class MoondreamVlModel(HuggingFaceModel, VlModel):
     _model_class = AutoModelForCausalLM
     default_config = MoondreamConfig  # type: ignore[assignment]
     config: MoondreamConfig  # type: ignore[assignment]
-
-    def stop(self) -> None:
-        """Release model and free GPU memory."""
-        # Call LocalModel.stop() which handles _model cleanup and cuda cache
-        HuggingFaceModel.stop(self)
-        VLModel.stop(self)
 
     @cached_property
     def _model(self) -> AutoModelForCausalLM:
--- a/dimos/models/vl/test_captioner.py
+++ b/dimos/models/vl/test_captioner.py
@@ -27,21 +27,21 @@ def test_image() -> Image:
     return Image.from_file(get_data("cafe.jpg")).to_rgb()
 
 
-def model_fixture(model_type: type[M]) -> Generator[M, None, None]:
-    model = model_type()
-    model.start()
-    yield model
-    model.stop()
+def generic_model_fixture(model_type: type[M]) -> Generator[M, None, None]:
+    model_instance = model_type()
+    model_instance.start()
+    yield model_instance
+    model_instance.stop()
 
 
 @pytest.fixture(params=[Florence2Model, MoondreamVlModel])
 def captioner_model(request: pytest.FixtureRequest) -> Generator[CaptionerModel, None, None]:
-    yield from model_fixture(request.param)
+    yield from generic_model_fixture(request.param)
 
 
 @pytest.fixture(params=[Florence2Model])
 def florence2_model(request: pytest.FixtureRequest) -> Generator[Florence2Model, None, None]:
-    yield from model_fixture(request.param)
+    yield from generic_model_fixture(request.param)
 
 
 @pytest.mark.gpu
--- a/dimos/models/vl/moondream_hosted.py
+++ b/dimos/models/vl/moondream_hosted.py
@@ -131,3 +131,6 @@ class MoondreamHostedVlModel(VlModel):
 
         return pixel_points
 
+    def stop(self) -> None:
+        pass
+


dimos/models/vl/base.py
```diff
@@ -93,12 +215,24 @@ def query_detections(self, image: Image, query: str, **kwargs) -> ImageDetection
 
         image_detections = ImageDetections2D(image)
 
+        # Get scale factor for coordinate rescaling
+        _, scale = self._prepare_image(image)
```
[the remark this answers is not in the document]
leshy: fixed

dimos/models/embedding/mobileclip.py
```diff
@@ -12,55 +12,55 @@
...
+    default_config = MobileCLIPModelConfig
+    config: MobileCLIPModelConfig
 
-        # Load model
-        pretrained = str(model_path) if model_path else None
-        self.model, _, self.preprocess = open_clip.create_model_and_transforms(
-            model_name, pretrained=pretrained
+    @cached_property
+    def _model_and_preprocess(self) -> tuple[Any, Any]:
+        model_path = get_data("models_mobileclip") / (self.config.model_name + ".pt")
```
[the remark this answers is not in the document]
leshy: true!

dimos/models/vl/florence.py
```diff
@@ -0,0 +1,170 @@
...
+    ) -> None:
+        """Initialize Florence-2 model.
+
+        Args:
+            model_name: HuggingFace model name. Options:
+                - "microsoft/Florence-2-base" (~0.2B, fastest)
+                - "microsoft/Florence-2-large" (~0.8B, better quality)
+            **kwargs: Additional config options (device, dtype, warmup, etc.)
+        """
+        HuggingFaceModel.__init__(self, model_name=model_name, **kwargs)
```
paul-nechifor: Personally, I don't like multiple inheritance because it leads to complicated hierarchies where everything operates in the same namespace. Why not compose instead of inherit:

```
super().__init__()
self._model = HuggingFaceModel(model_name=model_name, **kwargs)
```

leshy: In this PR multiple inheritance is not due to functionality, but typing.
here I'm specifying that this model is a Captioner (empty ABC)

(I pushed a commit to flip the deps so I can use `super().__init__`)

same here, `Captioner` and `Resource` are `ABC`s
```python
class VlModel(Captioner, Resource, Configurable[VlModelConfig]):
   ...
```

I could use `Protocol` also, but not totally sold on them, we can see what a recomendation for modern py is and set a policy for this globally

can have a policy on ABC class naming like AResource and ACaptioner so it's clear when looking at a class

dimos/models/depth/metric3d.py
```diff
@@ -12,32 +12,50 @@
...
+class Metric3DConfig(LocalModelConfig):
+    """Configuration for Metric3D depth estimation model."""
+
+    camera_intrinsics: list[float] = field(default_factory=lambda: [500.0, 500.0, 320.0, 240.0])
+    """Camera intrinsics [fx, fy, cx, cy]."""
+
+    gt_depth_scale: float = 256.0
+    """Scale factor for ground truth depth."""
+
+    device: str = "cuda"
```
paul-nechifor: Should this be `device: DeviceType = "cuda" if torch.cuda.is_available() else "cpu"`?

dimos/models/depth/test_metric3d.py
```diff
@@ -0,0 +1,87 @@
...
+    model = Metric3D(camera_intrinsics=sample_intrinsics)
+
+    with pytest.raises(ValueError, match="Intrinsic must be a list"):
+        model.update_intrinsic([1.0, 2.0])  # Only 2 values
+
+
+@pytest.mark.gpu
+def test_metric3d_infer_depth(sample_intrinsics: list[float]) -> None:
+    """Test depth inference on a sample image."""
+    model = Metric3D(camera_intrinsics=sample_intrinsics)
```
paul-nechifor, line 46: It would be nicer to make model a fixture in these tests so you can automatically close it even if the test fails.

dimos/models/vl/moondream.py
```diff
@@ -1,64 +1,68 @@
...
+
+class MoondreamVlModel(VlModel, HuggingFaceModel):
+    _model_class = AutoModelForCausalLM
+    default_config = MoondreamConfig  # type: ignore[assignment]
+    config: MoondreamConfig  # type: ignore[assignment]
+
+    def stop(self) -> None:
+        """Release model and free GPU memory."""
+        # Call LocalModel.stop() which handles _model cleanup and cuda cache
+        HuggingFaceModel.stop(self)
```
paul-nechifor: This doesn't call `VlModel`'s stop.

This is why I dislike multiple inheritance. It currently works because `VlModel.stop` doesn't do anything. But in the future someone might add something there. The code will work if it's only tested with `Florence2Model` because it calls `super().stop()`, but it won't work here.

dimos/utils/cli/plot.py
```diff
@@ -0,0 +1,282 @@
...
+import plotext as plt  # type: ignore[import-untyped]
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+
+def _default_size() -> tuple[int, int]:
+    """Return default plot size (terminal width, half terminal height)."""
+    # tw, th = plt.terminal_size()
+    # return tw, th // 2
```
paul-nechifor: Should this be uncommented?

dimos/models/vl/test_captioner.py
```diff
@@ -0,0 +1,89 @@
...
+    print(f"\n{model_name} captioning test passed!")
+
+
+@pytest.mark.gpu
+def test_florence2_detail_levels() -> None:
+    """Test Florence-2 different detail levels."""
+    image = Image.from_file(get_data("cafe.jpg")).to_rgb()
+
+    model = Florence2Model()
+    model.start()
```
paul-nechifor: The model should probably be a fixture. Also you should probably call `model.stop()`.
The image is used in multiple tests so it could be a fixture too.

dimos/models/vl/florence.py
```diff
@@ -0,0 +1,170 @@
...
+                num_beams=3,
+                do_sample=False,
+            )
+
+        # Decode all
+        generated_texts = self._processor.batch_decode(generated_ids, skip_special_tokens=False)
+
+        # Parse outputs
+        captions = []
+        for text, pil_img in zip(generated_texts, pil_images, strict=False):
```
paul-nechifor: `strict=False` is the default, but it silences errors if the two lists don't match. I think you want `strict=True`.

dimos/models/depth/test_metric3d.py
```diff
@@ -0,0 +1,87 @@
...
+    assert depth_map.min() >= 0  # Depth should be non-negative
+
+    print(f"Depth map shape: {depth_map.shape}")
+    print(f"Depth range: [{depth_map.min():.2f}, {depth_map.max():.2f}]")
+
+    model.stop()
+
+
+@pytest.mark.gpu
+def test_metric3d_multiple_inferences(sample_intrinsics: list[float]) -> None:
```
spomichter, line 69: Should also test metric3d across multiple processes. Generally need to figure out how to handle models with multithreading as each process will be loading its own version of the model by default

leshy: Yeah didn't want to touch metric3d much, (idk if we have a use for it also) just wanted to bring it up to speed in terms of correct typing/config. I'd expect to have a single module using a module 99% of the time, if not we need a way to easily deploy a model as a Module

dimos/models/depth/metric3d.py
leshy: yes

PR #829, opened 2025-12-12
DOCS: Modules, Transforms, Configuration, Tooling for documentation

leshy:
# 2 new tools:

- doclinks https://github.com/dimensionalOS/dimos/blob/4f1eed2181962f4589a2ff9342e5d7189c413a0f/dimos/utils/docs/doclinks.md
keeps links in docs up to date, runs via git-commit-hooks

- md-babel-py (sorry about my naming) https://github.com/leshy/md-babel-py
executes code in docs and appends output

# first pass
module, transforms, config, reactivex, documentation

new dir for agent documents at `docs/agent/` and first pass on docs-writing-agent instructions

# sikkk autogenerated graphics

### DIFF
--- a/dimos/core/stream.py
+++ b/dimos/core/stream.py
@@ -140,11 +140,9 @@ class Stream(Generic[T]):
 
 class Out(Stream[T], ObservableMixin[T]):
     _transport: Transport  # type: ignore[type-arg]
-    _local_subscribers: list[Callable[[T], None]]
 
     def __init__(self, *argv, **kwargs) -> None:  # type: ignore[no-untyped-def]
         super().__init__(*argv, **kwargs)
-        self._local_subscribers = []
         self._rerun_config: dict | None = None  # type: ignore[type-arg]
         self._rerun_last_log: float = 0.0
 
@@ -173,15 +171,7 @@ class Out(Stream[T], ObservableMixin[T]):
             ),
         )
 
-    def subscribe(self, cb) -> Callable[[], None]:
-        self._local_subscribers.append(cb)
-        return lambda: self.local_subscribers.remove(cb)
-
     def publish(self, msg) -> None:  # type: ignore[no-untyped-def]
-        if self._local_subscribers:
-            for cb in self._local_subscribers:
-                cb(msg)
-
         if not hasattr(self, "_transport") or self._transport is None:
             logger.warning(f"Trying to publish on Out {self} without a transport")
             return
@@ -191,6 +181,69 @@ class Out(Stream[T], ObservableMixin[T]):
             self._log_to_rerun(msg)
 
         self._transport.broadcast(self, msg)
+
+    def subscribe(self, cb) -> Callable[[], None]:  # type: ignore[no-untyped-def]
+        """Subscribe to this output stream.
+
+        Args:
+            cb: Callback function to receive messages
+
+        Returns:
+            Unsubscribe function
+        """
+        return self.transport.subscribe(cb, self)  # type: ignore[arg-type, func-returns-value, no-any-return]
+
+    def autolog_to_rerun(
+        self,
+        entity_path: str,
+        rate_limit: float | None = None,
+        **rerun_kwargs: Any,
+    ) -> None:
+        """Configure this output to auto-log to Rerun (fire-and-forget).
+
+        Call once in start() - messages auto-logged when published.
+
+        Args:
+            entity_path: Rerun entity path (e.g., "world/map")
+            rate_limit: Max Hz to log (None = unlimited)
+            **rerun_kwargs: Passed to msg.to_rerun() for rendering config
+                           (e.g., radii=0.02, colormap="turbo", colors=[255,0,0])
+
+        Example:
+            def start(self):
+                super().start()
+                # Just declare it - fire and forget!
+                self.global_map.autolog_to_rerun("world/map", rate_limit=5.0, radii=0.02)
+        """
+        self._rerun_config = {
+            "entity_path": entity_path,
+            "rate_limit": rate_limit,
+            "rerun_kwargs": rerun_kwargs,
+        }
+        self._rerun_last_log = 0.0
+
+    def _log_to_rerun(self, msg: T) -> None:
+        """Log message to Rerun with rate limiting."""
+        if not hasattr(msg, "to_rerun"):
+            return
+
+        if self._rerun_config is None:
+            return
+
+        import time
+
+        config = self._rerun_config
+
+        # Rate limiting
+        if config["rate_limit"] is not None:
+            now = time.monotonic()
+            min_interval = 1.0 / config["rate_limit"]
+            if now - self._rerun_last_log < min_interval:
+                return  # Skip - too soon
+            self._rerun_last_log = now
+
+        rerun_data = msg.to_rerun(**config["rerun_kwargs"])
+        rr.log(config["entity_path"], rerun_data)
 
 
 class RemoteStream(Stream[T]):
@@ -248,7 +301,8 @@ class In(Stream[T], ObservableMixin[T]):
         ...
 
     def connect(self, value: Out[T]) -> None:
-        value.subscribe(self.transport.publish)  # type: ignore[arg-type]
+        # just for type checking
+        ...
 
     @property
     def state(self) -> State:


dimos/core/introspection/blueprint/dot.py
```diff
@@ -0,0 +1,236 @@
+# Copyright 2025 Dimensional Inc.
```
paul-nechifor: Is `dot` used? Should it be deleted and `dot2` renamed to `dot`?

dimos/core/introspection/module/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+    "SkillInfo",
+    "StreamInfo",
+    "ansi",
+    "dot",
+    "extract_module_info",
+    "render_module_io",
+]
+
+
+def render_module_io(
```
paul-nechifor: I think `__init__.py` should be just for exporting, and not contain actual code.

But I also dislike re-exporting because it creates ambiguity with respect to where something should be imported from. For example `render_module_io` can be imported from `dimos.core.introspection.module`, but also from `dimos.core.introspection`.

This always leaves people with the choice of where to import from.

Importing from `dimos.core.introspection` gives people shorter imports, but I don't see the benefit of shorter imports over unambiguous imports.

paul-nechifor: > This always leaves people with the choice of where to import from.

I've also been burned by this when files have the same name as the thing they export.

For example, if `dimos/msg/ClassA.py` contains `class ClassA`, and you use `ClassA` in your code, in an IDE you're given the choice of importing:

```
from dimos.msg.ClassA import ClassA
# or
from dimos.msg import ClassA
```

Potentially, both work, if `dimos/msg/__init__.py` exports `ClassA` from `ClassA.py`. But if it doesn't, then `from dimos.msg import ClassA` imports the module instead of the class.

I've encountered this issue several times which forces me to always verify I've imported the right thing or always use the long import (`from dimos.msg.ClassA import ClassA`) as that's always safer.

---

Having multiple ways of importing is also a source of Git conflicts.

dimos/core/introspection/module/dot.py
```diff
@@ -0,0 +1,234 @@
...
+    h = int(hashlib.md5(s.encode()).hexdigest(), 16)
+    return colors[h % len(colors)]
+
+
+def _sanitize_id(s: str) -> str:
+    """Sanitize a string to be a valid graphviz node ID."""
+    return re.sub(r"[^a-zA-Z0-9_]", "_", s)
+
+
+# Colors for type nodes (same as dot2 for consistency)
```
paul-nechifor: `dot2`'s `TYPE_COLORS` has more colors.

Should these constants be moved to a `constants.py` file and used by both?

docs/api/configuration.md
```diff
@@ -0,0 +1,90 @@
+# Configuration
+
+Dimos provides a `Configurable` base class, see [`service/spec.py`](/dimos/protocol/service/spec.py#L22).
```
paul-nechifor, line 3: I've seen the previous docs also link to specific lines in the code. But how do you guarantee that this gets updated when the code changes? Is there a tool that monitors that the links are correct?

docs/api/configuration.md
```diff
@@ -0,0 +1,90 @@
...
+
+@dataclass
+class Config():
+    x: int = 3
+    hello: str = "world"
+
+class MyClass(Configurable):
+    default_config = Config
+    config: Config
+    def __init__(self, **kwargs: object) -> None:
```
paul-nechifor: Should be `**kwargs: Any` not `object`.

`Any` essentially disables the type checker. `object` is a specific type. Using `kwargs['x'] + 100` is probably a type error since object doesn't do additions, but with `Any` you can do addition because `kwargs['x']` *could* be a number.

docs/api/reactivex.md
```diff
@@ -0,0 +1,533 @@
...
+    ops.take(10),
+    ops.sample(0.2),
+    ops.to_list(),
+).run()
+print("sample() got:", results)
+```
+
+<!--Result:-->
+```
+sample() got: [2, 6, 9]
```
paul-nechifor, line 135: The comment below implies this returns [3, 7]:

> sample(0.2s) -> gets value at 200ms, 400ms marks -> [3, 7]

docs/api/transforms.md
```diff
@@ -0,0 +1,469 @@
...
+
+## The TF Service
+
+Every module has access to `self.tf`, a transform service that:
+
+- **Publishes** transforms to the system
+- **Looks up** transforms between any two frames
+- **Buffers** historical transforms for temporal queries
+
+The TF service is implemented in [`tf.py`](/dimos/protocol/tf/tf.py) and is lazily initialized on first access.
```
paul-nechifor, line 210: I've noticed that I have to do:

```
def start(...):
    _ = self.tf
```

so I have access to the position when I first need it:

```
@skill
def some_skill(...):
    tf = self.tf.get("world", "base_link")
```

If I don't, `.get` returns None.

paul-nechifor: Oh, you mentioned it too below.

docs/api/transforms.md
```diff
@@ -0,0 +1,469 @@
...
+┌▽──────────┐
+│camera_link│
+└┬──────────┘
+┌▽─────────────┐
+│camera_optical│
+└──────────────┘
+```
+
+
+You can also run `foxglove-bridge` in the next terminal (binary provided by dimos and should be in your py env) and `foxglove-studio` to view these transforms in 3D
```
paul-nechifor: Foxglove Studio used to be called Foxglove Bridge. Maybe we should call this command `foxglove-studio-bridge`.

docs/concepts/modules.md
```diff
@@ -0,0 +1,175 @@
+
+
+
+# Dimos Modules
+
+Module is a subsystem on a robot that operates autonomously and communicates to other subsystems using standardized messages
```
paul-nechifor:
```suggestion
Modules are subsystems on a robot that operate autonomously and communicate to other subsystems using standardized messages.
```

docs/concepts/modules.md
```diff
@@ -0,0 +1,175 @@
...
+
+Some examples of are:
```
paul-nechifor:
```suggestion
Some examples of modules are:
```

docs/concepts/modules.md
```diff
@@ -0,0 +1,175 @@
...
+
+- Webcam (outputs image)
+- Navigation (inputs a map and a target, outputs a path)
+- Detection (takes an image and a vision model like yolo, outputs a stream of detections)
```
paul-nechifor:
```suggestion
- Detection (takes an image and a vision model like YOLO, outputs a stream of detections)
```

docs/concepts/modules.md
```diff
@@ -0,0 +1,175 @@
...
+
+A common module structure for controling a robot looks something like this, black blocks are modules, colored lines are connections and message types, it's ok if this doesn't make sense now,
+it will by the end of this document.
```
paul-nechifor:
```suggestion
Below is an example of a structure for controlling a robot. Black blocks represent modules and colored lines are connections and message types. It's okay if this doesn't make sense now, it will by the end of this document.
```

docs/concepts/modules.md
```diff
@@ -0,0 +1,175 @@
...
+└┬─────────────┘
+ ├─ [93mcolor_image[0m: [92mImage[0m
+ ├─ [93mcamera_info[0m: [92mCameraInfo[0m
+ │
+ ├─ RPC [94mset_transport[0m(stream_name: [92mstr[0m, transport: [92mTransport[0m) -> [92mbool[0m
+ ├─ RPC [94mstart[0m() -> [92mstr[0m
+ ├─ RPC [94mstop[0m() -> [92mNone[0m
+ │
+ ├─ Skill [96mvideo_stream[0m (stream=passive, reducer=latest_reducer, output=image)
+```
```
paul-nechifor, line 54: Unfortunately, this is unreadable without the color codes being interpreted. I think you should strip the codes with `ansi2txt`, but that means the python print() doesn't produce the same output.

docs/concepts/modules.md
```diff
@@ -0,0 +1,175 @@
...
+We can see that camera module outputs two streams:
+
+- `color_image` with [sensor_msgs.Image](https://docs.ros.org/en/melodic/api/sensor_msgs/html/msg/Image.html) type
+- `camera_info` with [sensor_msgs.CameraInfo](https://docs.ros.org/en/melodic/api/sensor_msgs/html/msg/CameraInfo.html) type
+
+Offers two RPC calls, `start()` and `stop()`
+
+As well as an agentic [Skill][skills.md] called `video_stream` (more about this later, in [Skills Tutorial][skills.md])
+
+We can start this module and explore the output of it's streams in real time (this will use your webcam)
```
paul-nechifor:
```suggestion
We can start this module and explore the output of its streams in real time (this will use your webcam).
```

There are quite a few typos and punctuation issues in this file. You should pass it through an AI agent to fix them.

docs/api/configuration.md
leshy: Yes! I wrote `doclinks` tool which auto-finds stuff,

https://github.com/dimensionalOS/dimos/blob/da513d64a15a35d554c41ce81400c87aa61711b2/docs/agents/docs/doclinks.md

needs to go to pre-commit hooks & ci

docs/concepts/modules.md
leshy: aha yes in the next version I auto-strip, there is a config setting for my tool to remove these

docs/api/transforms.md
leshy: we might want to auto start tf service everywhere or toggle it via config, if this is a recurring issue

dimos/core/introspection/module/__init__.py
leshy: yeah agreed, this dot stuff is actually completely vibed and didn't even read the code, wasn't ready for review :D :D

dimos/core/introspection/blueprint/dot.py
leshy: yes

dimos/core/introspection/blueprint/dot.py
```diff
@@ -0,0 +1,253 @@
...
+            f'width=0, height=0, margin="0.1,0.05", fontsize=10];'
+        )
+
+    lines.append("")
+
+    # Add edges: producer -> type_node -> consumer
+    lines.append("    // Edges")
+    for key, color in sorted(
+        active_channels.items(), key=lambda x: f"{x[0][0]}:{x[0][1].__name__}"
+    ):
```
paul-nechifor, line 205: Same `sorted` as in the above for loop. Might as well save and reuse.

dimos/core/module.py
```diff
@@ -198,64 +203,124 @@ def inputs(self) -> dict[str, In]:  # type: ignore[type-arg]
...
 
-    @classmethod  # type: ignore[misc]
-    @property
-    def rpcs(cls) -> dict[str, Callable]:  # type: ignore[type-arg]
+    @classproperty
+    def rpcs(self) -> dict[str, Callable[..., Any]]:
         return {
-            name: getattr(cls, name)
-            for name in dir(cls)
+            name: getattr(self, name)
```
paul-nechifor, line 209: Why rename cls to self?

dimos/robot/unitree/connection/go2.py
```diff
@@ -287,28 +287,10 @@ def _odom_to_tf(cls, odom: PoseStamped) -> list[Transform]:
...
-            ts=odom.ts,
-        )
-
-        map_to_world = Transform(
-            translation=Vector3(0.0, 0.0, 0.0),
-            rotation=Quaternion(0.0, 0.0, 0.0, 1.0),
-            frame_id="map",
-            child_frame_id="world",
-            ts=time.time(),
-        )
```
paul-nechifor, line 304: I added this because it's required for `tf = self.tf.get("map", "base_link")` in `perception/spatial_perception.py`, I think.

docs/concepts/transports.md
```diff
@@ -0,0 +1,368 @@
...
+# The interface every transport must implement:
+import inspect
+print(inspect.getsource(PubSub.publish))
+print(inspect.getsource(PubSub.subscribe))
+```
+
+<!--Error:-->
+```
+Session process exited unexpectedly:
+/home/lesh/coding/dimos/.venv/bin/python3: No module named md_babel_py.session_server
```
paul-nechifor, line 24: This is unintentional, right?

docs/concepts/transports.md
```diff
@@ -0,0 +1,368 @@
...
+## Available Transports
+
+Dimos includes several transport implementations:
+
+| Transport | Use Case | Process Boundary | Network |
+|-----------|----------|------------------|---------|
+| `Memory` | Testing, single process | No | No |
+| `SharedMemory` | Multi-process on same machine | Yes | No |
+| `LCM` | Network communication (UDP multicast) | Yes | Yes |
+| `Redis` | Network communication via Redis server | Yes | Yes |
```
paul-nechifor, line 78: I saw we have this, but is it actually used? I don't see redis installed.

docs/concepts/modules.md
```diff
@@ -0,0 +1,176 @@
...
+ ├─ detected_image_0: Image
+ ├─ detected_image_1: Image
+ ├─ detected_image_2: Image
+ │
+ ├─ RPC set_transport(stream_name: str, transport: Transport) -> bool
+ ├─ RPC start() -> None
+ ├─ RPC stop() -> None
+```
+
+TODO: add easy way to print config
```
paul-nechifor, line 123: Do you want to do this before the merge?

docs/concepts/modules.md
```diff
@@ -0,0 +1,176 @@
...
+
+looks like detector just needs an image input, outputs some sort of detection and annotation messages, let's connect it to a camera.
+
+```pythonx ansi=false
```
paul-nechifor, line 127: What is a typo, or does pythonx do something different?

dimos/utils/docs/doclinks.md
```diff
@@ -0,0 +1,96 @@
...
+
+# GitHub links
+doclinks --link-mode github \
+  --github-url https://github.com/org/repo docs/
+
+# Dry run (preview changes)
+doclinks --dry-run docs/
+
+# CI check (exit 1 if changes needed)
+doclinks --check docs/
```
paul-nechifor, line 58: Is this actually added to CI?

PR #831, opened 2025-12-14
Generalized manipulator driver

mustafab0:
Introduces component-based architecture for manipulator drivers enabling rapid integration of new robot arms.

- **BaseManipulatorSDK**: Abstract interface for vendor SDK wrappers (standardized units: radians, meters)
- **BaseManipulatorDriver**: 2-thread architecture (100Hz control loop, 10Hz monitor) with DIMOS Module integration
- **Standard Components**: Reusable Motion, Servo, Status components with `@component_api` decorator for auto RPC exposure
- **Implementations**: XArm and Piper drivers using new architecture (~50 lines each)
- **Documentation**: How-to README and architecture reference

Adding new arms now requires only an SDK wrapper + thin driver assembly.

### DIFF
--- a/dimos/hardware/manipulators/test_integration_runner.py
+++ b/dimos/hardware/manipulators/test_integration_runner.py
@@ -173,7 +173,7 @@ class MockSDK(BaseManipulatorSDK):
 # =============================================================================
 
 
-def test_connection(driver, hardware: bool) -> bool:
+def check_connection(driver, hardware: bool) -> bool:
     """Test that driver connects to hardware/mock."""
     print("Testing connection...")
 
@@ -190,7 +190,7 @@ def test_connection(driver, hardware: bool) -> bool:
     return True
 
 
-def test_read_joint_state(driver, hardware: bool) -> bool:
+def check_read_joint_state(driver, hardware: bool) -> bool:
     """Test reading joint state."""
     print("Testing read joint state...")
 
@@ -215,7 +215,7 @@ def test_read_joint_state(driver, hardware: bool) -> bool:
     return True
 
 
-def test_get_robot_state(driver, hardware: bool) -> bool:
+def check_get_robot_state(driver, hardware: bool) -> bool:
     """Test getting robot state."""
     print("Testing robot state...")
 
@@ -232,7 +232,7 @@ def test_get_robot_state(driver, hardware: bool) -> bool:
     return True
 
 
-def test_servo_enable_disable(driver, hardware: bool) -> bool:
+def check_servo_enable_disable(driver, hardware: bool) -> bool:
     """Test enabling and disabling servos."""
     print("Testing servo enable/disable...")
 
@@ -269,7 +269,7 @@ def test_servo_enable_disable(driver, hardware: bool) -> bool:
     return True
 
 
-def test_joint_limits(driver, hardware: bool) -> bool:
+def check_joint_limits(driver, hardware: bool) -> bool:
     """Test getting joint limits."""
     print("Testing joint limits...")
 
@@ -292,7 +292,7 @@ def test_joint_limits(driver, hardware: bool) -> bool:
     return True
 
 
-def test_stop_motion(driver, hardware: bool) -> bool:
+def check_stop_motion(driver, hardware: bool) -> bool:
     """Test stop motion command."""
     print("Testing stop motion...")
 
@@ -313,7 +313,7 @@ def test_stop_motion(driver, hardware: bool) -> bool:
     return True
 
 
-def test_small_motion(driver, hardware: bool) -> bool:
+def check_small_motion(driver, hardware: bool) -> bool:
     """Test a small joint motion (5 degrees on joint 1).
 
     WARNING: With --hardware, this MOVES the real robot!
@@ -513,18 +513,18 @@ def run_tests(
 
     # Define tests (stop_motion last since it leaves arm in stopped state)
     tests = [
-        ("connection", test_connection),
-        ("read_state", test_read_joint_state),
-        ("robot_state", test_get_robot_state),
-        ("joint_limits", test_joint_limits),
-        # ("servo", test_servo_enable_disable),
+        ("connection", check_connection),
+        ("read_state", check_read_joint_state),
+        ("robot_state", check_get_robot_state),
+        ("joint_limits", check_joint_limits),
+        # ("servo", check_servo_enable_disable),
     ]
 
     if not skip_motion:
-        tests.append(("motion", test_small_motion))
+        tests.append(("motion", check_small_motion))
 
     # Stop test always last (leaves arm in stopped state)
-    tests.append(("stop", test_stop_motion))
+    tests.append(("stop", check_stop_motion))
 
     # Run tests
     results = {}


tests/test_xarm_driver_unit.py
```diff
@@ -0,0 +1,344 @@
+# Copyright 2025 Dimensional Inc.
```
spomichter: Tests/ is deprecated. move this and implement as integration test

tests/test_xarm_rt_driver.py
```diff
@@ -0,0 +1,385 @@
+# Copyright 2025 Dimensional Inc.
```
spomichter: Same here

mustafab0: Uploaded tests

dimos/manipulation/planning/trajectory_generator/joint_trajectory_generator.py
```diff
@@ -0,0 +1,453 @@
...
+        # Use simple scaling: scale both v and a by the same factor
+        # This preserves the profile shape
+        scale = t_opt / duration
+
+        # For a symmetric trapezoidal/triangular profile:
+        # If we scale time by k, we need to scale velocity by 1/k
+        # But we also need to ensure we travel the same distance
+
+        # Simpler approach: compute the average velocity needed
+        distance / duration
```
paul-nechifor, line 300: I assume this was `some_variable = distance / duration`, but `some_variable` was removed by the linter because it wasn't used. I guess the whole line should be removed.

dimos/manipulation/control/trajectory_setter.py
```diff
@@ -0,0 +1,446 @@
...
+
+        # Will be set dynamically from joint_state
+        self.num_joints: int | None = None
+        self.generator: JointTrajectoryGenerator | None = None
+
+        print(f"TrajectorySetter initialized for {self.arm_type.upper()}")
+        print("  Publishing to: /trajectory")
+        print(f"  Subscribing to: {joint_state_topic}")
+
+    def start(self) -> bool:
```
paul-nechifor, line 84: I think you should add a stop method as well were you can call `self.trajectory_pub.lcm.stop()`.

dimos/manipulation/control/trajectory_controller/spec.py
```diff
@@ -0,0 +1,88 @@
...
+
+def get_status():
+    """
+    Get the current status of the trajectory execution.
+    Returns a TrajectoryStatus message with details.
+     "state": "IDLE" | "EXECUTING" | "COMPLETED" | "ABORTED" | "FAULT",
+      "progress": float in [0,1],
+      "active_traj_id": Optional[str],
+      "error": Optional[str],
+    """
```
paul-nechifor: What are these empty functions for?

dimos/manipulation/control/trajectory_controller/spec.py
```diff
@@ -0,0 +1,88 @@
...
+
+        Returns:
+            Tuple of (q_ref, qd_ref):
+                - q_ref: Joint positions (radians)
+                - qd_ref: Joint velocities (rad/s)
+        """
+        ...
+
+
+class TrajectoryStatus(Protocol):
```
paul-nechifor: These are protocols for the message classes with the same name, right? But they are never used. Protocols are generally used if you have code that can use multiple class and you need to specify what they have in common. Since there's only one TrajectoryStatus, there's no need for a protocol.

dimos/manipulation/control/trajectory_controller/joint_trajectory_controller.py
```diff
@@ -0,0 +1,363 @@
...
+                            # Sample trajectory
+                            q_ref, qd_ref = self._trajectory.sample(t)
+
+                            # Create and publish command (outside lock would be better but simpler here)
+                            cmd = JointCommand(positions=q_ref, timestamp=time.time())
+
+                            # Publish - must release lock first for thread safety
+                            trajectory_active = True
+
+                if trajectory_active if "trajectory_active" in dir() else False:
```
paul-nechifor, line 341: If you add `trajectory_active = False` at the start of the method, you don't need the `dir()` check.

paul-nechifor: Actually, I don't think you need `trajectory_active`. Just add `cmd = None` at the start of the loop and change this line to `if cmd is not None`

dimos/manipulation/control/trajectory_controller/joint_trajectory_controller.py
```diff
@@ -0,0 +1,363 @@
...
+                        logger.error(f"Failed to publish joint command: {e}")
+                        with self._lock:
+                            self._state = TrajectoryState.FAULT
+                            self._error_message = f"Publish failed: {e}"
+
+                # Reset flag
+                trajectory_active = False
+
+                # Maintain loop frequency
+                time.sleep(period)
```
paul-nechifor, line 354: Technically, this doesn't achieve the desired frequency because it doesn't account for the time the loop takes to execute. It doesn't look like you're doing anything intensive, so it probably doesn't matter, but it's better practice to use either:

```python
while True:
    start_time = time.perf_counter()

    the_loop_work()

    elapsed = time.perf_counter() - start_time
    sleep_time = max(0.0, period - elapsed)
    time.sleep(sleep_time)
```

or

```python
next_time = time.perf_counter()
while True:
    the_loop_work()
    next_time += period
    time.sleep(max(0.0, next_time - time.perf_counter()))
```

dimos/manipulation/control/trajectory_controller/joint_trajectory_controller.py
```diff
@@ -0,0 +1,363 @@
...
+    # Callbacks
+    # =========================================================================
+
+    def _on_joint_state(self, msg: JointState) -> None:
+        """Callback for joint state feedback."""
+        self._latest_joint_state = msg
+
+    def _on_robot_state(self, msg: RobotState) -> None:
+        """Callback for robot state feedback."""
+        self._latest_robot_state = msg
```
paul-nechifor, line 285: It doesn't look like these are used.

dimos/manipulation/control/servo_control/cartesian_motion_controller.py
```diff
@@ -0,0 +1,715 @@
...
+                    time.sleep(period)
+                    continue
+
+                # Check for target timeout
+                if is_tracking and (current_time - last_target_time) > self.config.command_timeout:
+                    logger.warning("Target pose timeout - clearing target")
+                    with self._target_lock:
+                        self._target_pose_ = None
+                        self._is_tracking = False
+                    continue
```
paul-nechifor, line 469: You need `time.sleep` here too, right?

When I want to use continue, I usually start the loop with the sleep. That way I can continue without having to sleep first.

```python
while True:
    time.sleep(amount)
    ...
    if condition:
        continue
```

Also, like in the other comment, to achieve the correct frequency, you need to account for the time it takes to execute the loop.

dimos/manipulation/control/servo_control/cartesian_motion_controller.py
```diff
@@ -0,0 +1,715 @@
...
+    The controller is hardware-agnostic: it works with any arm driver that
+    implements the ArmDriverSpec protocol (provides IK/FK RPC methods).
+    """
+
+    default_config = CartesianMotionControllerConfig
+
+    # RPC methods to request from other modules (resolved at blueprint build time)
+    rpc_calls = [
+        "XArmDriver.get_forward_kinematics",
+        "XArmDriver.get_inverse_kinematics",
```
paul-nechifor, line 102: Is this supposed to work with the Piper arm as well? If so, instead of binding to XArmDriver, you can bind to an abstract base class and the blueprint system resolves to whichever arm driver is deployed in the blueprints. E.g.:

```
class ArmDriver(ABC):
    @abstractmethod
    def get_forward_kinematics(self, angles: list[float]) -> tuple[int, list[float] | None]:
        ...

...

class XArmDriver(BaseManipulatorDriver, ArmDriver):
    ... 
```

and use:

```
rpc_calls = [
    'ArmDriver.get_forward_kinematics',
    ...
]
```

You can check for how `NavigationInterface` is currently used in the codebase.

dimos/manipulation/control/trajectory_controller/joint_trajectory_controller.py
mustafab0: Not currently, but I have left them if/when we want to add error catching in the trajectory controller, it hasn't been a concern with xarm or piper yet. but if another arm needs our controller to run a control loop we can use these for error calculation.

dimos/manipulation/control/servo_control/cartesian_motion_controller.py
mustafab0: The is a function of ik and fk only being available for the xarm sdk right now. I decided to leave this as is, and address proper ik and fk in the manipulation planning. 

This controller only works with xarm, but will be addressed when I add a generic ik/fk solver.

dimos/manipulation/planning/trajectory_generator/joint_trajectory_generator.py
mustafab0: fixed

dimos/manipulation/control/trajectory_setter.py
mustafab0: added stop method

dimos/manipulation/control/trajectory_controller/spec.py
mustafab0: these were added as part of planning, but did not end up using them. 

removed now

dimos/manipulation/control/trajectory_controller/joint_trajectory_controller.py
mustafab0: Thanks, removed the complicated if statement and simplified with `cmd = None`

dimos/manipulation/control/trajectory_controller/joint_trajectory_controller.py
mustafab0: added a helper function to sleep to standardize sleep across the trajecory controller methods

dimos/hardware/manipulators/xarm/components/gripper_control.py
```diff
@@ -0,0 +1,363 @@
...
+- Robotiq gripper
+"""
+
+from dimos.core import rpc
+from dimos.utils.logging_config import setup_logger
+
+logger = setup_logger()
+
+
+class GripperControlComponent:
```
paul-nechifor, line 36: It doesn't look like this is used anywhere. Same for the piper one.

paul-nechifor: Many other components aren't used. Are they the old ones? Should they be deleted?

mustafab0: These  are for gripper controls, they need to be developed and tested further.

dimos/manipulation/control/servo_control/cartesian_motion_controller.py
mustafab0: Modified cartesian motion controller to have consistent frequency based on your  suggestion and sleeping at the start of the loop

pyproject.toml
```diff
@@ -281,7 +283,7 @@ python_version = "3.12"
 incremental = true
 strict = true
 warn_unused_ignores = false
-exclude = "^dimos/models/Detic(/|$)|.*/test_.|.*/conftest.py*"
+exclude = "^dimos/models/Detic(/|$)|.*/test_.|.*/conftest.py*|.*/example_.*\\.py$|.*integration_test.*\\.py$|.*/mock_.*\\.py$|^dimos/msgs/.*"
```
paul-nechifor: * `dimos/msgs/` should not be here
* `integration_test` can be renamed to `test_integration_` to fall under `test_`.
* `example_` should probably be type checked.
* if `mock_` contains test mocks, it should be defined in `conftest.py`

mustafab0: dimos/msgs/ should not be here - removed dimo/msgs/ from excldue
integration_test can be renamed to test_integration_ to fall under test_. - renamed file to test_integration_runner
example_ should probably be type checked. - removed example_ from exclude
if mock_ contains test mocks, it should be defined in conftest.py - renamed mock_sdk to conftest.py

PR #859, opened 2025-12-17
Point Detections

leshy:

### DIFF
--- a/dimos/perception/detection/type/detection2d/point.py
+++ b/dimos/perception/detection/type/detection2d/point.py
@@ -48,6 +48,7 @@ class Detection2DPoint(Detection2D):
     x: float
     y: float
     name: str
+    ts: float
     image: Image
     track_id: int = -1
     class_id: int = -1


dimos/models/vl/base.py
```diff
@@ -238,3 +289,56 @@ def query_detections(
...
+        format should be:
+        `[
+        [label, x, y]
+        ...
+        ]`
+
+        (etc, multiple matches are possible)
+
+        If there's no match return `[]`. Label is whatever you think is appropriate.
+        Only respond with the coordinates, no other text."""
```
paul-nechifor: I'd format this a little better. I've seen LLMs perform worse when given ambiguous text.
~~~suggestion
        full_query = f"""Show me point coordinates in pixels for this query: `{query}`

The format should be:
```json
[
    ["label 1", x, y],
    ["label 2", x, y],
    ...
]
```

Multiple matches are possible.

If there's no match return `[]`. The label should be whatever you think is appropriate.

Only respond with a JSON object of the labels and coordinates, no other text."""
~~~

leshy: cool, updated bbox query also

dimos/perception/detection/type/detection2d/point.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+@dataclass
+class Detection2DPoint(Detection2D):
+    """A 2D point detection, visualized as a circle."""
+
+    x: float
+    y: float
+    name: str
+    ts: float
```
paul-nechifor: This exists in Detection2D's superclass

PR #862, opened 2025-12-17
Cuda mapper

leshy:
```sh
dimos --replay run unitree-go2-newmapper
```

- 1-to-1 global mapper, runs on CUDA if possible but 6ms injest on CPU
- experiment with incline based costmap that we should move to (checked out Pauls costmap files)
- small structure for testing with moment.py (sketch for now)

removed stupid cylinder based map clearing, does perfect correct insertion of new frames


dimos/robot/unitree_webrtc/unitree_go2_blueprints.py
```diff
@@ -42,39 +43,20 @@
...
-)
+basic = autoconnect(
+    go2_connection(),
+    mapper(voxel_size=0.05),
+    astar_planner(),
+    holonomic_local_planner(),
+    behavior_tree_navigator(),
+    wavefront_frontier_explorer(),
+    websocket_vis(),
+    foxglove_bridge(),
```
paul-nechifor: There's some issue with `LCMTransport` on Mac if you send too much data. The current setup with pSHMTransport for `color_image` works. But if you switch to `LCMTransport` for `color_image` it breaks. Note that `LCMTransport` works in sim because the sim produces a much smaller image.

dimos/robot/unitree_webrtc/unitree_go2_blueprints.py
```diff
@@ -42,39 +43,20 @@
...
-    .global_config(n_dask_workers=4, robot_model="unitree_go2")
-    .transports(
-        # These are kept the same so that we don't have to change foxglove configs.
-        # Although we probably should.
-        {
-            ("color_image", Image): pSHMTransport(
-                "/go2/color_image", default_capacity=DEFAULT_CAPACITY_COLOR_IMAGE
-            ),
-            ("camera_pose", PoseStamped): LCMTransport("/go2/camera_pose", PoseStamped),
-            ("camera_info", CameraInfo): LCMTransport("/go2/camera_info", CameraInfo),
```
paul-nechifor, line 77: Should the foxglove config be updated, or is it not worth it now that we're moving to rerun?

dimos/utils/decorators/decorators.py
```diff
@@ -142,6 +153,13 @@ def getter(self):  # type: ignore[no-untyped-def]
                 setattr(self, attr_name, method(self))
             return getattr(self, attr_name)
 
+    def invalidate_cache(instance: Any) -> None:
+        """Clear the cached value for the given instance."""
+        if hasattr(instance, attr_name):
```
paul-nechifor: You need to acquire the lock here too.

dimos/mapping/pointclouds/occupancy.py
```diff
@@ -0,0 +1,418 @@
...
+
+    Costs are assigned based on the gradient magnitude of the terrain height.
+    Steeper slopes get higher costs, with max_step height change mapping to cost 100.
+    Cells without observations are marked unknown (-1).
+
+    Args:
+        cloud: PointCloud2 message containing 3D points
+        resolution: Grid resolution in meters/cell (default: 0.05)
+        can_pass_under: Max height to consider - points above are ignored (default: 0.6)
+        max_step: Height change in meters that maps to cost 100 (default: 0.15)
```
paul-nechifor: max_step -> can_climb

Although it would be more accurate to say lowest_height_which_cannot_be_climbed

dimos/mapping/pointclouds/occupancy.py
```diff
@@ -0,0 +1,418 @@
...
+        cost_float = (height_change_per_cell / can_climb) * 100.0
+        cost_float = np.clip(cost_float, 0, 100)
+
+        # Erode observed mask - only trust gradients where all neighbors are observed
+        # This prevents false high costs at boundaries with unknown regions
+        structure = ndimage.generate_binary_structure(2, 1)  # 4-connectivity
+        valid_gradient_mask = ndimage.binary_erosion(observed_mask, structure=structure)
+
+        # Convert to int8, marking cells without valid gradients as -1
+        cost = np.where(valid_gradient_mask, cost_float.astype(np.int8), -1).astype(np.int8)
```
paul-nechifor:
```suggestion
        cost = np.where(valid_gradient_mask, cost_float.astype(np.int8), -1)
```

It's already int8.

dimos/utils/metrics.py
```diff
@@ -0,0 +1,52 @@
...
+F = TypeVar("F", bound=Callable[..., Any])
+
+
+def timed(
+    transport: Callable[[F], Transport[Float32]] | Transport[Float32] | None = None,
+) -> Callable[[F], F]:
+    def timed_decorator(func: F) -> F:
+        t: Transport[Float32]
+        if transport is None:
+            t = LCMTransport(f"/metrics/{func.__name__}", Float32)
```
paul-nechifor, line 33: Idea: start the transport when the first call is made. If you do it in the warpper function you have access to `self`. You can check if `self._disposables` exists and add the cleanup function to it: `self._disposables.add(Disposable(t.stop))`

dimos/utils/testing/test_moment.py
```diff
@@ -0,0 +1,70 @@
...
+from dimos.msgs.sensor_msgs import CameraInfo, Image, PointCloud2
+from dimos.protocol.tf import TF
+from dimos.robot.unitree.connection import go2
+from dimos.utils.data import get_data
+from dimos.utils.testing.moment import Moment, SensorMoment
+
+data_dir = get_data("unitree_go2_office_walk2")
+
+
+class Go2Moment(Moment):
```
paul-nechifor, line 27: I'd move this out of this test file because it's a class what's used in other tests.

dimos/mapping/voxels.py
```diff
@@ -0,0 +1,276 @@
...
+        dev = (
+            o3c.Device(self.config.device)
+            if (self.config.device.startswith("CUDA") and o3c.cuda.is_available())
+            else o3c.Device("CPU:0")
+        )
+
+        print(f"SparseVoxelGridMapper using device: {dev}")
+
+        self.vbg = o3d.t.geometry.VoxelBlockGrid(
+            attr_names=("dummy",),
```
paul-nechifor, line 77: Why "dummy"? :)

dimos/mapping/voxels.py
```diff
@@ -0,0 +1,276 @@
...
+            # we are potentially moving out of CUDA here
+            ensure_legacy_pcd(self.get_global_pointcloud()),
+            frame_id=self.frame_id,
+            ts=time.time(),
+        )
+
+    @simple_mcache
+    def get_global_pointcloud(self) -> o3d.t.geometry.PointCloud:
+        voxel_coords, _ = self.vbg.voxel_coordinates_and_flattened_indices()
+        pts = voxel_coords + (self.config.voxel_size * 0.5)
```
paul-nechifor, line 199: What is the reason behind this? To move the points to the center of the voxel?

dimos/mapping/voxels.py
```diff
@@ -0,0 +1,276 @@
...
+        )
+
+        print(f"SparseVoxelGridMapper using device: {dev}")
+
+        self.vbg = o3d.t.geometry.VoxelBlockGrid(
+            attr_names=("dummy",),
+            attr_dtypes=(o3c.uint8,),
+            attr_channels=(o3c.SizeVector([1]),),
+            voxel_size=self.config.voxel_size,
+            block_resolution=1,
```
paul-nechifor, line 81: Our maps are quite contiguous. Isn't it faster to use the default of block_resolution=16.

dimos/mapping/voxels.py
```diff
@@ -0,0 +1,276 @@
...
+            self._hm.activate(keys_Nx3)
+
+        self.get_global_pointcloud.invalidate_cache(self)  # type: ignore[attr-defined]
+        self.get_global_pointcloud2.invalidate_cache(self)  # type: ignore[attr-defined]
+        self.get_global_occupancygrid.invalidate_cache(self)  # type: ignore[attr-defined]
+
+    def _carve_and_insert(self, new_keys: o3c.Tensor) -> None:
+        """Column carving: remove all existing voxels sharing (X,Y) with new_keys, then insert."""
+        if new_keys.shape[0] == 0:
+            self._hm.activate(new_keys)
```
paul-nechifor: I think the meaning of `_hm` is quite obscure at this level. Maybe `_voxel_hash_map`?

dimos/mapping/voxels.py
```diff
@@ -0,0 +1,276 @@
...
+    # @timed()  # TODO: fix thread leak in timed decorator
+    def add_frame(self, frame: LidarMessage) -> None:
+        # we are potentially moving into CUDA here
+        pcd = ensure_tensor_pcd(frame.pointcloud, self._dev)
+
+        if pcd.is_empty():
+            return
+
+        pts = pcd.point["positions"].to(self._dev, o3c.float32)
+        vox = (pts / self.config.voxel_size).floor().to(self._key_dtype)
```
paul-nechifor, line 133:
```suggestion
        vox = (pts / self.config.voxel_size).round().to(self._key_dtype)
```

I think this should be round. The data we get is already voxelized but because of floating point precision issues, we often get 0.05 as (for example) 0.04999999999997, which means floor turns it into 0. 

This is the issue I encountered here:

dimos/robot/unitree_webrtc/unitree_go2_blueprints.py
leshy: yes I will update the dash you are right, the dash changes because vis changes a bit for new mapper so I figured good time to remove this

dimos/utils/metrics.py
leshy: oh good idea, this is broken because of thread handling yes

dimos/utils/testing/test_moment.py
leshy: me too

dimos/mapping/voxels.py
leshy: yes you get a  drift on roundrip otherwise

dimos/mapping/voxels.py
leshy: need to check a bit, there was a good reason for this, generally the whole thing needs a more careful review

dimos/mapping/voxels.py
leshy: yeah this is all actually very terrible, the main issue is that o3d api at this layer is completely undocumented, and the way we use it is unusual, I will rename though

dimos/mapping/voxels.py
leshy: great catch also

dimos/robot/unitree_webrtc/unitree_go2_blueprints.py
leshy: I guess it's their network stack, now I have even more amunition to trash mac

dimos/mapping/voxels.py
leshy: we are abusing voxelblockgrid, it's supposed to carry information per voxel and doesn't work without setting these attributes, but we just using it to store 3d occupancy. I think probably this whole thing can be replaced with a CUDA hashmap instead of o3d, but this was faster to implement.

They do have some raytracing etc on top of voxelblockgrid so might turn out useful for projecting rgb or something

dimos/utils/metrics.py
leshy: decided to look into this in more detail in the next pr, proly implemented at module level

dimos/mapping/voxels.py
leshy: nope, round messes it up, I'm checking roundtrip (lidar -> global_map -> lidar) making sure it's the same, so algo is solid

dimos/robot/unitree_webrtc/unitree_go2_blueprints.py
leshy: added mac detection

PR #863, opened 2025-12-18
Restore MacOS Support (flake.nix)

jeff-hykin:


flake.nix
```diff
@@ -4,90 +4,418 @@
...
+
+          #
+          # python & setup
+          #
+          # automate the readme
+          dimos_setup() {
+            macos_version="$(sw_vers -productVersion 2>/dev/null || echo "0.0")"
+            macos_major_version="''${macos_version%%.*}"
+            echo "Making sure git lfs is installed..."
+            git lfs install || true
```
leshy: `git lfs install --skip-smudge`
to not download all the files but references

flake.nix
```diff
@@ -4,90 +4,418 @@
...
+            fi
+            echo
+            return 1 # failure
+          }
+
+          #
+          # python & setup
+          #
+          # automate the readme
+          dimos_setup() {
```
leshy: I like this directionally and can let this in, but dev env setup should be shared, already implemented for a dev docker image, so that we avoid maintaining two parallel systems from diverging.  I'd say priority to cleanup after rerun is done (pre commit hooks missing here but nvm, since will unify)

https://github.com/dimensionalOS/dimos/blob/dev/docs/development.md
https://github.com/dimensionalOS/dimos/blob/dev/.devcontainer/devcontainer.json
https://github.com/dimensionalOS/dimos/blob/dev/docker/dev/Dockerfile

(can discuss the details, atm `./bin/dev` is a single command that gets you up an running with dev env,
 can be extended to let you choose nix vs docker)

jeff-hykin: Sounds good. And yeah sorry about the hooks. I end up toggling them during development and forgetting to turn them back on.

PR #868, opened 2025-12-21
Rerun 0.1 Integration 

jeff-hykin:
## Status:
- [x] run yourself with `dimos --replay run unitree-go2` (tested on Mac and Linux and tested over ssh/remote)
- [x] pytests added (and passing)
- [x] rerun needs numpy >= 2.0 ... but dimos needs numpy < 2.0, this is blocking CI success. I'm using numpy 1.26 and haven't had issues
- [x] lint / dead code removal / greptile-apps feedback

# NOTE: 

This video example is running remote, over slow internet connection (<10Mbs), on a 10 year old i5 CPU (no CUDA)

https://github.com/user-attachments/assets/fbcdf198-6eae-45cc-aa72-c7809ce29160

# How do I use the Dashboard?

If you have a module like this (`CameraListener`):

```py
from dimos.core.blueprints import autoconnect
from dimos.core import In, Module, pSHMTransport
from dimos.hardware.camera.module import CameraModule
from dimos.manipulation.visual_servoing.manipulation_module import ManipulationModule
from dimos.msgs.sensor_msgs import Image

class CameraListener(Module):
    color_image: In[Image] = None  # type: ignore[assignment]

        self._count = 0

    @rpc
    def start(self) -> None:
        super().start()

        def _on_frame(img: Image) -> None:
            self._count += 1
            if self._count % 20 == 0:
                print(
                    f"[camera-listener] frame={self._count} ts={img.ts:.3f} "
                    f"shape={img.height}x{img.width}"
                )

        print("camera subscribing")
        unsub = self.color_image.subscribe(_on_frame)
        self._disposables.add(Disposable(unsub))

blueprint = (
    autoconnect(
        CameraModule.blueprint(),
        CameraListener.blueprint(),
    )
    .global_config(n_dask_workers=1)
)
coordinator = blueprint.build()
print("Webcam pipeline running. Press Ctrl+C to stop.")
coordinator.loop()
```

Then modify it to use rerun like this:

```py
from dimos.core import In, Module, pSHMTransport, autoconnect
from dimos.hardware.camera.module import CameraModule
from dimos.manipulation.visual_servoing.manipulation_module import ManipulationModule
from dimos.dashboard.module import Dashboard, RerunConnection
from dimos.msgs.sensor_msgs import Image

class CameraListener(Module):
    color_image: In[Image] = None  # type: ignore[assignment]

        self._count = 0

    @rpc
    def start(self) -> None:
        super().start()

        # 
        # HERE (1 of 3)
        #
        self.rc = RerunConnection() # one connection per process

        def _on_frame(img: Image) -> None:
            self._count += 1

            # 
            # HERE (2 of 3)
            #
            self.rc.log(f"/{self.__class__.__name__}/color_image", img.to_rerun())
            if self._count % 20 == 0:
                print(
                    f"[camera-listener] frame={self._count} ts={img.ts:.3f} "
                    f"shape={img.height}x{img.width}"
                )

        print("camera subscribing")
        unsub = self.color_image.subscribe(_on_frame)
        self._disposables.add(Disposable(unsub))

blueprint = (
    autoconnect(
        CameraModule.blueprint(),
        CameraListener.blueprint(),
        # 
        # HERE (3 of 3)
        #
        Dashboard.blueprint(
            auto_open=True,
            terminal_commands={
                "lcm-spy": "dimos lcmspy",
                "skill-spy": "dimos skillspy",
            },
        ),
    )
    .transports({("color_image", Image): pSHMTransport("/cam/image")})
    .global_config(n_dask_workers=1)
)
coordinator = blueprint.build()
print("Webcam pipeline running. Press Ctrl+C to stop.")
coordinator.loop()
```

### DIFF
--- a/dimos/dashboard/module.py
+++ b/dimos/dashboard/module.py
@@ -116,6 +116,9 @@ class Dashboard(Module):
         def _cleanup_dashboard_thread():
             dashboard_started.clean()
 
+    def stop(self) -> None:
+        self.logger.info("Stopping dashboard server")
+
 
 class RerunConnection:
     def __init__(self) -> None:


dimos/dashboard/README.md
```diff
@@ -0,0 +1,81 @@
...
+blueprint = (
+    autoconnect(
+        CameraModule.blueprint(),
+        ManipulationModule.blueprint(),
+    )
+    .global_config(n_dask_workers=1)
+)
+coordinator = blueprint.build()
+print("Webcam pipeline running. Press Ctrl+C to stop.")
+coordinator.loop()
```
paul-nechifor: This should be behind an `if __name__ == '__main__':` block. If you don't add it when Dask starts any worker, it will re-execute `blueprint.build()` as many times are there are workers.

dimos/dashboard/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+```py
+from dimos.core.blueprints import autoconnect
+from dimos.hardware.camera.module import CameraModule
+from dimos.manipulation.visual_servoing.manipulation_module import ManipulationModule
+from dimos.dashboard.module import Dashboard, RerunConnection
+from dimos.msgs.sensor_msgs import Image
+
+class CameraListener(Module):
+    color_image: In[Image] = None  # type: ignore[assignment]
```
paul-nechifor:
```suggestion
    color_image: In[Image]
```

This isn't needed anymore.

dimos/dashboard/module.py
```diff
@@ -0,0 +1,168 @@
...
+    zellij_token: str | None = os.environ.get("ZELLIJ_TOKEN")
+    zellij_url: str | None = None
+    zellij_session_name: str | None = "dimos-dashboard"
+    https_key_path: str | None = os.environ.get("HTTPS_KEY_PATH")
+    https_cert_path: str | None = os.environ.get("HTTPS_CERT_PATH")
+    logger: logging.Logger | None = None
+
+    def __init__(self, *args, **kwargs) -> None:  # type: ignore[no-untyped-def]
+        super().__init__()
+        self.__dict__.update(kwargs)
```
paul-nechifor: Personally, I'm not a fan of this. It does remove duplication, but also allows people to set anything, with no feedback when they're wrong. I think it's better to rely on tools like mypy to make sure we're passing all the right params.

dimos/dashboard/support/html_generation.py
```diff
@@ -0,0 +1,223 @@
...
+
+DEFAULT_SESSION_NAME = "dimos-dashboard"
+
+path_to_baseline_css = os.path.join(os.path.dirname(__file__), "css_baseline.css")
+with open(path_to_baseline_css) as f:
+    css_baseline_contents = f.read()
+
+
+# prevent any kind of html injection (even accidental)
+def escape_js_for_html(text):
```
paul-nechifor: Just use Jinja2. We already have it installed (although not as a direct dependency).

HTML and JS should not be defined inline like in this file.

dimos/dashboard/module.py
jeff-hykin: Sounds good. It was definitely fast development not necessarily good DX. 

I'll redo all the args cause the server args could use some cleanup as well.

dimos/dashboard/support/html_generation.py
jeff-hykin: Sounds good. I default to minimizing necessary dependencies, but happy to switch to jinja.

dimos/dashboard/support/zellij_tooling.py
```diff
@@ -0,0 +1,298 @@
...
+import shutil
+import subprocess
+from typing import Optional
+
+from aiohttp import ClientSession
+
+SESSION_LINE_RE = re.compile(r"^(.+?)\s+\[Created\s+(.+?)\s+ago\](.*)$")
+
+
+class ZellijManager:
```
paul-nechifor: About Zellij, why do we want this? This is to start agentspy, skillspy, etc, right?

But those can also be started as web apps (right @leshy ?). Would it not be better to embed them into the rerun web? Is that possible?

Or is Zellij just a temporary thing?

dimos/dashboard/support/zellij_tooling.py
```diff
@@ -0,0 +1,298 @@
...
+                for command in terminal_commands.values():
+                    sanitized_command = re.sub(r"[^A-Za-z\s_\-\=\*]", "", command)
+                    file_path = f"/tmp/{sanitized_command}.sh"
+                    with open(file_path, "w") as file:
+                        file.write(f"""
+                            . .envrc
+                            . source.ignore.sh
+                            . ./venv/bin/activate
+                            {command}
+                        """)
```
paul-nechifor: This looks a little sketchy. ( Might get us flagged by services which analyze code published to PyPI. :D )

With tmux you can orchestrate layouts by giving it commands (like `tmux split-window -t session:window.pane 'your_command'`) does that not work with Zellij so we don't leave odd files in /tmp?

dimos/dashboard/support/utils.py
```diff
@@ -0,0 +1,81 @@
...
+    for each in psutil.Process(os.getpid()).parents():
+        try:
+            with open(f"/tmp/{each.pid}.json") as infile:
+                return json.load(infile)
+        except Exception:
+            pass
+    # if none of the parents have a json file, make one
+    with open(f"/tmp/{os.getpid()}.json", "w") as outfile:
+        json.dump(json_data, outfile)
+    return json_data
```
paul-nechifor: I find this a bit odd. It seems a bit haphazardous to try all parents.

An alternative to communicating values to children is to set the values as environment variables, since the children inherit those (I'm not certain that works, maybe Dask messes with them).

Another would be to write the config to shared memory. The Mujoco process does something like this.

dimos/dashboard/support/zellij_tooling.py
```diff
@@ -0,0 +1,298 @@
...
+            raise RuntimeError(
+                f"zellij list-sessions failed (code {proc.returncode}): {stderr_text or stdout_text}"
+            )
+
+        sessions = ZellijManager.parse_zellij_sessions(stdout_text)
+        self.log.info("Found %s sessions", len(sessions))
+        return {"success": True, "sessions": sessions, "count": len(sessions)}
+
+    def init_zellij_session(
+        self: logging.Logger,
```
paul-nechifor: Is `self` supposed to be `logger` here? Also, is this a static class. It's used as one in:

```
           ZellijManager.init_zellij_session(
                self.log, self.session_name, self.terminal_commands, self.zellij_layout
            )
```

dimos/dashboard/module.py
```diff
@@ -0,0 +1,182 @@
...
+            )
+            self.stream.connect_grpc(rerun_info.url)
+            self._init_id = mp.current_process().pid
+
+        if self._init_id != mp.current_process().pid:
+            raise Exception(
+                """Looks like you are somehow using RerunConnection to log data to rerun. However, the process/thread where you init RerunConnection is different from where you are logging. A RerunConnection object needs to be created once per process/thread."""
+            )
+
+        self.stream.log(msg, value, **kwargs)
```
paul-nechifor, line 147: You need to use a lock to prevent the race condition where two threads try to initialize the stream.

But I don't understand the pid check. How can it ever be true?

dimos/dashboard/support/zellij_tooling.py
jeff-hykin: We don't have to have zellij but it is pretty flexible which I think is good for an inital release, ex: `htop` can run no problem. Interactive CLI tools work no problem. Also sessions can be re-connected. You can open the zellij window on your phone browser while the rerun stuff renders in the rerun app.

I don't really see how agentspy, skillspy, etc could be run as a webapp without building a custom webapp for them. Maybe there is some tool I don't know about that converts a `textual` app to an iframe? @leshy

dimos/dashboard/support/zellij_tooling.py
jeff-hykin: I am glad you flagged this. This gets into a disscussion about `pip install dimos`. When I wrote this PR I was only thinking of developement (git clone dimos). `source.ignore.sh` absolutely needs to be taken out, that was so I could get this to run with my nix environment.

Maybe for this PR I can just pull out the Zellij stuff so its faster to merge

dimos/dashboard/support/utils.py
jeff-hykin: This is almost certainly a bigger problem I'd like to talk to you @paul-nechifor  and @leshy about (shared state). 

I did use ENV vars for some values as a "cheat" but only as a temp fix, I think the overall design problem of trying to share data/state between modules is something we need to talk about before launch. In reality, these "constants" should be configuable in a module constructor, which is more complicated.

For shared memory, my understanding was that safe writes to shared memory require a lock, and locks require initilzation in the main module, and I didn't see clean way to have a main module hook for the `RerunConnection` class. Modules need to be able to import/use RerunConnection even if dashboard module is never used. Shared memory also seemed kinda overkill for something that doesn't change after module initilization.

dimos/dashboard/module.py
jeff-hykin: Actually there need to be multiple streams, at least one per worker but more is fine. The whole class exists to help enforce at least one strefam per worker.

Here is an example of when the check fails. Maybe just writing a pickle hook that fails would be a good equivlent. Arguably I should set `_init_id` in the constructor (which was how I was originally doing it before lazy-setting the stream)

```py
top_level_connection = RerunConnection()
top_level_connection.log("test") 
# ^this log call used to not be required

def p1(connection: RerunConnection) -> None:
    connection.log("from dashboard_process")

if __name__ == '__main__':
    mp.set_start_method("spawn", force=True)
    dashboard_proc = mp.Process(
        target=p1, args=(top_level_connection,), name="p1_like_dashboard"
    )
    dashboard_proc.start()
```

dimos/dashboard/support/utils.py
paul-nechifor: Yep, it's a bit complicated. Since we don't create the child processes it's not so easy to pass data to them. Dask has a way of passing config to workers ( https://docs.dask.org/en/stable/configuration.html ) but we might want to stop using Dask, so it's not worth it to switch to that now.

This should work for now then.

dimos/dashboard/support/zellij_tooling.py
paul-nechifor: There's textual-web ( https://textual.textualize.io/blog/2023/09/06/what-is-textual-web/ ) . I haven't used it so I don't know how feasible it is.

jeff-hykin: Looks like it'll require some hacking to get it into an iframe. I'll wait to do that in another PR

dimos/dashboard/support/zellij_tooling.py
jeff-hykin: Fixed this problem individually (for later), then also pulled all of Zellij out of this PR.

dimos/dashboard/support/utils.py
jeff-hykin: This idea/approach (which I now put into a little `FileBasedBoolean` helper) is use the PID heirarchy so that a filepath is the same across all workers, then that file itself can be used as a boolean shared state. AFAIK file existance should be threadsafe. It would be really nice if, as part of dimos, there was an always active / always included `state` module that acted as a manager of a python dict. The module could be passed to the init of all other modules so that they could set/get shared values.

Since autoconnect isnt mandatory, the `.start` of all modules would need to check if the `state` module had been started (and if not, they would need to start it first).

PR #885, opened 2025-12-27
Planner improvements

paul-nechifor:
* Separate out PD controller from local planner.
* Add P controller for sim.
* Allow unknown goals (does not try to make them safe).
* Reset retry attempts if traveled far enough (2 meters from where it was initially stuck).
* Added numba optimizations for occupancy. `simple_occupancy` went from `0.01057` to `0.00166`. `height_cost_occupancy` went from `0.04679` to `0.01817`.
* Changed /debug_navigation to only publish every second to not overwhelm the system.
* Merged the end-to-end navigation test in Mujoco ( https://github.com/dimensionalOS/dimos/pull/775 ) in here since navigation is so reliable now.


dimos/utils/trigonometry.py
```diff
@@ -0,0 +1,19 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import math
+
+
+def angle_diff(a: float, b: float) -> float:
```
leshy, line 18: Vector3 supports angle diffs, just noting, should add  Vector2 eventually

PR #887, opened 2025-12-27
Release planner

leshy:
https://github.com/dimensionalOS/dimos/pull/884
https://github.com/dimensionalOS/dimos/pull/885
https://github.com/dimensionalOS/dimos/pull/886

and foxglove dash all together

### DIFF
--- a/dimos/robot/test_ros_bridge.py
+++ b/dimos/robot/test_ros_bridge.py
@@ -120,6 +120,8 @@ class TestROSBridge(unittest.TestCase):
             self.assertAlmostEqual(msg.linear.y, float(i * 2), places=5)
             self.assertAlmostEqual(msg.angular.z, float(i * 0.1), places=5)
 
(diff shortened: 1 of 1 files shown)


dimos/e2e_tests/dimos_cli_call.py
```diff
@@ -0,0 +1,69 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import signal
+import subprocess
+import time
+
+
+class DimosCliCall:
```
spomichter, line 20: @paul-nechifor This seems like a hacky way to be running async `dimos run` we need a a pythonic equivalent method to run blueprints that don't require a bash subprocess. Any ideas. Ideally `blueprint.run()`

PR #894, opened 2025-12-28
Granular deps

leshy:
`uv pip install .[unitree]`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,8 +78,6 @@ dimos = "dimos.robot.cli.dimos:main"
 [project.optional-dependencies]
 misc = [
     # Core requirements
-    "openai",
-    "anthropic>=0.19.0",
     "cerebras-cloud-sdk",
     "yapf==0.40.2",
     "typeguard",
@@ -138,6 +136,7 @@ agents = [
     "anthropic>=0.19.0",
 
     # Audio
+    "openai",
     "openai-whisper",
     "sounddevice",
 
--- a/uv.lock
+++ b/uv.lock
@@ -1494,6 +1494,7 @@ agents = [
     { name = "langchain-text-splitters" },
     { name = "mcp" },
     { name = "ollama" },
+    { name = "openai" },
     { name = "openai-whisper" },
     { name = "sounddevice" },
 ]
@@ -1571,7 +1572,6 @@ manipulation = [
     { name = "trimesh" },
 ]
 misc = [
-    { name = "anthropic" },
     { name = "catkin-pkg" },
     { name = "cerebras-cloud-sdk" },
     { name = "clip" },
@@ -1583,7 +1583,6 @@ misc = [
     { name = "lark" },
     { name = "onnx" },
     { name = "open-clip-torch" },
-    { name = "openai" },
     { name = "opencv-contrib-python" },
     { name = "pygame" },
     { name = "python-multipart" },
@@ -1628,6 +1627,7 @@ unitree = [
     { name = "mcp" },
     { name = "moondream" },
     { name = "ollama" },
+    { name = "openai" },
     { name = "openai-whisper" },
     { name = "pillow" },
     { name = "rerun-sdk" },
@@ -1653,7 +1653,6 @@ web = [
 [package.metadata]
 requires-dist = [
     { name = "anthropic", marker = "extra == 'agents'", specifier = ">=0.19.0" },
-    { name = "anthropic", marker = "extra == 'misc'", specifier = ">=0.19.0" },
     { name = "asyncio", specifier = "==3.4.3" },
     { name = "bitsandbytes", marker = "sys_platform == 'linux' and extra == 'agents'", specifier = ">=0.48.2,<1.0" },
     { name = "catkin-pkg", marker = "extra == 'misc'" },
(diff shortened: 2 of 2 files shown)


dimos/agents/skills/navigation.py
```diff
@@ -201,75 +200,75 @@ def _navigate_to(self, pose: PoseStamped) -> bool:
...
-        logger.warning(f"Navigation to '{query}' timed out after {timeout}s")
-        stop_track_rpc()
-        return None
-
-    def _get_bbox_for_current_frame(self, query: str) -> BBox | None:
-        if self._latest_image is None:
-            return None
-
-        return get_object_bbox_from_image(self._vl_model, self._latest_image, query)
+    # def _navigate_to_object(self, query: str) -> str | None:
```
leshy: the reason why this is out, is that it uses deprecated agents, and old bbox class, and pulls a bunch of heavy deps with it that I didn't want to include in unitree install

pyproject.toml
```diff
@@ -22,84 +22,75 @@ description = "Powering agentive generalist robotics"
 requires-python = ">=3.10"
 
 dependencies = [
-    # Core requirements
-    "opencv-python",
+    # Transport Protocols
```
paul-nechifor, line 32: You need to run `uv lock` to generate the updated `uv.lock` file.

This has happened a few times. (In fact uv.lock on dev is out of date.) I've added https://github.com/dimensionalOS/dimos/pull/902 to check uv.lock in pre-commit.

pyproject.toml
```diff
@@ -22,84 +22,75 @@ description = "Powering agentive generalist robotics"
...
+skillspy = "dimos.utils.cli.skillspy.skillspy:main"
+agentspy = "dimos.utils.cli.agentspy.agentspy:main"
+humancli = "dimos.utils.cli.human.humanclianim:main"
+dimos = "dimos.robot.cli.dimos:main"
+
+[project.optional-dependencies]
+misc = [
+    # Core requirements
     "openai",
     "anthropic>=0.19.0",
```
paul-nechifor: Should these be in agents?

pyproject.toml
```diff
@@ -22,84 +22,75 @@ description = "Powering agentive generalist robotics"
...
-    "colorlog==6.9.0",
     "yapf==0.40.2",
     "typeguard",
     "empy==3.3.4",
     "catkin_pkg",
     "lark",
-    "plum-dispatch==2.5.7",
-    "ffmpeg-python",
     "tiktoken>=0.8.0",
     "Flask>=2.2",
```
paul-nechifor: LOL, kinda odd that we have both fastapi and flask. It's only used in `dimos/web/flask_server.py` and that's not used, so maybe this should be removed.

pyproject.toml
```diff
@@ -133,15 +113,50 @@ dependencies = [
...
+    "moondream",
+]
+
+unitree = [
+    "dimos[agents,web,perception]",
+    # WebRTC streaming
+    "aiortc==1.9.0",
+    "unitree-webrtc-connect",
+    "pycryptodome",
+    "pyaudio",
```
paul-nechifor: It doens't look like this is used anymore.

pyproject.toml
```diff
@@ -133,15 +113,50 @@ dependencies = [
...
+    "transformers[torch]==4.49.0",
+    "moondream",
+]
+
+unitree = [
+    "dimos[agents,web,perception]",
+    # WebRTC streaming
+    "aiortc==1.9.0",
+    "unitree-webrtc-connect",
+    "pycryptodome",
```
paul-nechifor: Looks like this provides the `Crypto` package, but it's not used anywhere.

pyproject.toml
leshy: yeah oai is for voice (tts) - we might want to split this differently, but for now I dumped it to agents
anthropic will be deleted once we remove agents_deprecated

pyproject.toml
leshy: yes had some connection issues and wasn't sure, removing

pyproject.toml
paul-nechifor: BTW, the worst thing about Python is the terrible package management. Just because pycryptodome is never imported doesn't mean it's not used. Because pycryptodome actually provides `Crypto`.

There are **a lot** of packages pypackages.toml which have 0 usages, but it's hard to tell if they're needed or not.

jeff-hykin: (Commenting because this has an impact on the reliability and difficulty of the installer). The number of apt-get installs needed just to make the non-optional pip modules work is massive. I'm all for minimizing them and breaking them up. 

I might be able to add a runtime pytest tracer to tell us at least what imports are being used for each test. Something like:

```py
import sys
import pytest

@pytest.fixture(scope="session", autouse=True)
def track_imports():
    before = set(sys.modules.keys())
    yield
    after = set(sys.modules.keys())
    imported = sorted(after - before)

    with open("imported_modules.txt", "w") as f:
        for m in imported:
            f.write(m + "\n")
```

and I'm sure theres some way to get a reverse mapping of pip module name to import name.

I know the tests dont have full coverage but I think per-test list of what pip module names are used could be pretty useful for finding dead pip dependencies

dimos/agents/skills/navigation.py
paul-nechifor: Added back to pass the tests.

paul-nechifor: It just uses qwen which is hosted

dimos/stream/video_operators.py
```diff
@@ -215,27 +214,6 @@ def with_optical_flow(
             ops.map(frame_processor.visualize_flow),
         )
 
-    @staticmethod
```
spomichter, line 218: What does this random change have to do with this PR @paul-nechifor

PR #900, opened 2025-12-29
Rerun distributed logging

### LINKED ISSUE #895: Rerun
we want equivalent of fg dashboard - needs to look good.

https://github.com/dimensionalOS/dimos/pull/869/ - rerun experiment

like for example, changing voxel size on the global mapper needs to not trash the rerun view
(`voxel_mapper(voxel_size=0.05),` in `unitree_go2_blueprints.py`)

we can always use `rr.log` within modules for super custom stuff (navigation algo renders poses, probabilities etc) but what modules very often want is just to publish their output streams, so we need an easy way for a module to specify what to output to rerun, and an easy way to render any our standard messages (dimos/msgs) to rerun.

proposal is to add `to_rerun()` method to messages and the return of this is what we can feed to `rr.log`, so

`rr.log(PointCloud2.to_rerun())`

# module api

Modules very often want to publish their outputs directly to rerun. this will be 99% of rerun usage

add a helper function

```python
class VoxelGridMapper(Module):
    global_map: Out[LidarMessage]

    def start(self):
        self.global_map.to_rerun({ some_config: ..., rate_limit: 0.2hz, color_map: ... })
        ...
```

above just  takes all messages coming into self.global_map and calls rr.log(msg, this_config_we_provided)

## better module api

we can add a way to configure outputs but needs module intervention (for later)

```python
class VoxelGridMapper(Module):
    global_map: Out[LidarMessage] = Rerun(rate=10hz)
```

# interactivity

rerun let's you click on some panels and you need some rust or something? idk, investigate, important

Nabla7:
## Distributed Rerun Logging

Modules now log directly via `rr.log()` instead of centralized subscription.

### Usage

```py
import rerun as rr
from dimos.dashboard import rerun_init

class MyModule(Module):
    def start(self):
        rr.log("world/entity", msg.to_rerun())Viewer modes: `RERUN_VIEWER=web|native|none` (default: web)
```

### Changes
- Deleted `RerunModule` (611 lines)
- Added `rerun_init.py` for multi-process initialization
- Modules log their own data: GO2Connection, VoxelGridMapper, CostMapper, ReplanningAStarPlanner

Fixes #895 - modules own visualization, no race conditions, config changes don't break view.

**Limitation:** No `Out.to_rerun({ rate_limit })` helper yet - modules call `rr.log()` manually. Can add later if needed.

### DIFF
--- a/dimos/msgs/nav_msgs/OccupancyGrid.py
+++ b/dimos/msgs/nav_msgs/OccupancyGrid.py
@@ -614,7 +614,7 @@ class OccupancyGrid(Timestamped):
         oy = self.origin.position.y
         r = self.resolution
 
-        for i, (y, x) in enumerate(zip(gy, gx)):
+        for i, (y, x) in enumerate(zip(gy, gx, strict=False)):
             # World position of cell corner
             wx = ox + x * r
             wy = oy + y * r


rerun_pr_comparison_summary.txt
```diff
@@ -0,0 +1,277 @@
+ .gitattributes                                     |    1 +
```
paul-nechifor: I think this should be deleted.

dimos/web/websocket_vis/websocket_vis_module.py
```diff
@@ -186,7 +186,119 @@ def _create_server(self) -> None:
         self.sio = socketio.AsyncServer(async_mode="asgi", cors_allowed_origins="*")
 
         async def serve_index(request):  # type: ignore[no-untyped-def]
-            return HTMLResponse("<html><body>Use the extension.</body></html>")
+            html = """<!DOCTYPE html>
```
paul-nechifor: I think it would be best to move this to an .html file

dimos/msgs/nav_msgs/OccupancyGrid.py
```diff
@@ -416,3 +416,247 @@ def cell_value(self, world_position: Vector3) -> int:
...
+
+        Returns:
+            Rerun archetype for logging (rr.Image, rr.Mesh3D, or rr.Points3D)
+
+        The visualization uses:
+        - Free space (value 0): white/blue
+        - Unknown space (value -1): gray/transparent
+        - Occupied space (value > 0): black/red with gradient
+        """
+        import rerun as rr
```
paul-nechifor, line 457: Please move all the imports inside functions to the top.

dimos/dashboard/rerun_init.py
```diff
@@ -0,0 +1,113 @@
...
+            f"Rerun: gRPC only on port {RERUN_GRPC_PORT}, "
+            f"connect with: rerun --connect {RERUN_GRPC_ADDR}"
+        )
+
+
+# Initialize at import time
+try:
+    _init_rerun()
+except Exception as e:
+    logger.warning(f"Failed to initialize Rerun: {e}")
```
paul-nechifor: Why convert an exception to a warning? Do we really want to continue if rerun failed?

dimos/dashboard/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Dashboard module for visualization and monitoring.
+
+Importing this module initializes Rerun (happens at import time, before multiprocessing).
+
+Usage in modules:
+    import rerun as rr
+    from dimos.dashboard import rerun_init  # triggers Rerun initialization
```
paul-nechifor: I don't like the idea of imports having side effects. (Ruff removes unused imports like this (but this disabled until I fix all the errors).) It would be better to make `_init_rerun` a public function and it should be called to initialize rerun.

dimos/dashboard/rerun_init.py
```diff
@@ -0,0 +1,113 @@
...
+        server_uri = rr.serve_grpc(grpc_port=RERUN_GRPC_PORT)
+        rr.serve_web_viewer(web_port=RERUN_WEB_PORT, open_browser=False, connect_to=server_uri)
+        logger.info(f"Rerun: web viewer on http://localhost:{RERUN_WEB_PORT}")
+    else:
+        # Just gRPC server, no viewer (connect externally)
+        rr.serve_grpc(grpc_port=RERUN_GRPC_PORT)
+        logger.info(
+            f"Rerun: gRPC only on port {RERUN_GRPC_PORT}, "
+            f"connect with: rerun --connect {RERUN_GRPC_ADDR}"
+        )
```
paul-nechifor, line 97: I think you have to use `rr.disconnect` to shut them down https://ref.rerun.io/docs/python/0.28.1/common/initialization_functions/#rerun.disconnect

dimos/dashboard/rerun_init.py
```diff
@@ -0,0 +1,113 @@
...
+    # before main process server is fully up)
+    for attempt in range(5):
+        if _is_port_in_use(RERUN_GRPC_PORT):
+            # Server already running (we're a worker) - connect to it
+            rr.connect_grpc(RERUN_GRPC_ADDR)
+            logger.info(f"Rerun: connected to server at {RERUN_GRPC_ADDR}")
+            return
+        if attempt == 0:
+            # First attempt failed - we're likely the main process, start server
+            break
```
paul-nechifor: This seems unsafe. Two processes can both think they're the main server.

An alternative would be to start the server in `core/blueprints.py`, and pass the info about the server to the workers which are started in there.

dimos/dashboard/rerun_init.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+# Initialize at import time
+try:
+    _init_rerun()
+except Exception as e:
+    logger.warning(f"Failed to initialize Rerun: {e}")
+
+
+__all__ = ["rr"]
```
paul-nechifor:
```suggestion
```

No need to re-export it. `import rerun as rr` is already short

dimos/msgs/nav_msgs/OccupancyGrid.py
```diff
@@ -416,3 +416,247 @@ def cell_value(self, world_position: Vector3) -> int:
...
+
+            # Occupied (100) = black, costs (1-99) = gradient
+            occupied_mask = self.grid > 0
+            if np.any(occupied_mask):
+                # Map 1-100 to 127-0 (darker = more occupied)
+                costs = self.grid[occupied_mask].astype(np.float32)
+                vis[occupied_mask] = (127 * (1 - costs / 100)).astype(np.uint8)
+
+            # Flip vertically to match world coordinates (y=0 at bottom)
+            return rr.Image(np.flipud(vis), color_model="L")
```
paul-nechifor: We already have `visualize_occupancy_grid(og, "turbo")`

dimos/msgs/nav_msgs/OccupancyGrid.py
```diff
@@ -416,3 +416,247 @@ def cell_value(self, world_position: Vector3) -> int:
...
+
+        points = np.column_stack([wx, wy, wz])
+
+        # Determine colors
+        if colormap is not None:
+            import matplotlib.pyplot as plt
+
+            # Normalize costs to 0-1 range
+            cost_norm = costs / 100.0
+            cmap = plt.get_cmap(colormap)
```
paul-nechifor: I think this should be cached

dimos/web/websocket_vis/websocket_vis_module.py
Nabla7: new commit it lives dimos/dimos/web/templates

dimos/msgs/sensor_msgs/image_impls/CudaImage.py
```diff
@@ -647,6 +647,35 @@ def resize(self, width: int, height: int, interpolation: int = cv2.INTER_LINEAR)
...
+        match self.format:
+            case ImageFormat.RGB:
+                return rr.Image(cpu_data, color_model="RGB")
+            case ImageFormat.RGBA:
+                return rr.Image(cpu_data, color_model="RGBA")
+            case ImageFormat.BGR:
+                return rr.Image(cpu_data, color_model="BGR")
+            case ImageFormat.BGRA:
+                return rr.Image(cpu_data, color_model="BGRA")
+            case ImageFormat.GRAY | ImageFormat.DEPTH:
```
paul-nechifor: I think ImageFormat.DEPTH should return rr.DepthImage

dimos/msgs/sensor_msgs/image_impls/NumpyImage.py
```diff
@@ -126,6 +126,26 @@ def to_grayscale(self) -> NumpyImage:
...
+            case ImageFormat.BGR:
+                return rr.Image(self.data, color_model="BGR")
+            case ImageFormat.BGRA:
+                return rr.Image(self.data, color_model="BGRA")
+            case ImageFormat.GRAY | ImageFormat.DEPTH:
+                return rr.Image(self.data, color_model="L")
+            case ImageFormat.GRAY16 | ImageFormat.DEPTH16:
+                return rr.Image(self.data, color_model="L")
+            case _:
+                raise ValueError(f"Unsupported format for Rerun: {self.format}")
```
paul-nechifor: This match should be extracted to a shared function because it's identical to the one in CudaImage.py.

dimos/navigation/replanning_a_star/module.py
```diff
@@ -62,7 +66,10 @@ def start(self) -> None:
         unsub = self.target.subscribe(self._planner.handle_goal_request)
         self._disposables.add(Disposable(unsub))
 
-        self._disposables.add(self._planner.path.subscribe(self.path.publish))
+        def _publish_path(nav_path: Path) -> None:
+            self.path.publish(nav_path)  # Auto-logs to Rerun
+
+        self._disposables.add(self._planner.path.subscribe(_publish_path))
```
paul-nechifor: Why is this change needed?

dimos/robot/unitree/connection/go2.py
```diff
@@ -193,6 +209,66 @@ def start(self) -> None:
...
+                        name="Costmap",
+                        origin="world/nav/costmap",
+                    ),
+                    row_shares=[2, 1],
+                ),
+                column_shares=[3, 1],
+            ),
+            rrb.TimePanel(state="collapsed"),
+        )
+        rr.send_blueprint(blueprint)
```
paul-nechifor, line 275: I think Rerun blueprints should be tied to Dimos blueprints.

If GO2Connection is in charge of initializing the blueprint, then it has to know about all the modules that are started together with it. For example, if a separate camera is added or detections are enabled, the Rerun blueprints need to be expanded to include that.

dimos/robot/unitree/connection/go2.py
```diff
@@ -248,10 +329,47 @@ def _odom_to_tf(cls, odom: PoseStamped) -> list[Transform]:
...
-        self.tf.publish(*self._odom_to_tf(msg))
+        transforms = self._odom_to_tf(msg)
+        self.tf.publish(*transforms)
         if self.odom.transport:
             self.odom.publish(msg)
 
+        # Log to Rerun: robot pose (relative to parent entity "world")
+        rr.log(
+            "world/robot",
+            rr.Transform3D(
```
leshy, line 340: you should use transform.to_rerun() not manually construct

dimos/robot/unitree/connection/go2.py
```diff
@@ -40,6 +43,9 @@
 
 logger = setup_logger(level=logging.INFO)
 
+# URDF path for Go2 robot
+_GO2_URDF = Path(__file__).parent.parent / "go2" / "go2.urdf"
```
leshy, line 47: is this used? for urdf you need low level leg positions, we don't collect those atm,

PR #917, opened 2025-12-31
Feat/rerun latency panels

Nabla7:
# Viewer Backends

Dimos supports three visualization backends: Rerun (web or native) and Foxglove.

## Quick Start

Choose your viewer backend with the `VIEWER_BACKEND` environment variable:

```bash
# Rerun web viewer (default) - Full dashboard in browser
VIEWER_BACKEND=rerun-web dimos run unitree-go2

# Rerun native viewer - Fast native window + control center in browser
VIEWER_BACKEND=rerun-native dimos run unitree-go2

# Foxglove - Use Foxglove Studio instead of Rerun
VIEWER_BACKEND=foxglove dimos run unitree-go2
```

### Rerun Web (`rerun-web`)

**What you get:**
- Full dashboard at http://localhost:7779
- Rerun 3D viewer + command center sidebar in one page
- Works in browser, no display required (headless-friendly)

---

### Rerun Native (`rerun-native`)

**What you get:**
- Native Rerun application (separate window opens automatically)
- Command center at http://localhost:7779
- Better performance with larger maps/higher resolution

---

### Foxglove (`foxglove`)

**What you get:**
- Foxglove bridge on ws://localhost:8765
- No Rerun (saves resources)
- Better performance with larger maps/higher resolution
- Open layout: `dimos/assets/foxglove_dashboards/go2.json`

---

### Symptom: Slow Map Updates

If you notice:
- Robot appears to "walk across empty space"
- Costmap updates lag behind the robot
- Visualization stutters or freezes

This happens on lower-end hardware (NUC, older laptops) with large maps.

### Increase Voxel Size

Edit [`dimos/robot/unitree_webrtc/unitree_go2_blueprints.py`](../dimos/robot/unitree_webrtc/unitree_go2_blueprints.py) line 82:

```python
# Before (high detail, slower on large maps)
voxel_mapper(voxel_size=0.05),  # 5cm voxels

# After (lower detail, 8x faster)
voxel_mapper(voxel_size=0.1),   # 10cm voxels
```

**Trade-off:**
- Larger voxels = fewer voxels = faster updates
- But slightly less detail in the map

### DIFF
--- a/dimos/mapping/costmapper.py
+++ b/dimos/mapping/costmapper.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 from dataclasses import asdict, dataclass, field
-import os
 import queue
 import threading
 import time
@@ -34,7 +33,6 @@ from dimos.mapping.pointclouds.occupancy import (
 from dimos.msgs.nav_msgs import OccupancyGrid
 from dimos.msgs.sensor_msgs import PointCloud2
 from dimos.utils.logging_config import setup_logger
-from dimos.utils.metrics import timed
 
 logger = setup_logger()
 
--- a/dimos/msgs/nav_msgs/OccupancyGrid.py
+++ b/dimos/msgs/nav_msgs/OccupancyGrid.py
@@ -521,23 +521,23 @@ class OccupancyGrid(Timestamped):
             return rr.Image(np.flipud(vis), color_model="RGBA")
 
         # Grayscale visualization (no colormap)
-        vis = np.zeros((self.height, self.width), dtype=np.uint8)
+        vis_gray = np.zeros((self.height, self.width), dtype=np.uint8)
 
         # Free space = white
-        vis[self.grid == 0] = 255
+        vis_gray[self.grid == 0] = 255
 
         # Unknown = gray
-        vis[self.grid == -1] = 128
+        vis_gray[self.grid == -1] = 128
 
         # Occupied (100) = black, costs (1-99) = gradient
         occupied_mask = self.grid > 0
         if np.any(occupied_mask):
             # Map 1-100 to 127-0 (darker = more occupied)
             costs = self.grid[occupied_mask].astype(np.float32)
-            vis[occupied_mask] = (127 * (1 - costs / 100)).astype(np.uint8)
+            vis_gray[occupied_mask] = (127 * (1 - costs / 100)).astype(np.uint8)
 
         # Flip vertically to match world coordinates (y=0 at bottom)
-        return rr.Image(np.flipud(vis), color_model="L")
+        return rr.Image(np.flipud(vis_gray), color_model="L")
 
     def _to_rerun_points(self, colormap: str | None = None, z_offset: float = 0.01):  # type: ignore[no-untyped-def]
         """Convert to 3D points for occupied cells."""
@@ -619,7 +619,7 @@ class OccupancyGrid(Timestamped):
         vertices[:, 3, 1] = wy + r
         vertices[:, 3, 2] = z_offset
         # Flatten to (n_cells*4, 3)
-        vertices = vertices.reshape(-1, 3)
+        flat_vertices = vertices.reshape(-1, 3)
 
         # === VECTORIZED INDEX GENERATION ===
         # Base vertex indices for each cell: [0, 4, 8, 12, ...]
@@ -633,7 +633,7 @@ class OccupancyGrid(Timestamped):
         indices[:, 1, 1] = base_v + 2
         indices[:, 1, 2] = base_v + 3
         # Flatten to (n_cells*2, 3)
-        indices = indices.reshape(-1, 3)
+        flat_indices = indices.reshape(-1, 3)
 
         # === VECTORIZED COLOR GENERATION ===
         cell_values = self.grid[gy, gx]  # Get all cell values at once
@@ -664,7 +664,7 @@ class OccupancyGrid(Timestamped):
         colors = np.repeat(colors_per_cell, 4, axis=0)  # (n_cells*4, 4)
 
         return rr.Mesh3D(
-            vertex_positions=vertices,
-            triangle_indices=indices,
+            vertex_positions=flat_vertices,
+            triangle_indices=flat_indices,
             vertex_colors=colors,
         )
--- a/dimos/perception/detection/moduleDB.py
+++ b/dimos/perception/detection/moduleDB.py
@@ -23,8 +23,7 @@ from dimos_lcm.foxglove_msgs.ImageAnnotations import (
 from lcm_msgs.foxglove_msgs import SceneUpdate  # type: ignore[import-not-found]
 from reactivex.observable import Observable
 
-from dimos import spec
-from dimos.core import DimosCluster, In, Out, rpc
+from dimos.core import In, Out, rpc
 from dimos.msgs.geometry_msgs import PoseStamped, Quaternion, Transform, Vector3
 from dimos.msgs.sensor_msgs import Image, PointCloud2
 from dimos.msgs.vision_msgs import Detection2DArray
--- a/dimos/robot/unitree_webrtc/unitree_go2_blueprints.py
+++ b/dimos/robot/unitree_webrtc/unitree_go2_blueprints.py
@@ -19,7 +19,6 @@ import platform
 from dimos_lcm.foxglove_msgs.ImageAnnotations import (  # type: ignore[import-untyped]
     ImageAnnotations,
 )
-from lcm_msgs.foxglove_msgs import SceneUpdate  # type: ignore[import-not-found]
 
 from dimos.agents.agent import llm_agent
 from dimos.agents.cli.human import human_input
@@ -41,7 +40,6 @@ from dimos.navigation.frontier_exploration import (
 from dimos.navigation.replanning_a_star.module import (
     replanning_a_star_planner,
 )
-from dimos.perception.detection.module3D import Detection3DModule, detection3d_module
 from dimos.perception.detection.moduleDB import ObjectDBModule, detectionDB_module
 from dimos.perception.spatial_perception import spatial_memory
 from dimos.robot.foxglove_bridge import foxglove_bridge


dimos/robot/unitree/connection/go2.py
```diff
@@ -40,6 +43,9 @@
 
 logger = setup_logger(level=logging.INFO)
 
+# URDF path for Go2 robot
+_GO2_URDF = Path(__file__).parent.parent / "go2" / "go2.urdf"
```
paul-nechifor, line 48: You also need this in pypackage.toml to include the urdf file in the built package:

```toml
[tool.setuptools.package-data]
"dimos.robot.unitree.go2" = ["*.urdf"]
```

dimos/robot/unitree/connection/go2.py
```diff
@@ -137,6 +143,18 @@ class GO2Connection(Module, spec.Camera, spec.Pointcloud):
     _global_config: GlobalConfig
     _camera_info_thread: Thread | None = None
 
+    @classmethod
+    def rerun_views(cls):  # type: ignore[no-untyped-def]
+        """Return Rerun view blueprints for GO2 camera visualization."""
+        import rerun.blueprint as rrb
```
paul-nechifor: Please move imports to the top.

dimos/robot/unitree/connection/go2.py
```diff
@@ -179,9 +197,20 @@ def start(self) -> None:
...
+            self.color_image.publish(image)
+            rr.log("world/robot/camera/rgb", image.to_rerun())
+
+        def onodom(odom: PoseStamped) -> None:
+            self._publish_tf(odom)
+
         self._disposables.add(self.connection.lidar_stream().subscribe(self.lidar.publish))
-        self._disposables.add(self.connection.odom_stream().subscribe(self._publish_tf))
-        self._disposables.add(self.connection.video_stream().subscribe(self.color_image.publish))
+        self._disposables.add(self.connection.odom_stream().subscribe(onodom))
```
paul-nechifor:
```suggestion
        self._disposables.add(self.connection.odom_stream().subscribe(self._publish_tf))
```
There's no need for `def onodom`

dimos/robot/foxglove_bridge.py
```diff
@@ -15,30 +15,44 @@
...
         self.jpeg_shm_channels = jpeg_shm_channels or []
+        self._global_config = global_config
 
     @rpc
     def start(self) -> None:
         super().start()
+        
+        # Skip if Rerun is the selected viewer backend
+        if self._global_config and self._global_config.viewer_backend.startswith("rerun"):
+            logger.info(f"Foxglove bridge skipped (viewer_backend={self._global_config.viewer_backend})")
```
paul-nechifor:
```suggestion
            logger.info("Foxglove bridge skipped", viewer_backend=self._global_config.viewer_backend)
```
More idiomatic with structlog.

dimos/msgs/sensor_msgs/CameraInfo.py
```diff
@@ -372,6 +372,30 @@ def __eq__(self, other) -> bool:  # type: ignore[no-untyped-def]
...
+    def to_rerun(self, image_plane_distance: float = 0.5):  # type: ignore[no-untyped-def]
+        """Convert to Rerun Pinhole archetype for camera frustum visualization.
+
+        Args:
+            image_plane_distance: Distance to draw the image plane in the frustum
+
+        Returns:
+            rr.Pinhole archetype for logging to Rerun
+        """
+        import rerun as rr
```
paul-nechifor: Please move to the top.

dimos/msgs/sensor_msgs/PointCloud2.py
```diff
@@ -44,6 +44,14 @@
 from dimos.types.timestamped import Timestamped
 
 
+@functools.lru_cache(maxsize=16)
+def _get_matplotlib_cmap(name: str):  # type: ignore[no-untyped-def]
+    """Get a matplotlib colormap by name (cached for performance)."""
+    import matplotlib.pyplot as plt
```
paul-nechifor: Please move all the imports out of functions.

dimos/core/stream.py
```diff
@@ -173,8 +174,75 @@ def publish(self, msg) -> None:  # type: ignore[no-untyped-def]
...
+
+        Args:
+            cb: Callback function to receive messages
+
+        Returns:
+            Unsubscribe function
+        """
+        return self.transport.subscribe(cb, self)  # type: ignore[arg-type, func-returns-value, no-any-return]
+
+    def to_rerun(
```
paul-nechifor: Personally, I was confused by seeing this below:

```
# Auto-log path to Rerun
self.path.to_rerun("world/nav/path")
```

I didn't understand how that autologs, it seemed like it just called `Path.to_rerun` once. I didn't realize that `Out.to_rerun` is different.

Maybe rename this method to `autolog_to_rerun` so it's not confused for `to_rerun` on messages.

dimos/core/global_config.py
```diff
@@ -29,6 +29,9 @@ class GlobalConfig(BaseSettings):
     robot_ip: str | None = None
     simulation: bool = False
     replay: bool = False
+    rerun_enabled: bool = True
+    rerun_server_addr: str | None = None
+    viewer_backend: str = "rerun-native"  # "rerun-web", "rerun-native", or "foxglove"
```
paul-nechifor: Use Literal, to avoid comments, do proper type checking, and have typer generate help examples. This is how `planner_strategy` does it:


---

```python
ViewerBackend: TypeAlias = Literal["rerun-web", "rerun-native", "foxglove"]
...
viewer_backend: ViewerBackend = "rerun-native"
```

dimos/mapping/costmapper.py
```diff
@@ -37,27 +47,122 @@ class CostMapper(Module):
...
+
+        return [
+            rrb.TimeSeriesView(
+                name="Costmap (ms)",
+                origin="/metrics/costmap",
+                contents=["+ /metrics/costmap/calc_ms"],
+            ),
+        ]
+
+    def __init__(self, **kwargs: object) -> None:
```
paul-nechifor:
```suggestion
    def __init__(self, **kwargs: Any) -> None:
```

dimos/mapping/costmapper.py
```diff
@@ -37,27 +47,122 @@ class CostMapper(Module):
...
+            except queue.Empty:
+                continue
 
     @rpc
     def start(self) -> None:
         super().start()
+        
+        # Only start Rerun logging if Rerun backend is selected
+        viewer_backend = os.environ.get("VIEWER_BACKEND", "rerun-web").lower()
+        if viewer_backend.startswith("rerun"):
```
paul-nechifor: VIEWER_BACKEND isn't necesarely passed an an env var, it can be passed as `dimos --viewer-backend rerun-web run ...`. All modules receive `global_config` in `__init__`.  Save it in `self._global_config` and use



```suggestion
        if self._global_config.viewer_backend.startswith("rerun"):
```

dimos/mapping/voxels.py
```diff
@@ -72,10 +101,50 @@ def __init__(self, **kwargs: object) -> None:
...
+                continue
 
     @rpc
     def start(self) -> None:
         super().start()
+        
+        # Only start Rerun logging if Rerun backend is selected
+        import os
+        viewer_backend = os.environ.get("VIEWER_BACKEND", "rerun-web").lower()
+        if viewer_backend.startswith("rerun"):
```
paul-nechifor: Same as above, use global_config.viewer_backend

dimos/msgs/nav_msgs/OccupancyGrid.py
```diff
@@ -29,6 +30,21 @@
 from dimos.msgs.geometry_msgs import Pose, Vector3, VectorLike
 from dimos.types.timestamped import Timestamped
 
+# Optional Rerun visualization support
+try:
+    import rerun as rr
+except ImportError:
+    rr = None  # type: ignore[assignment]
```
paul-nechifor: Rerun is listed as a core dependency. Why the try-except?

dimos/dashboard/rerun_init.py
```diff
@@ -0,0 +1,154 @@
...
+
+    _server_started = True
+
+    # Register shutdown handler
+    atexit.register(shutdown_rerun)
+
+    return RERUN_GRPC_ADDR
+
+
+def connect_rerun(server_addr: str | None = None) -> None:
```
paul-nechifor: connect_rerun is called from start() in modules, but start() runs in a thread and there can be multiple modules in the same worker.

Does rerun only need to be initialized once per process? If so, then this should use a lock.

dimos/dashboard/rerun_init.py
jeff-hykin, line 1: > Worker modules connect to the server via connect_rerun

Thread locks dont work for workers because workers can be using python multiprocessing. A multiprocessing lock needs to be used instead. Right now I think the code is working by coincidence (you could remove the lock entirely and get similar behavior).

Nabla7: When I removed the lock I would get errors where two modules would be trying to start the rerun server at the same time.

Can you provide the errors you are getting now?

jeff-hykin: The thread lock will prevent an error when its using threads (the lock is doing something, it just doesnt work under all conditions). Sometimes dask uses threads but sometimes it uses processeses. It depends how many modules are loaded and what `n_workers` is set to. If dask uses processes then the lock won't actually lock.

IMO the worst bugs are the ones that dont give errors. Even with just threads, adding a random time delay (instead of a lock) would make an error nearly impossible to reproduce, but doesnt fix the issue.

I havent run the code, I just wanted to give you a heads up that I think theres a logic issue thats going to cause unreliable behavior as more modules are added and as people set custom values for n_workers. 

If I were to try to make an error super likely, I would set n_workers=25, threads_per_worker=1, and then init rerun on 25 modules.

PR #918, opened 2025-12-31
Modular Manipulation Planning stack

mustafab0:
Adds a protocol-based manipulation planning system with Drake backend for motion planning, IK, and collision checking.

1. Uses a protocol for WorldSpec, KinematicsSpec, PlannerSpec
2. Factory builds implementations of the different specs, easy to swap different planners, and solvers
3. Drake implementation for physics, IK, and RRT-Connect planning
4. ManipulationModule for RPC-based planning operations
5. WorldMonitor for reactive state sync with real robots
6. Dynamic obstacle management (box, sphere, cylinder)

### DIFF
--- a/dimos/manipulation/planning/world/drake_world.py
+++ b/dimos/manipulation/planning/world/drake_world.py
@@ -53,7 +53,6 @@ from __future__ import annotations
 from contextlib import contextmanager
 import copy
 from dataclasses import dataclass
-import logging
 from pathlib import Path
 from threading import RLock
 from typing import TYPE_CHECKING, Any
@@ -66,6 +65,7 @@ from dimos.manipulation.planning.spec import (
     ObstacleType,
     RobotModelConfig,
 )
+from dimos.utils.logging_config import setup_logger
 
 if TYPE_CHECKING:
     from collections.abc import Generator
@@ -105,7 +105,7 @@ try:
 except ImportError:
     DRAKE_AVAILABLE = False
 
-logger = logging.getLogger(__name__)
+logger = setup_logger()
 
 
 @dataclass
--- a/dimos/manipulation/manipulation_skills.py
+++ b/dimos/manipulation/manipulation_skills.py
@@ -40,17 +40,17 @@ cluster.link(skills.rpc >> agent.register_skills)
 
 from __future__ import annotations
 
-import logging
 from typing import TYPE_CHECKING
 
 from dimos.core.core import rpc
 from dimos.core.skill_module import SkillModule
 from dimos.protocol.skill.skill import skill
+from dimos.utils.logging_config import setup_logger
 
 if TYPE_CHECKING:
     from dimos.core.rpc_client import RpcCall
 
-logger = logging.getLogger(__name__)
+logger = setup_logger()
 
 
 class ManipulationSkillContainer(SkillModule):
--- a/dimos/manipulation/planning/kinematics/drake_kinematics.py
+++ b/dimos/manipulation/planning/kinematics/drake_kinematics.py
@@ -37,7 +37,6 @@ Example:
 
 from __future__ import annotations
 
-import logging
 from typing import TYPE_CHECKING
 
 import numpy as np
@@ -49,6 +48,7 @@ from dimos.manipulation.planning.utils.kinematics_utils import (
     compute_pose_error,
     damped_pseudoinverse,
 )
+from dimos.utils.logging_config import setup_logger
 
 if TYPE_CHECKING:
     from numpy.typing import NDArray
@@ -62,7 +62,7 @@ try:
 except ImportError:
     DRAKE_AVAILABLE = False
 
-logger = logging.getLogger(__name__)
+logger = setup_logger()
 
 
 class DrakeKinematics:
--- a/dimos/manipulation/planning/mesh_utils.py
+++ b/dimos/manipulation/planning/mesh_utils.py
@@ -32,7 +32,6 @@ Example:
 from __future__ import annotations
 
 import hashlib
-import logging
 import os
 from pathlib import Path
 import re
@@ -40,7 +39,9 @@ import shutil
 import tempfile
 from typing import TYPE_CHECKING
 
-logger = logging.getLogger(__name__)
+from dimos.utils.logging_config import setup_logger
+
+logger = setup_logger()
 
 # Cache directory for processed URDFs
 _CACHE_DIR = Path(tempfile.gettempdir()) / "dimos_urdf_cache"
--- a/dimos/manipulation/planning/monitor/world_obstacle_monitor.py
+++ b/dimos/manipulation/planning/monitor/world_obstacle_monitor.py
@@ -26,7 +26,6 @@ Example:
 
 from __future__ import annotations
 
-import logging
 import time
 from typing import TYPE_CHECKING
 
@@ -36,6 +35,7 @@ from dimos.manipulation.planning.spec import (
     Obstacle,
     ObstacleType,
 )
+from dimos.utils.logging_config import setup_logger
 
 if TYPE_CHECKING:
     from collections.abc import Callable
@@ -46,7 +46,7 @@ if TYPE_CHECKING:
 
     from dimos.manipulation.planning.spec import WorldSpec
 
-logger = logging.getLogger(__name__)
+logger = setup_logger()
 
 
 class WorldObstacleMonitor:
--- a/dimos/manipulation/planning/planners/drake_planner.py
+++ b/dimos/manipulation/planning/planners/drake_planner.py
@@ -33,7 +33,6 @@ Example:
 from __future__ import annotations
 
 from dataclasses import dataclass, field
-import logging
 import time
 from typing import TYPE_CHECKING
 
@@ -45,11 +44,12 @@ from dimos.manipulation.planning.utils.path_utils import (
     interpolate_path,
     interpolate_segment,
 )
+from dimos.utils.logging_config import setup_logger
 
 if TYPE_CHECKING:
     from numpy.typing import NDArray
 
-logger = logging.getLogger(__name__)
+logger = setup_logger()
 
 
 @dataclass


dimos/hardware/manipulators/piper/piper_description/meshes/dae/base_link.dae
```diff
@@ -0,0 +1,320 @@
+<?xml version="1.0" encoding="utf-8"?>
```
paul-nechifor: All of these dae, stl, obj, etc files should be put into Git LFS. Even though many are text files, they are not source code, so they shouldn't be tracked by Git.

If you don't anticipate editing them individually, you can put them all in an archive and move them to `data`. There's the a helper script in `bin/lfs_push`.  I'm not sure where these files are used, but you can automatically download and extract them and using them in Python with the `get_data` function.

dimos/hardware/manipulators/integration_test_runner.py
```diff
@@ -0,0 +1,626 @@
...
+
+"""
+Integration test runner for manipulator drivers.
+
+This is a standalone script (NOT a pytest test file) that tests the common
+BaseManipulatorDriver interface that all arms implement.
+Supports both mock mode (for CI/CD) and hardware mode (for real testing).
+
+NOTE: This file is intentionally NOT named test_*.py to avoid pytest auto-discovery.
+For pytest-based unit tests, see: dimos/hardware/manipulators/base/tests/test_driver_unit.py
```
paul-nechifor, line 24: If it's not using pytest, how is it being executed in CI?

I've seen that we already have many tests which are never executed in CI. I'm not sure what value tests bring if they're not executed automatically for new code changes.

Is it possible to run them in GitHub?

dimos/hardware/manipulators/piper/piper_description/meshes/dae/base_link.dae
mustafab0: Understood, will update this!

dimos/hardware/manipulators/integration_test_runner.py
mustafab0: I agree with you. 

This file is part of the generalized_manipulator_driver it just runs a test to behave as a mock sdk, it can also be used to interface with real hardware if available.

This test is a manual test. Maybe not a good place for it to be? 

I have to redo the structure for the hardware manipulator driver anyway.

dimos/hardware/manipulators/piper/piper_description/meshes/dae/base_link.dae
mustafab0: Added all mesh files and their corresponding types to git lfs tracking.

The older commits still have the mesh files in the commit history (roughly ~132mb). It's okay to leave it as is for now, right?

paul-nechifor: I think you should squash the commits to remove the files from the history.

dimos/utils/cli/lcmspy/lcmspy.py
```diff
@@ -70,7 +70,8 @@ def _get_messages_in_window(self, time_window: float):  # type: ignore[no-untype
         """Get messages within the specified time window"""
         current_time = time.time()
         cutoff_time = current_time - time_window
-        return [(ts, size) for ts, size in self.message_history if ts >= cutoff_time]
+        # Copy to list to avoid RuntimeError: deque mutated during iteration
+        return [(ts, size) for ts, size in list(self.message_history) if ts >= cutoff_time]
```
paul-nechifor, line 74: I think the real issue here is that self.message_history is read/mutated without a lock.

dimos/simulation/mujoco/input_controller.py
```diff
@@ -18,6 +18,8 @@
 
 from numpy.typing import NDArray
 
+from dimos.hardware.end_effectors.end_effector import EndEffector
+
```
paul-nechifor:
```suggestion
```

Not used.

dimos/simulation/manipulators/mujoco_sim/model_utils.py
```diff
@@ -0,0 +1,103 @@
...
+        num_joints: Optional number of joints (for robots with multiple variants like xarm6, xarm7)
+
+    Returns:
+        Path to the scene.xml file
+
+    Raises:
+        FileNotFoundError: If the model file is not found
+    """
+    # Path: dimos/simulation/manipulators/mujoco_sim/ -> dimos/simulation/manipulators/data/
+    base_path = Path(__file__).parent.parent / "data"
```
paul-nechifor, line 44: I think all the data should be in `$repo/data/` not `$repo/dimos/`.

dimos/simulation/manipulators/mujoco_sim/bridge_base.py
```diff
@@ -0,0 +1,263 @@
...
+    THREAD_JOIN_TIMEOUT,
+    VELOCITY_STOP_THRESHOLD,
+)
+from dimos.simulation.manipulators.mujoco_sim.model_utils import load_manipulator_model
+from dimos.utils.logging_config import setup_logger
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+    pass
```
paul-nechifor:
```suggestion
```

dimos/simulation/manipulators/mujoco_sim/model_utils.py
leshy: if you just exec a `get_data('manipulation_piper')` you get an actual convinient Path object, there should never be a need for individual files to deal with pathing like this

dimos/simulation/manipulators/mujoco_sim/bridge_base.py
```diff
@@ -0,0 +1,263 @@
...
+            num_joints: Number of joints in the robot arm
+            control_frequency: Control frequency in Hz
+            model_path: Optional explicit path to MuJoCo model XML file.
+                       If None, will be found automatically.
+        """
+        self._robot_name = robot_name
+        self._num_joints = num_joints
+        self._control_frequency = (
+            control_frequency
+            if control_frequency > MIN_CONTROL_FREQUENCY
```
paul-nechifor:
```suggestion
            if control_frequency >= MIN_CONTROL_FREQUENCY
```

But I think you should throw an error.

As a general issue, if something is configured badly, it's better for the system to loudly exit rather than quietly override. Because the small problem is probably a symptom of a larger issue.

For example MIN_CONTROL_FREQUENCY is currently set to 0.01, and DEFAULT_CONTROL_FREQUENCY is set to 100. I can't imagine a situation where you'd want the frequency to be below 0.01, unless it's being confused with the period.

dimos/simulation/manipulators/data/xarm7/xarm7_nohand.xml
```diff
@@ -0,0 +1,106 @@
+<mujoco model="xarm7 nohand">
```
paul-nechifor, line 1: All the XML files are data files too. They should be in Git LFS in `data/`.

dimos/simulation/manipulators/data/xarm7/xarm7.png
paul-nechifor, line 1: This is a 1.6 MB file. It should be in LFS if used, but is it actually needed for anything? I think you should delete it.

dimos/manipulation/planning/world/drake_world.py
```diff
@@ -0,0 +1,948 @@
...
+        MultibodyPlant,
+    )
+    from pydrake.multibody.tree import JacobianWrtVariable
+    from pydrake.systems.framework import Context, DiagramBuilder
+
+    DRAKE_AVAILABLE = True
+except ImportError:
+    DRAKE_AVAILABLE = False
+
+logger = logging.getLogger(__name__)
```
paul-nechifor: why not `setup_logger()`

dimos/simulation/manipulators/data/piper/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
spomichter, line 1: what is this vibecoded shit we are not on an MIT license

paul-nechifor: This is the license of the copied in data. I did say we should put it all in an archive in `data/.lfs` since it's not actual source code for us.

dimos/manipulation/planning/world/drake_world.py
mustafab0: updated all manipulation files to use setup_logger.

dimos/simulation/manipulators/data/xarm7/xarm7_nohand.xml
mustafab0: As the simulation/manipulator and hardware/manipulator were developed independently we have these duplication.

dimos/hardware/manipulators/piper/piper_wrapper.py
```diff
@@ -43,53 +43,83 @@ def __init__(self) -> None:
...
+                self.logger.error(f"Unknown connection_type: {connection_type}")
                 return False
 
-        except ImportError:
-            self.logger.error("Piper SDK not installed. Please install piper_sdk")
+        except ImportError as e:
+            if connection_type == "hardware":
+                self.logger.error("Piper SDK not installed. Please install piper_sdk")
+            else:
+                self.logger.error(f"Import error: {e}")
```
paul-nechifor, line 122: The `try-except` should wrap just `from piper_sdk import C_PiperInterface_V2` at the start of the function. There's no need to place so much code inside the block.

dimos/hardware/manipulators/piper/piper_wrapper.py
```diff
@@ -43,53 +43,83 @@ def __init__(self) -> None:
...
-            )
+        connection_type = config.get("connection_type", "hardware")
 
-            # Connect to CAN port
-            self.native_sdk.ConnectPort(piper_init=True, start_thread=True)
-
-            # Wait for initialization
-            time.sleep(0.025)
+        try:
+            if connection_type == "hardware":
```
paul-nechifor, line 60: We have this problem different connection types in Unitree Go2 as well. But instead of handling both situations in each function by using an `if`, we handle each type in its own class.

```
    def start(self): 
        match self.connection_type:
            case "webrtc":
                self.connection = UnitreeWebRTCConnection(**self.connection_config)
            case "fake":
                self.connection = FakeRTC(**self.connection_config, seek=12.0)
            case "mujoco":
                from dimos.robot.unitree_webrtc.mujoco_connection import MujocoConnection

                self.connection = MujocoConnection(GlobalConfig())
                ...
            case _:
                raise ValueError(f"Unknown connection type: {self.connection_type}")
```

Note, we also support a third type of connection, "fake", for playing back recorded data.

This, BTW, also handles the import check you're having (you're handling the import for `from piper_sdk import C_PiperInterface_V2` in both cases, even though the piper sdk is not imported for `connection_type == "sim"`).

PR #937, opened 2026-01-04
WIP: User Install Experience

jeff-hykin:
- The user runs a curl command `sh <(curl -fsSL "https://raw.githubusercontent.com/dimensionalOS/dimos/refs/heads/master/install")`
- The installer has two stages:
- Stage 0 is `bin/install`: Its only job is to get python and run stage 1 (the python-based installer)
- Stage 1 is unders `installer/__main__.py`
    - The script figures out:
       1. What dimos features does the user want
       2. What system dependencies are needed (based on the selected features)
       3. How does the user want those system dependencies installed (nix, docker, native package manager)
       4. Is the user running in a project directory
       5. Is project environment setup correctly (venv, .env, .envrc)
       6. Did the pip-install actually work (post-install smoke tests)
    - The script is EITHER interactive (ask the user what features) or completely automated (no user input)
    - The script has 6 phases
        - ex: phases 0 and 1 are skipped by docker becaues they're purely interactive
    - The script assumes NOTHING about the user environment (no `git` or `grep`, not even python)
    - The dockerfile uses the script (this way there isn't two apt-get lists, its one unified installer)

Stage one is compiled into a pyz app (all pip dependencies are bundled)

To build the installer there are build scripts:
1. `build_pyz.py`
2. `refresh_system_dep_db.py` this creates a mapping from each pip module to what system deps it needs (apt, brew, nix). That mapping is what allows us to calculate what system dependencies a user would need based on what dimos features they want to enable.

Todo:
- .pyz files from history (will do on another branch)

### DIFF
--- a/installer/pyz_app/phases/phase00_logo_and_basic_checks.py
+++ b/installer/pyz_app/phases/phase00_logo_and_basic_checks.py
@@ -89,7 +89,7 @@ def phase0():
         time.sleep(timeout)
     logo.stop()
     p.clear_screen()
-    
+
     # ask user project question up front
     project_dir = get_project_directory()
     if installer_status.get("template_repo"):


installer/pyz_app/support/dimos_banner.py
```diff
@@ -0,0 +1,646 @@
...
+import random
+import re
+import shutil
+import sys
+import threading
+import time
+from typing import Any
+
+DEFAULT_BANNER = [
+    "██████╗ ██╗███╗   ███╗███████╗███╗   ██╗███████╗██╗ ██████╗ ███╗   ██╗ █████╗ ██╗           ██████╗ ███████╗",
```
spomichter, line 43: dimensional - drop the os

PR #938, opened 2026-01-04
EXAMPLES: Third party language interop

leshy:
Implements `simplerobot.py` - dimos robot Module in pygame that emits pose and takes twist (just for a light sample, unitree go2 real, replay or sim will work as well)

then writes 3 projects for diff languages that communicate to it and drive it:
- CPP
- LUA
- TS (+ web browser example)

### DIFF
--- a/examples/simplerobot/vis.py
+++ b/examples/simplerobot/vis.py
@@ -65,14 +65,14 @@ def run_visualization(robot, window_size=(800, 800), meters_per_pixel=0.02):
         pygame.draw.circle(screen, ROBOT, (rx, ry), 20)
 
         # Draw direction arrow
-        ax = rx + int(30 * math.cos(robot.theta))
-        ay = ry - int(30 * math.sin(robot.theta))
+        ax = rx + int(45 * math.cos(robot.theta))
+        ay = ry - int(45 * math.sin(robot.theta))
         pygame.draw.line(screen, ARROW, (rx, ry), (ax, ay), 3)
 
         # Arrowhead
         for sign in [-1, 1]:
-            hx = ax - int(10 * math.cos(robot.theta + sign * 2.5))
-            hy = ay + int(10 * math.sin(robot.theta + sign * 2.5))
+            hx = ax - int(10 * math.cos(robot.theta + sign * 0.5))
+            hy = ay + int(10 * math.sin(robot.theta + sign * 0.5))
             pygame.draw.line(screen, ARROW, (ax, ay), (hx, hy), 3)
 
         # Info text


examples/language-interop/cpp/main.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <cmath>
+#include <cstdio>
+#include <thread>
+#include <atomic>
+#include <chrono>
+
+#include "geometry_msgs/PoseStamped.hpp"
+#include "geometry_msgs/Twist.hpp"
+
+class RobotController {
```
spomichter, line 14: Super awesome

examples/language-interop/cpp/main.cpp
```diff
@@ -0,0 +1,68 @@
...
+public:
+    RobotController() : lcm_("udpm://239.255.76.67:7667?ttl=1"), running_(true) {}
```
spomichter, line 16: what is this IP address? @leshy

leshy: default LCM broadcast address, should be default and should delete this

examples/simplerobot/simplerobot.py
```diff
@@ -0,0 +1,193 @@
...
+                    self._linear_vel = Vector3()
+                    self._angular_vel = Vector3()
+                linear_vel = self._linear_vel
+                angular_vel = self._angular_vel
+
+            # Integrate velocity (unicycle model)
+            self.x += linear_vel.x * math.cos(self.theta) * dt
+            self.y += linear_vel.x * math.sin(self.theta) * dt
+            self.theta += angular_vel.z * dt
+            self.theta = math.atan2(math.sin(self.theta), math.cos(self.theta))
```
spomichter, line 108: This example would be much cleaner with some way to abstract safe velocity and control handling. So if we import one of our Controller classes that just takes in continuous Twist commands and internally handles threadsafe access and a heartbeat `if now - self._last_cmd_time > self.config.cmd_timeout`

spomichter: Downside here though is that maybe we want the user to understand what is happening with the controller....but i think most non-idiots can understand even if we abstract this a bit

examples/language-interop/README.md
```diff
@@ -0,0 +1,20 @@
+# Language Interop Examples
+
+Demonstrates controlling a dimos robot from non-Python languages.
+
+## Usage
+
+1. Start the robot (in another terminal):
+   ```bash
+   cd ../simplerobot
+   python main.py
```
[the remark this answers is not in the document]
spomichter: oh yeah this also @leshy

examples/simplerobot/simplerobot.py
leshy: agreed yeah, I hate this threading stuff, and we need better actual module examples, function of this is just to have a light thing to interface against in third party language integrations,
but will see what I can do yes

PR #942, opened 2026-01-04
Fix doclinks use git

spomichter:


docs/api/sensor_streams/reactivex.md
```diff
@@ -210,27 +210,8 @@ That's it. You create or receive an Observable, then subscribe to start receivin
 
 When you subscribe, data flows through a pipeline:
 
-<details>
```
leshy, line 213: just this deletion sneaked in, otherwise looking good

PR #955, opened 2026-01-07
temporal memory integration

ClaireBookworm:
Uses videoRAG https://github.com/shreyasrajesh0308/videorag (temporal branch) and integrates into the DimOS system. Works with @Skill calls the agent can use when referencing temporal information (Questions like "who woke up latest yesterday" or "who took out trash last week"). Creates a few jsons of entities, summaries, states, etc.

Todos:

add persistence (right now only keeps last 50 windows + entities & rolling summary), so just integrate simple sql db
support better time range queries (suppoed in original videorag for immediate context, fixed when adding db)
entity graphs (this would deprecate spatialmemory, currently, questions like "what objects are in the ktichen" require both modules).

### DIFF
--- a/dimos/agents/vlm_agent.py
+++ b/dimos/agents/vlm_agent.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from langchain_core.messages import AIMessage, HumanMessage, SystemMessage
+from langchain_core.messages import AIMessage, HumanMessage
 
 from dimos.agents.llm_init import build_llm, build_system_message
 from dimos.agents.spec import AgentSpec, AnyMessage
--- a/dimos/protocol/mcp/test_mcp_module.py
+++ b/dimos/protocol/mcp/test_mcp_module.py
@@ -21,7 +21,6 @@ from pathlib import Path
 import socket
 import subprocess
 import sys
-import threading
 
 import pytest
 
--- a/dimos/robot/unitree_webrtc/mujoco_connection.py
+++ b/dimos/robot/unitree_webrtc/mujoco_connection.py
@@ -61,9 +61,11 @@ class MujocoConnection:
         # Pre-download the mujoco_sim data.
         get_data("mujoco_sim")
 
-        # Trigger the download of the mujoco_menajerie package. This is so it
+        # Trigger the download of the mujoco_menagerie package. This is so it
         # doesn't trigger in the mujoco process where it can time out.
-        import mujoco_playground
+        from mujoco_playground._src import mjx_env
+
+        mjx_env.ensure_menagerie_exists()
 
         self.global_config = global_config
         self.process: subprocess.Popen[bytes] | None = None
--- a/dimos/stream/video_operators.py
+++ b/dimos/stream/video_operators.py
@@ -16,7 +16,7 @@ import base64
 from collections.abc import Callable
 from datetime import datetime, timedelta
 from enum import Enum
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING
 
 import cv2
 import numpy as np


dimos/models/vl/base.py
```diff
@@ -183,6 +183,32 @@ def _prepare_image(self, image: Image) -> tuple[Image, float]:
...
+            Response from the model
+        """
+        warnings.warn(
+            f"{self.__class__.__name__}.query_multi_images() is using default implementation "
+            "with only the first image. Override for true multi-image support.",
+            stacklevel=2,
+        )
+        if not images:
+            raise ValueError("Must provide at least one image")
+        return self.query(images[0], query, **kwargs)
```
paul-nechifor: `query_batch` has a default implementation because it's easy to just call `query` multiple times.

But this base implementation of `query_multi_images` drops data if someone uses multiple images. This is an error. A warning does not suffice. I think you should just remove the implementation or replace it with `raise NotImplementedError()`.

dimos/perception/temporal_memory.py
```diff
@@ -0,0 +1,780 @@
...
+        self.config = config or TemporalMemoryConfig()
+
+        # Verify API key if using OpenAIVlModel
+        if hasattr(vlm, "__class__") and "OpenAI" in vlm.__class__.__name__:
+            api_key = getattr(vlm.config, "api_key", None) if hasattr(vlm, "config") else None
+            if not api_key and not os.getenv("OPENAI_API_KEY"):
+                logger.warning(
+                    "OpenAIVlModel detected but OPENAI_API_KEY not found. "
+                    "Set OPENAI_API_KEY environment variable or pass api_key in config."
+                )
```
paul-nechifor: TemporalMemory should not be reaching into the properties of a particular class. What if QwenVlModel is passed?

You can define `VlModel.is_set_up` (or similar) and call it here. On OpenAIVlModel it can implement `is_set_up`, and throw an error (it shouldn't be a warning) if OPENAI_API_KEY is not defined.

dimos/perception/temporal_memory.py
```diff
@@ -0,0 +1,780 @@
...
+
+        # Override output_dir if provided
+        if output_dir is not None:
+            self.config.output_dir = output_dir
+
+        # Thread safety for state access
+        self.state_lock = threading.Lock()
```
paul-nechifor: More recently we've started prefixing private properties with `_`. Could you please prefix all properties which are not meant to be accessed from the outside with `_`?

dimos/perception/temporal_memory.py
```diff
@@ -0,0 +1,780 @@
...
+            api_key = getattr(vlm.config, "api_key", None) if hasattr(vlm, "config") else None
+            if not api_key and not os.getenv("OPENAI_API_KEY"):
+                logger.warning(
+                    "OpenAIVlModel detected but OPENAI_API_KEY not found. "
+                    "Set OPENAI_API_KEY environment variable or pass api_key in config."
+                )
+
+        # Override output_dir if provided
+        if output_dir is not None:
+            self.config.output_dir = output_dir
```
paul-nechifor: Why take output_dir if it's provided in `self.config`?

dimos/perception/temporal_memory.py
```diff
@@ -0,0 +1,780 @@
...
+        self.__dict__.update(state)
+        # Reinitialize runtime attributes
+        self.state_lock = threading.Lock()
+        # Reload environment variables on worker if needed for OpenAI VLM
+        if hasattr(self.vlm, "__class__") and "OpenAI" in self.vlm.__class__.__name__:
+            if hasattr(self.vlm, "config") and not getattr(self.vlm.config, "api_key", None):
+                # Try to load from environment on worker
+                api_key = _load_openai_api_key()
+                if api_key and hasattr(self.vlm, "config"):
+                    self.vlm.config.api_key = api_key
```
paul-nechifor: `self.vlm` should handle its own pickling. So you need to implement it on all subclasses of `VlModel`.

dimos/perception/temporal_memory.py
```diff
@@ -0,0 +1,780 @@
...
+        load_dotenv(env_file, override=True)
+        api_key_after_env = os.getenv("OPENAI_API_KEY")
+        if not api_key_after_env and default_env_file.exists():
+            load_dotenv(default_env_file, override=True)
+    elif default_env_file.exists():
+        load_dotenv(default_env_file, override=True)
+    else:
+        load_dotenv(override=True)
+
+    return os.getenv("OPENAI_API_KEY")
```
paul-nechifor: This should be deleted. Just call `load_dotenv()` in the blueprint file. (Or maybe it should be called in `robot/cli/dimos.py` so that no other file needs to call it.)

dimos/perception/temporal_memory.py
```diff
@@ -0,0 +1,780 @@
...
+
+
+# Blueprint for deployment
+temporal_memory = TemporalMemory.blueprint
+
+__all__ = [
+    "Frame",
+    "TemporalMemory",
+    "TemporalMemoryConfig",
+    "_load_openai_api_key",
```
paul-nechifor: Private functions should not be exported, but this function should be deleted anyway.

dimos/perception/temporal_memory_example.py
```diff
@@ -0,0 +1,130 @@
...
+        camera.stop()
+        print("\nTemporalMemory stopped")
+
+    finally:
+        if "temporal_memory" in locals():
+            temporal_memory.stop()
+        if "camera_module" in locals():
+            camera_module.stop()
+        if "dimos" in locals():
+            dimos.close_all()
```
paul-nechifor: Using `locals()` is bad practice. Set `temporal_memory = None` at the start and check if it's None here.

dimos/perception/test_temporal_memory_module.py
```diff
@@ -0,0 +1,237 @@
...
+
+        # Subscribe to the replay stream and publish to LCM
+        self._subscription = (
+            video_replay.stream()
+            .pipe(
+                ops.sample(1),  # Sample every 1 second
+                ops.take(10),  # Only take 10 frames total
+            )
+            .subscribe(self.video_out.publish)
+        )
```
paul-nechifor:
```suggestion
        self._disposables.add(
            video_replay.stream()
            .pipe(
                ops.sample(1),  # Sample every 1 second
                ops.take(10),  # Only take 10 frames total
            )
            .subscribe(self.video_out.publish)
        )
```

You don't need `_subscription`

dimos/perception/videorag_utils.py
```diff
@@ -0,0 +1,437 @@
...
+    """
+    # Try to extract JSON (handles code fences)
+    text = response_text.strip()
+    if text.startswith("```"):
+        parts = text.split("```")
+        if len(parts) >= 2:
+            candidate = parts[1].lstrip()
+            if candidate.startswith("json"):
+                candidate = candidate[4:].lstrip()
+            text = candidate.strip()
```
paul-nechifor: We have `extract_json` in `dimos/utils/llm_utils.py`

dimos/perception/test_temporal_memory_module.py
```diff
@@ -0,0 +1,237 @@
...
+            video_module.video_out.transport = core.LCMTransport("/test_video", Image)
+
+            # Temporal memory module
+            output_dir = os.path.join(temp_dir, "temporal_memory_output")
+            # Load API key before creating VLM to ensure it's available when pickled
+            from dimos.perception.temporal_memory import _load_openai_api_key
+
+            api_key = _load_openai_api_key()
+            if not api_key:
+                pytest.skip("OPENAI_API_KEY not found in environment")
```
paul-nechifor: Add this at the top:

```
@pytest.mark.skipif(not os.getenv("OPENAI_API_KEY"), reason="OPENAI_API_KEY not set.")
```

dimos/perception/test_temporal_memory_module.py
```diff
@@ -0,0 +1,237 @@
...
+            except Exception as e:
+                logger.warning(f"get_entity_roster failed: {e}")
+
+            # Test get_rolling_summary() RPC method
+            try:
+                summary = temporal_memory.get_rolling_summary()
+                logger.info(f"Rolling summary: {summary[:200] if summary else 'empty'}...")
+                assert isinstance(summary, str), "Rolling summary should be a string"
+            except Exception as e:
+                logger.warning(f"get_rolling_summary failed: {e}")
```
paul-nechifor: Please remove try-except.

dimos/perception/test_temporal_memory_module.py
```diff
@@ -0,0 +1,237 @@
...
+                    temporal_memory.stop()
+                except Exception:
+                    pass
+            if "video_module" in locals():
+                try:
+                    video_module.stop()
+                except Exception:
+                    pass
+            if "dimos" in locals():
+                dimos.close_all()  # Use close_all() instead of close() to shut down workers
```
paul-nechifor: Please don't use `locals()`.

I think you should define each of these as a fixture and to the cleanup in the fixture so it doesn't have to be done in the tests.

dimos/perception/temporal_memory.py
```diff
@@ -0,0 +1,780 @@
...
+                    for rec in frames_index:
+                        f.write(json.dumps(rec, ensure_ascii=False) + "\n")
+                logger.info(f"Saved {len(frames_index)} frames to {self.frames_index_file}")
+            return True
+        except Exception as e:
+            logger.error(f"Failed to save frames index: {e}", exc_info=True)
+            return False
+
+
+def _load_openai_api_key() -> str | None:
```
spomichter: Delete - use dotenv()

dimos/perception/temporal_memory.py
```diff
@@ -0,0 +1,780 @@
...
+    config: TemporalMemoryConfig | None = None,
+    output_dir: str | Path | None = None,
+) -> TemporalMemory:
+    if vlm is None:
+        from dimos.models.vl.openai import OpenAIVlModel
+
+        # Load API key from environment (ensures it's available when pickled to Dask workers)
+        api_key = _load_openai_api_key()
+
+        if not api_key:
```
spomichter: This check shouldnt be handled here. LLMs tend to add more Nonetype checks than is needed. Should keep this stuff tight so we're only adding try/catches or if/then error checks when there's actually a possibility of the object erroring out. Typechecking in python is supposed to also help and make much of these additional checks redundant

dimos/perception/temporal_memory.py
```diff
@@ -0,0 +1,780 @@
...
+            logger.error(
+                "OPENAI_API_KEY not found in environment. "
+                "Make sure default.env is loaded or set OPENAI_API_KEY environment variable."
+            )
+        else:
+            logger.info(f"OpenAI API key loaded (length: {len(api_key)})")
```
spomichter: Why does the length of the api key matter lmao

dimos/perception/temporal_memory.py
```diff
@@ -0,0 +1,780 @@
...
+            try:
+                answer = self.vlm.query(latest_frame.image, query_prompt)
+                return answer.strip()
+            except Exception as e:
+                logger.error(f"Query failed: {e}", exc_info=True)
+                return f"Error answering question: {e}"
+
+        return "No frames available to answer question"
+
+    @skill()
```
spomichter: great

dimos/perception/temporal_memory.py
ClaireBookworm: Yea oops, when i was using the Moondream API Key i was getting repeated weird "API key not found" issues so the key length check and this are remnants of debugging that I forgot to delete. Will do!

spomichter: ahh got it

dimos/models/vl/base.py
ClaireBookworm: Hmm this makes sense - the idea for query multiple images in one API call is for the model to see changes over time so it gets some form of temporal information, but can remove for now and add later (since that information isn't being very effectively used yet)?

dimos/protocol/rpc/spec.py
```diff
@@ -93,7 +93,12 @@ def serve_module_rpc(self, module: RPCInspectable, name: str | None = None) -> N
                 name = module.__class__.__name__
 
             def override_f(*args, fname=fname, **kwargs):  # type: ignore[no-untyped-def]
-                return getattr(module, fname)(*args, **kwargs)
+                # Get method from rpcs dict to avoid class attribute shadowing
+                # (e.g., when a stream named 'query' shadows an RPC method 'query')
+                method = module.rpcs[fname]
+                # Bind the method to the instance (module.rpcs returns unbound methods)
+                bound_method = method.__get__(module, type(module))
```
spomichter: What does this do? Avoid changes to core/ or protocol/ since you shouldn't need any changes to those for this feature without a very good reason

dimos/perception/temporal_memory.py
```diff
@@ -0,0 +1,771 @@
...
+
+def safe_float(value: Any, default: float) -> float:
+    """Safely convert value to float with fallback."""
+    try:
+        return float(value)
+    except (TypeError, ValueError):
+        return default
+
+
+class TemporalMemory(AgentSpec):
```
spomichter: Temporal memory isn't an agent, so it would just inherit from SkillModule. It would communicate with Agent via LCM and the connected query and answer topics.

dimos/models/vl/base.py
paul-nechifor: ~~Oh, I'm not saying you shouldn't have `query_multi_images`, just that it shouldn't have an implementation here and instead be an `@abstractmethod` like `query` is so that each actual model needs to implement it.~~ Oops, I was looking at an old version.

dimos/models/vl/base.py
```diff
@@ -228,6 +278,10 @@ def caption(self, image: Image) -> str:
         """Generate a caption by querying the VLM with a standard prompt."""
         return self.query(image, "Describe this image concisely.")
 
+    def is_set_up(self) -> None:
```
paul-nechifor:
```suggestion
    @abstractmethod
    def is_set_up(self) -> None:
```

dimos/models/vl/openai.py
```diff
@@ -0,0 +1,162 @@
...
+                self.config.api_key = api_key
+
+        # Verify setup (will raise ValueError if API key is still missing)
+        try:
+            self.is_set_up()
+        except ValueError as e:
+            # Log warning but don't fail - the error will be raised when _client is accessed
+            import logging
+            logger = logging.getLogger(__name__)
+            logger.warning(f"OpenAIVlModel setup verification failed after unpickling: {e}")
```
paul-nechifor: For logging, you should use `logger = setup_logger()` at the top. But in this case, if the model is not set up, it's unusable, so there's no point in continuing, right? Might as well let it raise an error.

dimos/models/vl/qwen.py
```diff
@@ -69,6 +69,52 @@ def query(self, image: Image | np.ndarray, query: str) -> str:  # type: ignore[o
 
         return response.choices[0].message.content  # type: ignore[return-value]
 
+    def query_multi_images(self, images: list[Image], query: str, response_format: dict | None = None) -> str:  # type: ignore[no-untyped-def, override]
```
paul-nechifor: You should implement `def is_set_up(self) -> None:` here as well and check for `ALIBABA_API_KEY`

dimos/perception/temporal_memory.py
```diff
@@ -0,0 +1,771 @@
...
+                try:
+                    self.color_image.transport.close()
+                except Exception as e:
+                    logger.warning(f"Error closing transport: {e}")
+            elif hasattr(self.color_image.transport, "dispose"):
+                try:
+                    self.color_image.transport.dispose()
+                except Exception as e:
+                    logger.warning(f"Error disposing transport: {e}")
+        logger.info("TemporalMemory stopped")
```
paul-nechifor: Transports are shared between modules so I'm not sure it makes sense to close them here.

dimos/perception/temporal_memory.py
```diff
@@ -0,0 +1,771 @@
...
+logger = setup_logger()
+
+
+@dataclass
+class Frame:
+    """Represents a single video frame with temporal metadata."""
+
+    frame_index: int
+    timestamp_s: float
+    timestamp: str  # Formatted as MM:SS.mmm
```
paul-nechifor: You can use the timestamp in `Image.ts`. And since the string representation can be derived, you probably don't need to store it.

dimos/perception/temporal_memory.py
```diff
@@ -0,0 +1,771 @@
...
+
+        Returns:
+            Dictionary containing entity roster, rolling summary, and statistics
+        """
+        # No lock needed - rx guarantees sequential execution
+        return {
+            "entity_count": len(self._state.get("entity_roster", [])),
+            "entities": list(self._state.get("entity_roster", [])),
+            "rolling_summary": str(self._state.get("rolling_summary", "")),
+            "frame_count": self._frame_count,
```
paul-nechifor: `_frame_count` is accessed here from a different thread than it's set in `set_image` function above.

A lot of our code is not very thread safe at the moment :confounded:, so it shouldn't be used as a model.

Please use a lock here and in all places where data is mutated/accessed from different threads.

dimos/perception/temporal_memory.py
```diff
@@ -0,0 +1,567 @@
...
+
+                frame = Frame(
+                    frame_index=self._frame_count,
+                    timestamp_s=timestamp_s,
+                    image=image,
+                )
+                self._frame_buffer.append(frame)
+                self._frame_count += 1
+
+        unsub_image = self.color_image.subscribe(on_frame)
```
spomichter: This is correct

dimos/perception/temporal_memory.py
```diff
@@ -0,0 +1,567 @@
...
+    """
+    builds temporal understanding of video streams using vlms.
+
+    processes frames reactively, maintains entity rosters, tracks temporal
+    relationships, builds rolling summaries. responds to queries about current
+    state and recent events.
+    """
+
+    color_image: In[Image]
+    query: In[HumanMessage]
```
spomichter: This is wrong query is OUT and answer would be IN. Since we're sending query OUT to the VLMAgent. But yeah as mentioned first pass we can just use the openAI(VLModel) so we dont need to get crazy with too many streams and we can revisit if we want to use AgentSpec

dimos/perception/temporal_memory.py
```diff
@@ -0,0 +1,567 @@
...
+                    timestamp_s=timestamp_s,
+                    image=image,
+                )
+                self._frame_buffer.append(frame)
+                self._frame_count += 1
+
+        unsub_image = self.color_image.subscribe(on_frame)
+        self._disposables.add(Disposable(unsub_image))
+
+        if self.query.transport is not None:
```
spomichter: Can delete now - will just have the color_image subscription

dimos/perception/temporal_memory.py
```diff
@@ -0,0 +1,567 @@
...
+
+        context = {
+            "entity_roster": entity_roster,
+            "rolling_summary": rolling_summary,
+            "currently_present_entities": sorted(currently_present),
+            "recent_windows_count": len(recent_windows),
+            "timestamp": time.time(),
+        }
+
+        prompt = f"""answer this question about the video stream using context.
```
spomichter: Shouldnt have the prompt dumped in line like this, in an external .txt file that you can import in is cleaner. Similar to how we do agent message in agents/

dimos/perception/temporal_memory.py
```diff
@@ -0,0 +1,567 @@
...
+    return {
+        "entity_roster": [],
+        "rolling_summary": "",
+        "chunk_buffer": [],
+        "next_summary_at_s": 0.0,
+        "last_present": [],
+    }
+
+
+class TemporalMemory(Module):
```
spomichter: Change to inherit from SkillModule which will allow you to tag things with @skill that you want the Agent to have access to

dimos/perception/temporal_memory.py
```diff
@@ -0,0 +1,567 @@
...
+    def append_history(self, *msgs: list[AIMessage | HumanMessage]) -> None:
+        with self._state_lock:
+            for msg in msgs:
+                self._query_history.append(msg)
+
+    def history(self) -> list[HumanMessage | AIMessage]:
+        with self._state_lock:
+            return list(self._query_history)
+
+    @rpc
```
spomichter: Can make this @skill

Issue #956, opened 2026-01-07
Rerun fixes


leshy:
## urdf 

```py
# URDF path for Go2 robot
_GO2_URDF = Path(__file__).parent.parent / "go2" / "go2.urdf"


        # Load robot URDF
        if _GO2_URDF.exists():
            rr.log_file_from_path(
                str(_GO2_URDF),
                entity_path_prefix="world/robot",
                static=True,
            )
            logger.info(f"Loaded URDF from {_GO2_URDF}")
```

this was done because foxglove is dumb and won't let me draw a box in 3d
with rerun you can, so no need for this fake box urdf, just draw the box.

## TF module

idk why we have in blueprints:
```py
    tf_rerun(),  # Auto-visualize all TF transforms in Rerun
```

but then we build/vis all the transforms manually
```py

        # Log camera transform (compose base_link -> camera_link -> camera_optical)
        # transforms[1] is camera_link, transforms[2] is camera_optical
        cam_tf = transforms[1] + transforms[2]  # compose transforms
        rr.log(
            "world/robot/camera",
            rr.Transform3D(
                translation=[cam_tf.translation.x, cam_tf.translation.y, cam_tf.translation.z],
                rotation=rr.Quaternion(
                    xyzw=[
                        cam_tf.rotation.x,
                        cam_tf.rotation.y,
                        cam_tf.rotation.z,
                        cam_tf.rotation.w,
                    ]
                ),
            ),
        )


        # Log to Rerun: robot pose (relative to parent entity "world")
        rr.log(
            "world/robot",
            rr.Transform3D(
                translation=[msg.x, msg.y, msg.z],
                rotation=rr.Quaternion(
                    xyzw=[
                        msg.orientation.x,
                        msg.orientation.y,
                        msg.orientation.z,
                        msg.orientation.w,
                    ]
                ),
            ),
        )
        # Log axes as a child entity for visualization
        rr.log("world/robot/axes", rr.TransformAxes3D(0.5))  # type: ignore[attr-defined]

```

we can't do any of this stuff in go2 connection.
at most (if not using a rerun tf module) it should be this:

```py
self.tf.publish(*transforms)

for transform in transforms:
    rr.log(.... transform.to_rerun())

```

actual rerun transform module should be similar to above also..

never build objects manually that are equivalent to dimos objects - create to_rerun() methods

## camera image distortion

camera image needs to be logged under a projection so that it takes camerainfo distortion into account 

## tf module

```
[...]

PR #959, opened 2026-01-07
Rerun issue956

Nabla7:
This PR addresses issue #956.
Changes:
TF visualization now polls the TF buffer (self.tf.buffers) at a configurable rate and logs a snapshot of the latest transforms. No direct LCM subscription inside the TF viz module.
GO2Connection no longer manually logs robot/camera transforms to Rerun (no composing transforms, no hand-built rr.Transform3D). It only publishes TF via self.tf.publish(*transforms) and logs static assets (URDF, axes, camera pinhole) + camera images.
TF logs are written under world/tf/{child_frame} so they’re visible under the default world origin without mixing semantic entity paths with TF.
Removed queue/thread “async rerun worker” patterns from CostMapper and VoxelGridMapper; logging is synchronous again. Freezing/perf is expected to be handled by voxel size (O(n) costmap/mesh generation), not background threads.
Added/updated documentation: inventory of all Rerun usage in the codebase and a short “how to use Rerun on dev” section with TF/entity path conventions.
Notes:
URDF loading remains (intentionally) for future robot model usage.
Camera distortion: Rerun pinhole uses intrinsics; distortion must be handled by undistorting images upstream if needed.


docs/VIEWER_BACKENDS.md
```diff
@@ -4,12 +4,21 @@ Dimos supports three visualization backends: Rerun (web or native) and Foxglove.
...
+Choose your viewer backend via the CLI (preferred):
 
 ```bash
-# Rerun native viewer (default) - Fast native window + control center
+# Rerun native viewer (default) - native Rerun window + teleop panel at http://localhost:7779
 dimos run unitree-go2
-# or explicitly:
+
+# Explicitly select the viewer backend:
+dimos --viewer-backend rerun-native run unitree-go2
```
spomichter, line 14: yeah cleaner for sure

dimos/dashboard/tf_rerun_module.py
```diff
@@ -78,33 +85,48 @@ def start(self) -> None:
...
+                        continue
+                    last_ts = self._last_ts_by_edge.get((parent, child))
+                    if last_ts is not None and latest.ts == last_ts:
+                        continue
+
+                    # Log under `world/tf/...` so it is visible under the default 3D view origin.
+                    rr.log(f"world/tf/{child}", latest.to_rerun())  # type: ignore[no-untyped-call]
+                    self._last_ts_by_edge[(parent, child)] = latest.ts
+            except Exception as e:
+                logger.warning("TFRerunModule: TF poll loop error", error=str(e))
```
paul-nechifor: What exception are you expecting here? It's probably not a good idea to wrap things in try-except just in case.

dimos/dashboard/tf_rerun_module.py
```diff
@@ -78,33 +85,48 @@ def start(self) -> None:
...
+            logger.info("TFRerunModule: started TF snapshot polling", poll_hz=self._poll_hz)
+
+    def _poll_loop(self) -> None:
+        assert self._stop_event is not None
+        period_s = 1.0 / max(self._poll_hz, 0.1)
+
+        while not self._stop_event.is_set():
+            try:
+                # Snapshot keys to avoid concurrent modification while TF buffer updates.
+                items = list(self.tf.buffers.items())  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
paul-nechifor: What greptile wrote makes sense, but I'm not sure what the proper fix for this is.

Maybe the best way would be to add a `get_buffers` method in `MultiTBuffer` which returns a clone, but such a method would require a lock as well and none of the code there uses locks, so it's a big change. But it looks like the only way.

dimos/dashboard/tf_rerun_module.py
Nabla7: will remove, no good reason indeed for being there.

dimos/dashboard/tf_rerun_module.py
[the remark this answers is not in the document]
Nabla7: > What greptile wrote makes sense, but I'm not sure what the proper fix for this is.
> 
> Maybe the best way would be to add a `get_buffers` method in `MultiTBuffer` which returns a clone, but such a method would require a lock as well and none of the code there uses locks, so it's a big change. But it looks like the only way.

I am also unsure, is there a public API that is available? before I was doing some hacky stuff but this may be equally hacky.

dimos/robot/unitree/connection/go2.py
```diff
@@ -120,7 +121,22 @@ def odom_stream(self):  # type: ignore[no-untyped-def]
     # we don't have raw video stream in the data set
     @simple_mcache
     def video_stream(self):  # type: ignore[no-untyped-def]
-        video_store = TimedSensorReplay(f"{self.dir_name}/video")  # type: ignore[var-annotated]
+        # Legacy Unitree recordings can have RGB bytes that were tagged/assumed as BGR.
+        # Fix at replay-time by coercing everything to RGB before publishing/logging.
+        def _autocast_video(x):  # type: ignore[no-untyped-def]
```
spomichter, line 121: Cant go here

spomichter (approved):

PR #960, opened 2026-01-07
class VLMAgent(AgentSpec, Module) for streamed VLM queries over Transport

spomichter:
To test: 
`dimos --replay run unitree-go2-vlm-stream-test`

### DIFF
--- a/dimos/agents/vlm_agent.py
+++ b/dimos/agents/vlm_agent.py
@@ -15,7 +15,7 @@
 from langchain_core.messages import AIMessage, HumanMessage, SystemMessage
 
 from dimos.agents.llm_init import build_llm, build_system_message
-from dimos.agents.spec import AgentSpec
+from dimos.agents.spec import AgentSpec, AnyMessage
 from dimos.core import rpc
 from dimos.core.stream import In, Out
 from dimos.msgs.sensor_msgs import Image
@@ -28,8 +28,8 @@ class VLMAgent(AgentSpec):
     """Stream-first agent for vision queries with optional RPC access."""
 
     color_image: In[Image]
-    query: In[HumanMessage]
-    answer: Out[AIMessage]
+    query_stream: In[HumanMessage]
+    answer_stream: Out[AIMessage]
 
     def __init__(self, *args, **kwargs) -> None:  # type: ignore[no-untyped-def]
         super().__init__(*args, **kwargs)
@@ -43,7 +43,7 @@ class VLMAgent(AgentSpec):
     def start(self) -> None:
         super().start()
         self._disposables.add(self.color_image.subscribe(self._on_image))  # type: ignore[arg-type]
-        self._disposables.add(self.query.subscribe(self._on_query))  # type: ignore[arg-type]
+        self._disposables.add(self.query_stream.subscribe(self._on_query))  # type: ignore[arg-type]
 
     @rpc
     def stop(self) -> None:
@@ -54,12 +54,12 @@ class VLMAgent(AgentSpec):
 
     def _on_query(self, msg: HumanMessage) -> None:
         if not self._latest_image:
-            self.answer.publish(AIMessage(content="No image available yet."))
+            self.answer_stream.publish(AIMessage(content="No image available yet."))
             return
 
         query_text = self._extract_text(msg)
         response = self._invoke_image(self._latest_image, query_text)
-        self.answer.publish(response)
+        self.answer_stream.publish(response)
 
     def _extract_text(self, msg: HumanMessage) -> str:
         content = msg.content
@@ -74,7 +74,7 @@ class VLMAgent(AgentSpec):
     def _invoke(self, msg: HumanMessage) -> AIMessage:
         messages = [self._system_message, msg]
         response = self._llm.invoke(messages)
-        self.append_history(msg, response)  # type: ignore[arg-type]
+        self.append_history([msg, response])  # type: ignore[arg-type]
         return response  # type: ignore[return-value]
 
     def _invoke_image(self, image: Image, query: str) -> AIMessage:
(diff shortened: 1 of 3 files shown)


dimos/agents/agent.py
```diff
@@ -362,6 +363,12 @@ def query(self, query: str):  # type: ignore[no-untyped-def]
         # return sync(self._loop, self.agent_loop, query)
         return asyncio.run_coroutine_threadsafe(self.agent_loop(query), self._loop).result()  # type: ignore[arg-type]
 
+    @rpc
+    def query_image(self, image: Image, query: str):  # type: ignore[no-untyped-def]
```
spomichter: This can be deleted - not used

dimos/agents/spec.py
```diff
@@ -190,6 +191,10 @@ def register_skills(
     @abstractmethod
     def query(self, query: str): ...  # type: ignore[no-untyped-def]
 
+    @rpc
+    @abstractmethod
```
spomichter: This will also be deleted

dimos/agents/modules/vlm_stream_tester.py
```diff
@@ -0,0 +1,156 @@
...
+                        content=f"{self._prompt} (stream query {idx + 1}/{self._num_queries})"
+                    )
+                )
+
+                try:
+                    rpc_query = self.get_rpc_calls("VLMAgent.query_image")
+                    response = rpc_query(
+                        self._latest_image,
+                        f"{self._prompt} (rpc query {idx + 1}/{self._num_queries})",
+                    )
```
paul-nechifor: If you want to query the latest image, why not publish to VLMAgent.query to avoid sending the latest image over RPC since VLMAgent already has the latest image?

dimos/agents/vlm_agent.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def _invoke(self, msg: HumanMessage) -> AIMessage:
+        messages = [self._system_message, msg]
+        response = self._llm.invoke(messages)
+        self.append_history(msg, response)  # type: ignore[arg-type]
+        return response  # type: ignore[return-value]
+
+    def _invoke_image(self, image: Image, query: str) -> AIMessage:
+        content = [{"type": "text", "text": query}, *image.agent_encode()]
+        return self._invoke(HumanMessage(content=content))
```
paul-nechifor, line 82: Yeah, this makes sense, but it only works if particular model supports images. For example if we use some Ollama model instead of hosted OpenAI, it might not work with images.

And we might want to call other VL models anyway. With this, we can't support a setup like this: use OpenAI for text or general image queries, use Qwen for map lookup, use Moondream for detection.

paul-nechifor: An option here would be to add an argument specifying which model should be used.

dimos/agents/vlm_agent.py
```diff
@@ -0,0 +1,154 @@
...
+
+    @rpc
+    def register_skills(  # type: ignore[no-untyped-def]
+        self, container, run_implicit_name: str | None = None
+    ) -> None:
+        logger.warning(
+            "VLMAgent does not manage skills; register_skills is a no-op",
+            container=str(container),
+            run_implicit_name=run_implicit_name,
+        )
```
paul-nechifor, line 105: Oh, so VLMAgent isn't a replacement for Agent? In that case would we be running two agent loops?

spomichter: VLMAgent in this case is meant to be run one-off yeah. But i guess we could alos change VLMAgent to inherit from Agent instead of AgentSpec, so then i would run a agent loop / skill coordinator

dimos/agents/modules/vlm_agent.py
```diff
@@ -0,0 +1,154 @@
...
+
+        if self.config.system_prompt:
+            if isinstance(self.config.system_prompt, str):
+                self._system_message = SystemMessage(self.config.system_prompt)
+            else:
+                self._system_message = self.config.system_prompt
+        else:
+            self._system_message = SystemMessage(SYSTEM_PROMPT)
+
+        self.publish(self._system_message)
```
paul-nechifor, line 40: This is very similar to Agent.__init__, it should be deduplicated somehow.

dimos/agents/vlm_agent.py
spomichter: Yeah can add this thanks

dimos/agents/modules/vlm_agent.py
spomichter: True

dimos/agents/vlm_agent.py
spomichter: Yeah so we need to do this because agent runs its own agent loop but listens to /skill. We haven't solved multi-agent yet. So if VLMAgent also inherited from Agent it would get bogged down by tool responses in parallel.

spomichter: Soon we will fix this so each agent also has its own skill coordinator topics and then we can run 5 agents that don't clash

dimos/agents/vlm_agent.py
spomichter: Going to defer this for now since Langchain doesnt have a standard way to handle this. Can probably add a supports_images: Bool param to the AgentConfig

dimos/agents/modules/vlm_stream_tester.py
spomichter: Good catch

dimos/agents/modules/vlm_agent.py
spomichter: Fixed with LLMInitMixin

dimos/agents/modules/vlm_stream_tester.py
spomichter: Ah sorry no this class was vibecoded - its showing an example of a one-off RPC call AND a stream query just as an example. Want to show how you can use VLMAgent both via a stream of humanMessage queries AND as a normal RPC call where you pass (query, image). FYI @ClaireBookworm

PR #962, opened 2026-01-08
Quest Teleop PR#1

ruthwikdasyam:
#### Features Included

- **WebSocket Support**: WebSocket server that supports VR headset and reads tracking data with passthrough view
- **Quest3TeleopModule**: Based on `BaseTeleopModule`, publishes: (Max - 90Hz)
  - 2x `PoseStamped` (left and right controller poses)
  - 2x `Float32` (gripper trigger values)
- **Calibrated Poses**: Quest3TeleopModule sends calibrated poses (delta position from initial position)
- **Button Controls**: Button press functionality to start and stop teleop
- **Visualization**: Visualization of controller movement and gripper topics on Rerun
- **TeleopRobotController**: Template module that:
  - Reads from `Quest3TeleopModule`
  - Publishes 2x `Pose` + 2x `Bool` as required by robot modules
  - Reads robot current position and sends absolute position for the robot to move
  - Module designed to support further add-ons such as IK, constraints, and safety features

#### Topics viz
Screenshot from 2026-01-07 17-47-46


dimos/teleop/base/base_teleop_module.py
```diff
@@ -0,0 +1,524 @@
...
+    - reset_calibration() -> dict: Reset calibration
+    - is_vr_calibrated() -> bool: Check calibration status
+    - get_status() -> dict: Get teleoperation status
+    """
+
+    default_config = BaseTeleopConfig
+    config: BaseTeleopConfig
+
+    # LCM Output topics
+    left_controller_delta: Out[PoseStamped] = None  # type: ignore[assignment]
```
spomichter: Why not just output a transform and then the left and right PoseStamped objects, such that much of your operations / transformations on pose can be done via a type rather core implementation

dimos/teleop/base/base_teleop_module.py
```diff
@@ -0,0 +1,524 @@
...
+        """Stop the teleoperation module."""
+        logger.info(f"Stopping {self.__class__.__name__}...")
+        super().stop()
+
+    # =========================================================================
+    # Calibration
+    # =========================================================================
+
+    @rpc
+    def calibrate_vr(self) -> dict[str, Any]:
```
spomichter: This should definitely not be here. base teleop needs to work with any teleop, even an xbox controller or an iphone imu

dimos/teleop/base/base_teleop_module.py
```diff
@@ -0,0 +1,524 @@
...
+        """
+        self._is_calibrated = False
+        self._left_controller_initial = None
+        self._right_controller_initial = None
+
+        logger.info("Calibration reset. Recalibrate to resume teleoperation...")
+        return {"success": True, "message": "Calibration reset - recalibrate to resume"}
+
+    @rpc
+    def is_vr_calibrated(self) -> bool:
```
spomichter: Nothing mentioning VR should be in the generic teleop module

dimos/teleop/base/base_teleop_module.py
```diff
@@ -0,0 +1,524 @@
...
+
+    @rpc
+    def get_status(self) -> dict[str, Any]:
+        """Get current teleoperation status.
+
+        Returns:
+            Dictionary with status information
+        """
+        return {
+            "is_calibrated": self._is_calibrated,
```
spomichter: This needs to be an Enum

dimos/teleop/base/base_teleop_module.py
```diff
@@ -0,0 +1,524 @@
...
+    def _compute_and_publish_deltas(self) -> None:
+        """Compute and publish delta poses (current - initial)."""
+        # Track publish count for logging
+        self._publish_count += 1
+
+        try:
+            current_time = time.time()
+
+            # Left controller delta
+            if self.config.enable_left_arm and self._left_controller_initial is not None:
```
spomichter: Slop

dimos/teleop/base/base_teleop_module.py
ruthwikdasyam: Planning to do something similar in the next PR. Having head pos along with left and right PostStamped.

Although, if I have the transforms, I won't have the VR_world to robot_base transform. I need to depend on relative motion of the controllers; thus, had it compute this way to keep it generalizable.
If go2, we can infer the same output as the magnitudes of velocities and move the robot. Without any frame dependency (for now)

dimos/teleop/base/base_teleop_module.py
ruthwikdasyam: Sorry, its in the right place, typo with the naming, changing it.

dimos/teleop/base/base_teleop_module.py
ruthwikdasyam: Similar issue, renaming methods

.gitignore
```diff
@@ -63,3 +63,6 @@ yolo11n.pt
 /.mypy_cache*
 
 *mobileclip*
+
+# Teleop SSL certificates
+dimos/teleop/web/certs/
```
paul-nechifor: `dimos/` should be just for source code. Maybe use `assets/`?

ruthwikdasyam: Moved certs to assets/

dimos/teleop/base/base_teleop_module.py
ruthwikdasyam: Done

dimos/teleop/base/base_teleop_module.py
ruthwikdasyam: Updated!

dimos/teleop/devices/base_teleop_module.py
```diff
@@ -0,0 +1,286 @@
...
+
+            for i, idx in enumerate(self._active_indices):
+                pose_matrix = self._all_poses.get(idx)
+                if pose_matrix is not None:
+                    pose_obj = matrix_to_pose(pose_matrix)
+                    self._initial_poses[idx] = pose_obj
+                    logger.info(f"Captured controller {self._get_label(idx)} initial: {pose_obj}")
+                else:
+                    return {
+                        "success": False,
```
spomichter: Needs type no json shit

dimos/teleop/devices/vr_teleop_module.py
```diff
@@ -0,0 +1,225 @@
...
+@dataclass
+class VRTeleopConfig(BaseTeleopConfig):
+    """Configuration for VR Teleoperation Module."""
+
+    output_types: list[type] = field(default_factory=lambda: [PoseStamped, PoseStamped])
+    input_labels: list[str] = field(default_factory=lambda: ["left_vr", "right_vr"])
+    control_loop_hz: float = 50.0
+
+
+class VRTeleopModule(BaseTeleopModule):
```
spomichter: Needs to be VRTeleopModule(BaseTeleopModule[VRTeleopModuleConfig]) or whatever the syntax is i forget

dimos/teleop/devices/vr_teleop_module.py
```diff
@@ -0,0 +1,225 @@
...
+    vr_right_transform: In[LCMTransform] = None  # type: ignore[assignment]
+    vr_trigger_0: In[Float32] = None  # type: ignore[assignment]
+    vr_trigger_1: In[Float32] = None  # type: ignore[assignment]
+    teleop_enable: In[Bool] = None  # type: ignore[assignment]  # X button calibration toggle
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+
+        self._lcm_lock = threading.Lock()
+        self._lcm_controller_poses: dict[int, NDArray[np.float64] | None] = {
```
spomichter: Needs a type for ControllerPose cant just have a dict of random floats here bc need cleanoperations on this tyoe

dimos/teleop/devices/vr_teleop_module.py
```diff
@@ -0,0 +1,225 @@
...
+
+        if self.vr_trigger_1 and self.vr_trigger_1.transport:
+            self.vr_trigger_1.subscribe(
+                lambda msg, idx=right_index: self._on_lcm_trigger(idx, msg)  # type: ignore[misc]
+            )
+
+        if self.teleop_enable and self.teleop_enable.transport:
+            self.teleop_enable.subscribe(self._on_lcm_teleop_enable)
+
+        logger.info("VR Teleoperation Module started")
```
spomichter: Clean up can use match here

dimos/utils/teleop_transforms.py
```diff
@@ -0,0 +1,235 @@
...
+        target_pose = delta_pose
+    return PoseStamped(
+        ts=time.time(),
+        frame_id="teleop_target",
+        position=target_pose.position,
+        orientation=target_pose.orientation,
+    )
+
+
+def _to_twist(
```
spomichter: This cannot be here needs to be in the type for the thing i said earlier we need which is ControllerPoseType which will wrap Vector3 and handle all conversion to vector3 very cleanly

dimos/teleop/devices/vr_teleop_module.py
spomichter: and then you can do stuff like vector3 type conversions there

dimos/msgs/geometry_msgs/Pose.py
```diff
@@ -222,6 +222,21 @@ def __add__(self, other: Pose | PoseConvertable | LCMTransform | Transform) -> P
 
         return Pose(new_position, new_orientation)
 
+    def __sub__(self, other: Pose | PoseConvertable) -> Pose:
```
spomichter: @leshy I assume there are issues with this due to other Pose operations. __add__ for example has some checks

dimos/teleop/devices/base_teleop_module.py
```diff
@@ -0,0 +1,286 @@
...
+
+    # Output topics: PoseStamped for arms (0,1), Twist for locomotion (2,3)
+    controller_delta_0: Out[PoseStamped] = None  # type: ignore
+    trigger_value_0: Out[Bool] = None  # type: ignore
+    controller_delta_1: Out[PoseStamped] = None  # type: ignore
+    trigger_value_1: Out[Bool] = None  # type: ignore
+    controller_delta_2: Out[Twist] = None  # type: ignore
+    trigger_value_2: Out[Bool] = None  # type: ignore
+    controller_delta_3: Out[Twist] = None  # type: ignore
+    trigger_value_3: Out[Bool] = None  # type: ignore
```
spomichter: What if we are teleoping without controllers?

dimos/teleop/devices/vr_teleop_module.py
ruthwikdasyam: Not exactly pattern matching over a value, but can use loop

dimos/teleop/devices/base_teleop_module.py
ruthwikdasyam: removed json type returns

dimos/teleop/devices/vr_teleop_module.py
ruthwikdasyam: Added ControllerPose msg type, shifted all math

dimos/teleop/devices/base_teleop_module.py
ruthwikdasyam: no controllers mean no subscribers to read data from vr.  Checks to make sure at least one `output_type` is specified

dimos/teleop/teleop_blueprints.py
```diff
@@ -0,0 +1,43 @@
...
+        output_types=[PoseStamped, Twist],
+        input_labels=["left_vr", "right_vr"],
+        visualize_in_rerun=True,
+    ),
+).transports(
+    {
+        ("vr_left_transform", LCMTransform): LCMTransport("/vr_left_transform", LCMTransform),
+        ("vr_right_transform", LCMTransform): LCMTransport("/vr_right_transform", LCMTransform),
+        ("vr_trigger_0", Float32): LCMTransport("/vr_trigger_0", Float32),
+        ("vr_trigger_1", Float32): LCMTransport("/vr_trigger_1", Float32),
```
paul-nechifor: These are the defaults, so you can delete these lines.

Also, if you rename teleop_enable to vr_teleop_enable in the Module, you don't need to change the topic here.

dimos/teleop/devices/vr_teleop_module.py
```diff
@@ -0,0 +1,216 @@
...
+    """
+
+    default_config = VRTeleopConfig
+
+    # LCM inputs from Deno bridge
+    vr_left_transform: In[LCMTransform] = None  # type: ignore[assignment]
+    vr_right_transform: In[LCMTransform] = None  # type: ignore[assignment]
+    vr_trigger_0: In[Float32] = None  # type: ignore[assignment]
+    vr_trigger_1: In[Float32] = None  # type: ignore[assignment]
+    teleop_enable: In[Bool] = None  # type: ignore[assignment]  # X button calibration toggle
```
paul-nechifor:
```suggestion
    vr_left_transform: In[LCMTransform]
    vr_right_transform: In[LCMTransform]
    vr_trigger_0: In[Float32]
    vr_trigger_1: In[Float32]
    teleop_enable: In[Bool]  # X button calibration toggle
```

`= None` isn't needed anymore

dimos/teleop/devices/base_teleop_module.py
```diff
@@ -0,0 +1,290 @@
...
```
paul-nechifor:
```suggestion
    controller_delta_0: Out[PoseStamped]
    trigger_value_0: Out[Bool]
    controller_delta_1: Out[PoseStamped]
    trigger_value_1: Out[Bool]
    controller_delta_2: Out[Twist]
    trigger_value_2: Out[Bool]
    controller_delta_3: Out[Twist]
    trigger_value_3: Out[Bool]
```

dimos/utils/teleop_transforms.py
```diff
@@ -0,0 +1,105 @@
...
+        available = OUTPUT_TYPE_INDICES.get(output_type, [])
+        for idx in available:
+            if idx not in used_indices:
+                indices.append(idx)
+                used_indices.add(idx)
+                break
+        else:
+            raise ValueError(f"No available index for output type {output_type.__name__}")
+
+    return indices
```
paul-nechifor: This docstring is quite puzzling.

My understanding now is that it maps to the first available output in `controller_delta_0, ..., _3` in BaseTeleopModule. Is this correct?

If so, it should be a method on that class as it's specific to it.

dimos/teleop/devices/vr_teleop_module.py
```diff
@@ -0,0 +1,216 @@
...
+                self.vr_right_transform,
+                lambda msg, idx=right_index: self._on_lcm_transform(idx, msg),
+            ),
+            (self.vr_trigger_0, lambda msg, idx=left_index: self._on_lcm_trigger(idx, msg)),
+            (self.vr_trigger_1, lambda msg, idx=right_index: self._on_lcm_trigger(idx, msg)),
+            (self.teleop_enable, self._on_lcm_teleop_enable),
+        ]
+        for stream, handler in subscriptions:
+            if stream and stream.transport:
+                stream.subscribe(handler)  # type: ignore[misc, arg-type]
```
paul-nechifor:
```suggestion
                self._disposables.add(Disposable(stream.subscribe(handler)))  # type: ignore[misc, arg-type]
```
You also have to import Disposable

dimos/msgs/geometry_msgs/ControllerPose.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+class ControllerPose(Pose):
+    """Controller pose with conversion utilities for teleop output types.
+
+    Extends Pose with methods to convert to PoseStamped or Twist commands.
+    """
+
+    @classmethod
+    def from_pose(cls, pose: Pose) -> ControllerPose:
```
paul-nechifor: Personally, I'm not a fan of this.

If you need ways to convert Pose to PoseStamped or Twist, you can just write standalone functions for that. There's no need to introduce a new type but with no new fields.

dimos/msgs/geometry_msgs/ControllerPose.py
```diff
@@ -0,0 +1,112 @@
...
+        euler = self.orientation.to_euler()
+        angular = Vector3(
+            clamp(euler.x * angular_scale, max_angular),
+            clamp(euler.y * angular_scale, max_angular),
+            clamp(euler.z * angular_scale, max_angular),
+        )
+
+        return Twist(linear=linear, angular=angular)
+
+    def __sub__(self, other: Pose | PoseConvertable) -> ControllerPose:  # type: ignore[override]
```
paul-nechifor: I'm not sure why you need this, but generally `# type: ignore[override]` is bad because it means you're not respecting the contract.

dimos/msgs/geometry_msgs/ControllerPose.py
leshy: Agreed, also shouldn't use Pose at all but generate PoseStamped straight from your hardware, we always want the pose with the frame_id and timestamp.

I also don't understand exactly why we are converting pose to twist? We should probably be generating twist in some intelligent way when holding a button or something, otherwise scratching your butt would make a robot go insane?

dimos/msgs/std_msgs/Float32.py
```diff
@@ -0,0 +1,58 @@
...
+# limitations under the License.
+
+"""Float32 message type."""
+
+from typing import ClassVar
+
+from dimos_lcm.std_msgs import Float32 as LCMFloat32
+
+try:
+    from std_msgs.msg import Float32 as ROSFloat32  # type: ignore[attr-defined]
```
leshy: why do we have ros stuff in this? don't just delete, pls explain, I want to understand why I keep seeing this in PRs

dimos/msgs/std_msgs/Float32.py
```diff
@@ -0,0 +1,58 @@
...
+    """ROS-compatible Float32 message."""
+
+    msg_name: ClassVar[str] = "std_msgs.Float32"
+
+    def __init__(self, data: float = 0.0) -> None:
+        """Initialize Float32 with data value."""
+        self.data = data
+
+    @classmethod
+    def from_ros_msg(cls, ros_msg: ROSFloat32) -> "Float32":
```
leshy: why did you write this conversion?

dimos/msgs/geometry_msgs/Pose.py
leshy: This might be ok if it passes tests.. but this `PoseConvertible` is a bit goofy code I wrote very very early, probably we should be strict and only __add__ and __sub__ between `Pose` and not try and do fancy auto type conversions.

idk if he can easily remove PoseConvertable and pass tests from both __add__ and __sub__ if yes, great. if not, make __sub__ only support `Pose` so that my wrong approach doesn't spread

dimos/msgs/std_msgs/Float32.py
ruthwikdasyam:
```
    @classmethod
    def from_ros_msg(cls, ros_msg: ROSBool) -> "Bool":
        """Create a Bool from a ROS std_msgs/Bool message.
```
```
    @classmethod
    def from_ros_msg(cls, ros_msg: ROSInt8) -> "Int8":
        """Create a Bool from a ROS std_msgs/Bool message.
```
Don't we want it? Int8, Bool types from std_msgs have this method. I assumed its a convention we follow

dimos/msgs/std_msgs/Float32.py
ruthwikdasyam: As mentioned below, took reference from other message type classes. If we don't need this, should we remove it from all those msg types?

dimos/utils/teleop_transforms.py
ruthwikdasyam: Yes, as we can't declare them at run time - initiating all potential streams we need, which are mapped later using this method - it can be 2xPoseStamped, 2xTwist, or one each.

mm... Implementation didn't feel too traditional to put it inside the class. 
But, yes, this is super specific to class, I will move it

dimos/teleop/devices/vr_teleop_module.py
ruthwikdasyam: Got it, updating all declarations!

dimos/msgs/std_msgs/Float32.py
spomichter: yes no ros needed

dimos/teleop/teleop_blueprints.py
ruthwikdasyam: Actually yeah. renamed and deleted mentions

dimos/msgs/geometry_msgs/ControllerPose.py
ruthwikdasyam: Okay, I can get timestamp and frame_id from the controller.

Reg Twist: I have start/stop (X button), so hoping this should be enough to control it well. Planning further to have spherical volume of say 5cm radius, around caliberated point - which would not send any twist pos (zeros), and moving controller further would start velocities.. something safe.

ruthwikdasyam: And, reg ControllerPose type...
As an alternative, I can have PoseStamped in place of ControllerPoseStamped (Now that we need to carry timestamps... It shall be ControllerPoseStamped) and use the 'to_twist' methods in utils. What do you suggest? @spomichter

dimos/msgs/geometry_msgs/Pose.py
ruthwikdasyam: Made sub only support `Pose` and it passes tests.

dimos/msgs/geometry_msgs/ControllerPose.py
ruthwikdasyam: Sorted!
Replaced `ControllerPose` with `Pose`, Added standalone functions in `utils/teleop_transform.py`.
frame_id and timestamps, added as received from hardware.

dimos/msgs/geometry_msgs/ControllerPose.py
ruthwikdasyam: without `-> ControllerPose` the method returns a `Pose`. ignore is added as this overrides parents `__sub__` with different return type.
 
 Anyways, I have removed entire ControllerPose msg type now

dimos/msgs/std_msgs/Float32.py
leshy: I understand why it seems so :) this is part of an older ros integration we are obsoleting

dimos/msgs/geometry_msgs/ControllerPose.py
leshy: so I still see ControllerPose file, this file shouldn't exist, geometry_msgs matches ros msgs, and ControllerPose isn't one

dimos/teleop/utils/teleop_transforms.py
leshy, line 1: this file seems teleop specific? pls put it in your teleop dir

dimos/teleop/utils/teleop_visualization.py
leshy, line 1: also seems like something that belongs in teleop dir

dimos/teleop/utils/teleop_transforms.py
ruthwikdasyam: Yes, they are. Maybe I can add them at `dimos/teleop/utils/`

dimos/msgs/geometry_msgs/ControllerPose.py
ruthwikdasyam: I did remove the file. I don't find it anywhere. Can you lmk where you see it

leshy: aha it's gone, was some github stale diff issue

PR #977, opened 2026-01-12
[Tiny] Add Webcam demo

jeff-hykin:
- `dimos run demo-camera`

### DIFF
--- a/dimos/agents/__init__.py
+++ b/dimos/agents/__init__.py
@@ -1,3 +1,12 @@
+from langchain_core.messages import (
+    AIMessage,
+    HumanMessage,
+    MessageLikeRepresentation,
+    SystemMessage,
+    ToolCall,
+    ToolMessage,
+)
+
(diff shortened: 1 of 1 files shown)


dimos/hardware/sensors/camera/module.py
```diff
@@ -114,4 +115,9 @@ def stop(self) -> None:
 
 camera_module = CameraModule.blueprint
 
+demo_camera = autoconnect(
```
spomichter, line 118: @paul-nechifor do we need to run autocorrect on a blueprint with one module?

jeff-hykin: Thats a good point, probably not. I'll test that tomorrow and simplify if it works without autoconnect.

paul-nechifor: No. It can just be `demo_camera = CameraModule.blueprint()`

PR #978, opened 2026-01-12
[Tiny] Add primary/main/core/std (whatever you want to call it) extras group for tutorial

jeff-hykin:
@spomichter `core` is aready used by `dimos.core` and these main/std/whatever deps are completely different than the deps that are required from `dimos.core` (which only needs `lcm` in principle), I would guess that lesh has some objections to this group being named "core"

Other than "not core" I don't care what the name is, but it probably will be locked in for a very long time

other changes: 
- the line change about rerun:
     - EITHER rerun needs to be in the base dependencies (which is what this PR does) OR (the better long term fix) the `dimos.core` python code needs to be edited to make rerun truly optional. Right now absolutely nothing can run with dimos without rerun installed (even if its never used)


pyproject.toml
```diff
@@ -64,6 +64,10 @@ dependencies = [
     # Used for calculating the occupancy map.
     "numba>=0.60.0",    # First version supporting Python 3.12
     "llvmlite>=0.42.0", # Required by numba 0.60+
+
+    # TODO: rerun shouldn't be required but rn its in core (there is NO WAY to use dimos without rerun rn)
+    # remove this once rerun is optional in core
+    "rerun-sdk>=0.20.0",
```
jeff-hykin, line 70: EITHER rerun needs to be in the base dependencies (which is what this PR does) OR (the better long term fix) the `dimos.core` python code needs to be edited to make rerun truly optional. Right now absolutely nothing can run with dimos without rerun installed (even if its never used)

I'm happy to make a PR for the other fix, but this one is faster and my actual goal is the go2 tutorial, this is just in the way

spomichter: Yeah so prob need to add lazy imports for rerun where needed. For now can probably approve this

pyproject.toml
```diff
@@ -271,6 +275,10 @@ drone = [
     "pymavlink"
 ]
 
+main = [
+    "dimos[agents,web,perception,visualization]",
```
spomichter: Redundant. You just added rerun to base dependencies so why would you also import visualization. and why is visualization extra even there

jeff-hykin: presumably rerun is going to be taken out of the required depnendcies soon (when that happens this change will be necessary). Also if someone adds something extra to visualization they might not know to add visulization here.

e.g. this line is just "if visulization has any deps, install them with 'base'"

spomichter: add sim

PR #1035, opened 2026-01-16
mujoco sim module

jca0:
mujoco simulation module

created
- mujoco backend to bring up mujoco
- mujoco manipulation interface to easily use with other robots
- mujoco driver to make it a module/determine topics to publish and subscribe to

mujoco parses XML file to determine robot and environment

command to test: `python3 -m dimos.simulation.sim_blueprints`

todos:
- fix mac mujoco viewer issue

### DIFF
--- a/dimos/robot/all_blueprints.py
+++ b/dimos/robot/all_blueprints.py
@@ -61,10 +61,10 @@ all_blueprints = {
     "unitree-go2-spatial": "dimos.robot.unitree_webrtc.unitree_go2_blueprints:unitree_go2_spatial",
     "unitree-go2-temporal-memory": "dimos.robot.unitree_webrtc.unitree_go2_blueprints:unitree_go2_temporal_memory",
     "unitree-go2-vlm-stream-test": "dimos.robot.unitree_webrtc.unitree_go2_blueprints:unitree_go2_vlm_stream_test",
-    "xarm7-trajectory-sim": "dimos.simulation.sim_blueprints:xarm7_trajectory_sim",
     "xarm-perception": "dimos.manipulation.manipulation_blueprints:xarm_perception",
     "xarm6-planner-only": "dimos.manipulation.manipulation_blueprints:xarm6_planner_only",
     "xarm7-planner-coordinator": "dimos.manipulation.manipulation_blueprints:xarm7_planner_coordinator",
+    "xarm7-trajectory-sim": "dimos.simulation.sim_blueprints:xarm7_trajectory_sim",
 }
 
 


dimos/simulation/engines/base.py
```diff
@@ -0,0 +1,97 @@
...
+    engine: str
+    asset: str | None = None
+    dof: int | None = None
+    joint_names: list[str] | None = None
+    limits: JointLimits | None = None
+    vendor: str | None = None
+    model: str | None = None
+
+
+class SimulationEngine(ABC):
```
spomichter, line 28: nice

dimos/simulation/engines/base.py
```diff
@@ -0,0 +1,97 @@
...
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from dimos.hardware.manipulators.spec import JointLimits
+
+
+@dataclass(frozen=True)
+class RobotSpec:
```
spomichter: Shouldn't this extend some existing general robot spec @mustafab0 ? Yeah pretty sure this should subclass, and shouldnt this take urdf @jca0

dimos/simulation/manipulators/sim_manip_interface.py
```diff
@@ -0,0 +1,188 @@
...
+    """Adapter wrapper around a simulation engine to provide a uniform manipulator API."""
+
+    def __init__(self, engine: SimulationEngine, spec: RobotSpec) -> None:
+        self.logger = logging.getLogger(self.__class__.__name__)
+        self._engine = engine
+        self._spec = spec
+        self._dof = int(spec.dof) if spec.dof is not None else 0
+        self._connected = False
+        self._servos_enabled = False
+        self._control_mode = ControlMode.POSITION
```
mustafab0, line 40: Why do we need ControlMode.Position?

dimos/simulation/engines/mujoco_engine.py
```diff
@@ -0,0 +1,217 @@
...
+    def read_joint_efforts(self) -> list[float]:
+        return self.joint_efforts
+
+    def write_joint_positions(self, positions: list[float]) -> None:
+        with self._lock:
+            limit = min(len(positions), self._num_joints)
+            for i in range(limit):
+                self._joint_position_targets[i] = float(positions[i])
+
+    def write_joint_velocities(self, velocities: list[float]) -> None:
```
spomichter: Needs to be changed to list of some velocity type. I assume Vector3

dimos/simulation/plugins/mujoco_xarm.py
mustafab0, line 1: This is not a scalable approach. Having to build plugin for each oem manufacturer + sim engine is not maintainable.

The robot arms should not be specific to any SimEngine

dimos/simulation/plugins/mujoco_xarm.py
```diff
@@ -0,0 +1,91 @@
...
+        upper_deg = [360, 118, 11, 180, 360]
+
+    lower_rad = [math.radians(d) for d in lower_deg[:dof]]
+    upper_rad = [math.radians(d) for d in upper_deg[:dof]]
+    max_vel_rad = math.radians(180.0)
+    return JointLimits(
+        position_lower=lower_rad,
+        position_upper=upper_rad,
+        velocity_max=[max_vel_rad] * dof,
+    )
```
mustafab0: The sim module should get a Robot config that has all its limits, assets etc. 

Robot shouldn't need a registry to be loaded, but should be defined in the blueprint

dimos/simulation/manipulators/sim_module.py
```diff
@@ -0,0 +1,278 @@
...
+            self._pending_positions = list(msg.positions)
+            self._pending_velocities = None
+
+    def _on_joint_velocity_command(self, msg: JointCommand) -> None:
+        with self._command_lock:
+            self._pending_velocities = list(msg.positions)
+            self._pending_positions = None
+
+    def _control_loop(self) -> None:
+        period = 1.0 / max(self._control_rate, 1.0)
```
spomichter: The second 1.0 here is the minimum control rate so should be an env var called MIN CONTROL RATE or something

dimos/simulation/manipulators/sim_module.py
```diff
@@ -0,0 +1,278 @@
...
+            if self._backend:
+                if positions is not None:
+                    self._backend.write_joint_positions(positions)
+                elif velocities is not None:
+                    self._backend.write_joint_velocities(velocities)
+
+            time.sleep(period)
+
+    def _monitor_loop(self) -> None:
+        period = 1.0 / max(self._monitor_rate, 1.0)
```
spomichter: Same as above here

dimos/simulation/manipulators/sim_module.py
```diff
@@ -0,0 +1,278 @@
...
+            time.sleep(period)
+
+    def _resolve_joint_names(self, dof: int) -> list[str]:
+        if self._spec and self._spec.joint_names:
+            if len(self._spec.joint_names) >= dof:
+                return list(self._spec.joint_names[:dof])
+        return [f"{self._joint_prefix}{i + 1}" for i in range(dof)]
+
+
+def get_blueprint() -> dict[str, Any]:
```
spomichter: Just return blueprint how we normally do idk if we support returning blueprint via dict like this or if converting this back to .blueprint() type even works.

dimos/simulation/manipulators/sim_module.py
```diff
@@ -0,0 +1,278 @@
...
+                    warn_code=0,
+                    cmdnum=0,
+                    mt_brake=0,
+                    mt_able=1 if self._backend.read_enabled() else 0,
+                    tcp_pose=[],
+                    tcp_offset=[],
+                    joints=[float(p) for p in positions],
+                )
+            )
+            time.sleep(period)
```
spomichter: WRONG

dimos/simulation/plugins/mujoco_xarm.py
```diff
@@ -0,0 +1,91 @@
...
+
+
```
spomichter: Delete

spomichter: bad

dimos/simulation/engines/mujoco_engine.py
jca0: doesn't make sense to. Vector3 is 3D, Twist is 6D, joints can only go in 2 directions

dimos/simulation/manipulators/sim_manip_interface.py
jca0: just a default until user changes it via `set_control_mode`

dimos/simulation/assets/xarm7/xarm7.png
mustafab0, line 1: Should also be in lfs

dimos/simulation/manipulators/sim_module.py
```diff
@@ -19,24 +19,20 @@
...
 import time
-from typing import TYPE_CHECKING, Any
+from typing import Any
 
 from reactivex.disposable import Disposable
 
 from dimos.core import In, Module, Out, rpc
 from dimos.core.module import ModuleConfig
 from dimos.msgs.sensor_msgs import JointCommand, JointState, RobotState
+from dimos.simulation.engines.mujoco_engine import MujocoEngine
```
mustafab0: sim_module should not have "Mujoco" references.

.gitattributes
```diff
@@ -16,3 +16,4 @@
 *.mov filter=lfs diff=lfs merge=lfs -text binary
 *.gif filter=lfs diff=lfs merge=lfs -text binary
 *.foxe filter=lfs diff=lfs merge=lfs -text binary
+dimos/simulation/assets/xarm7/** filter=lfs diff=lfs merge=lfs -text
```
paul-nechifor: `dimos/` is for source code. Please data files to `data/`. Fortunately, we have docs now: https://github.com/dimensionalOS/dimos/blob/main/docs/data.md

So you can use a single dir and it will be compressed and added to LFS.

dimos/simulation/utils/xml_parser.py
mustafab0, line 1: Should be in a utils folder

dimos/simulation/sim_blueprints.py
```diff
@@ -24,8 +26,7 @@
 
 xarm7_trajectory_sim = simulation(
     engine="mujoco",
-    robot="xarm7",
-    config_path=None,
+    config_path="dimos/simulation/assets/xarm7/scene.xml",
```
mustafab0: Hard coded paths are a No-No

please look at how the manipulation branch does it

dimos/e2e_tests/test_simulation_module.py
mustafab0, line 1: Please see test_control_module.py in e2e tests for reference. 

The e2e test should confirm an instance of Sim engine is running and it can initialize correctly, and the module can communicate on its respective streams.

dimos/simulation/engines/mujoco_engine.py
```diff
@@ -0,0 +1,252 @@
...
+
+    def disconnect(self) -> None:
+        logger.info(f"{self.__class__.__name__}: disconnect()")
+        with self._lock:
+            self._connected = False
+
+        self._stop_event.set()
+        if self._sim_thread and self._sim_thread.is_alive():
+            self._sim_thread.join(timeout=2.0)
+        self._sim_thread = None
```
mustafab0: These should return either a bool or an int to track if call was successful

dimos/simulation/engines/mujoco_engine.py
mustafab0: Ideally a JointState message

Both write_joint_position & write_joint_velocities can be refactored into 1 write_joint_command/etc and take in a standard JointState message that can set Pos/Vel/Effort

dimos/simulation/manipulators/sim_module.py
```diff
@@ -0,0 +1,257 @@
...
+
+            self.joint_state.publish(
+                JointState(
+                    frame_id=self.frame_id,
+                    name=names,
+                    position=positions,
+                    velocity=velocities,
+                    effort=efforts,
+                )
+            )
```
mustafab0: Should be part of the control loop, as joint states would be needed at the same control  frequency

dimos/simulation/manipulators/sim_module.py
mustafab0: time.sleep(period) ignores work time; loop runs slower than target.

dimos/simulation/sim_blueprints.py
```diff
@@ -0,0 +1,48 @@
...
+    JointState,
+    RobotState,
+)
+from dimos.msgs.trajectory_msgs import JointTrajectory
+from dimos.simulation.manipulators.sim_module import simulation
+from dimos.utils.data import get_data
+
+xarm7_trajectory_sim = simulation(
+    engine="mujoco",
+    config_path=str(get_data("xarm7") / "scene.xml"),
```
paul-nechifor: Keep `config_path` as a `Path`. In `MujocoEngine` you're converting it back to a Path anyway.

dimos/simulation/manipulators/sim_module.py
```diff
@@ -0,0 +1,257 @@
...
+        self._stop_event = threading.Event()
+        self._control_thread: threading.Thread | None = None
+        self._monitor_thread: threading.Thread | None = None
+        self._command_lock = threading.Lock()
+        self._pending_positions: list[float] | None = None
+        self._pending_velocities: list[float] | None = None
+
+    def _create_backend(self) -> SimManipInterface:
+        if not self.config.engine:
+            raise ValueError("engine is required for SimulationModule")
```
paul-nechifor: If `engine` is required, then its type should be `str`, not `str | None = None`.

There's no need to verify it here if you disallow the class from being instantiated with an invalid config.

dimos/simulation/manipulators/sim_module.py
```diff
@@ -0,0 +1,257 @@
...
+        if not self.config.config_path:
+            raise ValueError("config_path is required for SimulationModule")
+        engine_key = self.config.engine.lower()
+        if engine_key not in self._ENGINES:
+            raise ValueError(f"Unknown simulation engine: {self.config.engine}")
```
paul-nechifor: I think it's better to use the type system.

```python
EngineType = Literal["mujoco", "some_other"]
_ENGINES: dict[EngineType, SimulationEngine] = {
    "mujoco": MujocoEngine,
}

...

# No need for None check, lower(), or key existence.
engine_cls = self._ENGINES[self.config.engine]
```

dimos/simulation/manipulators/sim_module.py
```diff
@@ -0,0 +1,257 @@
...
+            ):
+                self._disposables.add(
+                    Disposable(
+                        self.joint_position_command.subscribe(self._on_joint_position_command)
+                    )
+                )
+        except Exception as exc:
+            import logging
+
+            logging.getLogger(self.__class__.__name__).warning(
```
paul-nechifor: We use `logger = setup_logger()` for logging.

But what exceptions are you expecting here? I would expect just this to work without `if` and `try`:

```
                self._disposables.add(
                    Disposable(
                        self.joint_position_command.subscribe(self._on_joint_position_command)
                    )
                )
```

If that doesn't work, it would be good to know why.

dimos/simulation/manipulators/sim_manip_interface.py
```diff
@@ -0,0 +1,192 @@
...
+        return {
+            "state": 1 if is_moving else 0,
+            "mode": mode_int,
+        }
+
+    def read_error(self) -> tuple[int, str]:
+        return self._error_code, self._error_message
+
+    def write_joint_positions(self, positions: list[float], velocity: float = 1.0) -> bool:
+        _ = velocity
```
paul-nechifor: Is this to silence the `velocity` parameter not being used? If so, it's better to rename it to `_velocity`. I assume it's required in the interface?

Oh, `velocity` isn't used at all. Just delete it?

dimos/simulation/manipulators/sim_manip_interface.py
```diff
@@ -0,0 +1,192 @@
...
+        if not self._servos_enabled:
+            return False
+        self._control_mode = ControlMode.VELOCITY
+        self._engine.write_joint_velocities(velocities[: self._dof])
+        return True
+
+    def write_stop(self) -> bool:
+        if hasattr(self._engine, "hold_current_position"):
+            self._engine.hold_current_position()  # type: ignore[attr-defined]
+            return True
```
paul-nechifor: Checking if a method exists with `hasattr` is often a sign that the code isn't ideal.

It looks like `hold_current_position` exists on MujocoEngine, but not on SimulationEngine and you're using `hasattr` to break the abstraction.

The better way to accomplish this (adding special handling just for Mujoco) without ignoring types is to check if `self._engine` is an instance of MujocoEngine .

But isn't it better to just add `hold_current_position` to SimulationEngine?

dimos/simulation/engines/mujoco_engine.py
```diff
@@ -0,0 +1,252 @@
...
+                / len(mapping.tendon_qpos_adrs)
+            )
+        if mapping.actuator_id is not None:
+            return float(self._data.actuator_length[mapping.actuator_id])
+        return 0.0
+
+    def _apply_control(self) -> None:
+        with self._lock:
+            pos_targets = list(self._joint_position_targets)
+        with self._lock:
```
paul-nechifor: Why two? Seems like you're releasing and acquiring unnecessary the second time.

dimos/simulation/engines/mujoco_engine.py
```diff
@@ -0,0 +1,252 @@
...
+        return self.joint_velocities
+
+    def read_joint_efforts(self) -> list[float]:
+        return self.joint_efforts
+
+    def write_joint_positions(self, positions: list[float]) -> None:
+        with self._lock:
+            limit = min(len(positions), self._num_joints)
+            for i in range(limit):
+                self._joint_position_targets[i] = float(positions[i])
```
paul-nechifor: I see this pattern a lot in the code where you're copying values only up to a point.

But I think this just silently ignores errors. Is there a valid reason why write_joint_positions would be sent more positions than can be stored? I think you should let the code throw.

dimos/e2e_tests/test_simulation_module.py
```diff
@@ -0,0 +1,85 @@
...
+    target: list[float],
+    tolerance: float,
+) -> bool:
+    if len(positions) < len(target):
+        return False
+    return all(abs(positions[i] - target[i]) <= tolerance for i in range(len(target)))
+
+
+@pytest.mark.skipif(bool(os.getenv("CI")), reason="LCM doesn't work in CI.")
+class TestSimulationModuleE2E:
```
paul-nechifor, line 36:
```suggestion
@pytest.mark.e2e
class TestSimulationModuleE2E:
```

spomichter (approved):

Issue #1033, opened 2026-01-16
AGIBOT Integration


leshy:
Writing a DDS transport

on bridge vs transport https://discord.com/channels/1341146487186391173/1460724184241148044/1461526788231659560

G1 First as test

then moving on to AGIBot

## DDSTransport

- pass `pubsub/test_spec.py` for your DDS pubsub implementation

- example `ROSTransport` is at https://github.com/dimensionalOS/dimos/pull/1034
it has `test_benchmark.py` that you might want to take and run on your end also. (we'll need to do this for the pubsub to be prod ready)

IDK what your encoding will be for g1, etc, for now you might want to just pickle for very-initial tests?

- write DDSTransport and use it for 2 dimos modules to talk to each other.. can you even deploy unitreeGO2 over dds? if not why not? - you'll need to find `LCMTransport` and write your own, it's a few lines of code, just a glue to your dds pubsub

- to connect 2 modules directly without blueprints can read this https://github.com/dimensionalOS/dimos/blob/main/docs/concepts/modules.md#connecting-modules or can ask Paul how to specify transports if using blueprints

- generally own DDS properly, does dds have QOS etc? what settings are optimal in which cases? read dds docs

PR #1036, opened 2026-01-16
Create DDSTransport DDSPubSubBase

Kaweees:
Related: #1033, #1058

```sh
uv run pytest -svm tool dimos/protocol/pubsub/benchmark/test_benchmark.py --override-ini="addopts=" -k "DDS"
uv run pytest -svm tool dimos/protocol/pubsub/benchmark/test_benchmark.py
```

### DIFF
--- a/dimos/protocol/pubsub/benchmark/test_benchmark.py
+++ b/dimos/protocol/pubsub/benchmark/test_benchmark.py
@@ -83,16 +83,7 @@ def pubsub_id(testcase: TestCase[Any, Any]) -> str:
 class Message(IdlStruct):
     """DDS message with binary data payload following IdlStruct format."""
 
-    payload: str
-
-    def dds_encode(self) -> bytes:
-        """Encode message to bytes for DDS transmission."""
-        return self.payload.encode("latin-1")
-
-    @classmethod
-    def dds_decode(cls, data: bytes) -> "Message":
-        """Decode bytes back to Message instance."""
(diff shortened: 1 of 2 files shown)


dimos/protocol/service/ddsservice.py
```diff
@@ -0,0 +1,101 @@
...
+    default_config = DDSConfig
+
+    def __init__(self, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self._participant_lock = threading.Lock()
+        self._started = False
+        # Support passing an existing DomainParticipant
+        self.participant: DomainParticipant | None = self.config.participant
+
+    def __getstate__(self) -> dict[str, Any]:
```
leshy: this is actually pretty weird, (not your fault I see it in LCM also) we shouldn't be pickling transports, I see Paul added this to LCM and will talk to him, you can remove __getstate__ and __setstate__ - all should work without it

Kaweees: Got it, removed

dimos/protocol/pubsub/benchmark/test_benchmark.py
```diff
@@ -0,0 +1,213 @@
+#!/usr/bin/env python3
```
paul-nechifor, line 1: This file in on `dev` now. You might want to rebase onto `dev` so that it it disappears from the diff of this PR.

PR #1038, opened 2026-01-16
Transport benchmarks + Raw ros transport

leshy:
implements transport protocol benchmark suite

as well as raw ros transport (missing part to use this for unitree go2 etc is Dimos <-> ros msg translator, coming soon)

`python -m pytest -svm tool dimos/protocol/pubsub/benchmark/test_benchmark.py`

to run in docker:
`./bin/dev python -m pytest -svm tool dimos/protocol/pubsub/benchmark/test_benchmark.py`

### DIFF
--- a/dimos/protocol/service/lcmservice.py
+++ b/dimos/protocol/service/lcmservice.py
@@ -108,7 +108,6 @@ def _set_net_value(commands_needed: list[str], sudo: str, name: str, value: int)
 
 
 TARGET_RMEM_SIZE = 67108864
-TARGET_RMEM_SIZE_MACOS = 2097152  # why not 67108864? this makes the transport worse
 TARGET_MAX_SOCKET_BUFFER_SIZE_MACOS = 8388608
 TARGET_MAX_DGRAM_SIZE_MACOS = 65535
 
@@ -134,7 +133,7 @@ def check_buffers() -> tuple[list[str], int | None]:
         current_max = _set_net_value(
             commands_needed, sudo, "kern.ipc.maxsockbuf", TARGET_MAX_SOCKET_BUFFER_SIZE_MACOS
         )
-        _set_net_value(commands_needed, sudo, "net.inet.udp.recvspace", TARGET_RMEM_SIZE_MACOS)
+        _set_net_value(commands_needed, sudo, "net.inet.udp.recvspace", TARGET_RMEM_SIZE)
         _set_net_value(commands_needed, sudo, "net.inet.udp.maxdgram", TARGET_MAX_DGRAM_SIZE_MACOS)
     else:
         # For other systems, skip buffer configuration
--- a/dimos/protocol/service/test_lcmservice.py
+++ b/dimos/protocol/service/test_lcmservice.py
@@ -22,7 +22,6 @@ from dimos.protocol.service.lcmservice import (
     TARGET_MAX_DGRAM_SIZE_MACOS,
     TARGET_MAX_SOCKET_BUFFER_SIZE_MACOS,
     TARGET_RMEM_SIZE,
-    TARGET_RMEM_SIZE_MACOS,
     autoconf,
     check_buffers,
     check_multicast,
@@ -390,7 +389,7 @@ def test_check_buffers_macos_needs_config() -> None:
             sudo = get_sudo_prefix()
             expected = [
                 f"{sudo}sysctl -w kern.ipc.maxsockbuf={TARGET_MAX_SOCKET_BUFFER_SIZE_MACOS}",
-                f"{sudo}sysctl -w net.inet.udp.recvspace={TARGET_RMEM_SIZE_MACOS}",
+                f"{sudo}sysctl -w net.inet.udp.recvspace={TARGET_RMEM_SIZE}",
                 f"{sudo}sysctl -w net.inet.udp.maxdgram={TARGET_MAX_DGRAM_SIZE_MACOS}",
             ]
             assert commands == expected


dimos/protocol/pubsub/benchmark/type.py
```diff
@@ -0,0 +1,277 @@
...
+        return f"{size_bytes / 1048576:.1f} MB"
+    if size_bytes >= 1024:
+        return f"{size_bytes / 1024:.1f} KB"
+    return f"{size_bytes} B"
+
+
+def _format_throughput(bytes_per_sec: float) -> str:
+    """Format throughput to human-readable string."""
+    if bytes_per_sec >= 1e9:
+        return f"{bytes_per_sec / 1e9:.2f} GB/s"
```
paul-nechifor, line 67: While this is the correct SI definition for GB, many still use GB to mean 1024\*\*3. It could be less confusing to use 1024\*\*3 while labeling it GiB. This way we don't confuse measurements.

dimos/protocol/pubsub/benchmark/test_benchmark.py
```diff
@@ -0,0 +1,175 @@
...
+@pytest.mark.parametrize("pubsub_context, msggen", testdata, ids=[pubsub_id(t) for t in testdata])
+def test_throughput(
+    pubsub_context: PubSubContext[Any, Any],
+    msggen: MsgGen[Any, Any],
+    msg_size: int,
+    benchmark_results: BenchmarkResults,
+) -> None:
+    """Measure throughput for publishing and receiving messages over a fixed duration."""
+    with pubsub_context() as pubsub:
+        topic, msg = msggen(msg_size)
```
paul-nechifor, line 98: I'm not sure if this is a realistic way to benchmark. By generating the message at the start of the loop, it's only measuring how long it takes to transport the message. That's fair, if we're only interested in measuring that aspect of message passing.

 But real messages also take time to encode, and encodings can be cached giving unrealistic real-world expectations.

I've tried moving `msggen` to the publish loop, and the results are very different.

Also, I think we should be sending unique messages to make sure no caching is happening.

dimos/protocol/pubsub/rospubsub.py
```diff
@@ -0,0 +1,269 @@
...
+        self._subscriptions.clear()
+        self._spin_thread = None
+
+    def _spin(self) -> None:
+        """Background thread for spinning the ROS executor."""
+        while self._running:
+            executor = self._executor
+            if executor is None:
+                break
+            executor.spin_once(timeout_sec=0)  # Non-blocking for max throughput
```
paul-nechifor, line 159: I've tried with something more sensible like 0.01, but it performs worse (in latency). Still, I don't think we should use 0, because that causes a busy loop wasting CPU even when no messages are being sent, right?

dimos/protocol/pubsub/benchmark/test_benchmark.py
leshy: Actual encoding is happening at the transport layer and that's what we want to measure. I'm generating a source `Image` once (this generation I wanted outside of the loop itself) But an image is passed into the Transport layer, the transport specific encoding is executed (so lcm_encode, pickle, whatever ros does etc)

This is visible in the next PR https://github.com/dimensionalOS/dimos/pull/1041 in which I implement bytes only transports (to benchmark encoding costs, for example pickle vs lcm byte encode) and get much higher performance out of raw bytes transports

dimos/protocol/pubsub/rospubsub.py
leshy: Yes, I'm not sure either, I will further review and optimize ros transport once it's ready to use, once message encoding is implemented (once we can pass dimos.msgs.* into it) here I was trying to give maximum possible edge to ROS since it was performing so badly, and it's used only in this test and nowhere else

dimos/protocol/pubsub/benchmark/test_benchmark.py
```diff
@@ -0,0 +1,175 @@
...
+def pubsub_id(testcase: TestCase[Any, Any]) -> str:
+    """Extract pubsub implementation name from context manager function name."""
+    name: str = testcase.pubsub_context.__name__
+    # Convert e.g. "lcm_pubsub_channel" -> "LCM", "memory_pubsub_channel" -> "Memory"
+    prefix = name.replace("_pubsub_channel", "").replace("_", " ")
+    return prefix.upper() if len(prefix) <= 3 else prefix.title().replace(" ", "")
+
+
+@pytest.fixture(scope="module")
+def benchmark_results() -> Generator[BenchmarkResults, None, None]:
```
spomichter, line 77: THis doesn't do anything any reason it needs to be a test

leshy: yeah we are using pytest as an easy way to specify new transports to test (via `testdata.py`) and a runner (runs a single test per transport/packet size combo)

and you can easily select which protocols to test via

`python -m pytest -svk "lcm or ros" -m tool dimos/protocol/pubsub/benchmark/test_benchmark.py`

this benchmark_results draws a graph for you for transports you chose (like images I've been attaching)

dimos/protocol/pubsub/rospubsub.py
```diff
@@ -0,0 +1,269 @@
...
+        Args:
+            msg: A ROS message (e.g., geometry_msgs.msg.Vector3)
+
+        Returns:
+            The corresponding LCM message (e.g., dimos_lcm.geometry_msgs.Vector3)
+        """
+        raise NotImplementedError("Decode method not implemented")
+
+
+class DimosROS(
```
spomichter, line 260: CHeck if PubSubMixIn needs to be before the RawROS otherwise `DimosROS().publish()` wont encode()

leshy: haha true, inactive atm but switched

PR #1041, opened 2026-01-16
SHM Transport basic fixes

leshy:
Easy wins for non-lame SHM depends on https://github.com/dimensionalOS/dimos/pull/1038

# Before:

# After:

(autogen py LCM encoding is pretty slow, but seems still much better then pickle)

# Raw bytes

Raw bytes are nuts and close to RAM limits (comparison with raw SHM vs UDP bytes) can use in very specific cases


bin/hooks/filter_commit_message.py
```diff
@@ -28,10 +28,16 @@ def main() -> int:
 
     lines = commit_msg_file.read_text().splitlines(keepends=True)
 
-    # Find the first line containing "Generated with" and truncate there
+    # Patterns that trigger truncation (everything from this line onwards is removed)
+    truncate_patterns = [
+        "Generated with",
+        "Co-Authored-By",
```
paul-nechifor, line 34: But what if the co-author is a human? :)

leshy: I take the all the credit
138275ff55bc91702be9d885cb951f93-1908044775

spomichter: Lolll yeah we should prob remove this because sometimes ill manually add co authored by to work we're pulling from as inspiration

leshy: just don't add dashes and exactly the same format? is that ok?

leshy: added this because claude started adding Co-Authored-By and passing "Generated With" filter :sweat:

dimos/protocol/pubsub/shmpubsub.py
```diff
@@ -178,15 +182,17 @@ def publish(self, topic: str, message: bytes) -> None:
 
         # Build host frame [len:4] + [uuid:16] + payload and publish
         # We embed the message UUID in the frame for echo suppression
-        host = np.zeros(st.shape, dtype=st.dtype)
+        # Reuse pre-allocated buffer to avoid allocation overhead
+        host = st.publish_buffer
```
paul-nechifor: Calls to publish are coming from different threads so I don't think you can reuse the same buffer.

paul-nechifor: Initially I was thinking this can be solved by using a thread local cache: https://docs.python.org/3/library/threading.html#thread-local-data

But it still needs to be invalidated when the shape changes. To do that reliably requires locks, defeating the purpose. Maybe allocate buffers which are large enough for any message? But if we do that, and we're sending images, that requires image size buffers per each thread, wasting space...

leshy: I think maybe transport level lock? I don't really mind this for SHM, it makes sense to lock? idk can think later

paul-nechifor (dismissed):

paul-nechifor (approved):

PR #1042, opened 2026-01-16
Person follow skill with EdgeTAM

paul-nechifor:
## Usage

```bash
uv sync --all-extras
uv run dimos --robot-ip 192.168.X.XXX run unitree-go2-agentic
uv run humancli
```

Write "follow the man in blue" or some description of yourself.

## E2E

Run the end-to-end test:

```bash
uv run pytest dimos/e2e_tests/test_person_follow.py
```

## How it works

* Uses EdgeTAM for detection.
* Estimates the person's distance based on the width of the detection. (I'm using width instead of height since people's height isn't always visible from the low angle of the robot.)
* Rotates or moves forward/back to keep the detection in the center.
* Does not check for free space. This is especially bad when you make it move backwards.
* Cannot follow you around tight corners, either because it loses track,or because it tries to move through the corner.


pyproject.toml
```diff
@@ -101,7 +101,10 @@ misc = [
     # Perception Dependencies
     "scikit-learn",
     "timm>=1.0.15",
+    "edgetam @ git+https://github.com/facebookresearch/EdgeTAM.git",
```
paul-nechifor: I need to replace this with a non-repo.

dimos/agents/skills/person_follow.py
```diff
@@ -0,0 +1,340 @@
...
+    def stop_following(self) -> str:
+        """Stop following the current person.
+
+        Returns:
+            Confirmation message.
+        """
+        self._stop_following()
+
+        try:
+            move_rpc = self.get_rpc_calls("GO2Connection.move")
```
leshy: any reason we aren't just emitting twist from this as OUT? since this would makae this work cross embodyment

dimos/models/segmentation/edge_tam.py
```diff
@@ -0,0 +1,262 @@
...
+logger = setup_logger()
+
+
+class SAM2InferenceState(TypedDict):
+    images: list[torch.Tensor | None]
+    num_frames: int
+    cached_features: dict[int, Any]
+
+
+class EdgeTAMProcessor(Detector):
```
leshy, line 52: isn't this some new model, for tracking only? so would be good to have maybe new dir for it, since it's not segmentation? and new type for trackers, since this is not strictly a `Detector` so we can change trackers? not sure, what do you think?

dimos/agents/skills/person_follow.py
```diff
@@ -0,0 +1,340 @@
...
+    _max_lost_frames: int = 15  # number of frames to wait before declaring person lost
+
+    # Distance is estimated by assuming the width of a person. We don't use height because
+    # when the robot is close to the person, it often cannot see his full height.
+    _assumed_person_width: float = 0.45  # meters (average shoulder width)
+
+    def __init__(self, camera_info: CameraInfo, global_config: GlobalConfig) -> None:
+        super().__init__()
+        self._global_config: GlobalConfig = global_config
+        self._latest_image: Image | None = None
```
leshy, line 69: we have this fancy system of hot/colr getters

https://github.com/dimensionalOS/dimos/blob/dev/docs/api/sensor_streams/advanced_streams.md#getter_hot---background-subscription-instant-reads

so you can do `getter = geter_hot(self.image().observable())`

idk if you want this, just letting you know it exists. also getter is not disposable, should make it one

dimos/agents/skills/person_follow.py
```diff
@@ -0,0 +1,340 @@
...
+        with self._lock:
+            if self._tracker is not None:
+                self._tracker.stop()
+                self._tracker = None
+
+        self._vl_model.stop()
+        super().stop()
+
+    @skill()
+    def follow_person(self, query: str) -> str:
```
leshy, line 106: generally I'd much prefer this as a reactive pipeline in which you

```python
self.image_frame.observable().pipe(
  ops.map(track)
  ops.map(detection_to_twist)
).subscribe(self.twist.publish)
```

so you don't need to do this self._image thing, thread management etc etc. much cleaner, just letting you know, this is an ok first test

leshy (dismissed):
looks ok as a first sketch, would like to understand global planner issues etc.
will need to plug in reid also, Alex found some good model

dimos/agents/skills/person_follow.py
```diff
@@ -0,0 +1,340 @@
...
+        bbox_width = bbox[2] - bbox[0]  # x2 - x1
+
+        if bbox_width <= 0:
+            return None
+
+        # Pinhole camera model: distance = (real_width * fx) / pixel_width
+        fx = self._camera_info.K[0]  # focal length x in pixels
+        estimated_distance = (self._assumed_person_width * fx) / bbox_width
+
+        return estimated_distance
```
leshy: again first sketch so all good, we should make sure that API for merging this with lidar and estimating segmentation distance (turning into detection3d) is super easy easy. detections/ stuff implements this for projecting 2d into 3d

`perception/detection/reid/module.py`

this is for re-assembling fancy Detection2D types from ros msgs + image, but you'd do the same for aligning lidar + camera image

```py

    def detections_stream(self) -> Observable[ImageDetections2D]:
        return backpressure(
            align_timestamped(
                self.image.pure_observable(),
                self.detections.pure_observable().pipe(
                    ops.filter(lambda d: d.detections_length > 0)  # type: ignore[attr-defined]
                ),
                match_tolerance=0.0,
                buffer_size=2.0,
            ).pipe(ops.map(lambda pair: ImageDetections2D.from_ros_detection2d_array(*pair)))  # type: ignore[misc, arg-type]
        )

```

documented at https://github.com/dimensionalOS/dimos/blob/dev/docs/api/sensor_streams/temporal_alignment.md

also this might be interesting
https://github.com/dimensionalOS/dimos/blob/dev/docs/api/sensor_streams/quality_filter.md

dimos/models/segmentation/edge_tam.py
alexlin2: Actually the way I implemented before was I conformed it to the same structure as YOLO and YOLOE, also in detection/types/detection3d/seg.py, there is a from_sam2_result() that is for this edgeTAM output. It should output ImageDetection2D[Detection2DSeg). I think it's better to put the general EdgeTAMTracker code (conforms to the same detector type) under perception/, keep the inference and model loading code as separate in models/ like I did before.

dimos/agents/skills/person_follow.py
paul-nechifor: It slipped my mind. :sweat_smile: Fixed now.

dimos/agents/skills/person_follow.py
```diff
@@ -0,0 +1,340 @@
...
+
+        if estimated_distance is None:
+            return Twist.zero()
+
+        # Calculate distance error (positive = too far, need to move forward)
+        distance_error = estimated_distance - self._target_distance
+
+        # Compute angular velocity (turn towards person)
+        # Negative because positive angular.z is counter-clockwise (left turn)
+        angular_z = -x_norm * 1.0  # Proportional gain (reduced from 2.0 to avoid oscillation)
```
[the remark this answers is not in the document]
paul-nechifor: Extracted

dimos/agents/skills/person_follow.py
paul-nechifor: But doing it that way ties the frequency of the cmd_vel to the frequency of the camera, right? And you still need some state like counting how many frames are without a detection, or knowing when to stop.

dimos/agents/skills/person_follow.py
paul-nechifor: Yeah, makes sense. I'll use the lidar to estimate the distance.

dimos/agents/skills/person_follow.py
leshy: there are nice tools for all of this - like `sample()` maybe? samples latest value at fixed hz, so you can emit twist at a faster fixed rate then input image

https://github.com/dimensionalOS/dimos/blob/dev/docs/api/sensor_streams/reactivex.md#sampleinterval---emit-latest-value-every-n-seconds

> you still need some state like counting how many frames are without a detection, or knowing when to stop.

yeah you either start propagading more complex types (worst case but still nice)
```py
{
  Image :image
  Detection2d: ...
  frames: n
}
```
and this gives you full power, and clear typing also, since you can transform between those complex types at each step

`def bla(in: ComplexType1) -> ComplexType2`

`ops.map(bla)` 

but as an also easier shortcut before going for complex types you can have a map that has a sideeffect that counts empty Detection2Ds and starts returning false/exiting early.

`detection_to_twist` in above example can have a sideeffect of counting empty detections it got, and starting to output twist(0,0,0) - after initial few twist(0,0,0) you can also choose to terminate tracking loop as well, observable just finishes


to be clear Im not asking for a rewrite, just mentioning how this can be structured in a way in which it's easy and fun to quickly iterate (add twist easing etc), rotation towards a person and walking towards a person can be separate maps() that layer onto the same Twist, etc

dimos/agents/skills/person_follow.py
leshy: no need for first merge as well, your call how to proceed with this, it could turn out that lidar method is less robust also

leshy (approved):

PR #1048, opened 2026-01-16
Fix split view on wide monitors

jeff-hykin:
test with: `dimos --viewer-backend rerun-web --replay unitree-go2`
- the printout was the wrong address (fixed that)
- creates a split view but on large monitors the command center is like 1inch wide / unusable (fixed that)

### DIFF
--- a/dimos/web/templates/rerun_dashboard.html
+++ b/dimos/web/templates/rerun_dashboard.html
@@ -8,6 +8,12 @@
         body { background: #0a0a0f; font-family: -apple-system, system-ui, sans-serif; }
         :root { --command-center-width: max(30vw, 35rem); }
         .container { display: flex; width: 100%; height: 100%; }
+        .command-center {
+            width: var(--command-center-width);
+            min-width: 16rem;
+            border: none;
+            border-right: 1px solid #333;
+        }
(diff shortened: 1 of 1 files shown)


dimos/web/templates/rerun_dashboard.html
```diff
@@ -6,15 +6,78 @@
...
+        .divider.dragging { background: #3a4458; }
+        body.dragging { user-select: none; cursor: col-resize; }
     </style>
 </head>
 <body>
     <div class="container">
-        <iframe class="rerun" src="http://localhost:9090/?url=rerun%2Bhttp%3A%2F%2Flocalhost%3A9876%2Fproxy"></iframe>
         <iframe class="command-center" src="http://localhost:7779/command-center"></iframe>
+        <div class="divider" role="separator" aria-label="Resize panels"></div>
+        <iframe class="rerun" src="http://localhost:9090/?url=rerun%2Bhttp%3A%2F%2Flocalhost%3A9876%2Fproxy"></iframe>
```
jeff-hykin, line 34: To be clear, after launch this whole file needs to be nuked. The ports are hardcoded, it's not integrated into the react build system, etc. This PR is just a minimal change to fix the UX

PR #1065, opened 2026-01-20
Fix System Config Values for LCM on MacOS and Refactor

jeff-hykin:
Unit test with `cd dimos/protocol/service;pytest`, integration test with `dimos --replay run unitree-go2`

I needed to fix something in lcmservice, and Ivan had mentioned that the system checks in lcmservice were messy. I agreed so I refactored it

REFACTOR CHANGES:
- system logic removed from lcmservce, put into `system_configurator.py`
- LCM decides which system config values (Multicast, Socket Buffer Size), but that is it
- Checking logic simplified

NEW STUFF (beyond refactor)
1. Ask the user for permission instead of just attempting to change a bunch of system values (only asks once and lists everything that would be changed)
2. The missing `ulimit -n` check for MacOS (default max number of open files is 256 on MacOS) was added.
3. Turns out the max buffer size on macos is in fact 8Mb (8_388_608). Me seeing that `67_108_864` (64Mb) "worked" was just the setter command failing with a warning and not being very loud (while the rest of dimensional still ran). It seems that 8_388_608 is a hardcoded limit in the MacOS kernel. The logic now is split where MacOS is using the 8_388_608 value, and Linux uses 67_108_864

NOTE: 
- `test_high_volume_messages` seems flakey on CI @leshy (on dev or this PR). Passes every time for me on a dimensional laptop. On my mac its not even close to passing (450's out of 1000)

### DIFF
--- a/dimos/protocol/service/test_system_configurator.py
+++ b/dimos/protocol/service/test_system_configurator.py
@@ -115,7 +115,7 @@ class TestWriteSysctlInt:
                     ["sudo", "sysctl", "-w", "net.core.rmem_max=67108864"],
                     check=True,
                     text=True,
-                    capture_output=True,
+                    capture_output=False,
                 )
 
 


dimos/protocol/service/system_configurator.py
```diff
@@ -0,0 +1,438 @@
...
+    TARGET_DGRAM_SIZE = MAX_POSSIBLE_DGRAM_SIZE
+
+    def __init__(self):
+        self.needs: list[tuple[str, int]] = []
+
+    def check(self) -> bool:
+        self.needs.clear()
+        for key, target in [
+            ("kern.ipc.maxsockbuf", self.TARGET_BUFFER_SIZE),
+            ("net.inet.udp.recvspace", self.MAX_POSSIBLE_RECVSPACE),
```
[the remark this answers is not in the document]
jeff-hykin: good catch

dimos/protocol/service/lcmservice.py
```diff
@@ -240,9 +72,8 @@ class LCMConfig:
     lcm: lcm.LCM | None = None
 
     def __post_init__(self) -> None:
-        if self.url is None and platform.system() == "Darwin":
-            # On macOS, use multicast with TTL=0 to keep traffic local
-            self.url = _DEFAULT_LCM_URL_MACOS
+        if self.url is None:
+            self.url = _DEFAULT_LCM_URL
```
jeff-hykin, line 79: this confirmed to be the default LCM for all OS's. I wanted to get it from lcm directly but it seems they don't expose it

dimos/protocol/service/system_configurator.py
```diff
@@ -0,0 +1,437 @@
...
+    @abstractmethod
+    def fix(self) -> None:
+        """Apply fixes (may attempt sudo, catch, and apply fallback measures if needed)."""
+        raise NotImplementedError
+
+
+# ----------------------------- generic enforcement of system configs -----------------------------
+
+
+def configure_system(checks: list[SystemConfigurator], check_only: bool = False) -> None:
```
jeff-hykin, line 94: this is generic enough that it maybe shouldn't live under protocol/service.

leshy: yes, can move later when other systems need it

PR #1080, opened 2026-01-21
Unified TimeSeriesStore with pluggable backends, global rewrite of timed event storage

leshy:
## Summary

This is a first pass on memory, cleans up the way we deal with timed events in dimos
Creates a `TimeSeriesStore[T]` abstraction for timestamped data with multiple backend implementations. 

Backend implementations are very simple and TBH I don't expect to use psql but sqlite, implemented them more as examples.

### `TimeSeriesStore[T]` Base Class (`dimos/memory/timeseries/base.py`)
- Generic abstract base — backends implement `_save`, `_load`, `_delete`, `_iter_items`, `_find_closest_timestamp`, `_count`, `_last_timestamp`, `_find_before`, `_find_after`

- API built on top: `save()`, `find_closest()`, `find_before()`, `find_after()`, `prune_old()`, `slice_by_time()`, `iterate()`, `iterate_realtime()`, `stream()`, `consume_stream()`

- `T` bound to `Timestamped` — timestamps come from `.ts` attribute

### Backend Implementations

| Backend | File | Use Case |
|---------|------|----------|
| `InMemoryStore` | `inmemory.py` | Live streaming, buffers, testing. O(log n) via SortedKeyList |
| `PickleDirStore` | `pickledir.py` | Simple file-based, one pickle per timestamp |
| `SqliteStore` | `sqlite.py` | Single-file DB with indexed queries |
| `PostgresStore` | `postgres.py` | Multi-user, implements `Resource` lifecycle |
| `LegacyPickleStore` | `legacy.py` | Backward compat with old `TimedSensorReplay` recordings |

### Replaced `TimestampedCollection`

`TimestampedCollection` and its subclass `TimestampedBufferCollection` were used for:
- Transform buffers in TF service (`TBuffer`)
- Message timestamp alignment (`align_timestamped`)
- Quality filtering (image sharpness windows)

All now use `InMemoryStore` directly. `TimestampedBufferCollection` is a thin wrapper adding auto-prune on insert. `TimestampedCollection` class deleted entirely.

### Replaced Pickle Sensor Replay System with standard time series store interface
future recording and reply can go via sqlite for example

### Replaced Transform service in-memory transform storage
they all depend on the same base, storing transforms in postgres (not that we want to) is a one line change

### Other changes
- Removed generic `Embedding` type — all embedding models return the same type

### Usage

```python
from dimos.memory.timeseries import InMemoryStore, SqliteStore, PostgresStore

# In-memory buffer
store = InMemoryStore()
# or
# SQLite persistent storage
# store = SqliteStore("recordings/lidar")
# or
# Postgres storage
# store = PostgresStore("my_lidar_table")

store.save(data)  # data.ts used as timestamp
latest = store.last()
closest = store.find_closest(target_ts, tolerance=0.5)

for item in store.iterate_realtime(speed=2.0):
    process(item)

# Observable streaming
store.stream(speed=1.0, seek=10.0).subscribe(on_next=process)
```


leshy:
> Personally, I don't like the API. It's similar to what we do with `LCM()` and others where we don't encode the lifecycle into the API. The API is simple, but doesn't give you enough control.

greptile is saying 5/5 paul is saying 0/5, this is a classic good cop bad cop situation

sounds good! will externalize the sessions

paul-nechifor (dismissed):

paul-nechifor (dismissed):

paul-nechifor (approved):

PR #1099, opened 2026-01-23
fix: default to rerun native viewer

Nabla7:
as the title implies

### DIFF
--- a/dimos/dashboard/rerun_init.py
+++ b/dimos/dashboard/rerun_init.py
@@ -20,8 +20,8 @@ Architecture:
     - All processes share the same Rerun recording stream
 
 Viewer modes (set via VIEWER_BACKEND config or environment variable):
-    - "rerun-web" (default): Web viewer on port 9090
-    - "rerun-native": Native Rerun viewer (requires display)
+    - "rerun-web": Web viewer on port 9090
+    - "rerun-native" (default): Native Rerun viewer (requires display)
     - "foxglove": Use Foxglove instead of Rerun
 
 Usage:


dimos/dashboard/rerun_init.py
```diff
@@ -64,7 +64,7 @@
 _rerun_init_lock = threading.Lock()
 
 
-def init_rerun_server(viewer_mode: str = "rerun-web", memory_limit: str = "4GB") -> str:
+def init_rerun_server(viewer_mode: str = "rerun-native", memory_limit: str = "4GB") -> str:
```
paul-nechifor:
```suggestion
def init_rerun_server(viewer_mode: str, memory_limit: str = "4GB") -> str:
```

I don't think this should have a default. It should just receive it from global_config.viewer_backend

dimos/dashboard/rerun_init.py
```diff
@@ -64,7 +64,7 @@
 _rerun_init_lock = threading.Lock()
 
 
-def init_rerun_server(viewer_mode: str = "rerun-web", memory_limit: str = "4GB") -> str:
+def init_rerun_server(viewer_mode: str, memory_limit: str = "4GB") -> str:
     """Initialize Rerun server in the main process.
```
spomichter, line 68: Comment is incorrect below. Still refernces viewer mode

spomichter: Greptile comment: The change is functionally correct and aligns with the documentation that describes rerun-native as offering better performance. However, the docstring comments in rerun_init.py still incorrectly reference rerun-web as the default in two places (lines 23 and 36).

PR #1100, opened 2026-01-23
autogenerate list of blueprints

paul-nechifor:
* Automatically updates `dimos/robot/all_blueprints.py` when you run `pytest dimos/robot/test_all_blueprints_generation.py`.
* Locally, it writes the file and you have to `git add` to confirm. The test fails until you commit the changes.
* In CI, it always fails if there are any differences.

### DIFF
--- a/dimos/robot/test_all_blueprints_generation.py
+++ b/dimos/robot/test_all_blueprints_generation.py
@@ -1,0 +1,214 @@
+# Copyright 2025-2026 Dimensional Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import ast
+from collections.abc import Generator
+import difflib
+import os
+from pathlib import Path
+import subprocess
+
+import pytest
+
+from dimos.constants import DIMOS_PROJECT_ROOT
+
+IGNORED_FILES: set[str] = {
+    "dimos/robot/all_blueprints.py",
+    "dimos/robot/get_all_blueprints.py",
+    "dimos/robot/test_all_blueprints.py",
+    "dimos/robot/test_all_blueprints_generation.py",
+    "dimos/core/blueprints.py",
+    "dimos/core/test_blueprints.py",
+}
+BLUEPRINT_METHODS = {"transports", "global_config", "remappings", "requirements"}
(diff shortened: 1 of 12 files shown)


dimos/robot/test_all_blueprints_generation.py
```diff
@@ -0,0 +1,214 @@
...
+    return ".".join(parts)
+
+
+def _find_blueprints_in_file(file_path: Path) -> tuple[list[str], list[str]]:
+    blueprint_vars: list[str] = []
+    module_vars: list[str] = []
+
+    try:
+        source = file_path.read_text(encoding="utf-8")
+        tree = ast.parse(source, filename=str(file_path))
```
jeff-hykin, line 163: To me parsing and traversing the AST feels too far (e.g. over-engineered). On the receiving end "my blueprint isn't being included, but idk why?" would be some painful debugging of this black-magic. I think a KISS approach would be like pytest's naming scheme: if the file ends in "_blueprint.py" and, when imported, has a `.blueprints` list, use that. No AST autoconnect-detecting node-traversal.

jeff-hykin: and instead of committing auto-generated code, and adding build step thats run on-commit, the simple "find _bluprints.py" could be done at runtime.

paul-nechifor: > I think a KISS approach would be like pytest's naming scheme: if the file ends in "_blueprint.py"

Yeah, that sounds great for locating the files. :+1: 

> when imported, has a .blueprints list, use that. No AST autoconnect-detecting node-traversal.

The issue with doing it that way is that you have to import it, which imports all the modules. This would make test slower. But it could be acceptable because we probably have already imported everything when running the test suite. The only other issue would be that all blueprints would have to be importable in all situations, even when dependencies are missing...

> and instead of committing auto-generated code, and adding build step thats run on-commit, the simple "find _bluprints.py" could be done at runtime.

Disagree with this because this then forces all modules to be imported on every run. The list of blueprints is intentionally listed as strings to be imported instead of actual imports because it's better to only import the blueprint you want to use.

jeff-hykin: > But it could be acceptable because we probably have already imported everything when running the test suite

We basically import everything just running `dimos --help` lol. Not that that is a good thing 

> Disagree with this because this then forces all modules to be imported on every run.

Ah that is a good point. I think I'll have a clean solution (prototype of lazy imports is tested and works) that avoid this but that probably won't be done for another week. 
So in the meantime, AST it is I suppose.

paul-nechifor: > We basically import everything just running dimos --help lol.

We import a lot when running `--help`. But there's so much more! :)

> when imported, has a .blueprints list, use that. No AST autoconnect-detecting node-traversal.

Tried it, but it doesn't work for the RealSenseCamera module because it imports `pyrealsense2` which is a system installed package. :(

leshy (approved):
clear improvement imo, without viable alternative proposals? can collect wishlist for this long term

PR #1108, opened 2026-01-25
Temporal Animation + Renaming Entities/Other Features

ClaireBookworm:
- added support for renaming entities, enabling a video example for showing graph 
- showing graph_viz_server, to show a way to see the graphed relationships between entiites


leshy (commented):
In experimental/ so happy to let through with minimal understanding, is that the plan?

jeff-hykin:
closeing cause stale

PR #1109, opened 2026-01-25
add observe skill to go2

paul-nechifor:
Run:

```bash
uv run dimos run unitree-go2-agentic
```

...and ask "what do you see" in 'humancli'`.

### DIFF
--- a/dimos/robot/unitree/connection/go2.py
+++ b/dimos/robot/unitree/connection/go2.py
@@ -25,6 +25,7 @@ import rerun.blueprint as rrb
 from dimos import spec
 from dimos.core import DimosCluster, In, LCMTransport, Module, Out, pSHMTransport, rpc
 from dimos.core.global_config import GlobalConfig
+from dimos.core.skill_module import SkillModule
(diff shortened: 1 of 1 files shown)


dimos/robot/unitree/connection/go2.py
```diff
@@ -212,6 +216,7 @@ def start(self) -> None:
 
         def onimage(image: Image) -> None:
             self.color_image.publish(image)
+            self._latest_video_frame = image
```
leshy, line 219: as long as we are familiar with this but choosing to do _latest_video_frame thing I'm ok. idk if this API is too obscuring

https://github.com/dimensionalOS/dimos/blob/main/docs/api/sensor_streams/advanced_streams.md#getting-values-synchronously

leshy (approved):
just an API note

dimos/robot/unitree/connection/go2.py
paul-nechifor: I did consider that (`getter_hot`), (since I reviewed the doc), but it seemed easier to just set the value (like the other skill does) since there already is a subscription for the image here.

PR #1114, opened 2026-01-26
Pubsub pattern subs

leshy:
## LCM can now do

```python
Topic(Glob("/sensor/*")),
Topic(Glob("/**/arm"), PoseStamped),
Topic(Glob("/**")),
Topic(re.compile(r"/sensor/.*")),
Topic(re.compile(r".*/arm"), PoseStamped),
Topic(re.compile(r".*/arm")),
Topic(re.compile(r".*/arm#geometry.*"))
```

And (bigger part) we have a spec/type structure now for any pubsub to support this. This is required for universal bridging

We need to write a rerun bridge asap

## Intro
Point of this is to construct a generic bridging language, ability to define mappers from one protocol topic/msg to another. 

This is easy - if you have a way to subscribe to all messages, this is what this PR lays the groundwork for

We need bridges mostly for spying/visualization - a rerun bridge atm

We want to do standard (and usually heaviest) `OUT` visualization related computation to be done outside of the modules themselves. Module OUTs can be tapped, then modules can `rr.log` individually only for specific specific fancy vis (projected pointcloud, planner algo rendering)

Rerun could be implemented as a funny (pub only) pubsub that can receive msgs that are automatically .to_rerun()-ed
.. or something like that

## reorg

- Move implementations to impl/ subdirectory (lcmpubsub, memory, shmpubsub, etc.)
- Extract encoder mixins to encoders.py (PubSubEncoderMixin, PickleEncoderMixin, LCMEncoderMixin, JpegEncoderMixin) fix typing
- Add grid tests for pattern matching subs `test_pattern_sub.py`
- Add docs on how to grid test in general (useful for agents) `docs/development/grid_testing.md` 

## AllPubSub and DiscoveryPubSub 
There are two ways in which a protocol can facilitate topic discovery, we implement base classes for both,

`AllPubsub` supports subscribing to all topics (Redis, LCM, MQTT)
`DiscoveryPubSub` supports discovering new topics (ROS)

These capabilities are orthogonal but they can implement one another and normalize the API

both implement `subscribe_all` function we care about.

## Extending LCM 
Topic class to support Regex or Glob patterns (as per the underlying proto)

## TODO
other protocols need similar extensions (I will only do SHM),

we just want to get to something like subscribe_all (in potentially different ways) and once there we can bridge

once subscribe_all is reached, potentially post-reception (expensive) glob/regex implementations can also be implemented implicitly

## Bridge
added preliminary bridge, mostly as a type spec to demonstrate the idea, shows how we could bridge different protocols once subscribe_all or glob/regex topics are implemented

## Tests issue
Some tests are failing because some other test, lcm msgs are flying around while these tests are running. I'll investigate this, didn't want to sweep under rug, likely this is causing flakyness with other LCM tests also


dimos/protocol/pubsub/impl/lcmpubsub.py
```diff
@@ -0,0 +1,221 @@
...
+        return f"{self.pattern}#{self.lcm_type.msg_name}"
+
+    @staticmethod
+    def from_channel_str(channel: str, default_lcm_type: type[DimosMsg] | None = None) -> Topic:
+        """Create Topic from channel string.
+
+        Channel format: /topic#module.ClassName
+        Falls back to default_lcm_type if type cannot be parsed.
+        """
+        from dimos.msgs import resolve_msg_type
```
paul-nechifor, line 124: Why an inline import?

dimos/protocol/pubsub/impl/lcmpubsub.py
```diff
@@ -0,0 +1,221 @@
...
+        def unsubscribe() -> None:
+            if self.l is None:
+                return
+            self.l.unsubscribe(lcm_subscription)
+
+        return unsubscribe
+
+
+class LCM(  # type: ignore[misc]
+    LCMEncoderMixin,  # type: ignore[type-arg]
```
paul-nechifor, line 197: Don't you have to add the type args?

dimos/protocol/pubsub/spec.py
```diff
@@ -50,101 +38,124 @@ class _Subscription:
...
-    @abstractmethod
-    def decode(self, msg: EncodingT, topic: TopicT) -> MsgT: ...
+    def subscribe_new_topics(self, callback: Callable[[TopicT], Any]) -> Callable[[], None]:
+        """Discover new topics by tracking seen topics from subscribe_all."""
+        seen: set[TopicT] = set()
 
-    def __init__(self, *args, **kwargs) -> None:  # type: ignore[no-untyped-def]
-        super().__init__(*args, **kwargs)
-        self._encode_callback_map: dict = {}  # type: ignore[type-arg]
+        def on_msg(msg: MsgT, topic: TopicT) -> None:
```
paul-nechifor, line 126: I think we need locks on modifying `seen`.

dimos/protocol/pubsub/spec.py
```diff
@@ -50,101 +38,124 @@ class _Subscription:
...
-class PickleEncoderMixin(PubSubEncoderMixin[TopicT, MsgT, bytes]):
-    def encode(self, msg: MsgT, *_: TopicT) -> bytes:  # type: ignore[return]
-        try:
-            return pickle.dumps(msg)
-        except Exception as e:
-            print("Pickle encoding error:", e)
-            import traceback
+    def subscribe_all(self, callback: Callable[[MsgT, TopicT], Any]) -> Callable[[], None]:
+        """Subscribe to all topics by subscribing to each discovered topic."""
+        subscriptions: list[Callable[[], None]] = []
```
paul-nechifor, line 148: Need a lock

dimos/protocol/pubsub/spec.py
```diff
@@ -50,101 +38,124 @@ class _Subscription:
...
 
-    # async iterator
     async def aiter(self, topic: TopicT, *, max_pending: int | None = None) -> AsyncIterator[MsgT]:
         q: asyncio.Queue[MsgT] = asyncio.Queue(maxsize=max_pending or 0)
 
         def _cb(msg: MsgT, topic: TopicT) -> None:
             q.put_nowait(msg)
 
-        unsubscribe_fn = self.subscribe(topic, _cb)
+        unsubscribe_fn = self.subscribe(topic, _cb)  # type: ignore[attr-defined]
```
paul-nechifor, line 57: How come this `# type: ignore` is needed now?

dimos/protocol/pubsub/bridge.py
```diff
@@ -0,0 +1,97 @@
...
+
+TopicT = TypeVar("TopicT")
+MsgT = TypeVar("MsgT")
+TopicFrom = TypeVar("TopicFrom")
+TopicTo = TypeVar("TopicTo")
+MsgFrom = TypeVar("MsgFrom")
+MsgTo = TypeVar("MsgTo")
+
+
+class Translator(Protocol[TopicFrom, TopicTo, MsgFrom, MsgTo]):  # type: ignore[misc]
```
paul-nechifor, line 38: Why `type: ignore` on a protocol declaration?

dimos/protocol/pubsub/impl/lcmpubsub.py
```diff
@@ -0,0 +1,223 @@
...
+        return self._glob
+
+    def __repr__(self) -> str:
+        return f"Glob({self._glob!r})"
+
+
+@dataclass
+class Topic:
+    topic: str | re.Pattern[str] | Glob
+    lcm_type: type[DimosMsg] | None = None
```
paul-nechifor, line 98: What about types for which we use pickling? Shouldn't this basically allow any class?

dimos/protocol/pubsub/impl/lcmpubsub.py
```diff
@@ -0,0 +1,223 @@
...
+
+        topic_str = str(topic) if isinstance(topic, Topic) else topic
+        self.l.publish(topic_str, message)
+
+    def subscribe_all(self, callback: Callable[[bytes, Topic], Any]) -> Callable[[], None]:
+        return self.subscribe(Topic(re.compile(".*")), callback)  # type: ignore[arg-type]
+
+    def subscribe(  # type: ignore[override]
+        self, topic: Topic | str, callback: Callable[[bytes, Topic | str], Any]
+    ) -> Callable[[], None]:
```
paul-nechifor, line 159: `# type: ignore[override]` is not nice. 

PubSub.subscribe has a different declaration:

```
    @abstractmethod
    def subscribe(
        self, topic: TopicT, callback: Callable[[MsgT, TopicT], None]
    ) -> Callable[[], None]:
```

Should it be changed there as well?

dimos/protocol/pubsub/impl/lcmpubsub.py
```diff
@@ -0,0 +1,223 @@
...
+            if not pattern_str.endswith("*"):
+                pattern_str = f"{pattern_str}(#.*)?"
+
+            lcm_subscription = self.l.subscribe(pattern_str, handler)
+        else:
+            topic_str = str(topic) if isinstance(topic, Topic) else topic
+            lcm_subscription = self.l.subscribe(topic_str, lambda _, msg: callback(msg, topic))
+
+        # Set queue capacity to 10000 to handle high-volume bursts
+        lcm_subscription.set_queue_capacity(10000)
```
paul-nechifor, line 186: Might want to make this a constant.

paul-nechifor (approved):
Just some comments on `type: ignore`s? Maybe fix in a second pass?

leshy:
> Just some comments on `type: ignore`s? Maybe fix in a second pass?

taking in follow up PR just so this large thing doesn't dangle, follow ups coming

dimos/protocol/pubsub/bridge.py
leshy: bridge to be totally rewritten, just a sketch to explain this PR

dimos/protocol/pubsub/impl/lcmpubsub.py
leshy: yes but this is a structural problem, I need to get rid of inheritance and switch to composition to have correct types and this sorted.. will do this later

PR #1116, opened 2026-01-26
feat:  add joint servo and cartesian controllers

mustafab0:
JointServoTask for streaming joint position control (teleop, playback) and 
JointVelocityTask for streaming joint velocity control (joystick) also, 
Added CartesianIKTask with internal Pinocchio IK solver for cartesian control on any joint-space hardware
Add joint_command and cartesian_command input streams to ControlCoordinator for real-time streaming control via LCM
Add unified task_invoke RPC to interact with any ControlTask method
Add pygame-based cartesian jogger example for testing IK control

closes DIM-352


dimos/control/blueprints.py
```diff
@@ -60,303 +198,194 @@ def _joint_names(hardware_id: str, dof: int) -> list[str]:
...
-        ),
-    ],
-).transports(
-    {
-        ("joint_state", JointState): LCMTransport("/coordinator/joint_state", JointState),
-    }
-)
+    hardware=[_mock_arm("arm", 7)],
+    tasks=[_trajectory_task("traj_arm", "arm", 7)],
+).transports(_standard_transports())
```
spomichter: isnt it clearer for documentation purposes to have the transports actually listed here instead of referenced from this method?

mustafab0: I reverted to old standalone blueprint style

dimos/control/examples/cartesian_ik_jogger.py
```diff
@@ -0,0 +1,349 @@
...
+        return PoseStamped(
+            ts=time.time(),
+            frame_id=task_name,  # Used for task routing
+            position=position,
+            orientation=orientation,
+        )
+
+
+# Jog speeds
+LINEAR_SPEED = 0.05  # m/s
```
spomichter, line 135: shouldnt these be at top and not buried down at bottom

dimos/control/examples/cartesian_ik_jogger.py
```diff
@@ -0,0 +1,349 @@
...
+
+# Task name for routing (must match blueprint config)
+TASK_NAME = "cartesian_ik_arm"
+
+
+def clamp(value: float, min_val: float, max_val: float) -> float:
+    return max(min_val, min(max_val, value))
+
+
+def _get_piper_model_path() -> str:
```
spomichter, line 151: Shouldnt be any arm specific stuff in generic methods. UNless this is just an example thing but seems like it may be used by other arms

spomichter: Shouldnt have piper in here

dimos/control/tasks/cartesian_ik_task.py
```diff
@@ -0,0 +1,472 @@
...
+                f"CartesianIKTask {self._name}: clamping joint motion - "
+                f"joint {self._joint_names_list[worst_joint]} delta "
+                f"{np.degrees(joint_deltas[worst_joint]):.1f}° -> "
+                f"{np.degrees(clamped_delta[worst_joint]):.1f}°"
+            )
+            return False, q_clamped
+
+        return True, q_new
+
+    def on_preempted(self, by_task: str, joints: frozenset[str]) -> None:
```
spomichter, line 327: Doesn't look like any of these `on_preempted` callbacks are used. Also not sure what they do its just for printing warning

dimos/control/tasks/servo_task.py
```diff
@@ -0,0 +1,242 @@
...
+                    self._active = False
+                    return None
+
+            return JointCommandOutput(
+                joint_names=self._joint_names_list,
+                positions=list(self._target),
+                mode=ControlMode.SERVO_POSITION,
+            )
+
+    def on_preempted(self, by_task: str, joints: frozenset[str]) -> None:
```
spomichter, line 152: Same here

dimos/control/tasks/velocity_task.py
```diff
@@ -0,0 +1,277 @@
...
+            # Reset timeout flag on successful output
+            self._timed_out = False
+
+            return JointCommandOutput(
+                joint_names=self._joint_names_list,
+                velocities=list(self._velocities),
+                mode=ControlMode.VELOCITY,
+            )
+
+    def on_preempted(self, by_task: str, joints: frozenset[str]) -> None:
```
spomichter, line 182: and here

dimos/control/examples/cartesian_ik_jogger.py
mustafab0: You are right. This was implemented as an example with piper. I have asked @ruthwikdasyam to work on generic examples and extending support to more arms.

dimos/control/examples/cartesian_ik_jogger.py
mustafab0: Yeah, can also make this user parameter or task parameter

dimos/control/tasks/cartesian_ik_task.py
mustafab0: When there are  multiple task controlling a hardware, and one task overrides another. then it logs a warning that this task was preempted, just so that it doesn't silently fail.

PR #1117, opened 2026-01-27
Fix - Webcam Image vis through rerun + Skill execution

ruthwikdasyam:
1. CameraModule - Rerun integration

- Added rerun logging support
- Checks `viewer_backend` config, connects if rerun enabled

2. G1 Skill Container - RPC and Topic Fixes

- RPC call names if  `G1ConnectionModule` -> `G1Connection`
- bug: arm commands now correctly use `rt/api/arm/request` topic instead of `rt/api/sport/request`
- Refactor '_execute_g1_command' to accept topic as parameter for clarity

Result:
- Able to visualize images from ZED wecam through rerun
- Able to perform gestures by sending commands from humancli

### DIFF
--- a/dimos/robot/unitree_webrtc/unitree_g1_blueprints.py
+++ b/dimos/robot/unitree_webrtc/unitree_g1_blueprints.py
@@ -69,7 +69,7 @@ _basic_no_nav = (
     autoconnect(
         camera_module(
             transform=Transform(
-                translation=Vector3(0.05, 0.0, 0.6),
+                translation=Vector3(0.05, 0.0, 0.6),  # height of camera on G1 robot
                 rotation=Quaternion.from_euler(Vector3(0.0, 0.2, 0.0)),
                 frame_id="sensor",
                 child_frame_id="camera_link",


dimos/hardware/sensors/camera/module.py
```diff
@@ -57,24 +60,38 @@ class CameraModule(Module[CameraModuleConfig], perception.Camera):
...
     def start(self) -> None:
+        super().start()
+
         if callable(self.config.hardware):
             self.hardware = self.config.hardware()
         else:
             self.hardware = self.config.hardware
 
+        # Connect to Rerun if enabled (cache flag for use in callbacks)
+        self._rerun_enabled = self._global_config.viewer_backend.startswith("rerun")
```
paul-nechifor, line 81: Why `self._rerun_enabled` instead of `rerun_enabled`? It's only used here.

ruthwikdasyam: Makes sense, Changing it

ruthwikdasyam:
### Update:

New addition:

- `g1sim.py` - G1SimConnnection camera_link transform is `(0.3,0.0,0.0)` which is probably for Go2. But for G1, this should be around `(0.05,0.0,0.6)` which is modified.
- `sensors/camera/module.py` - added pinhole logging to viz image in 3d with robot
- `unitree_g1_blueprints.py` - added tf logging, and corrected camera height from base_link

dimos/hardware/sensors/camera/module.py
ruthwikdasyam: reverted to property, as it's used in `publish_metadata()` method to log camera_info

paul-nechifor (dismissed):

paul-nechifor (dismissed):

leshy (dismissed):

leshy (approved):

PR #1143, opened 2026-01-28
RPC Rework: With Var Renames

jeff-hykin:


dimos/core/blueprints.py
```diff
@@ -210,36 +266,114 @@ def _deploy_all_modules(
...
+                    )
+                    raise Exception(
+                        f"""The {blueprint.module.__name__} has a module reference ({each_module_ref}) which requested a module that fills out the {each_module_ref.spec.__name__} spec. Some modules ({possible_module_candidates_str}) met the spec structurally but had a mismatch in type annotations\n"""
+                    )
+                # one valid candidate (and more than one structurally valid candidate)
+                else:
+                    mod_and_mod_ref_to_proxy[blueprint.module, each_module_ref.name] = (
+                        valid_module_candidates[0]
+                    )
+
```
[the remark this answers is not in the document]
jeff-hykin: Yeah no you're wrong. Those are two different lists

examples/rpc_calls.py
```diff
@@ -0,0 +1,56 @@
+# Copyright 2026 Dimensional Inc.
```
paul-nechifor, line 1: Should this example be in the docs? I don't think anyone will look at it here.

dimos/core/blueprints.py
```diff
@@ -202,47 +258,124 @@ def _verify_no_name_conflicts(self) -> None:
...
                 logger.info(
                     "Transport",
                     name=remapped_name,
                     original_name=original_name,
                     topic=str(getattr(transport, "topic", None)),
-                    type=f"{type.__module__}.{type.__qualname__}",
+                    type=f"{stream_type.__module__}.{stream_type.__qualname__}",
                     module=module.__name__,
                     transport=transport.__class__.__name__,
                 )
```
paul-nechifor: Maybe this should be called `_connect_streams()` and `connection` renamed to `stream` everywhere in this func to match the rest of the renames.

examples/rpc_calls.py
jeff-hykin: I should have updated blueprints.md for the docs and the had this as a standalone example for people who jump into the code. (I believe this dir is linked to somewhere in the docs)

paul-nechifor (dismissed):

dimos/core/blueprints.py
jeff-hykin: Okay, done! And updated the `blueprints.md` with an example

dimos/core/test_blueprints.py
```diff
@@ -377,3 +379,127 @@ def test_future_annotations_autoconnect() -> None:
...
+    @rpc
+    def compute1(self, a: int, b: int) -> int:
+        return a * b
+
+    @rpc
+    def compute2(self, a: float, b: float) -> float:
+        return a * b
+
+    @rpc
+    def start(self) -> None: ...
```
paul-nechifor, line 419: You need `super().start()` here and stop in the others.

PR #1144, opened 2026-01-28
Create DDS Transport Protocol

Kaweees:
This merge request implements DDS (Data Distribution Service) transport layer using CycloneDDS, providing a new high-performance pub/sub transport option.

## Quick Start

> [!NOTE]
> We currently use [Eclipse Cyclone DDS](https://cyclonedds.io/docs/cyclonedds-python/latest/) as our DDS implementation. Its [IdlStruct](https://cyclonedds.io/docs/cyclonedds-python/latest/cyclonedds.idl.html#cyclonedds.idl.IdlStruct) feature lets you [define DDS topic types in pure Python](https://cyclonedds.io/docs/cyclonedds-python/latest/idl.html#datatypes), eliminating the need for separate [IDL files](https://en.wikipedia.org/wiki/Interface_description_language), with automatic serialization support.

```python
from dataclasses import dataclass
from cyclonedds.idl import IdlStruct
from dimos.protocol.pubsub.ddspubsub import DDS, Topic

@dataclass
class SensorReading(IdlStruct):
    value: float

dds = DDS()
dds.start()

received = []
sensor_topic = Topic(name="sensors/temperature", data_type=SensorReading)

dds.subscribe(sensor_topic, lambda msg, t: received.append(msg))
dds.publish(sensor_topic, SensorReading(value=22.5))

import time
time.sleep(0.1)

print(f"Received: {received}")
dds.stop()
```

```
Received: [SensorReading(value=22.5)]
```

## Unit Tests/Benchmarks

```sh
uv run pytest -svm tool dimos/protocol/pubsub/benchmark/test_benchmark.py --override-ini="addopts=" -k "DDS"
uv run pytest -svm tool dimos/protocol/pubsub/benchmark/test_benchmark.py
```

image

This builds off of #1036, which was closed because the branch was renamed to `miguel/dds_transport`

### DIFF
--- a/dimos/core/transport.py
+++ b/dimos/core/transport.py
@@ -269,25 +269,27 @@ class DDSTransport(PubSubTransport[T]):
 
     def start(self) -> None:
         with self._start_lock:
-            self.dds.start()
-            self._started = True
+            if not self._started:
+                self.dds.start()
+                self._started = True
 
     def stop(self) -> None:
         with self._start_lock:
-            self.dds.stop()
-            self._started = False
+            if self._started:
+                self.dds.stop()
+                self._started = False
 
-    def broadcast(self, _, msg) -> None:  # type: ignore[no-untyped-def]
+    def broadcast(self, _, msg) -> None:
         with self._start_lock:
             if not self._started:
                 self.start()
-        self.dds.publish(self.topic, msg)
+            self.dds.publish(self.topic, msg)
 
     def subscribe(self, callback: Callable[[T], None], selfstream: In[T] = None) -> None:  # type: ignore[assignment, override]
         with self._start_lock:
             if not self._started:
                 self.start()
-        return self.dds.subscribe(self.topic, lambda msg, topic: callback(msg))  # type: ignore[return-value]
+            return self.dds.subscribe(self.topic, lambda msg, topic: callback(msg))  # type: ignore[return-value]
 
 
 class ZenohTransport(PubSubTransport[T]): ...
--- a/dimos/protocol/service/ddsservice.py
+++ b/dimos/protocol/service/ddsservice.py
@@ -60,6 +60,7 @@ class DDSService(Service[DDSConfig]):
         global _participant
         with _participant_lock:
             if _participant is not None:
+                _participant.close()
                 _participant = None
                 logger.info("DDS service stopped")
         super().stop()


dimos/core/transport.py
```diff
@@ -212,4 +213,31 @@ def start(self) -> None: ...
...
+        self._started = True
+
+    def stop(self) -> None:
+        self.dds.stop()
+        self._started = False
+
+    def broadcast(self, _, msg) -> None:  # type: ignore[no-untyped-def]
+        if not self._started:
+            self.start()
+        self.dds.publish(self.topic, msg)
```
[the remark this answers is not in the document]
Kaweees: @leshy should this also be applied to Transports affected by #1062

leshy:
This diverged from dev and needs a sync

Perforamnce issues currently

https://discord.com/channels/1341146487186391173/1462686235297579242/1466291772824227882

dimos/core/transport.py
```diff
@@ -258,4 +260,34 @@ def stop(self) -> None:
             self._ros = None
 
 
+class DDSTransport(PubSubTransport[T]):
+    _started: bool = False
+
+    def __init__(self, topic: str, type: type, **kwargs) -> None:  # type: ignore[no-untyped-def]
+        super().__init__(DDSTopic(topic, type))
+        if not hasattr(self, "dds"):
```
paul-nechifor: In what circumstance would `self.dds` already be present? I don't think this `if` is needed.

spomichter: yes delete

dimos/core/transport.py
```diff
@@ -258,4 +260,34 @@ def stop(self) -> None:
...
+            self.dds = DDS(**kwargs)
+        self._start_lock = threading.Lock()
```
paul-nechifor: Yeah, many of the above are using a boolean `self._started`. But that's not safe. You need to use the lock in `start/stop` too.

dimos/protocol/pubsub/ddspubsub.py
```diff
@@ -0,0 +1,166 @@
...
+
+class _DDSMessageListener(Listener):
+    """Listener for DataReader that dispatches messages to callbacks."""
+
+    __slots__ = ("callbacks", "topic")
+
+    def __init__(self, topic: Topic, callbacks_dict: dict[Topic, list[MessageCallback]]) -> None:
+        super().__init__()
+        self.topic = topic
+        self.callbacks = callbacks_dict[topic]
```
paul-nechifor: Please prefix with `_`.

paul-nechifor: There's no point in passing `callbacks_dict` if you only store `callbacks_dict[topic]`. Pass just the list to _DDSMessageListener.

dimos/protocol/pubsub/ddspubsub.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def on_data_available(self, reader: DDSDataReader) -> None:
+        """Called when data is available on the reader."""
+        try:
+            samples = reader.take()
+        except Exception as e:
+            logger.error(f"Error reading from topic {self.topic}: {e}")
+            return
+        callbacks = self.callbacks
+        topic = self.topic
```
paul-nechifor: This is not needed.

dimos/protocol/pubsub/ddspubsub.py
```diff
@@ -0,0 +1,166 @@
...
+            writer.write(message)
+        except Exception as e:
+            logger.error(f"Error publishing to topic {topic}: {e}")
+
+    def _get_reader(self, topic: Topic) -> DDSDataReader[Any]:
+        """Get or create a DataReader for the given topic with listener."""
+        with self._reader_lock:
+            if topic not in self._readers:
+                dds_topic = DDSTopic(self.participant, topic.name, topic.data_type)
+                listener = _DDSMessageListener(topic, self._callbacks)
```
paul-nechifor: This is unsafe. In _DDSMessageListener, you loop over `self._callbacks[topic]`. But it doesn't use the lock.

In DDSPubSubBase.subscribe you use the lock to append to `self._callbacks[topic]`, but `_DDSMessageListener` reads from the list without checking the lock.

Essentially, you either have to pass the lock, or have locks per list.

One way to do this is to colocate the list and the lock. Have the list of callbacks be stored in _DDSMessageListener and have a lock there too. Then expose a   `_DDSMessageListener.add_callback` method (which uses the lock) and is called in `DDSPubSubBase.subscribe`

docs/concepts/transports.md
```diff
@@ -288,6 +288,35 @@ shm.stop()
...
+
+@dataclass
+class SensorReading(IdlStruct):
+    value: float
+
+dds = DDS()
+dds.start()
+
+received = []
+sensor_topic = Topic(name="sensors/temperature", typename=SensorReading)
```
paul-nechifor: I don't think this is running. Topic has a `data_type` field, not `typename`.

You might want to check with @leshy how to execute this and get the error as feedback.

dimos/core/transport.py
Kaweees: Agreed, added the locks in start/stop

docs/concepts/transports.md
Kaweees: I forgot to update this example when I renamed `typename` to `data_type` in `Topic. After renaming this, I get the intended output:

```sh
extra-build-dependencies` option is experimental and may change without warning. Pass `--preview-features extra-build-dependencies` to disable this warning.
2026-02-01T04:48:43.538025Z [info     ] DDS service started with Cyclone DDS domain 0                [dimos/protocol/service/ddsservice.py]
Received: 22.5
```

dimos/core/transport.py
Kaweees: Removed

dimos/protocol/pubsub/ddspubsub.py
Kaweees: Assuming I implement the list approach via tuples (which are atomic) the lock also be used in `on_data_available()`?

Kaweees: I realized a lock is not needed in `on_data_available()`, since the callback would operate on an immutable tuple iterator that `add_callback`/`remove_callback` could not affect

dimos/core/transport.py
```diff
@@ -258,4 +260,33 @@ def stop(self) -> None:
...
+
+    def stop(self) -> None:
+        with self._start_lock:
+            self.dds.stop()
+            self._started = False
+
+    def broadcast(self, _, msg) -> None:  # type: ignore[no-untyped-def]
+        with self._start_lock:
+            if not self._started:
+                self.start()
```
paul-nechifor, line 285: This will deadlock. You have `with self._start_lock` both here and in `self.start`. You need to change `self._start_lock` to be a reentrant lock (`RLock()`)

dimos/protocol/pubsub/ddspubsub.py
```diff
@@ -0,0 +1,173 @@
...
+        except Exception as e:
+            logger.error(f"Error reading from topic {self._topic}: {e}")
+            return
+        for sample in samples:
+            if sample is not None:
+                for callback in self._callbacks:
+                    try:
+                        callback(sample, self._topic)
+                    except Exception as e:
+                        logger.error(f"Callback error on topic {self._topic}: {e}")
```
paul-nechifor:
```suggestion
                        logger.error(f"Callback error on topic {self._topic}: {e}", exc_info=True)
```

dimos/protocol/pubsub/ddspubsub.py
```diff
@@ -0,0 +1,173 @@
...
+        """Remove a callback from the listener."""
+        with self._lock:
+            self._callbacks = tuple(cb for cb in self._callbacks if cb is not callback)
+
+    def on_data_available(self, reader: DDSDataReader) -> None:
+        """Called when data is available on the reader."""
+        try:
+            samples = reader.take()
+        except Exception as e:
+            logger.error(f"Error reading from topic {self._topic}: {e}")
```
paul-nechifor:
```suggestion
            logger.error(f"Error reading from topic {self._topic}: {e}", exc_info=True)
```

dimos/protocol/pubsub/ddspubsub.py
```diff
@@ -0,0 +1,173 @@
...
+                self._writers[topic] = DDSDataWriter(self.participant, dds_topic, qos=self._qos)
+            return self._writers[topic]
+
+    def publish(self, topic: Topic, message: Any) -> None:
+        """Publish a message to a DDS topic."""
+        writer = self._get_writer(topic)
+        try:
+            writer.write(message)
+        except Exception as e:
+            logger.error(f"Error publishing to topic {topic}: {e}")
```
paul-nechifor:
```suggestion
            logger.error(f"Error publishing to topic {topic}: {e}, exc_info=True")
```

dimos/protocol/pubsub/ddspubsub.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def _unsubscribe_callback(self, topic: Topic, callback: MessageCallback) -> None:
+        """Unsubscribe a callback from a topic."""
+        with self._reader_lock:
+            listener = self._listeners.get(topic)
+        if listener:
+            listener.remove_callback(callback)
+
+
+class DDS(DDSPubSubBase): ...
```
paul-nechifor: If `DDSPubSubBase` is the same as `DDS`, why not just rename `DDSPubSubBase` to `DDS`?

dimos/protocol/service/ddsservice.py
```diff
@@ -0,0 +1,72 @@
...
+                _participant = DomainParticipant(self.config.domain_id)
+                logger.info(f"DDS service started with Cyclone DDS domain {self.config.domain_id}")
+
+    def start(self) -> None:
+        """Start the DDS service."""
+        pass
+
+    def stop(self) -> None:
+        """Stop the DDS service."""
+        pass
```
paul-nechifor: When you override methods, you almost always want to call `super().<method_name>()`.

dimos/protocol/pubsub/ddspubsub.py
Kaweees: Originally I wanted to write an abstract base class which would have implementations of DDS for CycloneDDS and FastDDS, and this is a relic of that. I have since realized that was premature optimization. I’ll rename DDSPubSubBase to DDS

dimos/protocol/pubsub/ddspubsub.py
paul-nechifor: :+1:

dimos/core/transport.py
Kaweees: Got it, changed it to a reentrant lock (`RLock()`)

dimos/core/transport.py
```diff
@@ -258,4 +260,36 @@ def stop(self) -> None:
             self._ros = None
 
 
+class DDSTransport(PubSubTransport[T]):
+    def __init__(self, topic: str, type: type, **kwargs) -> None:  # type: ignore[no-untyped-def]
+        super().__init__(DDSTopic(topic, type))
+        self.dds = DDS(**kwargs)
+        self._started: bool = False
+        self._start_lock = threading.RLock()
```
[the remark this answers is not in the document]
Kaweees: DDS will not have picking support, as CycloneDDS automatically handles encoding/decoding

Kaweees:
@greptile

Kaweees:
Too many conflicts with `dev`, I'll open a new branch

PR #1145, opened 2026-01-28
RPC Rework: no var renames

jeff-hykin:
Reverts the _BlueprintAtom/Blueprint rename

### DIFF
--- a/dimos/core/module.py
+++ b/dimos/core/module.py
@@ -31,7 +31,7 @@ if TYPE_CHECKING:
     from collections.abc import Callable
 
     from dimos.core.introspection.module import ModuleInfo
-    from dimos.core.rpc_client import RpcCall, RPCClient
+    from dimos.core.rpc_client import RPCClient
 
 from dask.distributed import Actor, get_worker
 from reactivex.disposable import CompositeDisposable
@@ -41,6 +41,7 @@ from dimos.core import colors
 from dimos.core.core import T, rpc
 from dimos.core.introspection.module import extract_module_info, render_module_io
 from dimos.core.resource import Resource
+from dimos.core.rpc_client import RpcCall  # noqa: TC001
 from dimos.core.stream import In, Out, RemoteIn, RemoteOut, Transport
 from dimos.protocol.rpc import LCMRPC, RPCSpec
 from dimos.protocol.service import Configurable  # type: ignore[attr-defined]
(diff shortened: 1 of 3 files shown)


dimos/core/introspection/blueprint/dot.py
```diff
@@ -96,9 +96,9 @@ def render(
             remapped_name = blueprint_set.remapping_map.get((bp.module, conn.name), conn.name)
             key = (remapped_name, conn.type)
             if conn.direction == "out":
-                producers[key].append(bp.module)
+                producers[key].append(bp.module)  # type: ignore[index]
```
jeff-hykin, line 99: For connections (streams) the replacement will only ever be a str, even though the remapping type annotation has `str | type[Spec]` because replacement of module ref's will be with Specs

dimos/core/module.py
```diff
@@ -11,8 +11,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from __future__ import annotations
```
jeff-hykin, line 14: I've seen some places where we use annotations, some places where we don't. If you guys have a preference let me know and I'll use that.

dimos/core/module.py
```diff
@@ -339,6 +342,10 @@ def set_rpc_method(self, method: str, callable: RpcCall) -> None:
         callable.set_rpc(self.rpc)  # type: ignore[arg-type]
         self._bound_rpc_calls[method] = callable
 
+    @rpc
+    def set_module_ref(self, name: str, module_ref: RPCClient) -> None:
+        setattr(self, name, module_ref)
```
jeff-hykin, line 348: Important bit: lets us connect the module reference after the module is running/deployed

dimos/core/test_blueprints.py
```diff
@@ -22,8 +25,7 @@
 from dimos.core.blueprints import (
     ModuleBlueprint,
     ModuleBlueprintSet,
-    ModuleConnection,
-    _make_module_blueprint,
+    StreamRef,
```
jeff-hykin, line 28: It made sense to have `ModuleRef` for this PR, and then to be consistent in naming I changed `ModuleConnection` (which is now easy to confuse with ModuleRef) to StreamRef to be less confusing

PR #1147, opened 2026-01-28
Fix/benchmark use iec

leshy:
benchmark should use IEC units since that's what ppl are used to

```sh
./bin/dev python -m pytest -svm tool -k "not bytes" dimos/protocol/pubsub/benchmark/test_benchmark.py
```

### DIFF
--- a/dimos/protocol/pubsub/benchmark/type.py
+++ b/dimos/protocol/pubsub/benchmark/type.py
@@ -41,6 +41,21 @@ class Case(Generic[TopicT, MsgT]):
 TestData = Sequence[Case[Any, Any]]
 
 
+def _format_mib(value: float) -> str:
+    """Format bytes as MiB with intelligent rounding.
+
+    >= 10 MiB: integer (e.g., "42")
+    1-10 MiB: 1 decimal (e.g., "2.5")
+    < 1 MiB: 2 decimals (e.g., "0.07")
+    """
+    mib = value / (1024**2)
+    if mib >= 10:
+        return f"{mib:.0f}"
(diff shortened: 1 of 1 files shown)


dimos/protocol/pubsub/benchmark/type.py
```diff
@@ -41,24 +41,16 @@ def __len__(self) -> int:
...
-def _format_throughput(bytes_per_sec: float) -> str:
-    """Format throughput to human-readable string."""
-    if bytes_per_sec >= 1e9:
-        return f"{bytes_per_sec / 1e9:.2f} GB/s"
-    if bytes_per_sec >= 1e6:
-        return f"{bytes_per_sec / 1e6:.2f} MB/s"
-    if bytes_per_sec >= 1e3:
-        return f"{bytes_per_sec / 1e3:.2f} KB/s"
-    return f"{bytes_per_sec:.2f} B/s"
+def _format_iec(value: float, concise: bool = False, decimals: int = 2) -> str:
```
paul-nechifor, line 59: Personally, I think using a single unit (maybe MB/s) is the best option because it's easier to compare

Image

It's much easier to visualize as:

```
  68 MB/s
 844 MB/s
2110 MB/s
0.98 MB/s
```

paul-nechifor: ...actually, you can just omit the unit and put it in the table header.

leshy: done

paul-nechifor (approved):

PR #1162, opened 2026-01-30
nix: Skip pre-commit install if hooks already exist

leshy:
Avoids running pre-commit install on every shell entry when hooks are already set up.


jeff-hykin (approved):
LGTM

paul-nechifor (approved):

PR #1170, opened 2026-02-03
Switch ./bin/dev to use ros-dev image

leshy:
Switch ./bin/dev to use ros-dev image

@paul-nechifor why wasn't this enough to have same mypy results as in CI?

### DIFF
--- a/bin/dev
+++ b/bin/dev
@@ -1,26 +1,26 @@
 #!/usr/bin/env bash
 REPO_ROOT=$(git rev-parse --show-toplevel)
 REL_PATH=$(realpath --relative-to="$REPO_ROOT" "$(pwd)")
-IMAGE="ghcr.io/dimensionalos/dev"
-IMAGE_NAME="$IMAGE:dev"
+IMAGE="ghcr.io/dimensionalos/ros-dev"
+IMAGE_TAG="$IMAGE:dev"
 
 image_exists() {
-    docker image inspect "${IMAGE_NAME}" &>/dev/null
+    docker image inspect "${IMAGE_TAG}" &>/dev/null
 }
 
 image_pull() {
-    docker pull "$IMAGE_NAME"
+    docker pull "$IMAGE_TAG"
 }
 
 ensure_image_downloaded() {
     if ! image_exists "$1"; then
-        echo "Image ${IMAGE_NAME} not found. Pulling..."
+        echo "Image ${IMAGE_TAG} not found. Pulling..."
         image_pull "$1"
     fi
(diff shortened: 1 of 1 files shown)


bin/dev
```diff
@@ -1,43 +1,43 @@
 #!/usr/bin/env bash
 REPO_ROOT=$(git rev-parse --show-toplevel)
 REL_PATH=$(realpath --relative-to="$REPO_ROOT" "$(pwd)")
-IMAGE="ghcr.io/dimensionalos/dev"
-IMAGE_NAME="$IMAGE:dev"
+IMAGE="ghcr.io/dimensionalos/ros-dev"
+IMAGE_TAG="$IMAGE:dev"
```
paul-nechifor, line 5: It's not really just the tag, so why not:

```suggestion
IMAGE="ghcr.io/dimensionalos/ros-dev:dev"
```

bin/dev
```diff
@@ -58,16 +58,16 @@ get_tag() {
...
 build_image() {
     local image_tag
-    image_tag=$(get_tag)
+    image_tag=$(get_branch_tag)
 
     docker build \
         --build-arg GIT_COMMIT=$(git rev-parse --short HEAD) \
         --build-arg GIT_BRANCH=$(git rev-parse --abbrev-ref HEAD) \
-        -t "ghcr.io/dimensionalos/dev:${image_tag}" -f docker/dev/Dockerfile .
+        -t "ghcr.io/dimensionalos/ros-dev:${image_tag}" -f docker/dev/Dockerfile .
```
paul-nechifor, line 66: Oh, I see. This should probably use `$IMAGE` then...

paul-nechifor (approved):

bin/dev
leshy: yeah this is a bit messy/confusing, the thing is that you have images related to just your branch that you can build locally via `./bin/dev build` so we have this discovery of tags - CI is or is supposed to do the same for PRs

at the same time we have pulling from :dev tag if you do just `./bin/dev` needs a rethink/rewrite, but just wanted a quick update to ros

bin/dev
```diff
@@ -1,43 +1,43 @@
 #!/usr/bin/env bash
 REPO_ROOT=$(git rev-parse --show-toplevel)
 REL_PATH=$(realpath --relative-to="$REPO_ROOT" "$(pwd)")
```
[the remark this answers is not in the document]
leshy: idk who wrote this docker compose why we have it etc, I think obsolete?

PR #1211, opened 2026-02-07
Agent refactor

paul-nechifor:
## Problem

We need a simpler base for agents so we can build long running tasks.

Closes DIM-447 ( https://linear.app/dimensional/issue/DIM-447/refactor-agents )

## Solution

  - Switched from init_chat_model with custom tool handling to create_agent (the recommended way for the latest version of langchain)
  - Skill registration is no longer required. The agent now discovers tools automatically by calling get_skills() on all launched modules at startup via the new on_system_modules callback from ModuleCoordinator
  - Removed AgentSpec class and `llm_init.py`. The Agent is now a plain Module with a threaded message loop instead of an async coroutine-based one
  - Module no longer inherits from SkillContainer. Skill discovery uses the __skill__ attribute set by the `@skill` decorator
  - Removed HumanInput module. Human input now comes in via the human_input stream on the Agent directly
  - Moved agents/cli/web.py to agents/web_human_input.py
  - Added AgentTestRunner module that feeds messages into the agent and waits for idle, used by the new test fixtures
  - Rewrote agent tests
  - Camera video_stream skill (which was a passive generator) replaced with take_a_picture that just returns the latest frame
  - Model/provider enums removed. Model is now just a string like `"gpt-4o"` or `"ollama:qwen3:8b"` or `"huggingface:Qwen/Qwen2.5-1.5B-Instruct"`
  - Blueprints updated: `llm_agent(), human_input() `replaced with just `agent()`. Added demo-agent and demo-agent-camera blueprints

## Breaking Changes

Quite a few breaking changes.

  - `@skill` now does not support any arguments. All skills marked with it must return in a short time. They must return a string or an Image like before. If they want to return multiple things, they can return None and then use `self.agent_spec.add_message(HumanMessage(...))` to add a new message at any time.
  - SkillModule is removed. Use Module instead.
  - SkillContainer, SkillCoordinator, and everything in dimos/protocol/skill/ is removed.
  - HumanInput module is removed.
  - llm_agent blueprint is removed, use agent instead.
  - AgentSpec (the old base class) is removed. The new AgentSpec is just a protocol with add_message (for use with the new RPC system).
  - Model is now a plain string, not an enum. Provider is encoded in the model string prefix.

## How to Test

```bash
uv run dimos run unitree-go2-agentic
uv run humancli
```
...and prompt it anything like before.


leshy (approved):
this nukes protocol skill/ enitrely and ideas behind this, but letting Paul own this, proper MCP will reintroduce all old features anyway
need to update others on skillspy, agentspy etc not being available anymore?

PR #1216, opened 2026-02-08
Remove `tofix` 

paul-nechifor:
* The `tofix` tests actually work (locally and in CI) and are useful because they test CLIP which is used in spatial memory.


dimos/agents_deprecated/memory/test_image_embedding.py
```diff
@@ -28,10 +28,10 @@
 
 
 @pytest.mark.heavy
+@pytest.mark.skip_macos
```
jeff-hykin: probably just remove `skip_macos` cause these actually all pass on macos. Not sure why `tofix` was on there in the first place

paul-nechifor: Done.

leshy (dismissed):

leshy (approved):

PR #1217, opened 2026-02-08
Feature: Detection3D and Object support to Manipulation Module  

mustafab0:
### Problem
Detected 3D objects from perception cannot be used as collision obstacles for motion planning.
ManipulationModule lacks TF publishing needed for eye-in-hand camera transform chains.

### Solution
ObjectSceneRegistration publishes deduplicated objects via new objects output port with pointclouds.
ManipulationModule subscribes async via observable() and caches objects in WorldObstacleMonitor.

Dedicated 10Hz TF thread publishes EE and tf_extra_links poses using scratch_context().
Eager self.tf init prevents autoconf() blocking in Dask workers.

WorldObstacleMonitor syncs BOX obstacles to Drake on refresh() with opt-in convex hull meshes.
pointcloud_to_convex_hull_obj centers points at origin to avoid double-transform with pose.

ManipulationClient adds refresh(), detections(), goto_object(), and perception status commands.
xarm_perception blueprint wires xArm7 + eye-in-hand RealSense + ObjectSceneRegistration + Foxglove.
### 
Breaking Changes
None.

### How to Test

```
pytest dimos/manipulation/test_manipulation_unit.py -v
dimos run xarm_perception
# In client: refresh(5), goto_object("cup", dz=0.1), preview(), execute()
```

closes DIM-353
closes DIM-397
closes DIM-396

### DIFF
--- a/dimos/manipulation/manipulation_module.py
+++ b/dimos/manipulation/manipulation_module.py
@@ -382,6 +382,15 @@ class ManipulationModule(Module):
         self._error_message = msg
         return False
 
+    def _dismiss_preview(self, robot_id: WorldRobotID) -> None:
+        """Hide the preview ghost if the world supports it."""
+        if self._world_monitor is None:
+            return
+        world = self._world_monitor.world
+        if hasattr(world, "hide_preview"):
+            world.hide_preview(robot_id)  # type: ignore[attr-defined]
+            world.publish_visualization()
+
     @rpc
     def plan_to_pose(self, pose: Pose, robot_name: RobotName | None = None) -> bool:
         """Plan motion to pose. Use preview_path() then execute().
@@ -442,6 +451,7 @@ class ManipulationModule(Module):
     ) -> bool:
         """Plan path from current position to goal, store result."""
         assert self._world_monitor and self._planner  # guaranteed by _begin_planning
+        self._dismiss_preview(robot_id)
         start = self._world_monitor.get_current_joint_state(robot_id)
         if start is None:
             return self._fail("No joint state")
@@ -492,7 +502,7 @@ class ManipulationModule(Module):
             return False
 
         # Interpolate and animate
-        interpolated = interpolate_path(planned_path, resolution=0.02)
+        interpolated = interpolate_path(planned_path, resolution=0.1)
         self._world_monitor.world.animate_path(robot_id, interpolated, duration)
         return True
 
--- a/dimos/manipulation/planning/spec/protocols.py
+++ b/dimos/manipulation/planning/spec/protocols.py
@@ -178,6 +178,10 @@ class WorldSpec(Protocol):
         """Animate a path in visualization."""
         ...
 
+    def close(self) -> None:
+        """Release visualization resources."""
+        ...
+
 
 @runtime_checkable
 class KinematicsSpec(Protocol):
--- a/dimos/manipulation/planning/monitor/world_monitor.py
+++ b/dimos/manipulation/planning/monitor/world_monitor.py
@@ -178,6 +178,8 @@ class WorldMonitor:
 
             logger.info("All monitors stopped")
 
+        self._world.close()
+
     # ============= Message Handlers =============
 
     def on_joint_state(self, msg: JointState, robot_id: WorldRobotID | None = None) -> None:
--- a/dimos/manipulation/planning/world/drake_world.py
+++ b/dimos/manipulation/planning/world/drake_world.py
@@ -16,10 +16,11 @@
 
 from __future__ import annotations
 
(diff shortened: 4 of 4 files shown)


dimos/manipulation/planning/utils/mesh_utils.py
```diff
@@ -271,6 +271,77 @@ def convert_mesh(match: re.Match[str]) -> str:
...
+    try:
+        hull, _ = pcd.compute_convex_hull()
+    except Exception as e:
+        logger.warning(f"Convex hull computation failed: {e}")
+        return None
+
+    if output_path is None:
+        hull_dir = _CACHE_DIR / "convex_hulls"
+        hull_dir.mkdir(parents=True, exist_ok=True)
+        output_path = hull_dir / f"hull_{id(points):x}.obj"
```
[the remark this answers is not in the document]
mustafab0: We don't want cache reuse here because each refresh() call gets a fresh pointcloud snapshot and should overwrite with the latest hull. Content hashing would accumulate stale files. The id(points) just needs to be unique within a single refresh cycle.

dimos/manipulation/planning/examples/manipulation_client.py
```diff
@@ -260,6 +261,104 @@ def close_gripper(self, robot_name: str | None = None) -> bool:
...
+    def perception(self) -> dict[str, int] | None:
+        """Get perception status (cached/added counts)."""
+        return cast("dict[str, int] | None", self._call("get_perception_status"))
+
+    def detections(self) -> list[dict[str, object]] | None:
+        """List cached detections from perception."""
+        result = cast("list[dict[str, object]] | None", self._call("list_cached_detections"))
+        if result:
+            for i, det in enumerate(result):
+                center = cast("list[float]", det.get("center", [0, 0, 0]))
```
paul-nechifor, line 277: Is the function supposed to return `list[dict[str, list[float]]] | None`?

mustafab0: since we are working with object types, we are just creating a summary.

We are just displaying this to the user, so it could just be `list[dict[str, list[float]]] | None` if we wanted. But is there a reason why we would want that?

dimos/manipulation/planning/utils/mesh_utils.py
```diff
@@ -271,6 +276,77 @@ def convert_mesh(match: re.Match[str]) -> str:
     return re.sub(pattern, convert_mesh, urdf_content)
 
 
+def pointcloud_to_convex_hull_obj(
+    points: NDArray[np.float64],
```
leshy, line 280: why are you keeping points as numpy arrays? should use PointCloud2 which is stored in o3d, which has natural very fast convex hull calculations so this function would be one line or can add .to_convex_hull on PointCloud2 class?

dimos/manipulation/manipulation_module.py
```diff
@@ -41,6 +41,7 @@
 # These must be imported at runtime (not TYPE_CHECKING) for In/Out port creation
 from dimos.msgs.sensor_msgs import JointState
 from dimos.msgs.trajectory_msgs import JointTrajectory
+from dimos.perception.detection.type.detection3d.object import Object as DetObject  # noqa: TC001
```
leshy, line 44: not a huge fan of this Object type, as this seems to be a current top level interface for the world this is a potential issue, but not reviewing it in detail, feel free to own/rewrite

dimos/manipulation/planning/spec/protocols.py
```diff
@@ -155,6 +155,12 @@ def get_ee_pose(self, ctx: Any, robot_id: WorldRobotID) -> PoseStamped:
         """Get end-effector pose."""
         ...
 
+    def get_link_pose(
+        self, ctx: Any, robot_id: WorldRobotID, link_name: str
+    ) -> NDArray[np.float64]:
```
leshy, line 160: why are you returning np arrays and not poses/transforms? as long as your external interface is dimos types I don't care, but u will suffer

dimos/manipulation/planning/spec/config.py
```diff
@@ -74,6 +74,8 @@ class RobotModelConfig:
     joint_name_mapping: dict[str, str] = field(default_factory=dict)
     coordinator_task_name: str | None = None
     gripper_hardware_id: str | None = None
+    # TF publishing for extra links (e.g., camera mount)
+    tf_extra_links: list[str] = field(default_factory=list)
```
leshy, line 78: shouldn't this be a list of actual transforms? from a particular existing arm transform to a new one? 
(like Joint3 -> Joint_Camera)

I can have multiple cameras etc, why strings, they don't encode actual spatial relationships

PR #1220, opened 2026-02-08
Pytest CUDA Marker

jeff-hykin:
GPU tests are different than cuda tests (most GPU tests run on MacOS, its only the cuda ones that dont)

### DIFF
--- a/dimos/models/depth/test_metric3d.py
+++ b/dimos/models/depth/test_metric3d.py
@@ -11,7 +11,7 @@ def sample_intrinsics() -> list[float]:
     """Sample camera intrinsics [fx, fy, cx, cy]."""
     return [500.0, 500.0, 320.0, 240.0]
 
-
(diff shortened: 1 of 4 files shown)


dimos/models/depth/test_metric3d.py
```diff
@@ -30,7 +30,6 @@ def test_metric3d_update_intrinsic(sample_intrinsics: list[float]) -> None:
     model.update_intrinsic(new_intrinsics)
     assert model.intrinsic == new_intrinsics
 
-
 @pytest.mark.gpu
```
jeff-hykin, line 33: NOTE: this one doesnt require cuda, even though its still metric3d (only some of the metric3d tests need cuda)

paul-nechifor (dismissed):

paul-nechifor (approved):

PR #1227, opened 2026-02-10
Bugfix: Trajectory preview randomly times out in drake meshcat visualizer

mustafab0:
## Problem

`preview_path()` RPC randomly times out. Drake's Meshcat throws `SystemExit` (not `Exception`) when called from any thread other than its creator. RPC handlers run on a 50-worker thread pool, so many Meshcat call dies silently — no response sent, client sees timeout.

Issue: #

---

## Solution

Route all Meshcat calls through a `ThreadPoolExecutor(max_workers=1)` that creates and owns the `Meshcat` instance. `_on_viz_thread(fn, *args)` submits work to this executor, with re-entrancy detection to avoid deadlocks when `animate_path`'s nested calls are already on the viz thread.

Also reduced interpolation resolution from `0.02` to `0.1` for faster preview animations.

---

## Breaking Changes

None

---

## How to Test

1. Run a blueprint xarm-perception, coordinator-mock and manipulation_client.py
2. Open Meshcat in browser
3. Call `plan_to_joints()` then `preview_path()` via RPC
4. Yellow preview robot animates the path, live robot stays in place, no timeout

closes DIM-443

### DIFF
--- a/dimos/manipulation/planning/world/drake_world.py
+++ b/dimos/manipulation/planning/world/drake_world.py
@@ -106,6 +106,50 @@ class _ObstacleData:
     source_id: Any  # SourceId
 
 
+class _ThreadSafeMeshcat:
+    """Wraps Drake Meshcat so all calls run on the creator thread.
+
+    Drake throws SystemExit from non-creator threads for every Meshcat operation.
+    This class creates a single-thread executor, constructs Meshcat on it,
+    and proxies all calls through it.
+    """
+
+    def __init__(self) -> None:
+        self._executor = ThreadPoolExecutor(max_workers=1, thread_name_prefix="meshcat")
+        self._thread = self._executor.submit(current_thread).result()
+        self._inner: Meshcat = self._executor.submit(Meshcat).result()
+
+    def _call(self, fn: Any, *args: Any, **kwargs: Any) -> Any:
+        if current_thread() is self._thread:
+            return fn(*args, **kwargs)
+        return self._executor.submit(fn, *args, **kwargs).result()
+
+    # --- Meshcat proxies ---
+
+    def SetObject(self, *args: Any, **kwargs: Any) -> Any:
+        return self._call(self._inner.SetObject, *args, **kwargs)
+
+    def SetTransform(self, *args: Any, **kwargs: Any) -> Any:
+        return self._call(self._inner.SetTransform, *args, **kwargs)
+
+    def SetProperty(self, *args: Any, **kwargs: Any) -> Any:
+        return self._call(self._inner.SetProperty, *args, **kwargs)
+
+    def Delete(self, *args: Any, **kwargs: Any) -> Any:
+        return self._call(self._inner.Delete, *args, **kwargs)
+
+    def web_url(self) -> str:
+        result: str = self._call(self._inner.web_url)
+        return result
+
+    def forced_publish(self, visualizer: Any, viz_ctx: Any) -> None:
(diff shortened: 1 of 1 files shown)


dimos/manipulation/planning/world/drake_world.py
```diff
@@ -124,11 +129,17 @@ def __init__(self, time_step: float = 0.0, enable_viz: bool = False):
...
+        # A single-thread executor ensures every viz call runs on that thread.
+        self._viz_executor: ThreadPoolExecutor | None = None
+        self._viz_thread: Any = None  # Thread object for re-entrancy detection
         self._meshcat: Meshcat | None = None
         self._meshcat_visualizer: MeshcatVisualizer | None = None
         if enable_viz:
-            self._meshcat = Meshcat()
+            self._viz_executor = ThreadPoolExecutor(max_workers=1, thread_name_prefix="meshcat")
+            self._viz_thread = self._viz_executor.submit(current_thread).result()
+            self._meshcat = self._viz_executor.submit(Meshcat).result()
```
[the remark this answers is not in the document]
mustafab0: Added close method to shutdown executor

paul-nechifor: Yes, looks good.

The only suggestion I'd have is to wrap this.

DrakeWorld doesn't care about the executor or the thread, it just needs to call meshcat, so it could be simpler from DrakeWorld's perspective if you'd have a SingleThreadedMeshcat (which contains all of this)

dimos/manipulation/planning/world/drake_world.py
```diff
@@ -860,44 +933,85 @@ def publish_visualization(self, ctx: Context | None = None) -> None:
...
+            self.show_preview(robot_id)
+            dt = duration / (len(path) - 1)
+            try:
+                for joint_state in path:
+                    positions = np.array(joint_state.position, dtype=np.float64)
+                    with self._lock:
+                        assert self._plant_context is not None
+                        self._set_preview_positions(self._plant_context, robot_id, positions)
+                    self.publish_visualization()
+                    time.sleep(dt)
```
paul-nechifor: Sleeping on the meshcat thread blocks other calls. You should sleep on the calling thread.

dimos/manipulation/manipulation_module.py
```diff
@@ -492,7 +492,7 @@ def preview_path(self, duration: float = 3.0, robot_name: RobotName | None = Non
             return False
 
         # Interpolate and animate
-        interpolated = interpolate_path(planned_path, resolution=0.02)
+        interpolated = interpolate_path(planned_path, resolution=0.1)
```
paul-nechifor, line 505: Why the resolution change?

mustafab0: The interpolation happens only for animation. As I was having too many timeouts, one theory was that the resolution was too fine. It doesn't seem to be the case. I can test to see if 0.2 is good enough. 

But 0.1 seems to work great for preview, and less computational overhead.

dimos/manipulation/planning/world/drake_world.py
mustafab0: removed sleep from meshcat thread. sleep now lives in the calling thread

dimos/manipulation/planning/world/drake_world.py
mustafab0: Wrapped the meshcat in a ThreadSafeMeshcat class.

Implementation is much more cleaner. Thanks !

PR #1234, opened 2026-02-11
Integrate graspgen in manipulationmodule as a service

### LINKED ISSUE #1130: Integrate GraspGen

JalajShuklaSS:
## Summary
Adds ``@rpc generate_grasps(pointcloud, scene_pointcloud)`` -> ``PoseArray`` to ``ManipulationModule`` that lazily spins up a ``DockerModule(GraspGenModule)`` on first call, managing the full Docker ``lifecycle``

How it works
ManipulationModule lazily creates and starts a GraspGen Docker container on first ``generate_grasps()`` call, to the point cloud to GraspGenModule via LCMRPC, and returns ranked grasp poses as PoseArray.

Changes
``dimos/manipulation/manipulation_module.py``

Closes #1130 DIM-390


dimos/manipulation/manipulation_module.py
```diff
@@ -598,6 +611,52 @@ def get_trajectory_status(
...
+
+        from dimos.core.docker_runner import DockerModule
+        from dimos.manipulation.grasping.graspgen_module import GraspGenModule
+        from dimos.utils.data import get_data
+        from dimos.utils.path_utils import get_project_root
+
+        # Ensure GraspGen model checkpoints are pulled from LFS
+        get_data("models_graspgen")
+
+        project_root = get_project_root()
```
paul-nechifor: Use `from dimos.constants import DIMOS_PROJECT_ROOT`

dimos/manipulation/manipulation_module.py
```diff
@@ -598,6 +611,52 @@ def get_trajectory_status(
...
 
+    def _get_graspgen(self) -> DockerModule | None:
+        """Get or create GraspGen Docker module (lazy init)."""
+        if self._graspgen is not None:
+            return self._graspgen
+
+        from dimos.core.docker_runner import DockerModule
+        from dimos.manipulation.grasping.graspgen_module import GraspGenModule
+        from dimos.utils.data import get_data
+        from dimos.utils.path_utils import get_project_root
```
paul-nechifor: Please don't use inline imports unless they're to avoid circular imports.

dimos/manipulation/manipulation_module.py
```diff
@@ -598,6 +611,52 @@ def get_trajectory_status(
...
+
+        # Ensure GraspGen model checkpoints are pulled from LFS
+        get_data("models_graspgen")
+
+        project_root = get_project_root()
+        docker_file = (
+            project_root / "dimos" / "manipulation" / "grasping" / "docker_context" / "Dockerfile"
+        )
+        self._graspgen = DockerModule(
```
paul-nechifor: I thought it was decided to rename this because `DockerModule` implies it's a `dimos.core.module.Module`.

dimos/manipulation/manipulation_module.py
```diff
@@ -598,6 +611,52 @@ def get_trajectory_status(
         status = client.get_trajectory_status(config.coordinator_task_name)
         return dict(status) if status else None
 
+    def _get_graspgen(self) -> DockerModule | None:
+        """Get or create GraspGen Docker module (lazy init)."""
+        if self._graspgen is not None:
+            return self._graspgen
```
paul-nechifor, line 710: You have to put this behind a lock because calls to it can come from different threads.

dimos/manipulation/manipulation_module.py
```diff
@@ -83,6 +85,16 @@ class ManipulationModuleConfig(ModuleConfig):
...
 
+    # GraspGen Docker settings (optional)
+    graspgen_docker_image: str = "dimos-graspgen:latest"
+    graspgen_gripper_type: str = "robotiq_2f_140"
+    graspgen_num_grasps: int = 400
+    graspgen_topk_num_grasps: int = 100
+    graspgen_grasp_threshold: float = -1.0
+    graspgen_filter_collisions: bool = False
+    graspgen_save_visualization_data: bool = False
+    graspgen_visualization_output_path: str = "/tmp/grasp_visualization.json"
```
paul-nechifor: This path is inside docker, right? It's not really a temp file. If it's meant to be accessed from the outside, it should be in a place where it's easy to access. Maybe `/var/lib/graspgen/visualization.json` so you can mount the dir?

dimos/manipulation/manipulation_module.py
```diff
@@ -596,6 +611,51 @@ def get_trajectory_status(
...
+            gripper_type=self.config.graspgen_gripper_type,
+            num_grasps=self.config.graspgen_num_grasps,
+            topk_num_grasps=self.config.graspgen_topk_num_grasps,
+            grasp_threshold=self.config.graspgen_grasp_threshold,
+            filter_collisions=self.config.graspgen_filter_collisions,
+            save_visualization_data=self.config.graspgen_save_visualization_data,
+            visualization_output_path=self.config.graspgen_visualization_output_path,
+        )
+        self._graspgen.start()
+        return self._graspgen
```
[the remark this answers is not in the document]
paul-nechifor: No.

dimos/manipulation/manipulation_module.py
JalajShuklaSS: Fixed, the container now writes to a fixed internal path ``/output/graspgen/visualization.json``, and when ``save_visualization_data `` is enabled, the host directory is auto-mounted so the file appears at ``~/.dimos/graspgen/visualization.json``. Went with ``~/.dimos/graspgen/ `` over ``/var/lib/graspgen/`` since ``/var/lib/`` requires root to create while ~``/.dimos/`` is always user-writable (same pattern as ``~/.docker/, ~/.kube/)``

Thanks

dimos/manipulation/manipulation_module.py
JalajShuklaSS: Thank you for this, I have fixed by implementing double-checked locking pattern in both `` _get_graspgen() ``and  ``stop()``.

The pattern uses threading .Lock context manager which guarantees lock release even on exception, preventing deadlock.

dimos/manipulation/manipulation_module.py
JalajShuklaSS: There is no circular import here, these were inline for lazy loading optimization (don't import Docker dependencies if ``GraspGen`` is never used). Though now have moved all four imports to the top of the file

dimos/manipulation/manipulation_module.py
JalajShuklaSS: Thank you and have that fixed that.

dimos/manipulation/manipulation_module.py
JalajShuklaSS: In the previous PR I followed the convention for the naming of the class ``(DockerModule)`` and the file name being docker_runner as the docker module works in that format, for now in order to avoid any confusion I have renamed it here as ``DockerRunner`` instead of ``DockerModule`` so if you wish to change the naming of the ``DockerModule`` class itself then let me know and can rename it from the core directory. Considering the consistency I have followed the naming consistency. Thank you.

paul-nechifor (approved):

mustafab0 (approved):
Looks Ok

PR #1235, opened 2026-02-11
Feature: nativemodule livox and fastlio2

leshy:
## NativeModule arch, Livox Mid-360 LiDAR + FAST-LIO2 Integration, no ros

I have to add screenshots look how cool this sensor is

## How do I review?

docs are at https://github.com/dimensionalOS/dimos/blob/3142b5bceb232222982d39d7960319f3d5ec3cf6/docs/usage/native_modules.md

I think NativeModule stuff is important, maybe the way I handle builds via nix, livox/fastlio is just a dumb glue that seems fine in terms of config and computational efficiency so wouldn't bother too much 

C++ voxel mapper is just a hashmap. shouldn't look at it, this is vibed and will go away once I write voxel deletion

### Summary

NativeModule framework, enables integration third party executables into the dimos module/blueprint system

Native C++ Dimos Module for the Livox mid360
Native C++ Dimos Module for FAST-LIO2

## NativeModule
(`dimos/core/native_module.py`)

should this be named externalmodule or something like that?

Hosts third party process, passes topics to use via CLI to the binary.
Passes also config settings as CLI args

Assumes binary speaks LCM, in the future we can pass also the protocol.
Doesn't support RPC

C++ binaries participate in blueprints like any Python module

## Livox Mid-360 driver 
`dimos/hardware/sensors/lidar/livox/`

- Direct Livox SDK2 C API integration (no ROS dependency)
- C++ binary publishes PointCloud2 and IMU at configurable rates over LCM
- Python wrapper: `Mid360Module` with network config (host/lidar IPs, ports)

```python
    pointcloud: Out[PointCloud2]
    imu: Out[Imu]
```

```python
mid360 = autoconnect(
    Mid360Module.blueprint(),
    rerun_bridge(),
).global_config(n_dask_workers=2, robot_model="mid360")
```

### Testing

```
cd dimos/hardware/sensors/lidar/livox/cpp
nix build .#mid360_native
dimos run mid360
```

for native build check cpp/readme.md

## FAST-LIO2
(`dimos/hardware/sensors/lidar/fastlio2/`)

- Custom non-ros FAST-LIO2 rewrite https://github.com/leshy/FAST-LIO-NON-ROS/tree/dimos-integration
- Point cloud filtering pipeline: range filter, voxel downsampling, statistical outlier removal
- Outputs: world-frame registered clouds, odometry with covariance, optional global voxel map

#### Fastlio binary has livox SDK baked in, talks to mid360 directly, LCM topics lose full spectrum of mid360 data which fastlio depends on.

raw mid360 data is so raw that I assume we'll always use it through fastlio. (for example lidar hits need compensation for rotation state of the scanner and motion of the lidar in space)

```python
    lidar: Out[PointCloud2]
    odometry: Out[Odometry]
    global_map: Out[PointCloud2]
```

```python
mid360_fastlio = autoconnect(
    FastLio2.blueprint()
    rerun_bridge(),
)
```

```sh
cd dimos/hardware/sensors/lidar/fastlio2/cpp
nix build .#fastlio2_native

# raw fastlio
dimos run mid360-fastlio

# original mapper
dimos run mid360-fastlio-voxels

# cpp mapper
dimos run mid360-fastlio-voxels-native
```

for native build check cpp/readme.md

## Voxel mapper tests
fastlio has three blueprint configs: 
basic ODOM + Pointcloud, ODOM + GlobalMap, ODOM + CPPGlobalMap,

CPP voxel mapper uses PCL library https://pointclouds.org/
and is 2x more efficient then already pretty efficient py/cuda mapper

### no voxel deleteons yet
needs a few days of experimenting

## Nix build infrastructure

- Hermetic builds for both `mid360_native` and `fastlio2_native` via Nix flakes
- Livox SDK2 compiled from source, FAST-LIO fetched from GitHub

## Blueprints

- `mid360` — raw LiDAR
- `mid360-fastlio` — SLAM with 0.15m voxel filtering
- `mid360-fastlio-voxels` — SLAM + downstream VoxelGridMapper
- `mid360-fastlio-voxels-native` — SLAM + C++ global voxel map at 3Hz


leshy:
@greptileai can you review

dimos/hardware/sensors/lidar/common/dimos_native_module.hpp
```diff
@@ -0,0 +1,64 @@
...
+    float arg_float(const std::string& key, float default_val = 0.0f) const {
+        auto it = args_.find(key);
+        return it != args_.end() ? std::stof(it->second) : default_val;
+    }
+
+    /// Get an int arg value, or a default if not present.
+    int arg_int(const std::string& key, int default_val = 0) const {
+        auto it = args_.find(key);
+        return it != args_.end() ? std::stoi(it->second) : default_val;
+    }
```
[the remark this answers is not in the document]
leshy: yeah cpp could use a proper cmdline arg parser, can check this stuff later, would be nice if we had a cpp person with their favorite libs for this etc

dimos/core/tests/native_echo.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python3
+# Copyright 2026 Dimensional Inc.
```
leshy, line 2: this is a test file for "external executable" for native modules.. but placement is super dumb, where do I put it?

jeff-hykin: Given this is a test fixture I'd put it under a test dir inside core `core/tests/`. We've been lucky to not need too many fixtures (we can do tests inline) but that'll probably change the more cross-language support is added. 

Could you have the heredoc explain why this file exists, like why is this a good way to test external executables? In contrast to a one-liner bash script like
```sh
#/bin/bash
timeout ${NATIVE_ECHO_DIE_AFTER}s echo $# "$@" > $NATIVE_ECHO_OUTPUT
```

dimos/core/native_module.py
```diff
@@ -0,0 +1,246 @@
...
+            logger.info("Stopping native process", pid=self._process.pid)
+            self._process.send_signal(signal.SIGTERM)
+            try:
+                self._process.wait(timeout=self.config.shutdown_timeout)
+            except subprocess.TimeoutExpired:
+                logger.warning(
+                    "Native process did not exit, sending SIGKILL", pid=self._process.pid
+                )
+                self._process.kill()
+                self._process.wait(timeout=5)
```
jeff-hykin, line 176: The timeout feels a bit weird being hardcoded (some systems are slower than others) but I agree it doesnt really feel like a config value. Maybe having it as a private class parameter would be good.

dimos/hardware/sensors/lidar/fastlio2/cpp/flake.nix
```diff
@@ -0,0 +1,59 @@
+{
```
jeff-hykin, line 1: This looks great. I'll test on macos. 

In some distant future when we have a dimos package repository, I would want to make it a requirement that published DimOS C++ modules have a flake.nix. We could have that rule internally.

dimos/hardware/sensors/lidar/fastlio2/cpp/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Build
+
+### Nix (recommended)
+
+```bash
+cd dimos/hardware/sensors/lidar/fastlio2/cpp
+nix build .#fastlio2_native
+```
+
```
jeff-hykin, line 16: I know I've pushed towards dimos users to not being forced to use nix, but for C++ and similar individual modules I think its fair to say "Either use the prebuilt binary, use nix, or figure it out yourself". E.g. for future C++ modules I don't think we need to support the native install instructions.

Why I feel these are okay to force is cause its for a build of an exe instead of a devShell.

jeff-hykin (dismissed):
I think the only thing missing is docs (all my other feedback is super minor). 

I can ask Jalaj to do the docs for this (after its merged) cause Jala already needs to add docs for his docker module, and this native module is way more simple. 

That said, in a future I think it would be good for native modules to support automated build hook for a missing binary.

dimos/hardware/sensors/lidar/fastlio2/cpp/README.md
leshy: yeah I agree, maintaining build systems would be horrible otherwise. nix can be used for wheel builds also

dimos/core/native_module.py
leshy: SIGKILL is non optional, this is an operating system level termination, that's why only SIGTERM timeout is configurable

jeff-hykin: I meant more why have a timeout on the wait? Why not set it to 0? Probably cause it takes some time. How much time? Well that probably depends if its an under clocked raspberry pi or Pim's supercomputer. 

Is 5 a sane number to cover all cases? Probably. But then why are the t.join's hardcoded to 2 sec instead of 5? Idk. Feels kinda weird, but not a big deal.

leshy:
> I think the only thing missing is docs (all my other feedback is super minor).
> 
> I can ask Jalaj to do the docs for this (after its merged) cause Jala already needs to add docs for his docker module, and this native module is way more simple.
> 
> That said, in a future I think it would be good for native modules to support automated build hook for a missing binary.

true, this can host docker very easily.. wrote some docs, and yeah autobuild sounds good, just wanted to keep the PR smaller, pretty big already

dimos/hardware/sensors/lidar/fastlio2/cpp/README.md
leshy: actually I added a build system it was just a few lines and makes it much more ergonomic

dimos/core/tests/native_echo.py
leshy: moved the file

reason why this isn't bash is that wasn't sure how deeply I wanted to test this when I wrote it, also sigterm handling (exit 0) some small things missing from bash, it is replacable by short script in it's current state but would leave py just in case we want to test more as this system grows

dimos/core/native_module.py
leshy: yeah I think it's a balance between configurability and ergonomics, this config is my people facing interface, so the fact that a person would need to read and think about this value is not worth it - given likely no one ever actually needs to read and think about this.

this value is orders of magnitude above what's expected by normal system operation even on weakest machines, so if it takes more then 5 seconds to SIGKILL - don't worry about that part, your kernel isn't working and you are unplugging the power anyway, you are not thinking of reconfiguring native_module.py

if above assumption proves wrong can add or change

docs/usage/native_modules.md
```diff
@@ -0,0 +1,282 @@
...
+    executable: str = "./build/my_module" # relative or absolute path to your executable
+    host_ip: str = "192.168.1.5"     # becomes --host_ip 192.168.1.5
+    frequency: float = 10.0           # becomes --frequency 10.0
+    enable_imu: bool = True           # becomes --enable_imu true
+    filters: list[str] = field(default_factory=lambda: ["a", "b"])  # becomes --filters a,b
+```
+
+- `None` values are skipped.
+- Booleans are lowercased (`true`/`false`).
+- Lists are comma-joined.
```
paul-nechifor, line 116: It's probably safer to pass it as JSON.

dimos/spec/utils.py
```diff
@@ -99,6 +100,47 @@ def foo(self) -> int: ...
...
+        assert_implements_protocol(Good, MyProto)  # passes
+    """
+    proto_hints = typing.get_type_hints(protocol, include_extras=True)
+    cls_hints = _own_hints(cls)
+
+    for name, expected_type in proto_hints.items():
+        assert name in cls_hints, f"{cls.__name__} missing '{name}' required by {protocol.__name__}"
+        assert cls_hints[name] == expected_type, (
+            f"{cls.__name__}.{name}: expected {expected_type}, got {cls_hints[name]}"
+        )
```
paul-nechifor: This shouldn't be here because it's not related to RPC `Spec`s. But also because it's intended for tests and we shouldn't be shipping such code.

(Looks like I forgot to ignore `test_` files when building the code. I've added an issue to do that: https://linear.app/dimensional/issue/DIM-539/do-not-ship-test-code )

docs/usage/native_modules.md
leshy: Yeah or ENV actually, given this would make it compatible with Docker modules, need to coordinate this a bit, IMO this is a POC, wanted to create an issue and discuss

dimos/core/tests/native_echo.py
paul-nechifor: Since it is a fixture, not a test, I'd put it in `dimos/core/fixtures/native_echo.py`.

dimos/core/native_module.py
```diff
@@ -0,0 +1,297 @@
...
+        if proc.returncode != 0:
+            raise RuntimeError(
+                f"Build command failed (exit {proc.returncode}): {self.config.build_command}"
+            )
+        if not exe.exists():
+            raise FileNotFoundError(
+                f"Build command succeeded but executable still not found: {exe}"
+            )
+
+    def _collect_topics(self) -> dict[str, str]:
```
paul-nechifor: This doesn't quite work because the names could have been renamed in the blueprints. I think the names must come from the blueprints.

dimos/core/native_module.py
```diff
@@ -0,0 +1,297 @@
...
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+        logger.info("Native process started", pid=self._process.pid)
+
+        self._stopping = False
+        self._io_threads = [
+            self._start_reader(self._process.stdout, "info"),
+            self._start_reader(self._process.stderr, "warning"),
+        ]
```
paul-nechifor: It's a bit wasteful to start two extra threads just to pipe two streams. You can do everything from the watchdog thread.

dimos/core/test_native_module.py
```diff
@@ -0,0 +1,180 @@
...
+def args_file():
+    """Temp file where native_echo.py writes the CLI args it received."""
+    fd, path = tempfile.mkstemp(suffix=".json", prefix="native_echo_")
+    os.close(fd)
+    os.unlink(path)
+    try:
+        yield path
+    finally:
+        if os.path.exists(path):
+            os.unlink(path)
```
paul-nechifor: You don't need this. The `tmp_path` pytest fixture gives you a dir and cleans it up.

paul-nechifor (dismissed):

dimos/spec/utils.py
leshy: it's also a stupid way to check protocol compliance, fixed, mypy covers this

docs/usage/native_modules.md
leshy: https://github.com/dimensionalOS/dimos/issues/1266

dimos/core/native_module.py
leshy: I've put management of these threads in the watchdog, but the code gets really ugly if I try to process everything in the same thread, so keeping split for ease of undertanding now

dimos/core/test_native_module.py
leshy: yeah removed this

paul-nechifor (dismissed):

paul-nechifor (approved):

PR #1237, opened 2026-02-11
Feature: Basic Agentic Pick and Place - reimplemented manipulation skills

mustafab0:
## Problem

ManipulationModule had no skill layer — only low-level RPCs requiring a custom IPython client.
No way for an LLM agent to invoke pick/place/move skills through the framework.
RRT planner rejected valid joint states due to floating-point drift past URDF limits.

---

## Solution

Added 11 `@skill()` methods to ManipulationModule (pick, place, move_to_pose, scan_objects, etc.).
Changed base class to `SkillModule` so skills auto-register with agents via `autoconnect()`.
Created `xarm-perception-agent` blueprint composing xarm_perception + llm_agent + human_input.
Added detection snapshot so pick uses stable labels instead of volatile live cache.
Added `limit_eps=1e-3` tolerance to RRT planner joint limit validation.
Removed ManipulationClient and `run_*` RPC wrappers — agent CLI replaces them.
CoordinatorClient updated to route execution through `task_invoke` instead of removed RPCs.

---

## Breaking Changes

`ManipulationClient` deleted — use `dimos run xarm-perception-agent` or direct RPC.
`run_pick`/`run_place`/`run_place_back`/`run_go_init` RPCs removed from ManipulationModule.

---

## How to Test

1. Run `dimos run coordinator-mock` then `dimos run xarm-perception-agent`
2. Type `scan the scene` — verify objects listed with 3D positions
3. Type `pick up the <object>` — verify approach, grasp, retract sequence
4. Type `place it back` — verify placement at original pick position
5. Verify skills appear in `dimos agentspy`

closes DIM-351
closes DIM-419

### DIFF
--- a/dimos/manipulation/manipulation_module.py
+++ b/dimos/manipulation/manipulation_module.py
@@ -30,6 +30,7 @@ import threading
 import time
 from typing import TYPE_CHECKING, Any, TypeAlias
 
+from dimos.agents.annotation import skill
 from dimos.constants import DIMOS_PROJECT_ROOT
 from dimos.core import In, Module, rpc
 from dimos.core.docker_runner import DockerModule as DockerRunner


dimos/manipulation/control/coordinator_client.py
```diff
@@ -121,20 +121,25 @@ def get_joint_positions(self) -> dict[str, float]:
         return self._rpc.get_joint_positions() or {}
 
     def get_trajectory_status(self, task_name: str) -> dict[str, Any]:
-        """Get status of a trajectory task."""
-        return self._rpc.get_trajectory_status(task_name) or {}
+        """Get status of a trajectory task via task_invoke."""
+        result = self._rpc.task_invoke(task_name, "get_status", {})
+        if isinstance(result, dict):
+            return result
+        return {}
```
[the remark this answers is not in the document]
mustafab0: fixed type

dimos/manipulation/manipulation_module.py
```diff
@@ -12,17 +12,26 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Manipulation Module - Motion planning with ControlCoordinator execution."""
+"""Manipulation Module - Motion planning with ControlCoordinator execution.
+
+Skills are organized in tiers:
+- Tier 1 (@rpc): Low-level building blocks (plan_to_pose, execute, etc.)
+- Tier 2 (@skill): Individually testable actions (move_to_pose, open_gripper, etc.)
+- Tier 3 (@skill): Composed behaviors (pick, place, pick_and_place)
```
spomichter: Semantially not sure the difference between how these are explained. both can be called by agents. Maybe better to clarify one long horizon and one short horizon "functions" given that there will rarely be failure during an open gripper action, but more potential failure for a full pick and place

dimos/manipulation/manipulation_module.py
```diff
@@ -99,6 +117,11 @@ class ManipulationModule(Module):
     # Input: Objects from perception (for obstacle integration)
     objects: In[list[DetObject]]
 
+    # RPC calls for GraspGen integration (resolved at runtime if modules are deployed)
+    rpc_calls: list[str] = [
```
spomichter: This should be deprecated with @jeff-hykin 's RPC stuff but will let it pass for now & in interest of release timeline

dimos/manipulation/manipulation_module.py
```diff
@@ -122,6 +145,18 @@ def __init__(self, *args: object, **kwargs: object) -> None:
         # Coordinator integration (lazy initialized)
         self._coordinator_client: RPCClient | None = None
 
+        # Init joints: captured from first joint state received, used by go_init
+        self._init_joints: list[float] | None = None
+        self._init_joints_captured = False
+
+        # Last pick position: stored during pick so place_back() can return the object
+        self._last_pick_position: tuple[float, float, float] | None = None
```
spomichter: Any reason these aren't typed. Makes development harder without guaranteed types / typechecking

dimos/manipulation/manipulation_module.py
```diff
@@ -122,6 +145,18 @@ def __init__(self, *args: object, **kwargs: object) -> None:
...
+
+        # Snapshotted detections from the last scan_objects/refresh call.
+        # The live detection cache is volatile (labels change every frame),
+        # so pick/place use this stable snapshot instead.
+        self._detection_snapshot: list[dict[str, object]] = []
```
spomichter: here also idk what object means

dimos/manipulation/manipulation_module.py
```diff
@@ -819,23 +906,664 @@ def get_gripper(self, robot_name: RobotName | None = None) -> float | None:
...
+
+    def _make_pose(
+        self, x: float, y: float, z: float, roll: float, pitch: float, yaw: float
+    ) -> Pose:
+        """Construct a Pose from position and Euler angles (radians)."""
+        from dimos.msgs.geometry_msgs import Pose, Vector3
+        from dimos.utils.transform_utils import euler_to_quaternion
+
+        orientation = euler_to_quaternion(Vector3(roll, pitch, yaw))
+        return Pose(position=Vector3(x, y, z), orientation=orientation)
```
spomichter: Why are we doing this

dimos/manipulation/manipulation_module.py
```diff
@@ -819,23 +906,664 @@ def get_gripper(self, robot_name: RobotName | None = None) -> float | None:
...
+        rname, _, config, _ = robot
+        client = self._get_coordinator_client()
+
+        if client is None or not config.coordinator_task_name:
+            # No coordinator — wait for trajectory duration as fallback
+            traj = self._planned_trajectories.get(rname)
+            if traj is not None:
+                logger.info(f"No coordinator status — waiting {traj.duration:.1f}s for trajectory")
+                time.sleep(traj.duration + 0.5)
+            return True
```
spomichter, line 1054: This should be handled by @paul-nechifor skill coordinator stuff. We should have a universal and Generic set of "Skills" or "Processes" within dimOS and all governed by the same coordinator and manager. Fine for now since not done yet. 

1. dimOS skills are functions that agents can call
2. dimOS processes are short-mid-long running tasks typically triggered by skill calls (or directly)

dimos/manipulation/manipulation_module.py
```diff
@@ -819,23 +906,664 @@ def get_gripper(self, robot_name: RobotName | None = None) -> float | None:
...
+        for det in detections:
+            if object_id and str(det.get("object_id", "")) == str(object_id):
+                return dict(det)
+            name = str(det.get("name", "")).lower()
+            if object_name.lower() in name or name in object_name.lower():
+                return dict(det)
+
+        available = [str(d.get("name", "?")) for d in detections]
+        logger.warning(f"Object '{object_name}' not found in snapshot. Available: {available}")
+        return None
```
spomichter, line 1151: This is pretty horrible and caused by the lack of type i flagged above on this detections dict list thing. We have a detection type that you would be able to just do this in one line. No need for dict because 2DDetection has a field `string` so you can for any detection check `detection.string` 

```
detections_list = List[2DDetection]
for det in detections_list:
     ...
``` 

And even better we just define a type `2DDetectionList` of `2DDetectionDB` which has search and `__iter__` methods. 

Only reason this may be acceptable to PR is that the above is undergoing a refactor from @leshy so idk the current state.

dimos/manipulation/manipulation_module.py
```diff
@@ -819,23 +906,664 @@ def get_gripper(self, robot_name: RobotName | None = None) -> float | None:
...
+            yield "Error: No home_joints configured for this robot"
+            return
+
+        yield "Opening gripper..."
+        self._set_gripper_position(0.85, robot_name)
+        time.sleep(0.5)
+
+        yield "Planning motion to home position..."
+        if not self.plan_to_joints(config.home_joints, robot_name):
+            yield "Error: Failed to plan path to home position"
```
spomichter: Yeah i assume @paul-nechifor 's stuff will solve for this generator / yielding requirement right now so fine for now

dimos/manipulation/manipulation_module.py
```diff
@@ -819,23 +906,664 @@ def get_gripper(self, robot_name: RobotName | None = None) -> float | None:
...
+        robot_name: str | None = None,
+    ) -> Generator[str, None, None]:
+        """Pick up an object by name using grasp planning and motion execution.
+
+        Scans the scene, generates grasp poses (via GraspGen or heuristic fallback),
+        plans collision-free approach/grasp/retract motions, and executes them.
+
+        Args:
+            object_name: Name of the object to pick (e.g. "cup", "bottle", "can").
+            object_id: Optional unique object ID from perception for precise identification.
```
spomichter, line 1412: @paul-nechifor Flagging for you since skills like this are helpful to see what we're solving for on agents

dimos/manipulation/manipulation_module.py
mustafab0: Makes sense. Short -  Long Horizon is a better classification.

The distinction here was just for readability and code organization.

dimos/manipulation/manipulation_module.py
mustafab0: Coordinator client interfaces with hardware much loser. I will add proper type checking to the ControlCoordinator API

dimos/manipulation/manipulation_module.py
mustafab0: This should have been an Object. A confusion with object vs Object. will fix.

dimos/manipulation/manipulation_module.py
paul-nechifor: @jeff-hykin's RPC stuff is merged. There's no reason to use the old stuff.

dimos/manipulation/manipulation_module.py
paul-nechifor: There's an issue open to rename all of them https://linear.app/dimensional/issue/DIM-452/rename-object-to-something-else-entity-body-etc

dimos/manipulation/manipulation_module.py
paul-nechifor: The plan is (discussed with @leshy ) to expose everything as an MCP server. So the tool coordination will happen through the MCP client. The MCP client can be claude code, but we'll also write our own client.

The tools/resources (replacing the skill terminology) will be defined the way MCP expects them. But we can have helper classes for running long running tools in our code.

Multiple async tool calls can happen through MCP and it supports updates from the server to the client.

Long running tools calls will push a stream of updates, like generator skills. This is a little more complex in MCP, but we'll have helper code to do this.

dimos/manipulation/manipulation_module.py
```diff
@@ -819,23 +906,673 @@ def get_gripper(self, robot_name: RobotName | None = None) -> float | None:
...
+            else:
+                # GraspGen succeeded — get poses from the grasps port or RPC
+                logger.info(f"GraspGen result: {result_str}")
+                # Try to get the grasp poses via RPC
+                try:
+                    get_grasps = self.get_rpc_calls("GraspingModule.get_latest_grasps")
+                    grasp_poses: PoseArray | None = get_grasps()
+                    if grasp_poses and len(grasp_poses.poses) > 0:
+                        return list(grasp_poses.poses)
+                except Exception:
```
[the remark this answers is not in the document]
mustafab0: The dead get_latest_grasps call and the PoseArray import are gone.

dimos/manipulation/manipulation_module.py
mustafab0: I will address it in this PR itself. @paul-nechifor 👍

dimos/manipulation/manipulation_module.py
mustafab0: Fixed, replaced list with standard DimOS types i.e Vector3 and JointState

dimos/manipulation/manipulation_module.py
mustafab0: Changed _detection_snapshot from list[dict[str, object]] to list[DetObject] — stores actual Object instances instead of serialized dicts

pyproject.toml
JalajShuklaSS, line 179: Should you add something like this ``pyrealsense2>=2.50.0`` dependency here or in perception part?

dimos/manipulation/manipulation_module.py
mustafab0: just a helper. removed as unnecesary part of legacy code before type checking.

dimos/manipulation/manipulation_module.py
mustafab0: replaced object data type with Any to avoid confusion in the interim.

dimos/manipulation/manipulation_module.py
```diff
@@ -583,8 +626,46 @@ def get_robot_info(self, robot_name: RobotName | None = None) -> dict[str, objec
             "max_acceleration": config.max_acceleration,
             "has_joint_name_mapping": bool(config.joint_name_mapping),
             "coordinator_task_name": config.coordinator_task_name,
+            "home_joints": config.home_joints,
+            "pre_grasp_offset": config.pre_grasp_offset,
+            "init_joints": self._init_joints,
         }
 
+    @rpc
```
JalajShuklaSS, line 647: For these three RPC methods, I guess its speculative API design its **not a bug** but wanted clarity on the followup: 

The pipeline the actually runs is:
Robot boots -> ``_on_joint_state fires`` -> ``self._init_joints`` auto-captured
here this now proceeds further and, 
LLM agent calls ``go_init`` skill -> reads ``self._init_joints`` directly -> plans & moves

No RPC call involved, ``go_init`` reads the field directly on ``self`` , not through ``get_init_joints()``

The only scenario would be an external script using RPCClient, something like this:
``client = RPCClient("manipulation")`` 
``client.set_init_joints([0.0, -0.5, 0.0, 0.8, 0.0, 1.0, 0.0])  # override``
``client.go_init()  # now goes to custom pose instead of startup pose``

But that pattern doesn't exist anywhere in the codebase. No script, no blueprint, no test, no agent does this.

dimos/manipulation/manipulation_module.py
mustafab0: removed the rpc call here, since it is dead code as graspgen is not integrated/implemented yet.

dimos/manipulation/manipulation_module.py
```diff
@@ -819,23 +906,664 @@ def get_gripper(self, robot_name: RobotName | None = None) -> float | None:
...
-    def close_gripper(self, robot_name: RobotName | None = None) -> bool:
-        """Close gripper fully.
+    @skill()
+    def open_gripper(self, robot_name: str | None = None) -> str:
+        """Open the robot gripper fully.
+
+        Args:
+            robot_name: Robot to control (only needed for multi-arm setups).
+        """
+        if self._set_gripper_position(0.85, robot_name):
```
JalajShuklaSS, line 1007: What I feel should exist considering modularity but it does not exist for now is that ``RobotModelConfig`` has ``gripper_hardware_id`` but has no ``gripper_open_position`` or ``gripper_close_position fields``. The adapter knows its own range (Piper has ``GRIPPER_MAX_OPENING_M``), but that information never flows up to ``ManipulationModule``.

So I feel this can be solved by two config fields to ``RobotModelConfig`` and the blueprint sets them per robot.
Or
You can normalize it.

Thank you.

dimos/manipulation/manipulation_module.py
```diff
@@ -819,23 +906,664 @@ def get_gripper(self, robot_name: RobotName | None = None) -> float | None:
...
+                    if state_val in (0, 2):  # IDLE or COMPLETED
+                        return True
+                    if state_val in (3, 4):  # ABORTED or FAULT
+                        logger.warning(f"Trajectory failed: state={state}")
+                        return False
+                    # state_val == 1 means EXECUTING, keep polling
+                else:
+                    # task_invoke returned None — task not found, assume done
+                    return True
+            except Exception:
```
JalajShuklaSS, line 1073: Is there a reason we are not catching the exception here, because if the ``task_invoke`` throws (e.g. coordinator crash), it accepts it and then returns True, so the skill thinks the trajectory succeeded and continues to the next step even though the arm may not have moved.

dimos/manipulation/manipulation_module.py
```diff
@@ -819,23 +906,664 @@ def get_gripper(self, robot_name: RobotName | None = None) -> float | None:
...
+
+    # =========================================================================
+    # Skill Helpers (internal)
+    # =========================================================================
+
+    def _make_pose(
+        self, x: float, y: float, z: float, roll: float, pitch: float, yaw: float
+    ) -> Pose:
+        """Construct a Pose from position and Euler angles (radians)."""
+        from dimos.msgs.geometry_msgs import Pose, Vector3
```
JalajShuklaSS: Just a suggestion that is it possible to put these imports with ``JointState`` and above as the ``geometry_msg`` is a lightweight msg module already imported under ``TYPE_CHECKING`` so there is no circular imports or heavy import costs.
Thank you.

dimos/manipulation/manipulation_module.py
```diff
@@ -819,23 +906,664 @@ def get_gripper(self, robot_name: RobotName | None = None) -> float | None:
...
+        if GraspGen is unavailable.
+
+        Args:
+            object_name: Name of the object
+            object_id: Optional object ID
+
+        Returns:
+            List of grasp poses (best first), or None if object not found
+        """
+        # Try GraspGen via rpc_calls if available
```
JalajShuklaSS: Here two grasping methods are integrated, however the GraspGen API is a bit misaligned, the ``GraspGenModule``, ``Docker context``, and ``generate_grasps(PointCloud2)`` RPC are all already available in ``dimos/manipulation/grasping/``  and following the same integration pattern for heuristic grasp for graspgen would get it up, can refer to ``demo_grasping.py`` to wire the API up, and keep the existing heuristic as the fallback.

dimos/manipulation/manipulation_module.py
```diff
@@ -819,23 +906,664 @@ def get_gripper(self, robot_name: RobotName | None = None) -> float | None:
...
+        pre_grasp_offset = config.pre_grasp_offset
+
+        # 1. Generate grasps (uses already-cached detections — call scan_objects first)
+        yield f"Generating grasp poses for '{object_name}'..."
+        grasp_poses = self._generate_grasps_for_pick(object_name, object_id)
+        if not grasp_poses:
+            yield f"Error: No grasp poses found for '{object_name}'. Object may not be detected."
+            return
+
+        # 3. Try each grasp candidate
```
JalajShuklaSS: Not serious but still mentioning, comment miss match, 1...3...4

dimos/manipulation/manipulation_module.py
```diff
@@ -819,23 +906,664 @@ def get_gripper(self, robot_name: RobotName | None = None) -> float | None:
...
+        max_attempts = min(len(grasp_poses), 5)
```
JalajShuklaSS, line 1428: Here I feel that the hard coded 5 caps how many ``GraspGen`` poses to attempt should be a config sourced from the ``graspgen`` config itself. ``GraspGen`` ranks grasps by grasp quality alone, it has no knowledge of the robot's current configuration, so many top-ranked grasps may be IK-infeasible or require awkward joint configurations. Instead of sequentially trying candidates (IK -> RRT -> fail -> next), a better approach I feel is to solve IK for the top N grasps in parallel (Drake IK is stateless and thread-safe), filter to reachable ones, then RRT-plan only the best feasible candidate. This avoids wasting time on unreachable grasps and lets us consider far more candidates (e.g. 100 instead of 5).
Again **not a bug** and can even extend the range from 5 to 20 or so as a quick fix but a simple suggestion.
Thank you.

dimos/manipulation/manipulation_module.py
```diff
@@ -819,23 +906,664 @@ def get_gripper(self, robot_name: RobotName | None = None) -> float | None:
...
+            object_id: Optional unique object ID from perception.
+            robot_name: Robot to use (only needed for multi-arm setups).
         """
-        return bool(self.set_gripper(0.0, robot_name))
+        # Pick phase
+        yield f"Starting pick and place: pick '{object_name}' → place at ({place_x:.3f}, {place_y:.3f}, {place_z:.3f})"
+        for msg in self.pick(object_name, object_id, robot_name):
+            yield msg
+            # Check if pick failed
+            if msg.startswith("Error:"):
```
JalajShuklaSS: Its just my personal opinion that I do not like string matching failure catch, in my opinion string-based error detection is fragile, if someone adds a new failure path like "Failure: ..." without the "Error:..." prefix, but as the code stands right now, it is correctly mapping the returns Error prefix, so functionally correct. Small concern.
Thank you.

dimos/manipulation/manipulation_blueprints.py
```diff
@@ -232,6 +232,8 @@ def _make_xarm7_config(
         coordinator_task_name=coordinator_task,
         gripper_hardware_id=gripper_hardware_id,
         tf_extra_links=tf_extra_links or [],
+        # Home configuration: arm extended forward, elbow up (safe observe pose)
+        home_joints=[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
```
JalajShuklaSS, line 236: Hey, I think we need to add the ``home_joints`` to the other config also for xarm6 and piper. If this addition is intentional only for xarm7 then its fine.
Thank you.

dimos/manipulation/manipulation_module.py
```diff
@@ -819,23 +898,627 @@ def get_gripper(self, robot_name: RobotName | None = None) -> float | None:
...
+        if url:
+            lines.append(f"Visualization: {url}")
+
+        # State
+        lines.append(f"State: {self.get_state()}")
+
+        return "\n".join(lines)
+
+    @skill()
+    def scan_objects(self, min_duration: float = 1.0, robot_name: str | None = None) -> str:
```
JalajShuklaSS, line 1311: The ``robot_name`` is does not seem to be used in the method body. Is it there for API consistency? or does it serve a different purpose? Just need some clarification.

dimos/manipulation/manipulation_module.py
```diff
@@ -819,23 +898,627 @@ def get_gripper(self, robot_name: RobotName | None = None) -> float | None:
...
+        """Scan the scene and list detected objects with their 3D positions.
+
+        Refreshes perception obstacles from the latest sensor data and returns
+        a formatted list of all detected objects.
+
+        Args:
+            min_duration: Minimum time in seconds to wait for stable detections.
+            robot_name: Robot context (only needed for multi-arm setups).
+        """
+        obstacles = self.refresh_obstacles(min_duration)
```
JalajShuklaSS, line 1321: Here I just wanted to understand that ``refresh_obstacles`` snapshots ``detections`` into ``self._detection_snapshot`` (stable, frozen at that moment). But then ``scan_objects`` doesn't use ``self._detection_snapshot``,  it calls ``self.list_cached_detections()`` which reads the live volatile cache from ``WorldMonitor`` also I am aware that 99% of the time these return the same data since they are called milliseconds apart. But the whole point of `` _detection_snapshot `` (as the comment says: "the live cache is volatile") was to avoid reading from the live cache. ``scan_objects`` ignores that and reads the live cache anyway, so again just a concern about the approach.

dimos/manipulation/control/coordinator_client.py
JalajShuklaSS, line 327: Here I presume that the``wait_for_completion()`` is broken after the ``task_invoke`` migration. ``get_trajectory_status`` now returns ``{"state": int, "task": str}``, but this function still reads active, progress, and here it expects a state as a string and then it exits on first poll reporting failure. I feel that it can be resolved by either update it to match the new return format or remove it and the ``ManipulationModule._wait_for_trajectory_completion()`` already handles this correctly.

Thank you.

dimos/manipulation/manipulation_module.py
mustafab0: Yes this is correct.

dimos/manipulation/manipulation_module.py
mustafab0: This is  a fair point. we can note this for yaml config files.

For now normalization is something we should start with

dimos/manipulation/manipulation_module.py
mustafab0: You are right. 

For more specific tasks we need better and robust error handling. Since this is the first pass at the module, i added exception, as i am testing with specific tasks.

dimos/manipulation/manipulation_module.py
mustafab0: removed the graspgen api has that was a bit presumptuous

dimos/manipulation/manipulation_module.py
mustafab0: Makes sense, will make it a config parameter with graspgen integration.

dimos/manipulation/manipulation_module.py
mustafab0: Agreed, I think the more robust behaviour tree will be useful here

dimos/manipulation/manipulation_blueprints.py
mustafab0: Just an example now. but will generalize these :+1:

dimos/manipulation/manipulation_module.py
mustafab0: For API consistency. Since the example only supports one robot we aren't using the robot id but would be useful for multi robot planning

dimos/manipulation/manipulation_module.py
mustafab0: This has been addressed thanks for checking it too.

dimos/manipulation/control/coordinator_client.py
mustafab0: fixed :+1:

dimos/manipulation/manipulation_blueprints.py
mustafab0: added

paul-nechifor (dismissed):

spomichter (approved):

spomichter (approved):

PR #1243, opened 2026-02-12
Jeff/jetson/1

spomichter:


bin/test
```diff
@@ -0,0 +1,384 @@
+#!/usr/bin/env -S deno run --allow-all --no-lock
```
spomichter, line 1: what is this file assume delete

PR #1229, opened 2026-02-10
Add ARM Support

jeff-hykin:
test on an Arm system with:
```
uv pip install "dimos[unitree] @ git+https://github.com/dimensionalOS/dimos.git@jeff/arm_support"
dimos --replay run unitree-go2
```
Just uses the stop-gap pypi package published at: https://github.com/jeff-hykin/open3d_jetson
Which is just pre-build wheel files for arm based on the open3d repo

closes DIM-405

PR #1244, opened 2026-02-12
Jeff/jetson/1 rebased

spomichter:


pyproject.toml
```diff
@@ -206,11 +209,23 @@ cpu = [
...
+    # Detic GPU stack
+    "mss",
+    "dataclasses",
+    "ftfy",
+    "regex",
+    "fasttext",
+    "lvis",
+    "nltk",
+    "clip",
+    "detectron2",
```
spomichter, line 228: @jeff-hykin why is dectron added back here

jeff-hykin: Cause I have an actual "Add Arm Support" PR #1229 is the latest and up to date with dev

This branch was never suppost to be a PR and its stale, it has junk in it and was just so you could do a quick test

PR #1246, opened 2026-02-12
Feat: Arm teleop with Pinocchio - single & dual arm

### LINKED ISSUE #1186: Refactor Teleop to match new control and task architecture
Use current joint servoing implementation

ruthwikdasyam:
## Feature
- Arm teleop support for Quest VR stack - any arm on DimOS can be teleoperated with Quest controllers
- Blueprints added to test with Piper, XArm6, or both (dual-arm)

## Problems
- Safety Clamping is not a good idea, It might give out joint_positions that aren't IK valid. rejecting is better.
- pinnocchio IK must be defined seperately in dimos/manipulation/. Can be used for any IK computation

### Issue - Linear
closes DIM-446
closes DIM-421

closes #1186 

## Approach/Solution

1. **`TeleopIKTask`** - new control task
   Accepts streaming cartesian delta poses from VR controllers, solves IK internally via Pinocchio at 100Hz.
   - Captures EE pose on engage; incoming poses applied as deltas from that reference
   - `on_buttons()` listens to `QuestButtons` - hold primary to track, release to stop
   - Safety: rejects solutions exceeding `max_joint_delta_deg` per tick, auto-disengages on stream timeout
   - Per-hand config (`"left"` / `"right"` / both) for dual-arm setups
2. **`PinocchioIK`** - new standalone IK solver
   Lightweight Pinocchio-based solver for real-time control (unlike `JacobianIK` which requires full `WorldSpec`). Provides FK, damped least-squares IK, pose conversion helpers, and safety utilities (`check_joint_delta`, `get_worst_joint_delta`).
3. **`CartesianIKTask`** refactored to use `PinocchioIK` - replaced inline IK/safety/pose code
4. **`ControlCoordinator`** - new `buttons: In[QuestButtons]` stream, `_on_buttons()` handler, `teleop_ik` task type in factory, `hand` field on `TaskConfig`
5. **Teleop coordinator blueprints** - `coordinator_teleop_xarm6`, `coordinator_teleop_piper`, `coordinator_teleop_dual`
6. **End-to-end blueprints** - `arm_teleop_xarm6`, `arm_teleop_piper`, `arm_teleop_dual`
7. **`ArmTeleopModule`** reworked - new `ArmTeleopConfig.task_names` for per-hand routing via `frame_id`, removed old toggle engage
8. **XArm6 URDF** added to LFS (Piper already existed)

## Breaking Changes/Info

- Renamed `CartesianIKTask.set_target_pose()` -> `on_cartesian_command()`
- Renamed `CartesianIKTask.set_target_pose_dict()` -> `on_cartesian_command_dict()`
- `CartesianIKTaskConfig` removed: `ik_max_iter`, `ik_eps`, `ik_damp`, `ik_dt`, `max_velocity`.
  `CartesianIKTask` internals `_solve_ik()`, `_safety_check()`, `_pose_to_se3()` moved to `PinocchioIK`
- `CartesianIKTask._pose_to_se3()` removed - Replaced by standalone `pose_to_se3()` in `dimos.manipulation.planning.kinematics.pinocchio_ik.` Import path changed.
- CartesianIKTask safety: clamp -> reject - Previously, large joint deltas were clamped (interpolated toward target). Now the entire solution is rejected (returns None). Arm freezes instead of moving partially when delta exceeds limit.
- All tasks inherit BaseControlTask instead of ControlTask

### Blueprint verification (required)
```bash
# Start the WebXR bridge (in a separate terminal)
./dimos/teleop/quest/web/teleop_server.ts

Step - 2
# Single XArm6 arm teleop
dimos run arm-teleop-xarm6

# Single Piper arm teleop
dimos run arm-teleop-piper

# Dual arm teleop (XArm6 + Piper)
dimos run arm-teleop-dual
```

### Manual verification
1. Start `teleop_server.ts` and launch a teleop blueprint above with hardware connected
2. Open Quest 3 headset, navigate to WebXR teleop page
3. Hold primary button (A on right / X on left) - arm should track controller deltas
4. Release button - arm should freeze at last position
5. Verify timeout: disconnect Quest mid-hold - arm should stop after 0.5s
6. Dual arm: verify each controller independently drives its assigned arm

### DIFF
--- a/dimos/utils/data.py
+++ b/dimos/utils/data.py
@@ -250,7 +250,18 @@ def get_data(name: str | Path) -> Path:
     if file_path.exists():
         return file_path
 
-    return _decompress_archive(_pull_lfs_archive(filename))
+    # extract archive root (first path component) and nested path
+    path_parts = Path(name).parts
+    archive_name = path_parts[0]
+    nested_path = Path(*path_parts[1:]) if len(path_parts) > 1 else None
+
+    # download and decompress the archive root
+    archive_path = _decompress_archive(_pull_lfs_archive(archive_name))
+
+    # return full path including nested components
+    if nested_path:
+        return archive_path / nested_path
+    return archive_path
 
 
 class LfsPath(type(Path())):  # type: ignore[misc]
@@ -323,15 +334,3 @@ class LfsPath(type(Path())):  # type: ignore[misc]
     def __rtruediv__(self, other: object) -> Path:
         """Reverse path division operator."""
         return other / self._ensure_downloaded()  # type: ignore[operator, return-value]
-    # extract archive root (first path component) and nested path
-    path_parts = Path(name).parts
-    archive_name = path_parts[0]
-    nested_path = Path(*path_parts[1:]) if len(path_parts) > 1 else None
-
-    # download and decompress the archive root
-    archive_path = _decompress_archive(_pull_lfs_archive(archive_name))
-
-    # return full path including nested components
-    if nested_path:
-        return archive_path / nested_path
-    return archive_path
--- a/dimos/utils/test_data.py
+++ b/dimos/utils/test_data.py
@@ -225,7 +225,7 @@ def test_lfs_path_with_real_file() -> None:
 def test_lfs_path_unload_and_reload() -> None:
     """Test unloading and reloading an LFS file."""
     filename = "three_paths.png"
-    data_dir = data._get_data_dir()
+    data_dir = data.get_data_dir()
     file_path = data_dir / filename
 
     # Clean up if file already exists


dimos/control/coordinator.py
```diff
@@ -464,11 +487,14 @@ def _on_cartesian_command(self, msg: PoseStamped) -> None:
                 logger.warning(f"Cartesian command for unknown task: {task_name}")
                 return
 
-            # Route to CartesianIKTask
-            if hasattr(task, "set_target_pose"):
-                task.set_target_pose(msg, t_now)  # type: ignore[attr-defined]
-            else:
-                logger.warning(f"Task {task_name} does not support set_target_pose")
+            task.on_cartesian_command(msg, t_now)  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
ruthwikdasyam: Fixed - Added error handling!

dimos/teleop/quest/quest_extensions.py
```diff
@@ -75,36 +79,54 @@ def _publish_msg(self, hand: Hand, output_msg: PoseStamped) -> None:
...
         - right_controller_output: PoseStamped (inherited)
         - buttons: QuestButtons (inherited)
     """
 
+    default_config = ArmTeleopConfig
+
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         super().__init__(*args, **kwargs)
-        self._prev_primary: dict[Hand, bool] = {Hand.LEFT: False, Hand.RIGHT: False}
+        cfg: ArmTeleopConfig = self.config  # type: ignore[assignment]
```
paul-nechifor: This ignore is not needed. You have to define `config: ArmTeleopConfig` after `default_config = ArmTeleopConfig`. Then remove this line.

dimos/manipulation/planning/kinematics/pinocchio_ik.py
```diff
@@ -0,0 +1,298 @@
...
+        rotation = pinocchio.Quaternion(quat.w, quat.x, quat.y, quat.z).toRotationMatrix()
+        return pinocchio.SE3(rotation, position)
+
+    @staticmethod
+    def pose_dict_to_se3(pose: dict[str, float]) -> pinocchio.SE3:
+        """Convert a pose dict with RPY to pinocchio SE3"""
+
+        position = np.array([pose["x"], pose["y"], pose["z"]])
+        rotation = pinocchio.rpy.rpyToMatrix(pose["roll"], pose["pitch"], pose["yaw"])
+        return pinocchio.SE3(rotation, position)
```
paul-nechifor: These should be standalone functions too. They don't use PinocchioIK.

dimos/control/coordinator.py
```diff
@@ -575,22 +604,30 @@ def start(self) -> None:
...
             except Exception as e:
                 logger.warning(f"Could not subscribe to cartesian_command: {e}")
 
+            # Also subscribe to buttons for engage/disengage
+            try:
+                if self.buttons.transport:
+                    self._buttons_unsub = self.buttons.subscribe(self._on_buttons)
+                    logger.info("Subscribed to buttons for engage/disengage")
+            except Exception as e:
+                logger.debug(f"Could not subscribe to buttons: {e}")
```
paul-nechifor: I see this a lot (checking if transport is truthy) and I don't think it's good.

In this case, if the button transport isn't working, the application is completely broken, no? Does it actually happen? If it does, that's a problem and we need to fix it.

But also, it's not good to catch an exception and log it as debug. It should be at least a warning.

dimos/control/coordinator.py
```diff
@@ -464,11 +487,17 @@ def _on_cartesian_command(self, msg: PoseStamped) -> None:
...
+                task.on_cartesian_command(msg, t_now)  # type: ignore[attr-defined]
             else:
-                logger.warning(f"Task {task_name} does not support set_target_pose")
+                logger.warning(f"Task {task_name} does not support on_cartesian_command")
+
+    def _on_buttons(self, msg: QuestButtons) -> None:
+        """Forward button state to all tasks that accept it."""
+        with self._task_lock:
+            for task in self._tasks.values():
+                if hasattr(task, "on_buttons"):
```
paul-nechifor: `hasattr` is almost always bad. Same with `type: ignore[attr-defined]`

The code is essentially saying "on the off chance that this object has a `on_buttons` method, call it with some random param. If the method is changed to require an additional param, `mypy` will never tell you because it doesn't know about it.

The simplest way to fix this is to define `on_buttons` in the `ControlTask` Protocol. This forces you to implement it on all tasks. In most cases, you can just implement the method as `pass`. (If this is too much code, you can define a BaseControlTask which implements all empty methods for such listeners, and if you inherit from it, only TeleopIKTask has to actually implement `on_buttons`.)

dimos/control/coordinator.py
```diff
@@ -464,11 +487,17 @@ def _on_cartesian_command(self, msg: PoseStamped) -> None:
                 logger.warning(f"Cartesian command for unknown task: {task_name}")
                 return
 
-            # Route to CartesianIKTask
-            if hasattr(task, "set_target_pose"):
-                task.set_target_pose(msg, t_now)  # type: ignore[attr-defined]
+            if hasattr(task, "on_cartesian_command"):
+                task.on_cartesian_command(msg, t_now)  # type: ignore[attr-defined]
```
paul-nechifor: Same as bellow.

dimos/control/coordinator.py
ruthwikdasyam: yeah, makes sense. So, remove all 'hasattr' from coordinator, which means implementing for all methods in the callbacks in either of ways you mentioned? Implementing this.

Added a BaseTask that implements all methods, and taks overrides and implements respective methods.

dimos/manipulation/planning/kinematics/pinocchio_ik.py
ruthwikdasyam: can write them outside the class. can they be in the same file, along with other standalone methods (check_joint_delta(), get_worst_joint_delta()), or should they go to any utils file, as this is a type conversion?

dimos/control/blueprints.py
```diff
@@ -35,6 +35,13 @@
 from dimos.core.transport import LCMTransport
 from dimos.msgs.geometry_msgs import PoseStamped
 from dimos.msgs.sensor_msgs import JointState
+from dimos.teleop.quest.quest_types import QuestButtons
+from dimos.utils.data import get_data
+
+_PIPER_MODEL_PATH = str(
+    get_data("piper_description") / "mujoco_model" / "piper_no_gripper_description.xml"
```
paul-nechifor: I didn't notice this before, but this calls `get_data` at import time. Meaning if the user doesn't have the repo, the program blocks here until the repo is downloaded.

I think someone else handled this case before by passing a lambda which is evaluated by the module. @leshy What do you think?

ruthwikdasyam: This was the previous implementation. But, this creates multiple methods everytime. What's the best way

`def _get_piper_model_path() -> str`
`    from dimos.utils.data import get_data`
`    piper_path = get_data("piper_description")`
`    return str(piper_path / "mujoco_model" / "piper_no_gripper_description.xml")`

jeff-hykin:
My bad @ruthwikdasyam looks like the LfsPath I added causes some breakage. I'll work with you on this today so we can it merged.

dimos/manipulation/planning/kinematics/pinocchio_ik.py
ruthwikdasyam: Changed it to standalone method

dimos/control/coordinator.py
ruthwikdasyam: This check happens in `if has_cartesian_ik or teleop_task`. `self.buttons.transport` returns true for `teleop_task` and false for `cartesian_ik` (here, it should not fail, coz we dont need `buttons` for `cartesian_ik`).

I changed the structure now. Writing a different check for `if has_teleop_task`. So that it fails if there's no `self.buttons.transport`.

ruthwikdasyam: There are three input streams. Joint_command, cartesian_command and buttons.

missing `.transport` in xase of `self.joint_command.transport` and `self.cartesian_command.transport` doesn't mean its broken coz the task methods can be also called via `task_invoke` RPC, without needing a stream at all. But, if using `.transport` is bad, I can directly try to subscribe and return a warning in case its missing (and not fail).

But, this should fail for `buttons` as there is no RPC alternative. implemented this.
@mustafab0 please confirm this won’t affect the manipulation module.

I verified it and nothing should break.

dimos/control/coordinator.py
```diff
@@ -145,6 +148,9 @@ class ControlCoordinator(Module[ControlCoordinatorConfig]):
     # Uses frame_id as task name for routing
     cartesian_command: In[PoseStamped]
 
+    # Input: Teleop buttons for engage/disengage signaling
+    buttons: In[QuestButtons]
```
jeff-hykin, line 152: Quest = MetaQuest buttons right? I think this needs to be generic. What would it look like if we were to add PicoButtons?

ruthwikdasyam: buttons stream name is generic, but the message type is QuestButtons since that's the only controller we support right now. Once we add a second device (Pico or VisionPro), the right generic interface will be clearer. abstracting before, without a second device/case usually leads to the wrong API. Keeping as-is for now.

dimos/control/blueprints.py
ruthwikdasyam: Fixed. used @jeff-hykin `LfsPath` for lazy loading - `get_data()` now defers to build time. Fixed bugs in implementation and verified import no longer triggers download.

paul-nechifor (approved):

PR #1256, opened 2026-02-13
Fix: Simplify Odometry Message 

leshy:
I looked at the file and suffered, so this kicks out Plum dispatch from Odometry, fixes types


paul-nechifor (approved):
This is beautiful.

PR #1280, opened 2026-02-17
Feat: Phone teleoperation stack with Go2 example

ruthwikdasyam:
## Phone Teleoperation for DimOS

Adding teleoperation via smartphone motion sensors. Tilt the phone to drive.
This extends the teleop subsystem introduced in the Quest PR, reusing the same architecture (WebSocket -> Deno LCM bridge -> Python module) but adapted for phone DeviceOrientation/DeviceMotion APIs

### Issue - Linear

closes DIM-393
closes DIM-535

### Approach

**Web layer**
- Mobile web app (`index.html`) reads DeviceOrientation (roll, pitch, yaw) and DeviceMotion (gyro rates) at native sensor frequency
- Encodes sensor data as `TwistStamped` and button state as `Bool`, sends over WebSocket at 50 Hz

**Python module**
- `PhoneTeleopModule` subscribes to `/phone/sensors` (TwistStamped) and `/phone/button` (Bool) LCM channels
- On engage (button hold), captures home orientation. Control loop at 50 Hz computes orientation delta from home, applies configurable gains, publishes `TwistStamped` velocity commands
- Implements `TeleopProtocol` - same interface as `QuestTeleopModule`
- Subclass hooks: `_handle_engage()`, `_should_publish()`, `_get_output_twist()`, `_publish_msg()`
- Auto-launches/stops the Deno bridge subprocess with proper SIGTERM/SIGKILL shutdown

**Also in this PR (teleop modules' improvements)**
- Deno bridge auto-launches with the module - no second terminal needed. 
- Robust subprocess lifecycle for both phone and quest teleop modules.

### Breaking Changes

None

### How to Test

With a Unitree Go2:

1. Set `ROBOT_IP` env var to the Go2's IP
2. Run `dimos run phone-go2-teleop`
3. On phone, open `https://<host-ip>:8444` and accept the self-signed certificate
4. Tap "Start Sensors" - grant orientation/motion permissions
5. Tap "Connect" - WS badge should turn blue
6. Press and hold "HOLD TO TELEOP" - tilt phone to drive the Go2
7. Release button - robot stops
8. Tap "Disconnect"


dimos/teleop/phone/phone_teleop_module.py
```diff
@@ -0,0 +1,369 @@
...
+        try:
+            self._server_process.wait(timeout=3)
+        except subprocess.TimeoutExpired:
+            logger.warning(
+                "Deno bridge did not exit, sending SIGKILL", pid=self._server_process.pid
+            )
+            self._server_process.kill()
+            self._server_process.wait(timeout=5)
+        logger.info("Deno bridge server stopped")
+        self._server_process = None
```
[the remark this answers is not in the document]
ruthwikdasyam: This Deno bridge setup is temporary - planning to move the WebSocket server directly into the Python module and decode incoming LCM bytes in-process. 
That removes the Deno dependency entirely and makes the bridge code go away, so extracting a shared mixin now would be throwaway work. Will clean up further.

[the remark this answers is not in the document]
ruthwikdasyam: `SIGKILL` is unconditional. timeout never happens - but, added try/except

dimos/teleop/phone/blueprints.py
```diff
@@ -0,0 +1,55 @@
...
+    phone_go2_teleop_module,
+    simple_phone_teleop_module,
+)
+
+# -----------------------------------------------------------------------------
+# Phone Teleop Blueprints
+# -----------------------------------------------------------------------------
+
+# Simple phone teleop (mobile base axis filtering)
+simple_phone_teleop = autoconnect(
```
spomichter, line 21: nice

dimos/teleop/phone/phone_teleop_module.py
```diff
@@ -0,0 +1,372 @@
...
+    def _on_button(self, msg: Bool) -> None:
+        """Callback for teleop button state."""
+        with self._lock:
+            self._teleop_button = bool(msg.data)
+
+    # -------------------------------------------------------------------------
+    # Deno Bridge Server
+    # -------------------------------------------------------------------------
+
+    def _start_server(self) -> None:
```
spomichter, line 151: This is not reccomended look at how its done here.

```
class WebInput(Module):
    _web_interface: RobotWebInterface | None = None
    _thread: Thread | None = None
    _human_transport: pLCMTransport[str] | None = None

    @rpc
    def start(self) -> None:
        super().start()

        self._human_transport = pLCMTransport("/human_input")

        audio_subject: rx.subject.Subject[AudioEvent] = rx.subject.Subject()

        self._web_interface = RobotWebInterface(
            port=5555,
            text_streams={"agent_responses": rx.subject.Subject()},
            audio_subject=audio_subject,
        )```

**Review by spomichter — CHANGES_REQUESTED** [review]

File: dimos/teleop/phone/phone_teleop_module.py (hunk 3 above)
Comment (ruthwikdasyam) [reply dimos/teleop/phone/phone_teleop_module.py:151]: Agreed -  planning to move the web server into the module itself and drop the Deno subprocess. Will address this in a follow-up PR tracked in #1222. This touches both Quest and phone teleop stacks so keeping it separate

File: dimos/teleop/quest/quest_teleop_module.py [hunk 4]
```diff
@@ -225,6 +234,52 @@ def _on_joy(self, hand: Hand, joy: Joy) -> None:
...
+            "run",
+            "--allow-net",
+            "--allow-read",
+            "--allow-run",
+            "--allow-write",
+            "--unstable-net",
+            script,
+        ]
+        try:
+            self._server_process = subprocess.Popen(cmd)
```
paul-nechifor, line 243: Before running this, I think you should check if `deno` is installed on the system and show how to install it.

paul-nechifor: Added this issue to discuss Deno, but it should be okay to keep it in this PR: https://linear.app/dimensional/issue/DIM-552/remove-deno

dimos/teleop/phone/README.md
```diff
@@ -0,0 +1,74 @@
...
+| `/teleop/twist` | TwistStamped | Output velocity command |
+
+## Running
+
+```bash
+dimos run phone-go2-teleop     # Go2
+dimos run simple-phone-teleop  # Generic ground robot
+```
+
+Server starts on port `8444`. Open `https://<host-ip>:8444` on phone, accept the self-signed certificate, allow sensor permissions, connect, hold button to drive.
```
paul-nechifor, line 58: Why do we need HTTPS and a certificate? Why not use HTTP?

dimos/teleop/phone/phone_teleop_module.py
```diff
@@ -0,0 +1,369 @@
...
+
+        input_streams: dict[str, tuple[Any, Any]] = {
+            "phone_sensors": (self.phone_sensors, self._on_sensors),
+            "phone_button": (self.phone_button, self._on_button),
+        }
+        connected = []
+        for name, (stream, handler) in input_streams.items():
+            if not (stream and stream.transport):  # type: ignore[attr-defined]
+                logger.warning(f"Stream '{name}' has no transport — skipping")
+                continue
```
paul-nechifor: Why is this check needed? Does the module still work if the subscription is never made?

dimos/teleop/phone/phone_teleop_module.py
```diff
@@ -0,0 +1,369 @@
...
+
+        logger.info("PhoneTeleopModule initialized")
+
+    # -------------------------------------------------------------------------
+    # Public RPC Methods
+    # -------------------------------------------------------------------------
+
+    @rpc
+    def start(self) -> None:
+        """Start the phone teleoperation module."""
```
paul-nechifor: I've mentioned this in https://discord.com/channels/1341146487186391173/1357833666751107174/1471684363253973125

Can you please remove docstrins and log messages which don't add any value like:

```
"""Start the phone teleoperation module."""
...
        if connected:
            logger.info(f"Subscribed to: {', '.join(connected)}")
...
logger.info("Phone Teleoperation Module started")
...
"""Stop the phone teleoperation module."""
logger.info("Stopping Phone Teleoperation Module...")
```

dimos/teleop/phone/phone_teleop_module.py
```diff
@@ -0,0 +1,369 @@
...
+        if self._control_loop_thread is not None:
+            self._control_loop_thread.join(timeout=1.0)
+            self._control_loop_thread = None
+        logger.info("Control loop stopped")
+
+    def _control_loop(self) -> None:
+        """Main control loop: handle engagement, compute deltas, publish twist"""
+        period = 1.0 / self.config.control_loop_hz
+
+        while self._control_loop_running:
```
paul-nechifor: Booleans aren't safe without a lock. Please use a `threading.Event`.

dimos/teleop/phone/phone_teleop_module.py
ruthwikdasyam: No, not here. In case of phone teleop, we just have 2 msgs coming, and we need both of them. removing the check.

dimos/teleop/phone/phone_teleop_module.py
```diff
@@ -0,0 +1,369 @@
...
+            try:
+                with self._lock:
+                    self._handle_engage()
+
+                    if self._should_publish():
+                        output_twist = self._get_output_twist()
+                        if output_twist is not None:
+                            self._publish_msg(output_twist)
+            except Exception:
+                logger.exception("Error in phone teleop control loop")
```
paul-nechifor: What exception do you anticipate happening here? Please only wrap in try-except the part that you expect to fail.

dimos/teleop/phone/phone_teleop_module.py
ruthwikdasyam: Thankyou! Was always skeptical about having them, included them to maintain uniformity across stack.
Noted, will do this from now!

dimos/teleop/phone/web/static/index.html
```diff
@@ -0,0 +1,393 @@
+<!DOCTYPE html>
```
paul-nechifor, line 1: You might want to move the "HOLD TO TELEOP" button to the top because it's partially hidden if I have my address bar showing.

dimos/teleop/phone/blueprints.py
```diff
@@ -0,0 +1,55 @@
...
+)
+
+# Phone teleop wired to Unitree Go2
+phone_go2_teleop = autoconnect(
+    phone_go2_teleop_module(),
+    unitree_go2_basic,
+).transports(
+    {
+        ("phone_sensors", TwistStamped): LCMTransport("/phone/sensors", TwistStamped),
+        ("phone_button", Bool): LCMTransport("/phone/button", Bool),
```
paul-nechifor: Why not use the default topics of `/phone_sensors` and `/phone_button` so you don't need `.transports(...)`?

dimos/teleop/phone/phone_extensions.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+
+    cmd_vel: Out[Twist]
+
+    def _publish_msg(self, output_msg: TwistStamped) -> None:
+        """Publish as Twist on cmd_vel to match GO2Connection.
+
+        Intentionally bypasses the base twist_output stream — only cmd_vel is used.
+        """
+        self.cmd_vel.publish(Twist(linear=output_msg.linear, angular=output_msg.angular))
```
paul-nechifor: Why is this specific to Go2? It should work on G1 too, right?

ruthwikdasyam: Oh ideally yes! Anything that takes twist should work. I tested for Go2, so named it explicitly. I will rename/generalize.

dimos/teleop/phone/phone_teleop_module.py
```diff
@@ -0,0 +1,372 @@
...
+
+    control_loop_hz: float = 50.0
+
+    # Gain: maps degrees of tilt to m/s.  30 deg tilt -> 1.0 m/s
+    linear_gain: float = 1.0 / 30.0
+    # Gain: maps degrees of yaw delta to rad/s.  30 deg -> 1.0 rad/s
+    angular_gain: float = 1.0 / 30.0
+
+
+class PhoneTeleopModule(Module[PhoneTeleopConfig], TeleopProtocol):
```
paul-nechifor: I haven't noticed TeleopProtocol before, but it seems unnecessary.

The purpose of `Protocol` is to enforce a contract between one class using another one. It's not to restrict inheritance. 

Protocols do not restrict implementation in the absence of something which uses the protocol. In other words, it's useless to define a Protocol and create classes which implement that protocol if the protocol isn't used anywhere.

Here, as far as I can see, the `@rpc` methods `engage`/`disengage` are never used. They're only meant to be used internally when responding to `self.phone_button`. And `start`/`stop` are only used by the deployment system, not other modules.

I think you should delete the protocol and remove `engage`/`disengage` .

@leshy Do you agree?

paul-nechifor: If you want to restrict inheritance, you have to use an `ABC`. With them, you get errors if you don't implement the abstract methods.

dimos/teleop/phone/phone_teleop_module.py
```diff
@@ -0,0 +1,372 @@
...
+        linear = (roll, pitch, yaw) in degrees
+        angular = (gyro_x, gyro_y, gyro_z) in deg/s
+        """
+        with self._lock:
+            self._current_sensors = msg
+
+    def _on_button(self, msg: Bool) -> None:
+        """Callback for teleop button state."""
+        with self._lock:
+            self._teleop_button = bool(msg.data)
```
paul-nechifor, line 145: Why set `_teleop_button`? Can't you just call `_engage`/`_disengage` from here?

dimos/teleop/phone/phone_extensions.py
paul-nechifor: Also, in my opinion, there's no need to have a three class structure to control cmd_vel. It should be a single class. (Unless you plan to expand this, it seems premature to split it like this.)

dimos/teleop/phone/phone_teleop_module.py
ruthwikdasyam: The Protocol is here to unify all teleop modules across devices (Quest, Phone, future). There's no shared implementation between them - just a common interface contract which i think is exactly what Protocol is for.
Can be removed if having no baseclass nor protocol between deviceteleopmodules is fine.

`engage()`/`disengage()` are exposed as RPCs so other modules can control teleoperation - Quest already uses them with the hand parameter. The plan is for a manipulation/recording module to call these when collecting trajectories - for something like imitation learning.

phone_teleop has very simple implementation compared to other devices. So, `engage`/`disengage` seems easy to remove. But, they carry more weight with other device implementations

dimos/teleop/phone/phone_teleop_module.py
ruthwikdasyam: They are defined in `_handle_engage()`, so that subclasses can override them. Saywe are controlling a different robot, and If i dont want press&hold, but a toggle behavior (press once to engage and again to disengage), I can modify `handle_engage()` method in the subclass.

Did not want subclasses to touch callback methods.

dimos/teleop/phone/README.md
ruthwikdasyam: seems that mobile browsers require HTTPS to access device sensors. The DeviceOrientation and DeviceMotion APIs won't work over plain HTTP and the browser silently blocks them

dimos/teleop/phone/phone_extensions.py
ruthwikdasyam: I am actually making simple_phone_teleop to work with go2, so removing third class. BasePhoneModule returning `TwistStamped` like it should, and one sub-class for `Twist` as Go2 needs.

Further, someone can expand this to arm, which potentially adds a slider or additional buttons on the webpage. Leaving room for expansion, and trying to maintain consistent format between device teleop modules.

I understand that, as this is straight forward for phone example, and looks over structured. I will simplify a bit trading off with design.

dimos/teleop/phone/README.md
spomichter: This is true

PR #1293, opened 2026-02-18
Fix CI Divergence (Mostly `mypy`)

jeff-hykin:
### Why does the new CI workflow have a ARM/MacOS runner?  <br>Why are there modifications to pyproject.toml?
  - Mypy divergence has been a miserable rabbit hole, so please bear with me
  - whether mypy fails/passes depends on EVERY version of EVERY dependency (because it inspects their source code)
      - Easy fix right? `uv sync --freeze <ci stuff>`
      - no. Because of the `<ci stuff>`
  - In principle `uv sync <ci stuff>` shouldn't fail on any system -- b/c in our pyproject toml if something only works on X system we should adding conditionals like "drake; only if linux". But currently we don't because we don't have any checks (and not everyone should get a jetson and a mac just to test their changes to pyproject.toml work on all systems)
  - This PR should prevent those pyproject.toml issues from getting on dev
  - Once `uv sync <ci stuff>` runs, our local mypy won't diverge from CI

### Why is `setup.py` modified?
- Turns out, when testing with the macos-15 github runner, there's a MacOS install issue with dimos. On MacOS the python executable can be "universal" (one executable for x86 and arm). The python executable for the uv python action on github runners is universal. In our setup.py, we add C-compiler optimization arguments (e.g. `-march=native`). This causes `uv sync` to break because clang can only perform that optimization if it targets specifically arm64 xor x86 MacOS (can't do it when targeting for MacOS's universal executable format). The setup.py keeps the optimization whenever possible. For people who use nix, they get the optimization.

### Why is there a `fast-docs.yml`?
- Right now when I push doc changes, nothing actually checks the doc links or that diagrams are up to date (see https://github.com/dimensionalOS/dimos/actions/runs/22152132097/job/64046102043?pr=1291 for an example). This should fix that and test diagrams (which I don't think we were checking before)


.github/workflows/readme.md
```diff
@@ -1,7 +1,7 @@
 # general structure of workflows
 
-Docker.yml checks for releavant file changes and re-builds required images
-Currently images have a dependancy chain of ros -> python -> dev (in the future this might be a tree and can fork)
+Docker.yml checks for relevant file changes and re-builds required images
+Currently images have a dependency chain of ros -> python -> dev (in the future this might be a tree and can fork)
```
jeff-hykin, line 4: Spelling fix

.github/workflows/fast-docs.yml
```diff
@@ -0,0 +1,34 @@
...
+jobs:
+  check_links:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
```
jeff-hykin, line 24: In this workflow (documentation checks) the version shouldn't matter

.github/workflows/fast-docs.yml
```diff
@@ -0,0 +1,34 @@
...
+
+      - name: Run doclinks # Note: doesn't need uv python, just needs some python to run the script
+        run: |
+            python -m dimos.utils.docs.doclinks docs/
+
+      - uses: leshy/md-babel-py@main
+        with:
+          files: 'docs/**/*.md'
+          fail-on-change: true
+          # maybe later we can enforce python, for now just do diagram langs
```
jeff-hykin, line 34: Previously we weren't checking diagram generation AFAIK, this adds that

.github/workflows/fast-test.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Setup uv
+        uses: astral-sh/setup-uv@v4
+        with:
+          enable-cache: true
+          python-version: "3.12"
+          cache-suffix: ${{ runner.os }}-${{ runner.arch }}-py3.12
+
+      - name: Install System dependencies
+        run: |
+          # these lines should stay exactly in sync with docs/development/README.md
```
jeff-hykin, line 38: This is an important change: we should have something on CI that matches our docs, and this new test does that

.github/workflows/fast-test.yml
```diff
@@ -0,0 +1,52 @@
...
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  pytest_fast:
+    strategy:
+      matrix:
+        os: [ubuntu-24.04, macos-15, ubuntu-24.04-arm]
```
jeff-hykin, line 23: *Discussion needed*: do we want to host our own macos / linux-arm testers? (or keep this as-is and run on github)

bin/ci-fast
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
```
jeff-hykin, line 1: This file is the overlap between local and CI: this file gets executed by CI and is executed locally (when running `bin/check-pr`)

bin/ci-fast
```diff
@@ -0,0 +1,26 @@
...
+
+echo "Ensuring pip versions are correct"
+time uv sync --frozen --extra misc --extra visualization --extra agents --extra web --extra perception --extra unitree --extra manipulation --extra cpu --extra dev --extra sim --extra drone --extra base
+
+echo "Running pre-commit checks"
+time pre-commit run --all-files
+
+echo "Running mypy checks"
+time uv run mypy --python-version 3.10 dimos
+time uv run mypy --python-version 3.12 dimos
```
jeff-hykin, line 19: This could be faster by excuting in parallel, but I opted for simplicity. No cache ~5min on github runners, so its pretty fast

bin/gen-diagrams
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+set -euo pipefail
```
jeff-hykin, line 2: Merge https://github.com/dimensionalOS/dimos/pull/1291 (very small) first and this diff will go away

bin/pr-check
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env python3
```
jeff-hykin, line 1: This is what everyone should run locally before making a pull request.
It
1. checks branch name (you can ignore the error and keep going)
2. runs only doclink checks if only markdown is edited
3. runs `bin/ci-fast` if more than just markdown is edited

A lot of good enhancements could be added, but I wanted to start simple

bin/pr-name-check
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env bash
```
jeff-hykin, line 1: I don't care which branch-naming-scheme we use, I just implemented a check for the one that is listed on [our wiki](https://github.com/dimensionalOS/wiki/wiki) `<name>/<type>/<description>`

Right now its not a forceful/blocking check it'll just yell at you if it doesnt match

docker/python/Dockerfile
```diff
@@ -48,4 +48,9 @@ COPY . /app/
 
 # Install dependencies with UV (10-100x faster than pip)
 RUN uv pip install --upgrade 'pip>=24' 'setuptools>=70' 'wheel' 'packaging>=24' && \
-    uv pip install '.[misc,cpu,sim,drone,unitree,web,perception,visualization]'
+    uv sync --frozen --extra misc --extra visualization --extra agents --extra web --extra perception --extra unitree --extra manipulation --extra cpu --extra dev --extra sim --extra drone --extra base
```
jeff-hykin, line 51: # Most Important Change

`--frozen` instead of "close enough" install

docs/development/writing_docs.md
```diff
@@ -3,5 +3,5 @@
 1. Where to put your docs:
     - If it only matters to people who contribute to dimos (like this doc), put them in `docs/development`
     - Otherwise put them in `docs/usage`
-2. Run `bin/gen_diagrams` to generate the svg's for your diagrams. We use [pikchr](https://pikchr.org/home/doc/trunk/doc/userman.md) as a diagram language.
+2. Run `bin/gen-diagrams` to generate the svg's for your diagrams. We use [mermaid](https://mermaid.js.org/intro/) (no generation needed) and [pikchr](https://pikchr.org/home/doc/trunk/doc/userman.md) as diagrams languages.
```
jeff-hykin, line 6: Merge https://github.com/dimensionalOS/dimos/pull/1291 (very small) first and this diff will go away

.github/workflows/docker.yml
```diff
@@ -292,41 +292,6 @@ jobs:
       cmd: "MYPYPATH=/opt/ros/humble/lib/python3.10/site-packages mypy dimos"
       dev-image: ros-dev:${{ (needs.check-changes.outputs.python == 'true' || needs.check-changes.outputs.dev == 'true' || needs.check-changes.outputs.ros == 'true') && needs.ros-dev.result == 'success' && needs.check-changes.outputs.branch-tag || 'dev' }}
 
-  # Run module tests directly to avoid pytest forking issues
```
jeff-hykin, line 295: If we need to restore old code, lets use git as it was meant to be (instead of having giant blocks of commented out code

.github/workflows/docker.yml
```diff
@@ -243,7 +243,7 @@ jobs:
     uses: ./.github/workflows/tests.yml
     secrets: inherit
     with:
-      cmd: "pytest -m heavy"
+      cmd: "uv run pytest -m heavy"
       dev-image: dev:${{ (needs.check-changes.outputs.python == 'true' || needs.check-changes.outputs.dev == 'true') && needs.dev.result == 'success' && needs.check-changes.outputs.branch-tag || 'dev' }}
```
jeff-hykin, line 247: I think this is a nessary change, based on: https://github.com/dimensionalOS/dimos/actions/runs/22160016061/job/64077594237?pr=1293

pyproject.toml
```diff
@@ -172,7 +171,8 @@ unitree = [
 
 manipulation = [
     # Planning (Drake)
-    "drake>=1.40.0; platform_machine != 'aarch64'",
+    "drake==1.45.0; sys_platform == 'darwin' and platform_machine != 'aarch64'",
+    "drake>=1.40.0; sys_platform != 'darwin' and platform_machine != 'aarch64'",
```
jeff-hykin, line 175: This is both a fix for MacOS and Arm. On MacOS drake 1.45.0 is needed, on x86 arm anything >1.40 is okay. 

The drake team says they're interested in supporting Arm linux so hopefully that'll come soon

pyproject.toml
```diff
@@ -191,7 +191,8 @@ manipulation = [
 
 cpu = [
     # CPU inference backends
-    "onnxruntime",
+    "onnxruntime<1.24; python_version < '3.11'",
```
jeff-hykin, line 194: CI caught this which is great: it was impossible for python 3.10 to have the version of onnx we had specified in the lock file. This guard fixes that

setup.py
```diff
@@ -13,20 +13,63 @@
...
 import os
+from pathlib import Path
+import struct
+import sys
 
 from pybind11.setup_helpers import Pybind11Extension, build_ext
 from setuptools import find_packages, setup
 
+
+def python_is_macos_universal_binary(executable: str | None = None) -> bool:
```
jeff-hykin, line 24: without this change macos-15 github uv python runner can't `uv sync`

.github/workflows/fast-test.yml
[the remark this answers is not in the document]
jeff-hykin: I'm running `ubuntu-24.04-arm` on a different repo. Its valid

.github/workflows/fast-test.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Install System dependencies
+        run: |
+          # these lines should stay exactly in sync with docs/development/README.md
+          if [ "$OSTYPE" = "linux-gnu" ]; then
+            sudo apt-get update
+            sudo apt-get install -y curl g++ portaudio19-dev git-lfs libturbojpeg python3-dev pre-commit
+          # On macOS (12.6 or newer)
+          elif [ "$(uname)" = "Darwin" ]; then
+            # install homebrew
+            /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
```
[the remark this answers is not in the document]
jeff-hykin: huge fantastic

bin/ci-slow
```diff
@@ -0,0 +1,41 @@
...
+echo "== dev: pytest -m heavy =="
+run_in_image "dev" "pytest -m heavy"
+
+echo
+echo "== dev: pytest -m lcm =="
+run_in_image "dev" "pytest -m lcm"
+
+echo
+echo "== dev: pytest -m integration =="
+run_in_image "dev" "pytest -m integration"
```
paul-nechifor, line 41: If you're running all of them sequentially, you might as well run a single command:

```
run_in_image "dev" "-m 'not (tool or module or neverending or mujoco)'"
```

bin/pr-name-check
```diff
@@ -0,0 +1,68 @@
...
+
+set -euo pipefail
+
+branch="$(git rev-parse --abbrev-ref HEAD)"
+
+# based on: https://github.com/dimensionalOS/wiki/wiki
+allowed_types="feat fix chore refactor docs"
+allowed_names="stash ivan pauln alexl mustafa miguel christie ruthwik jalaj yashas yash matt jing juan jeff unknown"
```
paul-nechifor:
```suggestion
allowed_names="stash ivan paul alexl mustafa miguel christie ruthwik jalaj yashas yash matt jing juan jeff unknown"
```
I switched to `paul` because everyone was using a single name.

bin/pr-name-check
```diff
@@ -0,0 +1,68 @@
...
+
+if [[ "$name_ok" -ne 1 || "$type_ok" -ne 1 ]]; then
+  echo
+  echo
+  echo
+  echo
+  echo
+  echo "Invalid branch name: '$branch'"
+  echo
+  echo "    Expected format: <your_name>/<type>/<description>"
```
paul-nechifor, line 54:
```suggestion
  echo "    Expected format: <your_name>/<type>/<description>"
  echo "            Example: jeff/fix/ci-divergence"
```

docker/python/Dockerfile
paul-nechifor: An issue I see here is that this forces us to always run with `uv run ...`. This should be fixable by adding some of the variables which activating gives you. This could be enough:

```
ENV VIRTUAL_ENV=/app/.venv
ENV PATH="/app/.venv/bin:$PATH"
```

Another issue that worries me is that `uv sync` always installs in `.venv`. The previous command installed in the system python. By installing in `/app` it means we can't mount our local dir as `/app` because it would use the `.venv` we have locally instead of the one in the docker image.

Not sure what the solution is here. (Maybe copy `pyproject.toml` to `/app-deps` and install from that dir so the packages are installed outside `/app`?

.github/workflows/docker.yml
```diff
@@ -212,7 +212,7 @@ jobs:
     uses: ./.github/workflows/tests.yml
     secrets: inherit
     with:
-      cmd: "pytest && pytest -m ros" # run tests that depend on ros as well
+      cmd: "uv run pytest && uv run pytest -m ros" # run tests that depend on ros as well
```
paul-nechifor: By changing environment variables we can make the `uv` python be the default python so that we don't need to prefix with `uv run` (mentioned in a comment below)

bin/ci-fast
```diff
@@ -0,0 +1,26 @@
...
+
+echo "# "
+echo "# "
+echo "# pytest fast"
+echo "# "
+echo "# "
+time uv run bin/pytest-fast
```
paul-nechifor, line 26: pytest-fast has this:

```
. .venv/bin/activate
exec pytest "$@" dimos
```

You probably just want to run `pytest` without any `-m`, not `./bin/pytest-fast`.

bin/ci-fast
```diff
@@ -0,0 +1,26 @@
...
+#
+# note this command is used by CI directly, and also is used by bin/pr-check
+# this is intentional: to keep them in sync
+# if something should only be done locally, change bin/pr-check
+# if something should only be done on CI, change .github/workflows/fast-test.yml
+#
+set -euo pipefail
+
+echo "Ensuring pip versions are correct"
+time uv sync --frozen --extra misc --extra visualization --extra agents --extra web --extra perception --extra unitree --extra manipulation --extra cpu --extra dev --extra sim --extra drone --extra base
```
paul-nechifor, line 12: I'm not sure this should be here. In CI, the packages should have been installed in the python step, no?

And locally this removes packages. (I do `uv sync --all-extras` locally.)

bin/pr-check
paul-nechifor: Why just the fast tests? Shouldn't people run all the tests? (The non-fast ones are more commonly broken on dev especially since a few of them don't run in CI at all.)

.github/workflows/fast-docs.yml
```diff
@@ -0,0 +1,34 @@
...
+    runs-on: [self-hosted, Linux]
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
+      - name: Run doclinks # Note: doesn't need uv python, just needs some python to run the script
```
leshy, line 26: git-commit hooks run this, so this is already executed by our `code-cleanup` action in CI

bin/ci-fast
jeff-hykin: No, AFAIK the github uv python action doesn install with --frozen which is what is needed to prevent mypy from diverging. Just run the workflow and you'll see this command does cause installs. 

Running the command when everything is already installed is almost instant (no reinstall, just performs a check). 

As for locally, its still basically necessary. I agree uninstallation isnt ideal but I don't think theres a way to say "keep all the features I selected (dont uninstall anything) but use the frozen versions from UV". If you dont want to run the command thats fine but IMO we should at least have one command that exactly replicates CI

.github/workflows/fast-docs.yml
jeff-hykin: I was thinking we would change it where this is the only thing that would run on .md file edits (to make it really fast to merge doc changes). If we keep code cleanup though, yeah I guess it would be good to remove it.

I guess there are those whitespace checks in pre-commit that still effect docs

bin/ci-fast
jeff-hykin: I can try to make a "use frozen versions of everything but dont uninstall anything" command. If its not too hacky maybe that would be useful

bin/ci-fast
jeff-hykin: My thought was, if we edit pytest-fast (to exclude something, or add something) we would want that change to be in sync with CI.

.github/workflows/fast-docs.yml
leshy: not sure if missing something, why add complexity for speed gains if pre commit stuff runs in less then a second?

docker/python/Dockerfile
leshy: I don't mind .venv install (every one of us locally uses venv afaik anyway) and if it brings us up to standard with modern uv locks etc.

our direnv files optionally help with these vars also.

only worry is how this vibes with ros that uses system python? but if tests are passing we are good?

jeff-hykin:
while I still want some of these changes, I'm closing because CI has changed so much and now the benefits of these changes are more marginal

PR #1295, opened 2026-02-19
Ivan/maindocs2

leshy:
First pass on the main readme & doc structure
High line change count is because I restored docs/agents (docs for agents) that were deleted previously

### DIFF
--- a/docs/usage/transports/index.md
+++ b/docs/usage/transports/index.md
@@ -9,7 +9,7 @@ Transports connect **module streams** across **process boundaries** and/or **net
 
 Each edge in the graph is a **transported stream** (potentially different protocols). Each node is a **module**:
 
-![go2_nav](assets/go2_nav.svg)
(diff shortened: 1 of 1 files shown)


.github/workflows/code-cleanup.yml
```diff
@@ -1,7 +1,8 @@
 name: code-cleanup
 on:
-  push:
   pull_request:
+    paths-ignore:
```
spomichter, line 4: FYI hesitant to fuck with CI any more since its all working.

leshy: this was so that pushes to a branch don't trigger CI if there is no PR, couldn't do my 3000 commits without this

leshy: but generally yes I'm also scared, I think this one is ok

spomichter: ok

PR #1305, opened 2026-02-19
fix: broken tests

ruthwikdasyam:
### Summary fixes
- Fix manipulation tests to pass `JointState` instead of raw lists to `plan_to_joints()`
- Increase e2e coordinator test timeout from 10s to 15s
- Fix lcmspy global totals test flaking from async LCM discovery packets

### Tests verified, no errors
- `time uv run pytest --durations=0 dimos`
- `time ./bin/pytest-slow --durations=0`

### FYI - Potential fail

Tests require `rclpy`, which is tied to a specific Python version per ROS distro:
- Ubuntu 22.04 + ROS Humble -> Python 3.10
- Ubuntu 24.04 + ROS Jazzy -> Python 3.12

If the python version doesn't match (python 3.12 via uv on Ubuntu 22.04), `rclpy` can't be installed and the tests will fail with `ModuleNotFoundError`

But, if someone already has these versions, can make this change
 `pytest.importorskip("rclpy")` so they skip cleanly instead of erroring out in file `dimos/protocol/pubsub/impl/test_rospubsub.py`

### DIFF
--- a/dimos/utils/cli/lcmspy/test_lcmspy.py
+++ b/dimos/utils/cli/lcmspy/test_lcmspy.py
@@ -176,7 +176,8 @@ def test_lcmspy_global_totals() -> None:
     spy.msg("/imu", b"imu data")
 
     # The spy itself should have accumulated all messages
-    assert len(spy.message_history) == 3
+    # (may be > 3 due to async LCM discovery packets on the subscription thread)
+    assert len(spy.message_history) >= 3
 
     # Check global statistics
     global_freq = spy.freq(1.0)
--- a/dimos/utils/cli/lcmspy/test_lcmspy.py
+++ b/dimos/utils/cli/lcmspy/test_lcmspy.py
@@ -175,9 +175,9 @@ def test_lcmspy_global_totals() -> None:
     spy.msg("/odom", b"odometry data")
     spy.msg("/imu", b"imu data")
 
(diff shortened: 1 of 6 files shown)


dimos/utils/cli/lcmspy/test_lcmspy.py
```diff
@@ -176,7 +176,8 @@ def test_lcmspy_global_totals() -> None:
     spy.msg("/imu", b"imu data")
 
     # The spy itself should have accumulated all messages
-    assert len(spy.message_history) == 3
+    # (may be > 3 due to async LCM discovery packets on the subscription thread)
+    assert len(spy.message_history) >= 3
```
paul-nechifor: I think the solution here is to filter out the discovery messages and assert that the remaining messages are 3.

ruthwikdasyam: As an alternative, can check per topic specifically, say

    for t in ("/video", "/odom", "/imu"):
        assert len(spy.topic[t].message_history) == 1

Can we go with this?

paul-nechifor: Yes, sounds good.

spomichter:
ok yeah manipulation needs to be added to docker extra

spomichter (approved):

PR #1307, opened 2026-02-20
fix(unitree-go2): minor issues

paul-nechifor:
* Change Unitree Go2 to use voxel cubes instead of points.
* When told to stop exploring, stop faster.
* Fix Mujoco end-to-end tests.
* Fix location tagging (self.tf is not available).
* Fix stand up on start and stand down on exit. Note that the Dask signal handler prevented calling `stop` on modules. I've removed that and ModuleCoordinator will call stop on all modules.
* Also fix some mypy issues on local runs.

### DIFF
--- a/dimos/robot/unitree/connection.py
+++ b/dimos/robot/unitree/connection.py
@@ -278,7 +278,9 @@ class UnitreeWebRTCConnection(Resource):
         return bool(self.publish_request(RTC_TOPIC["SPORT_MOD"], {"api_id": SPORT_CMD["StandUp"]}))
 
     def liedown(self) -> bool:
-        return bool(self.publish_request(RTC_TOPIC["SPORT_MOD"], {"api_id": SPORT_CMD["StandDown"]}))
+        return bool(
+            self.publish_request(RTC_TOPIC["SPORT_MOD"], {"api_id": SPORT_CMD["StandDown"]})
+        )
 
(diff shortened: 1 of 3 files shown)


dimos/robot/unitree/connection.py
```diff
@@ -275,33 +274,20 @@ def video_stream(self) -> Observable[Image]:
     def lowstate_stream(self) -> Observable[LowStateMsg]:
         return backpressure(self.unitree_sub_stream(RTC_TOPIC["LOW_STATE"]))
 
-    def standup_ai(self) -> bool:
-        return self.publish_request(RTC_TOPIC["SPORT_MOD"], {"api_id": SPORT_CMD["BalanceStand"]})  # type: ignore[no-any-return]
```
paul-nechifor, line 279: This doesn't work for me. `StandUp` works fine.

dimos/core/__init__.py
```diff
@@ -266,31 +265,6 @@ def start(n: int | None = None, memory_limit: str = "auto") -> DimosCluster:
     )
 
     patched_client = patchdask(client, cluster)
-    patched_client._shutting_down = False  # type: ignore[attr-defined]
```
paul-nechifor, line 269: Stopping Dask prevented the ModuleCoordinator from stopping modules so `GO2Connection.stop` never called `self.liedown`.

PR #1318, opened 2026-02-20
fix(deps): exclude dds extra from uv sync --all-extras

spomichter:
## Summary
`cyclonedds` requires building from source with `CYCLONEDDS_HOME` set, which fails on standard dev machines when running `uv sync --all-extras`.

## Changes
- Updated install docs (README, ubuntu, osx, nix) to use `uv sync --all-extras --no-extra dds`
- Added explanatory comment to the `dds` extra in pyproject.toml

## How to Test
```bash
uv sync --all-extras --no-extra dds
```
Should resolve without attempting to build cyclonedds.

### DIFF
--- a/docs/installation/osx.md
+++ b/docs/installation/osx.md
@@ -31,10 +31,7 @@ export GIT_LFS_SKIP_SMUDGE=1
 git clone -b dev https://github.com/dimensionalOS/dimos.git
 cd dimos
 
-uv venv --python 3.12
-source .venv/bin/activate
-
-uv sync --all-extras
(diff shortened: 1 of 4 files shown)


docs/installation/osx.md
```diff
@@ -34,7 +34,7 @@ cd dimos
 uv venv --python 3.12
 source .venv/bin/activate
```
paul-nechifor: Please remove this again:

```
uv venv --python 3.12
source .venv/bin/activate
```

uv sync creates it's own venv.

PR #1343, opened 2026-02-22
fix(tests): simplify testing

paul-nechifor:
## Problem

Our testing setup is quite complicated.

Closes DIM-559

## Solution

* Remove all marks except: `tool`, `slow` (replacement for `heavy`/`integration`/`e2e`), and `mujoco`.
* Have a single testing job which runs all `pytest` tests.
* Fix broken `neverending` and `module` tests.
* Add custom skipif markers like `skipif_in_ci`, `skipif_no_openai` to lessen duplication.
* Fix the remaining tests which didn't close threads.
* Update the docs.

## Breaking Changes

None

## How to Test

Run the commands from https://github.com/dimensionalOS/dimos/blob/paul/fix/testing-docs-2/docs/development/testing.md .
 

dimos/core/test_core.py
```diff
@@ -136,4 +137,7 @@ def test_basic_deployment(dimos) -> None:
     assert nav.odom_msg_count >= 8
     assert nav.lidar_msg_count >= 8
 
-    dimos.shutdown()
+    nav.stop()
+    nav.stop_rpc_client()
+    robot.stop_rpc_client()
+    dimos.close_all()
```
jeff-hykin, line 143: Is this typical/recommended over dimos.shutdown?

dimos/core/test_stream.py
```diff
@@ -90,14 +110,14 @@ class RXPYSubscriber(SubscriberBase):
...
+        self.unsub = self.odom.observable().subscribe(self._sub1_callback)
 
     @rpc
     def sub2(self) -> None:
-        self.unsub2 = self.odom.observable().subscribe(self.sub2_msgs.append)
+        self.unsub2 = self.odom.observable().subscribe(self._sub2_callback)
 
     @rpc
-    def stop(self) -> None:
+    def unsub_all(self) -> None:
```
jeff-hykin, line 120: Seems kinda nice for stop to be a universal method that includes stuff like unsub_all.

jeff-hykin (commented):
I love the color red when it comes to PR's. My only questions are:
- is cuda still fine? (E.g. non cuda systems just don't add the cuda flag, that's why we don't need a has_cuda check)
- I'm a bit unsure of our policy for `.stop `/close/shutdown/unsub. It feels like there should be a consistent naming system. Like stop is always stop-all or some similar rule.

dimos/core/test_stream.py
paul-nechifor: I agree. But this test module defined `stop` whilst not actually calling `super().stop()` and then calling `dimos.close_all()`. It felt easier at the time to just rename but I should probably fix it the right way. 😅

dimos/core/test_core.py
paul-nechifor: Yes, dimos.close_all does a few other things, but we won't be using Dask for long as I'm removing it.

leshy: I guess we can have whatever our forking coordinator is - conform to Resource so stop() recursively goes through active modules/processes?

.github/workflows/docker.yml
```diff
@@ -292,43 +231,8 @@ jobs:
...
-  #
-  #     - name: Run module tests
-  #       env:
-  #         CI: "true"
-  #       run: |
-  #         /entrypoint.sh bash -c "pytest -m module"
-
   ci-complete:
-    needs: [check-changes, ros, python, ros-python, dev, ros-dev, run-tests, run-heavy-tests, run-lcm-tests, run-integration-tests, run-ros-tests, run-mypy]
+    needs: [check-changes, ros, python, ros-python, dev, ros-dev, run-ros-tests, run-mypy]
```
leshy, line 235: just to know what the plan was with ros/non ros parallel tests - I planned to treat ros build as a separate OS, since it does heavy intervention into the OS itslef, and wanted to make sure that unguarded ros imports don't crash non-ros machines, so potentially we want to re-introduce

leshy (dismissed):
This is so amazing, thanks Paul

leshy:
Paul is codebase Jesus, suffering for our sins

leshy (approved):

.github/workflows/docker.yml
paul-nechifor: It makes sense to test that, but maybe not on every PR? Maybe we can have a periodic runner which runs all the tests on a variety of environments like Ubuntu 22.04/24.04, with ROS/without ROS, MacOS, etc.

PR #1346, opened 2026-02-22
Add ClockSyncConfigurator for Unitree WebRTC

### LINKED ISSUE #1344: Add clock sync system configurator for Unitree WebRTC
## Problem

Unitree WebRTC connections require tight clock synchronization between the control computer and robot (within 100ms). Clock divergence affects:
- Command timestamps
- Sensor data correlation
- Control loop timing
- Telemetry synchronization

Currently, there's no automated check or sync for clock drift before connecting to hardware.

---

## Proposed Solution

Add `ClockSyncConfigurator` to the system configurator framework (same pattern as LCM's multicast/buffer checks).

### Implementation

**File:** `dimos/protocol/service/system_configurator.py`

```python
class ClockSyncConfigurator(SystemConfigurator):
    """Check and sync local clock with NTP server."""
    
    critical = False
    MAX_DIVERGENCE_SECONDS = 0.1  # 100ms for WebRTC
    NTP_SERVER = "time.windows.com"
    
    def check(self) -> bool:
        # Use ntpdate -q to query clock offset
        # Return False if divergence > threshold
    
    def explanation(self) -> str | None:
        # Show offset in ms and sync command
    
    def fix(self) -> None:
        # Run: sudo ntpdate <server>
```

**Reference implementation:** Available in proposal docs (attached or can provide)

---

## Integration Challenge

**Problem:** Modules/connections run in **separate runner processes without STDIO access**. Can't prompt for user input during module/connection initialization.

**Current LCM pattern:**
- `lcmservice.autoconf()` runs at **CLI startup** (before blueprints)
- Has STDIO access for user prompts
- Global/early execution

**Naive approach (won't work):**
```python
class UnitreeWebRTCConnection:
    def __init__(self, ip):
        self.autoconf()  # ❌ No STDIO in runner process
        # ... connect ...
```

**Required approach:**
- Collect system config requirements **before spawning runner processes**
- Run all configurators with user prompts **at CLI level**
- Then proceed with blueprint execution

---

### Option 1: Blueprint-level system config registration

Add `system_config` field to blueprints:

```python
# In unitree_go2_blueprints.py
def unitree_go2(ip=None, sim=False, ...):
    system_checks = []
    if not sim and ip not in ["fake", "mock", "replay", "mujoco"]:
        system_checks.append(
            ClockSyncConfigurator(max_divergence=0.1)
        )
    
    return Blueprint(
        modules=[GO2Connection(ip=ip), ...],
        system_config=system_checks  # ← new field
    )
```

Then in blueprint startup (before runner spawns):
```python
configure_system(blueprint.system_config)  # has STDIO, can prompt
# ... spawn runners, execute blueprint ...
```

**Pros:**
- Declarative (clear what system config a blueprint needs)
- Runs before runner processes spawn
- Has STDIO for user prompts
- Conditional per blueprint (hardware vs sim)

**Cons:**
- Requires Blueprint infrastructure changes
- Need to add system_config field + collection/execution logic

---

### Option 2: Module metadata + collection phase

Add classmethod to modules that need system config:

```python
class UnitreeWebRTCConnection:
    @classmethod
    def system_requirements(cls) -> list[SystemConfigurator]:
        return [ClockSyncConfigurator(max_divergence=0.1)]
```

Before executing blueprint, scan modules and collect requirements:
```python
# In blueprint execution
requirements = []
for module in blueprint.modules:
    if hasattr(module.__class__, 'system_requirements'):
        requirements.extend(module.__class__.system_requirements())

configure_system(requirements)  # run before spawning runners
```

**Pros:**
- Less invasive (no Blueprint schema changes)
- Modules declare their own requirements

**Cons:**
- Harder to make conditional (hardware vs sim)
- Requires scanning all modules

---

## Acceptance Criteria

- [ ] `ClockSyncConfigurator` implemented in `system_configurator.py`
- [ ] Unit tests for clock sync check/fix
- [ ] Integration strategy chosen and implemented
- [ ] System config runs **before** runner processes spawn
- [ ] User prompted with STDIO access
- [ ] Works for hardware Go2/G1 (skipped for sim/replay)
- [ ] Tested on actual hardware with intentional clock skew

---

## Related

- Existing system config framework: `dimos/protocol/service/system_configurator.py`
- LCM autoconf example: `dimos/protocol/service/lcmservice.py`
- Runner process architecture: (need docs reference)

---

## Implementation Status

- ✅ `ClockSyncConfigurator` class designed and documented
- ⏸️  Integration blocked on runner process STDIO constraint
- ⏸️  Needs blueprint-level or pre-execution collection strategy

leshy:
## Problem

Closes #1344
Closes DIM-575

When running Unitree robots over WebRTC, clock drift between the host and robot causes timestamp mismatches. There was no automated check or fix for this.

## Solution

- `ClockSyncConfigurator`  pure py SNTP query detects clock offset (threshold: ±100ms). NTP unreachable -> gracefully warns (works offline). tries `ntpdate` > `sntp` > `date -s` fallback.

- `system_checks()` bridge - wraps `SystemConfigurator`(s) into a `Blueprint.requirements()` compatible callable

- `system_configurator` refactored into a package - `base.py` (framework), `lcm.py` (multicast/buffer), `clock_sync.py` (new).

- `dimos/utils/human.py` - consolidated 6 scattered duration/byte/number formatters into `human_duration()`, `human_bytes()`, `human_number()`.

- Integrated into `unitree_go2_basic` and `unitree_g1_basic` hardware blueprints.

## Breaking Changes

None

## How to Test

```bash
# offset system clock, run configurator
sudo date -s "+1 second"
dimos --replay run unitree-go2
```

## Issues

configurator API needs some improvements
- always print comands that will be executed in a consistent way (no per-configurator str building)
- OS checks belong in the configurator itself, inconvinent to check for OS and use different configurators externally
- configurators should ideally be functions or worst case callables so we can compose etc naturally (feed into blueprint.requires for example)


dimos/core/blueprints.py
```diff
@@ -203,16 +208,35 @@ def _is_name_unique(self, name: str) -> bool:
...
+        if configurators:
+            try:
+                configure_system(configurators)
+            except SystemExit:
+                labels = [type(c).__name__ for c in configurators]
+                print(
+                    f"Required system configuration was declined: {', '.join(labels)}",
+                    file=sys.stderr,
+                )
+                sys.exit(1)
```
paul-nechifor: configuring and checking requirements seem different. I think you should add:

```
autoconnect(
    ...
).configurators(SystemConfigurator())
```

dimos/protocol/service/system_configurator/base.py
```diff
@@ -0,0 +1,136 @@
...
+
+    if check_only:
+        return
+
+    try:
+        answer = input("Apply these changes now? [y/N]: ").strip().lower()
+    except (EOFError, KeyboardInterrupt):
+        answer = ""
+
+    if answer not in ("y", "yes"):
```
paul-nechifor:
```suggestion
    if typer.confirm("Apply these changes now?"):
```

dimos/protocol/service/system_configurator/base.py
```diff
@@ -0,0 +1,136 @@
...
+        if any(check.critical for check in failing):
+            raise SystemExit(1)
+        return
+
+    for check in failing:
+        try:
+            check.fix()
+        except subprocess.CalledProcessError as error:
+            if check.critical:
+                print(f"Critical fix failed rc={error.returncode}")
```
paul-nechifor: I think you should use logger not `print` in this file.

paul-nechifor (dismissed):

dimos/core/blueprints.py
leshy: I considered this and it would have been easier but I always like to have less categories then more.

I consider this a category of "function you call - it does either checks or config, if it errors out stuff is not good to run" so this can be pings, clock, network checks, LFS checks etc. From a blueprint perspective that's the way you want to think about the category imo, no need to know what's inside the functions.

Only ugliness here is that configurators are not functions but these classes. I considered keeping the wrapper to turn them into functions external to blueprints.py but this complicates the user facing API.

like 
```py
blueprint.requirements(configurator(Clock, Network)) 
```

There are a few more things with configurators to refactor, I'd like to turn them info functions, so I expect this code to go away from blueprints.. how does this sound?

paul-nechifor (dismissed):

dimos/core/blueprints.py
leshy: Paul owns blueprints so his taste decides here, I didn't sell my idea so implemented his ask :D

paul-nechifor (dismissed):

paul-nechifor (approved):

paul-nechifor (approved):

PR #1356, opened 2026-02-24
Tiny Fix: Handle the non-tty case for sysconf

jeff-hykin:
## Problem

Can break inside docker and other automated envs

## Solution

5 line change to check if stdin is tty

## Breaking Changes

None

## How to Test

Reboot or change multicast, then run any dimos blueprint with lcm


paul-nechifor (dismissed):

dimos/protocol/service/test_system_configurator.py
```diff
@@ -163,31 +163,39 @@ def test_check_only_mode_does_not_fix(self) -> None:
...
         with patch.dict(os.environ, {"CI": ""}, clear=False):
             mock_check = MockConfigurator(passes=False)
-            with patch("builtins.input", side_effect=EOFError):
-                configure_system([mock_check])
-                assert not mock_check.fix_called
+            with patch("sys.stdin") as mock_stdin:
+                mock_stdin.isatty.return_value = True
+                with patch("builtins.input", side_effect=EOFError):
+                    configure_system([mock_check])
+                    assert not mock_check.fix_called
```
paul-nechifor: You can use the `mocker` fixture to not need to indent so much. It automatically cleans up when it's done.

```suggestion
    def test_handles_eof_error_on_input(self, mocker) -> None:
        mocker.patch.dict(os.environ, {"CI": ""}, clear=False)
        mock_check = MockConfigurator(passes=False)
        mock_stdin = mocker.patch("sys.stdin")
        mock_stdin.isatty.return_value = True
        mocker.patch("builtins.input", side_effect=EOFError)

        configure_system([mock_check])

        assert not mock_check.fix_called
```

Also, all of them use `mocker.patch.dict(os.environ, {"CI": ""}, clear=False)` so you can put that in a common mock so you don't have to add it in each test.

leshy (approved):

PR #1357, opened 2026-02-24
Feature: go2 control coordinator TwistBase adapter

### LINKED ISSUE #1323: Write Go2 ControlCoordinator Adapter

### LINKED ISSUE #1354: Rule of three: if you have three duplicate things, it's best to generalize. You can add:
Rule of three: if you have three duplicate things, it's best to generalize. You can add:

```
self._disposables = CompositeDisposable()
```
in `__init__` and do `self._disposables.dispose()` here.

_Originally posted by @paul-nechifor in https://github.com/dimensionalOS/dimos/pull/1277#discussion_r2842866406_

mustafab0:
## Problem

The ControlCoordinator ocannot control unitree go2 using Twist (cmd_vel)

---

## Solution

Added `UnitreeGo2Adapter` implementing the `TwistBaseAdapter` protocol, enabling the Go2 quadruped to be driven through the ControlCoordinator. Key changes:

- **`UnitreeGo2Adapter`**: Wraps Unitree SDK2 `SportClient` with DDS state feedback. Reads actual velocities from `SportModeState` (not command echo). Handles the Go2's init sequence (StandUp → FreeWalk → Move).
- **`HardwareAdapter` base protocol**: Extracted `connect`/`disconnect`/`is_connected` into a shared base so `ManipulatorAdapter` and `TwistBaseAdapter` share a common ancestor, eliminating `type: ignore[override]` on the adapter property.
- **`unitree-go2-keyboard-teleop` blueprint**: Dual-mode — simulation uses `go2_connection` (MujocoConnection), real hardware uses ControlCoordinator with the Go2 adapter.
- **Coordinator cleanup**: Refactored subscription management into `_subscribe_if()` with list-based cleanup. Gripper methods use `isinstance` narrowing for `ManipulatorAdapter`.

---

## Breaking Changes

None

---

## How to Test

**Simulation ( ONLY verifies blueprint wiring, keyboard input, MuJoCo physics — does NOT exercise the coordinator or Go2 adapter):**

1. `dimos --simulation run unitree-go2-keyboard-teleop`
2. Use WASD/QE to drive the simulated Go2. Verify robot responds to input.

**Real hardware (verifies the full coordinator → adapter → SDK path):**

1. Install CycloneDDS C library (required by `unitree-sdk2py`):
   ```bash
   sudo apt install cmake gcc g++
   git clone https://github.com/eclipse-cyclonedds/cyclonedds.git /tmp/cyclonedds
   cd /tmp/cyclonedds && git checkout releases/0.10.x
   mkdir build && cd build
   cmake .. -DCMAKE_INSTALL_PREFIX=/usr/local
   cmake --build . --parallel
   sudo cmake --install .
   
2. `dimos run unitree-go2-keyboard-teleop`
3. Use WASD/QE to drive the simulated Go2. Verify robot responds to input.
4. to verify if topics are published correctly
```
dimos topic echo /coordinator/joint_state
dimos topic echo /cmd_vel
```

closes DIM-550


dimos/control/coordinator.py
```diff
@@ -628,12 +629,29 @@ def get_gripper_position(self, hardware_id: str) -> float | None:
...
+            return
+        try:
+            unsub = port.subscribe(callback)
+            self._subscriptions.append(unsub)
+            logger.info(f"Subscribed to {name}")
+        except Exception:
+            logger.warning(
+                f"Could not subscribe to {name}. "
+                "Use task_invoke RPC or set transport via blueprint."
+            )
```
paul-nechifor, line 642: I see this pattern a lot, but I don't understand it. If the subscription fails to happen, the entire system is borked, no? There's no point in continuing, so you might as well let the exception happen.

dimos/robot/unitree/go2/README.md
```diff
@@ -0,0 +1,147 @@
...
+The recommended approach is via pip — the wheel bundles the C++ library:
+
+```bash
+uv pip install "cyclonedds>=0.10.5"
+```
+
+If the prebuilt wheel doesn't work on your system (e.g. unsupported
+architecture), fall back to building from source:
+
+```bash
```
paul-nechifor: Instructions exist in docs/usage/transports/dds.md . This shouldn't be in Go2.

dimos/hardware/drive_trains/unitree_go2/adapter.py
```diff
@@ -0,0 +1,348 @@
...
+
+import logging
+import threading
+import time
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from dimos.hardware.drive_trains.registry import TwistBaseAdapterRegistry
+
+logger = logging.getLogger(__name__)
```
paul-nechifor:
```suggestion
logger = setup_logger()
```

dimos/control/examples/echo_cmd_vel.py
```diff
@@ -0,0 +1,55 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Echo Twist messages on /cmd_vel.
+
+Usage:
+    python -m dimos.control.examples.echo_cmd_vel
+    python -m dimos.control.examples.echo_cmd_vel --topic /my_cmd_vel
```
paul-nechifor: This file is not needed. We already have this.

dimos/hardware/drive_trains/unitree_go2/adapter.py
```diff
@@ -0,0 +1,348 @@
...
+
+
+class UnitreeGo2Adapter:
```
paul-nechifor: A lot of this code is quite repetitive and complex because it has to deal with the same situation over and over again.

The problem is that UnitreeGo2Adapter has `self._client`, has `self._connected`, etc. But these variables don't change independently.

If UnitreeGo2Adapter were to take a UnitreeGo2Session (containing a client that's already connected) it wouldn't have to deal with the client not existing in every method.

But that would mean changing the established API...

But you can still simplify code a bit by moving all the session variables together. Something like:

```python
  @dataclass                                                                                                                                                                                                                                                                                                                                                                 
  class _Session:                                                                                                                                                                                                                                                                                                                                                            
      client: SportClient                                                                                                                                                                                                                                                                                                                                                    
      state_sub: ChannelSubscriber                                                                                                                                                                                                                                                                                                                                           
      lock: threading.Lock                                                                                                                                                                                                                                                                                                                                                   
      latest_state: SportModeState_ | None = None                                                                                                                                                                                                                                                                                                                            
      enabled: bool = False                                                                                                                                                                                                                                                                                                                                                  
      locomotion_ready: bool = False                                                                                                                                                                                                                                                                                                                                         
      last_velocities: list[float] = field(default_factory=lambda: [0.0, 0.0, 0.0])   
```

Then you can have:

```python
  class UnitreeGo2Adapter:                                                                                                                                                                                                                                                                                                                                                   
      def __init__(...):                                                                                                                                                                                                                                                                                                                                                     
          self._session: _Session | None = None                                                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                                                                             
      def connect(...):
          self._session = ...                                                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                             
      def _get_session(self) -> _Session:                                                                                                                                                                                                                                                                                                                                
          if self._session is None:                                                                                                                                                                                                                                                                                                                                          
              raise RuntimeError("Go2 not connected")                                                                                                                                                                                                                                                                                                                        
          return self._session  
```

And you don't have to check if session is set:

```python
    def _send_velocity(self, vx: float, vy: float, wz: float) -> bool:
        ret = self._get_session().client.Move(vx, vy, wz)
        if ret != 0:
            logger.warning(f"Move() returned error code {ret}")
            return False
```

dimos/hardware/drive_trains/unitree_go2/adapter.py
mustafab0: My bad!

dimos/control/examples/echo_cmd_vel.py
mustafab0: Oh yes, I removed it in the previous PR, but it was rebased off of that, and forgot to delete in this one.

dimos/hardware/drive_trains/unitree_go2/adapter.py
```diff
@@ -0,0 +1,348 @@
...
+                return [0.0, 0.0, 0.0]
+            try:
+                state = self._latest_state
+                return [
+                    float(state.velocity[0]),  # vx
+                    float(state.velocity[1]),  # vy
+                    float(state.imu_state.gyroscope[2]),  # wz (yaw rate)
+                ]
+            except Exception:
+                return [0.0, 0.0, 0.0]
```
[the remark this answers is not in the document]
mustafab0: Fixed

dimos/control/coordinator.py
mustafab0: Removed exception, now coordinator catches subscription failure and does not start.

dimos/robot/unitree/go2/README.md
mustafab0: I updated readme to point to dds.md and cut it down quite a bit.

wanted this to be a quickstart readme. Should I delete completely?

paul-nechifor: No, I think you can keep it, but what I'm not sure about is this being the main readme for Go2. The homepage links to this page and most of our Go2 stuff uses WebRTC.

Maybe put this in `go2/dds.md`? What do you think @leshy ?

mustafab0: removed all dds related things to the existing dds.md in the docs.

This readme is not just a quick starter to load the keyboard teleop for the go2

leshy: Yeah I find it pretty confusing, this looks like a main readme for go2 given the dir and name, and then just starts with  "Velocity control for the Unitree Go2 quadruped via the ControlCoordinator."

And goes into DDS and is effectively a dead end for 99% of users that want dimos and go2.  (no lidar, requires root etc, claims "Required for DDS communication with the Go2.")

(our actual go2 readme this might be mistaken with is at https://github.com/dimensionalOS/dimos/blob/main/docs/platforms/quadruped/go2/index.md)

So I'd move all of this into some experimental dir or different filename, related to DDS, velocity control and go2? 

Actually, you could start a new pro/edu go2 readme, mark it very clearly at the start that this is a draft/beginning/wip, later we can fill in with other details, like backpack 3d print files, livox vs native etc

mustafab0: updated readme to mark as experimental and also moved to docs/development/

dimos/robot/unitree/go2/blueprints/basic/unitree_go2_keyboard_teleop.py
```diff
@@ -0,0 +1,84 @@
...
+from dimos.robot.unitree.keyboard_teleop import keyboard_teleop
+
+if global_config.simulation:
+    from dimos.robot.unitree.go2.connection import go2_connection
+
+    # go2_connection automatically uses MujocoConnection when simulation=True.
+    # keyboard_teleop.cmd_vel is wired directly to go2_connection.cmd_vel
+    # by autoconnect — no LCM needed.
+    _go2 = go2_connection()
+    _teleop = keyboard_teleop()
```
paul-nechifor: This already exists as `dimos --simulation run go2-connection keyboard-teleop`.

Running `dimos --simulation run unitree-go2-keyboard-teleop` implies that the `unitree-go2-keyboard-teleop` modules are being simulated, but it's just running an entirely different set of modules.

I would just remove this, or add a sim for `control_coordinator`.

dimos/hardware/manipulators/spec.py
```diff
@@ -89,7 +90,7 @@ def default_base_transform() -> Transform:
 
 
 @runtime_checkable
-class ManipulatorAdapter(Protocol):
+class ManipulatorAdapter(HardwareAdapter, Protocol):
```
paul-nechifor, line 88: Since it now inherits `HardwareAdapter` you can remove this from here:

```
    # --- Connection ---

    def connect(self) -> bool:
        """Connect to hardware. Returns True on success."""
        ...

    def disconnect(self) -> None:
        """Disconnect from hardware."""
        ...

    def is_connected(self) -> bool:
        """Check if connected."""
        ...
```

dimos/hardware/drive_trains/spec.py
```diff
@@ -25,7 +25,24 @@
...
+        """Disconnect from hardware."""
+        ...
+
+    def is_connected(self) -> bool:
+        """Check if connected."""
+        ...
+
+
+@runtime_checkable
+class TwistBaseAdapter(HardwareAdapter, Protocol):
```
paul-nechifor, line 45: Also remove connect/disconnect/is_connected from here.

dimos/hardware/drive_trains/unitree_go2/adapter.py
```diff
@@ -0,0 +1,341 @@
...
+    from dimos.hardware.drive_trains.registry import TwistBaseAdapterRegistry
+
+logger = setup_logger()
+
+
+@dataclass
+class _Session:
+    """Active connection state for a Go2."""
+
+    client: Any  # SportClient
```
paul-nechifor: Why Any instead of SportClient?

dimos/hardware/drive_trains/unitree_go2/adapter.py
```diff
@@ -0,0 +1,341 @@
...
+        if session is not None:
+            try:
+                session.client.StopMove()
+                time.sleep(0.5)
+
+                logger.info("Standing down Go2...")
+                session.client.StandDown()
+                time.sleep(2)
+            except Exception as e:
+                logger.error(f"Error during disconnect: {e}")
```
paul-nechifor, line 159: I think you're missing:

```
          if session.state_sub is not None:
              try:
                  session.state_sub.Close()
              except Exception as e:
                  logger.error(f"Error closing state subscriber: {e}")
```

dimos/hardware/manipulators/spec.py
mustafab0: I see! Done

dimos/robot/unitree/go2/blueprints/basic/unitree_go2_keyboard_teleop.py
mustafab0: removed

dimos/hardware/drive_trains/unitree_go2/adapter.py
mustafab0: Because unitree sdk is an optional dependency.

but makes sense to be more explicit. Will update

jeff-hykin:
> Simulation ( ONLY verifies blueprint wiring, keyboard input, MuJoCo physics — does NOT exercise the coordinator or Go2 adapter):

Is this because the control coordinator is basically lower level than cmd_vel (which is where the simulator abstracts away from hardware)?

mustafab0:
> > Simulation ( ONLY verifies blueprint wiring, keyboard input, MuJoCo physics — does NOT exercise the coordinator or Go2 adapter):
> 
> Is this because the control coordinator is basically lower level than cmd_vel (which is where the simulator abstracts away from hardware)?
Not in this case, we are still using the high level API.

The current MuJoCo implementation directly uses the UnitreeWebRTC to send commands to the simulated robot. It would need a bit of rework for it to work with the control coordinator. That would be out of scope for this PR (in the interest of keeping the PR succinct) 

I would like to address this as a bigger overhaul on the simulation development.

paul-nechifor:
> The current MuJoCo implementation directly uses the UnitreeWebRTC to send commands to the simulated robot.

No, MujocoConnection is a drop in replacement for UnitreeWebRTCConnection. It does not use it.

mustafab0:
> No, MujocoConnection is a drop in replacement for UnitreeWebRTCConnection. It does not use it.

Right, this makes sense. its been sometime working on this, I misremembered.

dimos/control/coordinator.py
```diff
@@ -639,12 +638,34 @@ def get_gripper_position(self, hardware_id: str) -> float | None:
...
+        self, condition: bool, port: In[Any], callback: Callable[..., object], name: str
+    ) -> None:
+        """Subscribe to a port if condition is met, tracking the unsub handle."""
+        if not condition:
+            return
+        try:
+            unsub = port.subscribe(callback)
+            self._subscriptions.append(unsub)
+            logger.info(f"Subscribed to {name}")
+        except Exception:
```
paul-nechifor, line 635: If "subscribe() can fail when no transport is wired to the port" then that should be the condition. You shouldn't just catch any exception and assume it's because 'no transport is wired'. Just check for that condition.

jeff-hykin (commented):
After in-person comments and changes, looks good to me!

dimos/control/coordinator.py
jeff-hykin: but the try-catch intentionally handles any kind of problem though, even as the implementation of .subscribe changes. Otherwise if there's an error I think it just gets eaten

paul-nechifor: > but the try-catch handles any kind of problem though

But, yeah, that's the issue. This hides problems even when it should work.

Example, if `self._subscriptions` is changed to be a CompositeDisposable, then `self._subscriptions.append` will fail with method doesn't exist (since you have to use `.add`), but this will work like there's no issue. We'll just wonder why shutdown doesn't happen correctly.

paul-nechifor:

PR #1362, opened 2026-02-25
Feature:  go2 webrtc TwistBase adapter for control coordinator

### LINKED ISSUE #1360: Go2 webRTC adapter for control coordinator

mustafab0:
### Problem
The ControlCoordinator cannot be used with the Unitree Go2 over WebRTC because the single WebRTC connection is shared between sensor streaming (GO2Connection) and actuation. The previous workaround used a mock adapter, which broke the coordinator's deterministic write contract.

### Solution
Added a TransportTwistAdapter — a generic, robot-agnostic adapter that connects the coordinator to any driver module via LCM pub/sub topics. The coordinator writes cmd_vel and reads odom through LCM transports instead of calling hardware directly.

For the Go2, GO2Connection owns the WebRTC connection and handles both sensor publishing and actuation. The coordinator communicates with it through LCM topics:

Keyboard/Nav/Agent → /cmd_vel → Coordinator → TransportTwistAdapter
                                → /go2/cmd_vel → GO2Connection → connection.move()
### Key changes:

New adapter: dimos/hardware/drive_trains/transport/adapter.py — registered as "transport_lcm" (and "transport_ros" when ROS is available). Topics derived from hardware_id: /{hardware_id}/cmd_vel, /{hardware_id}/odom.
New blueprints: unitree_go2_coordinator and unitree_go2_webrtc_keyboard_teleop — wire GO2Connection + coordinator with transport adapter, using remappings to prevent autoconnect from bypassing the coordinator.
MuJoCo fix: dimos/simulation/mujoco/policy.py — ONNX falls back to CPU when GPU providers fail.
Connection fixes: free_walk(), stop_movement() rename, balance_stand() docstring on UnitreeWebRTCConnection.

### Breaking Changes
None

### How to Test
**Real robot (WebRTC)**:
export ROBOT_IP=192.168.12.1
dimos run unitree-go2-webrtc-keyboard-teleop

**MuJoCo simulation (no robot):**
dimos --simulation run unitree-go2-webrtc-keyboard-teleop

Coordinator only (no teleop):
dimos run unitree-go2-coordinator

closes DIM-599


mustafab0:
@greptile New architectural changes have been made. Can you redraw the diagram?

dimos/robot/unitree/go2/blueprints/basic/unitree_go2_coordinator.py
```diff
@@ -0,0 +1,76 @@
...
+        ]
+    )
+    .transports(
+        {
+            ("cmd_vel", Twist): LCMTransport("/cmd_vel", Twist),
+            ("twist_command", Twist): LCMTransport("/cmd_vel", Twist),
+            ("go2_cmd_vel", Twist): LCMTransport("/go2/cmd_vel", Twist),
+            ("go2_odom", PoseStamped): LCMTransport("/go2/odom", PoseStamped),
+            ("joint_state", JointState): LCMTransport("/coordinator/joint_state", JointState),
+        }
```
paul-nechifor, line 71: I don't think any of these need to be changed except `joint_state` maybe? `/cmd_vel` and `/odom` work just fine as defaults. There's no need to explicitly state them/remap them.

Also, I'm not sure why there are 3 cmd_vels here. Two emit to `/cmd_vel` and one to `/go2/cmd_vel`.

dimos/hardware/drive_trains/transport/adapter.py
```diff
@@ -0,0 +1,195 @@
...
+            self._cmd_vel_transport = self._transport_cls(cmd_vel_topic, Twist)
+            self._odom_transport = self._transport_cls(odom_topic, PoseStamped)
+            self._odom_unsub = self._odom_transport.subscribe(self._on_odom)
+
+            self._connected = True
+            logger.info(
+                f"TransportTwistAdapter connected: cmd_vel={cmd_vel_topic}, odom={odom_topic}"
+            )
+            return True
+        except Exception as e:
```
paul-nechifor: It's not clear what error you're expecting to have here. Certain things definitely won't fail, like:

```
            cmd_vel_topic = f"/{self._prefix}/cmd_vel"
            odom_topic = f"/{self._prefix}/odom"
```

And this is just initializing objects:

```
            self._cmd_vel_transport = self._transport_cls(cmd_vel_topic, Twist)
            self._odom_transport = self._transport_cls(odom_topic, PoseStamped)
```

So I assume only this line is expected to fail:

```
self._odom_unsub = self._odom_transport.subscribe(self._on_odom)
```

But with what error?

dimos/hardware/drive_trains/transport/adapter.py
```diff
@@ -0,0 +1,195 @@
...
+
+        with self._lock:
+            self._last_velocities = list(velocities)
+
+        twist = Twist(
+            linear=Vector3(x=velocities[0], y=velocities[1] if self._dof > 1 else 0.0, z=0.0),
+            angular=Vector3(x=0.0, y=0.0, z=velocities[2] if self._dof > 2 else 0.0),
+        )
+        try:
+            self._cmd_vel_transport.broadcast(None, twist)
```
paul-nechifor: I don't know what None in broadcast is supposed to do. You're probably better off just calling:
```suggestion
            self._cmd_vel_transport.publish(twist)
```

dimos/hardware/drive_trains/transport/adapter.py
```diff
@@ -0,0 +1,195 @@
...
+            return True
+        except Exception as e:
```
paul-nechifor: What error are you expecting to happen?

If an error does occur when publishing, how does this handle it?

If the error is unhandled, the system crashes. But when you handle it and just log `logger.error`, the condition for the failure will probably stay the same and it will just end up spamming `TransportTwistAdapter write error:` in a loop.

I think it's better to either:

* do not handle the error
* handle it and take appropriate measures, like shutting down the system.

---

As far as I can see, `write_velocities` is called in ConnectedTwistBase.write_command, and it is called here:

```python
    def _write_all_hardware(
        self,
        hw_commands: dict[str, tuple[dict[str, float], ControlMode]],
    ) -> None:
        """Write commands to all hardware interfaces."""
        with self._hardware_lock:
            for hw_id, (positions, mode) in hw_commands.items():
                if hw_id in self._hardware:
                    try:
                        self._hardware[hw_id].write_command(positions, mode)
                    except Exception as e:
                        logger.error(f"Failed to write to {hw_id}: {e}")
```

The return value of `write_command` is never used. But it is surrounded by `try-except Exception`. But that will never happen because every possible `Exception` is already captured here.

dimos/hardware/drive_trains/transport/adapter.py
```diff
@@ -0,0 +1,195 @@
...
+    registry.register("transport_lcm", partial(TransportTwistAdapter, transport_cls=LCMTransport))
+
+    try:
+        from dimos.core.transport import ROSTransport
+
+        registry.register(
+            "transport_ros", partial(TransportTwistAdapter, transport_cls=ROSTransport)
+        )
+    except ImportError:
+        pass
```
paul-nechifor: An import error seems quite bad. I don't think you should just ignore it.

As far as I can see, only `DDSTransport` does not work if a package is missing. But looks like `ROSTransport` should always import.

dimos/hardware/drive_trains/transport/adapter.py
```diff
@@ -0,0 +1,195 @@
...
+            self._odom_unsub = None
+
+        for attr in ("_cmd_vel_transport", "_odom_transport"):
+            transport = getattr(self, attr, None)
+            if transport is not None:
+                try:
+                    transport.stop()
+                except Exception:
+                    pass
+                setattr(self, attr, None)
```
paul-nechifor: I think that `getattr` (and the others) should be used sparingly because it's too dynamic. You lose the help that the type checker provides.

```suggestion
        if self._cmd_vel_transport is not None:
            self._cmd_vel_transport.stop()
            self._cmd_vel_transport = None
        if self._odom_transport is not None:
            self._odom_transport.stop()
            self._odom_transport = None
```

dimos/simulation/mujoco/policy.py
```diff
@@ -40,7 +40,12 @@ def __init__(
         drift_compensation: list[float] | None = None,
     ) -> None:
         self._output_names = ["continuous_actions"]
-        self._policy = ort.InferenceSession(policy_path, providers=ort.get_available_providers())
+        providers = ort.get_available_providers()
+        try:
+            self._policy = ort.InferenceSession(policy_path, providers=providers)
+        except RuntimeError:
+            logger.warning("GPU providers failed, falling back to CPUExecutionProvider")
+            self._policy = ort.InferenceSession(policy_path, providers=["CPUExecutionProvider"])
```
paul-nechifor, line 48: Isn't this done automatically? I get this normally:

```
*************** EP Error ***************
EP Error /onnxruntime_src/onnxruntime/python/onnxruntime_pybind_state.cc:539 void onnxruntime::python::RegisterTensorRTPluginsAsCustomOps(PySessionOptions&, const onnxruntime::ProviderOptions&) Please install TensorRT libraries as mentioned in the GPU requirements page, make sure they're in the PATH or LD_LIBRARY_PATH, and that your GPU is supported.
 when using ['TensorrtExecutionProvider', 'CUDAExecutionProvider', 'CPUExecutionProvider']
Falling back to ['CUDAExecutionProvider', 'CPUExecutionProvider'] and retrying.
****************************************
```

On your branch, I get the same text because the code does not raise `RuntimeError`. Does it do that for you?

dimos/robot/unitree/go2/blueprints/basic/unitree_go2_webrtc_keyboard_teleop.py
```diff
@@ -0,0 +1,36 @@
...
+
+from __future__ import annotations
+
+from dimos.core.blueprints import autoconnect
+from dimos.robot.unitree.go2.blueprints.basic.unitree_go2_coordinator import (
+    unitree_go2_coordinator,
+)
+from dimos.robot.unitree.keyboard_teleop import KeyboardTeleop
+
+unitree_go2_webrtc_keyboard_teleop = autoconnect(
```
paul-nechifor, line 31: Shouldn't this be `unitree_go2_coordinator_keyboard_teleop`?

dimos/hardware/drive_trains/transport/adapter.py
```diff
@@ -0,0 +1,195 @@
...
+            qx = float(msg.orientation.x)
+            qy = float(msg.orientation.y)
+            qz = float(msg.orientation.z)
+            qw = float(msg.orientation.w)
+            yaw = math.atan2(2.0 * (qw * qz + qx * qy), 1.0 - 2.0 * (qy * qy + qz * qz))
+
+            with self._lock:
+                self._latest_odom = [x, y, yaw]
+        except Exception as e:
+            logger.warning(f"TransportTwistAdapter odom error: {e}")
```
paul-nechifor:
```suggestion
        with self._lock:
            self._latest_odom = [msg.x, msg.y, msg.yaw]
```

This is why doing code reviews is useful, BTW. I found out about `.yaw` existing from seeing code that used that.

dimos/robot/unitree/go2/blueprints/basic/unitree_go2_coordinator.py
```diff
@@ -0,0 +1,76 @@
...
+from dimos.msgs.geometry_msgs.Twist import Twist
+from dimos.msgs.sensor_msgs.JointState import JointState
+from dimos.robot.unitree.go2.connection import GO2Connection
+
+_go2_joints = make_twist_base_joints("go2")
+
+unitree_go2_coordinator = (
+    autoconnect(
+        GO2Connection.blueprint(),
+        ControlCoordinator.blueprint(
```
paul-nechifor, line 39: I don't understand what ControlCoordinator actually does here. GO2Connection already listens to `/cmd_vel`.

dimos/robot/unitree/go2/blueprints/basic/unitree_go2_coordinator.py
mustafab0: If I don't use explicit remapping then auto connect simply connects /cmd_vel from keyboard tele directly to the GO2Connection.

For the /odom, there is actually no need to remap it and it can work as is, but I am just experimenting here with remapping as this would come in handy when we are controlling 2+ robot using a single blueprint. For swarm robotics applications etc.

dimos/robot/unitree/go2/blueprints/basic/unitree_go2_webrtc_keyboard_teleop.py
mustafab0: No this is specifically for the webrtc.

United WebRTC implementation does not allow multiple connections. Since the webRTC connection needs to be managed seperately as it handles sensor communications and motor commands using a single client.

dimos/hardware/drive_trains/transport/adapter.py
mustafab0: Fixed

dimos/hardware/drive_trains/transport/adapter.py
mustafab0: removed all try/except blocks

mustafab0:
@greptile review

dimos/hardware/drive_trains/transport/adapter.py
mustafab0: removed try/except

dimos/hardware/drive_trains/transport/adapter.py
mustafab0: explicit rostransport imported and remove try except

dimos/hardware/drive_trains/transport/adapter.py
mustafab0: removed getattr and used stop

dimos/hardware/drive_trains/transport/adapter.py
mustafab0: Simplified this as per suggestion

dimos/robot/unitree/go2/blueprints/basic/unitree_go2_coordinator.py
mustafab0: GO2Connection alone works for simple navigation, but the coordinator adds task arbitration — multiple tasks (navigation, manipulation, safety stops) can compete for the same hardware and the coordinator resolves conflicts per-joint by priority. Without it, you'd need manual FSM logic to prevent tasks from fighting over cmd_vel.

That said, whether this complexity is needed for the Go2 specifically is still an open question. This PR makes it possible to use the coordinator with WebRTC, it doesn't mandate it.

dimos/simulation/mujoco/policy.py
mustafab0: I get the runtime error: 
`RuntimeError: CUDA failure 35: CUDA driver version is insufficient for CUDA runtime version ; GPU=-1
`
My coda version is much older I am running a 4090 and haven't updated the Nvidia driver. Mujoco fails to launch in mine, and the CUDA to CPU fallback never kicks in.

dimos/robot/unitree/go2/blueprints/basic/unitree_go2_webrtc_keyboard_teleop.py
paul-nechifor: So then `unitree_go2_coordinator` needs to be renamed to `unitree_go2_webrtc`, right?

paul-nechifor (approved):

dimos/robot/unitree/go2/blueprints/basic/unitree_go2_webrtc_keyboard_teleop.py
mustafab0: You are right.

let me fix this in the follow up PR. I will be adding more go2 coordinator blueprints. will rename the blueprints more descriptively

PR #1364, opened 2026-02-25
Task: Create base manipulation module

mustafab0:
Problem

ManipulationModule is a 1600-line monolith mixing core planning infrastructure with pick/place behaviors, GraspGen integration, and perception subscriptions. This makes it impossible to use the planner without pulling in perception and grasp generation dependencies.

Additionally, _BlueprintAtom.create in blueprints.py fails to resolve In/Out port annotations for modules using from __future__ import annotations with TYPE_CHECKING guards (e.g. RealSenseCamera), silently breaking autoconnect transport wiring.

Issue: DIM-451
Solution

Split ManipulationModule into a base class and a PickAndPlaceModule subclass:

    ManipulationModule (base): Core planning infrastructure, motion execution, gripper control, state machine. Skills: move_to_pose, move_to_joints, go_home, go_init, get_robot_state, reset.
    PickAndPlaceModule(ManipulationModule): Adds perception integration (objects port, obstacle monitor, scan_objects, get_scene_info), GraspGen, and long-horizon skills (pick, place, place_back, pick_and_place, clear_perception_obstacles).

PickAndPlaceModuleConfig extends ManipulationModuleConfig with GraspGen fields. Only xarm_perception and xarm_perception_agent blueprints switch to pick_and_place_module; all other blueprints use the base manipulation_module unchanged.

New blueprints: xarm7_planner_coordinator_agent for testing base module with an LLM agent.

Bug fixes:

    move_to_pose now preserves current EE orientation when roll/pitch/yaw are omitted (previously forced identity quaternion).
    reset changed from @rpc to @skill so agents can clear faults.
    clear_perception_obstacles changed from @rpc to @skill for agent recovery from COLLISION_AT_START.
    _BlueprintAtom.create now uses reversed-MRO namespace resolution for get_type_hints, fixing autoconnect transport wiring for all modules with from __future__ import annotations.

Breaking Changes

None. ManipulationModule retains all base skills and RPCs. PickAndPlaceModule is a drop-in replacement where perception/pick-place is needed. Both are exported from dimos.manipulation.
How to Test

    Test base agent: dimos run coordinator-mock then dimos run xarm7-planner-coordinator-agent
    Test perception agent: dimos run coordinator-mock then dimos run xarm-perception-agent

closes DIM-451

### DIFF
--- a/dimos/manipulation/manipulation_module.py
+++ b/dimos/manipulation/manipulation_module.py
@@ -22,11 +22,13 @@ Subclass PickAndPlaceModule (pick_and_place_module.py) adds perception integrati
 (scan_objects, get_scene_info) and long-horizon skills (pick, place, pick_and_place).
 """
 
+from __future__ import annotations
+
 from dataclasses import dataclass, field
 from enum import Enum
 import threading
 import time
-from typing import Any, TypeAlias
+from typing import TYPE_CHECKING, Any, TypeAlias
 
 from dimos.agents.annotation import skill
 from dimos.core.core import rpc


dimos/core/blueprints.py
```diff
@@ -63,9 +63,16 @@ def create(
...
 
-        # Use get_type_hints() to properly resolve string annotations.
+        # Resolve annotations using namespaces from the full MRO chain so that
+        # In/Out behind TYPE_CHECKING + `from __future__ import annotations` work.
+        # Iterate reversed MRO so the most specific class's namespace wins when
+        # parent modules shadow names (e.g. spec.perception.Image vs sensor_msgs.Image).
+        globalns: dict[str, Any] = {}
+        for c in reversed(module.__mro__):
+            if c.__module__ in sys.modules:
+                globalns.update(sys.modules[c.__module__].__dict__)
```
paul-nechifor, line 76: This looks quite hack-y. What is the issue? I thought I fixed `from __future__ import annotations`. Can you show an example where it fails?

mustafab0: > PR #1309 from Feb 20 added from __future__ import annotations + TYPE_CHECKING guard for Out to the RealSenseCamera — to lazy-load pyrealsense2. That broke _BlueprintAtom.create because it couldn't resolve the camera's string annotations. 
> 
> The fix I just applied makes _BlueprintAtom.create use MRO-based namespace resolution (with reversed order so the defining class wins).


The object scene registration module does not capture the camera image streams. Above quote is from Claude, after workaround was implemented.

I agree with it being hacky >.<

dimos/manipulation/manipulation_module.py
```diff
@@ -1107,103 +949,80 @@ def _preview_execute_wait(
...
+
+        # If no orientation specified, preserve the current EE orientation
+        if roll is None and pitch is None and yaw is None:
+            current_pose = self.get_ee_pose(robot_name)
+            if current_pose is not None:
+                orientation = current_pose.orientation
+            else:
+                orientation = Quaternion(0, 0, 0, 1)  # identity fallback
+        else:
+            orientation = Quaternion.from_euler(Vector3(roll or 0.0, pitch or 0.0, yaw or 0.0))
```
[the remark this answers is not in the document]
mustafab0: fixed

dimos/core/blueprints.py
mustafab0: @paul-nechifor what do you suggest

paul-nechifor: @mustafab0 You can merge this. I've created https://linear.app/dimensional/issue/DIM-636/fix-get-type-hints-usage to handle this later.

mustafab0: Thank you!

Issue #1082, opened 2026-01-22
Config System


leshy:
- each module receives both global settings and module based settings
- subclassable, my_detector_module can subclass detector_module config and it's defaults
- defaults/module config defined per module in module
- typed, with automatic conversion (e.g. if I pass --mujoco-start-point '3.1 4.5' it should set the list [3.1, 4.5])
- pydantic or dataclasses?
- be able to have a local overrides file (e.g. like .env currently, but should be pluggable, .json overlay .yaml overlay stuff like this)
- be able to override each setting in a command line option
- easy way to see what the actual config of a blueprint is (get config from each module in the blueprint, assemble global yaml/json with values) 
- after I import global config from somewhere `from dimos.core.global_config import GlobalConfig` this already parsed everything (cli args etc) - since atm this is just a defaults dict modified dynamically by cli tooling

overlay example, multiple systems overlaying into the same config tree

base_config -> json dir -> env vars

I have my base config

```
camera:
   fps: 30
   dev: /dev/video0
```

I run the system with

`CAMERA_FPS=44 dimos run ....`
or
`dimos run --camera-fps=44 ....`

config is overlayed

```
camera:
   fps: 44
   dev: /dev/video0
```

we also want control over which overlays are used, like apply "mac" dir if mac is detected (to override transports etc)

alexlin2:
It should be dynamically reconfigurable like the https://wiki.ros.org/dynamic_reconfigure/Tutorials/SettingUpDynamicReconfigureForANode%28python%29

Kaweees:
Thoughts on using [Tyro](https://brentyi.github.io/tyro/) for command-line configuration?

leshy:
> Thoughts on using [Tyro](https://brentyi.github.io/tyro/) for command-line configuration?

Not against it on a brief look, I think hydra already does this as well, owner can think in detail and decide

> It should be dynamically reconfigurable like the https://wiki.ros.org/dynamic_reconfigure/Tutorials/SettingUpDynamicReconfigureForANode%28python%29

can be done as a subsystem above for later, not for the first pass since lots to think about even without this

PR #1369, opened 2026-02-25
Config adjustments

Dreamsorcerer:
## Problem

Working towards #1082, this helps standardise setting in modules which can then be expanded upon to enable further things such as CLI arguments to be injected. They are also now type checked at runtime via Pydantic.

## Solution

Primary change to developers is that the `__init__()` method now uses a fixed signature and all configurable options appear in the config. That config is now a pydantic model, so it gets verified when the module is instantiated.

A lot of typing fixes were also made by changing incorrect references to Module to ModuleBase and similar adjustments.

## Breaking Changes

Subclasses of Module must now match the `__init__()` signature of the Module class.


dimos/core/module.py
```diff
@@ -426,7 +435,7 @@ def __init_subclass__(cls, **kwargs: Any) -> None:
                 if not hasattr(cls, name) or getattr(cls, name) is None:
                     setattr(cls, name, None)
 
-    def __init__(self, *args, **kwargs) -> None:  # type: ignore[no-untyped-def]
+    def __init__(self, global_config: GlobalConfig = global_config, **kwargs: Any):
```
Dreamsorcerer, line 422: As an alternative to asking subclasses (that need a custom `__init__()`) to include global_config here, I could move it into the ModuleConfig itself? Then subclasses will just pass kwargs alone.

dimos/agents/skills/person_follow.py
```diff
@@ -60,38 +64,31 @@ class PersonFollowSkillContainer(Module):
...
-    ) -> None:
-        super().__init__()
-        self._global_config: GlobalConfig = cfg
-        self._use_3d_navigation: bool = use_3d_navigation
+    def __init__(self, global_config: GlobalConfig = global_config, **kwargs: Any) -> None:
+        super().__init__(global_config, **kwargs)
         self._latest_image: Image | None = None
         self._latest_pointcloud: PointCloud2 | None = None
-        self._vl_model: VlModel = QwenVlModel()
+        self._vl_model = QwenVlModel()
```
paul-nechifor: The reason for using `: VlModel` is so people use it through that Protocol. Otherwise they might call methods which exist on `QwenVlModel` but not `VlModel`.

dimos/navigation/visual/query.py
```diff
@@ -20,7 +21,7 @@
 
 
 def get_object_bbox_from_image(
-    vl_model: VlModel, image: Image, object_description: str
+    vl_model: VlModel[Any], image: Image, object_description: str
```
paul-nechifor, line 24: I don't know what you think about this, but personally I'm not fond of the way some code is generic when it doesn't have to be. In this case, `VlModel` is forced to be generic because it extends `Configurable[_VlConfig]`. `VlModelConfig` just takes a single key: `auto_resize`.

I think we would be better off if we used:

```
class VlModel(Captioner, Resource):
    @abstractmethod
    def __init__(self, auto_resize: tuple[int,int] | None = None): ...
```

If subclasses extend `VlModel` and need more configuration options, they can just add more params onto `__init__`.

One of the reasons I don't like the existing system is because if we want a subclass of VlModel to have more configs we have to create a config class, and we pass it through `Configurable.__init__`, in a superclass of VlModel. I think VlModel (or its parents) shouldn't care about what config its children needs.

What do you think? (I'm not saying you need to change anything, just asking about what the vision for config is.)

Dreamsorcerer: So the reasoning is that we want to be able to document/validate/extend all the config options. So, by forcing everything through the same config in all subclasses, this becomes a much simpler, more explicit way to achieve that.

Currently, code has a mix of config, init parameters etc. which makes this unmanageable. init parameters are also not going to get type checked at runtime, and in most cases these parameters are going to be passed dynamically (blueprint(), CLI, config file), so it's pretty important to have runtime type checking.

> If subclasses extend `VlModel` and need more configuration options, they can just add more params onto `__init__`.

I've tried to make this as simple as possible. So, adding a config option is simply:
```
class NewConfig(SuperConfig):
    new_parameter: int
```
And adding NewConfig to the subclass. The `__init__` method does not need to be defined at all in the subclass for these simple cases.

dimos/agents/skills/person_follow.py
Dreamsorcerer: Seems odd if it's hardcoded to that specific model, but OK.
```suggestion
        # Use VlModel to keep usage in this class generic
        self._vl_model: VlModel = QwenVlModel()
```

dimos/navigation/visual/query.py
Dreamsorcerer: I get that the generic part is slightly awkward, but I've played with several ideas and this is the only one that seems to hold up. (I'd also note that developers who don't care about type safety can ignore the generic part and simply assign to `default_config`, and subclasses which change the config defaults without adding a new parameter can do that with type safety too).

paul-nechifor: Fair enough.

paul-nechifor (dismissed):

paul-nechifor:
The tests are still failing. Note that you have to cherry-pick this to fix protobuf: https://github.com/dimensionalOS/dimos/pull/1372/changes

pyproject.toml
```diff
@@ -322,8 +322,12 @@ exclude = [
 
 [tool.ruff.lint]
 extend-select = ["E", "W", "F", "B", "UP", "N", "I", "C90", "A", "RUF", "TCH"]
-# TODO: All of these should be fixed, but it's easier commit autofixes first
-ignore = ["A001", "A002", "B008", "B017", "B019", "B024", "B026", "B904", "C901", "E402", "E501", "E721", "E722", "E741", "F811", "F821", "F821", "F821", "N801", "N802", "N803", "N806", "N817", "N999", "RUF003", "RUF009", "RUF012", "RUF034", "RUF043", "RUF059", "UP007"]
+ignore = [
+    # TODO: All of these should be fixed, but it's easier commit autofixes first
+    "A001", "A002", "B008", "B017", "B019", "B024", "B026", "B904", "C901", "E402", "E501", "E721", "E722", "E741", "F811", "F821", "F821", "F821", "N801", "N802", "N803", "N806", "N817", "N999", "RUF003", "RUF009", "RUF012", "RUF034", "RUF043", "RUF059", "UP007",
+    # This breaks runtime type checking (both for us, and users introspecting our APIs)
+    "TC001", "TC002", "TC003"
```
paul-nechifor, line 331: I fixed this recently. The linter needs to know which classes need to have their annotations at run time. Are there more classes besides Module ones now?

```
[tool.ruff.lint.flake8-type-checking]
runtime-evaluated-base-classes = ["dimos.core.module.Module"]
```

dimos/core/module.py
```diff
@@ -70,20 +75,26 @@ def get_loop() -> tuple[asyncio.AbstractEventLoop, threading.Thread | None]:
         return loop, thr
 
 
-@dataclass
-class ModuleConfig:
+class ModuleConfig(BaseConfig):
     rpc_transport: type[RPCSpec] = LCMRPC
-    tf_transport: type[TFSpec] = LCMTF
+    tf_transport: type[TFSpec] = LCMTF  # type: ignore[type-arg]
```
paul-nechifor, line 80: Are you sure `# type: ignore[type-arg]` is needed? It wasn't before.

paul-nechifor (dismissed):

leshy:
hey for the future for easier reviews, usually we ask for oneliners to test, but this is an API change so could you show us the new API with example code blocks, like what does this change mean for us building modules.. either this or docs update link we can read

leshy:
Also do we need to accept global_config as arg even if not using global config in our module? why not pass kwargs directly?

pyproject.toml
Dreamsorcerer: I guess it's mainly ModuleConfig now. But, that may reference other classes which are not subclassed from ModuleConfig, creating a lot of extra false positives. I still get test failures after adding ModuleConfig and letting ruff fix things.

This rule is one of my pet peeves, as it achieves nothing (all these modules are getting imported at runtime somewhere, so there's going to be no noticeable improvement to import time), and thus only causing problems and making code slightly less readable. As noted in the comment, there's a small chance users may want to introspect our APIs at runtime and this would break any attempts to do that.

dimos/core/module.py
Dreamsorcerer: Yeah, pydantic won't let me resolve it:
```
Subscripting `type[]` with an already prametrized type is not supported.
```

It's because I made it generic, so mypy wants it subscripted.

paul-nechifor (approved):

PR #1371, opened 2026-02-26
feat: integrate gripper into coordinator tick loop

ruthwikdasyam:
### Problem

Gripper control bypasses the coordinator tick loop - `set_gripper_position` and `get_gripper_position` RPCs call the adapter directly, so gripper commands aren't synchronized with arm joints, or published in JointState. This makes gripper control from VR triggers during teleop impossible.

Closes DIM-598
Closes DIM-537

### Solution

Integrate gripper into the tick loop as a first-class hardware component.

- `ConnectedGripper` wraps the parent arm's adapter, routing `read_state()`/`write_command()` through `read_gripper_position()`/`write_gripper_position()` so the tick loop handles it uniformly alongside arm joints.
- `HardwareComponent.parent_hardware_id` lets gripper components reference their parent arm - no duplicate adapter connections.
- Analog trigger packing in `Buttons` - VR trigger floats packed as 7-bit values in bits 16-29 of the existing UInt32 message.
- `TeleopIKTask` gripper support - new `gripper_joint`, `gripper_open_pos`, `gripper_closed_pos` config. `on_gripper_trigger()` maps analog trigger [0,1] → gripper position

### Breaking Changes

None

### How to Test

- Hardware test (requires xArm7 + Quest controllers)
- Squeeze right trigger -- gripper should close proportionally

`dimos run arm-teleop-xarm7`


dimos/control/blueprints.py
```diff
@@ -715,6 +726,7 @@
     "coordinator_teleop_dual",
     "coordinator_teleop_piper",
     "coordinator_teleop_xarm6",
+    "coordinator_teleop_xarm7",
```
mustafab0, line 724: This is okay and not a problem at all. But let's thin out control blueprints in the future to only have 1 example of each use cases and only with mock hardware

For example control blueprints should have coordinator_teleop_mock, coordinator_visualservo-mock, coordinator-dual-mock.

dimos/control/hardware_interface.py
```diff
@@ -287,7 +287,53 @@ def write_command(self, commands: dict[str, float], _mode: ControlMode) -> bool:
...
+class ConnectedGripper(ConnectedHardware):
+    """Runtime wrapper for a gripper attached to a manipulator.
+
+    Registered as a separate HardwareComponent with hardware_type=GRIPPER
+    so the tick loop treats it uniformly alongside arm joints.
+    """
+
+    def __init__(
+        self,
+        adapter: ManipulatorAdapter,
```
mustafab0: Why do we need  ManipulatorAdapter here?

dimos/control/hardware_interface.py
```diff
@@ -287,7 +287,53 @@ def write_command(self, commands: dict[str, float], _mode: ControlMode) -> bool:
         return self._twist_adapter.write_velocities(ordered)
 
 
+class ConnectedGripper(ConnectedHardware):
+    """Runtime wrapper for a gripper attached to a manipulator.
+
+    Registered as a separate HardwareComponent with hardware_type=GRIPPER
+    so the tick loop treats it uniformly alongside arm joints.
+    """
```
mustafab0: ConnectedGripper Hardware Interface is overkill for this application.

If the gripper only opens and closes, it can be abstracted as an additional joint.

such that the ConnectedHardware (xarm7) has 7dof + 1dof and we just send 8 joint values in every JointState Command.

dimos/control/hardware_interface.py
```diff
@@ -287,7 +287,53 @@ def write_command(self, commands: dict[str, float], _mode: ControlMode) -> bool:
...
+        joint_name = self._joint_names[0]
+        if joint_name in commands:
+            self._last_commanded[joint_name] = commands[joint_name]
+        return self._adapter.write_gripper_position(self._last_commanded[joint_name])
+
+
 __all__ = [
+    "ConnectedGripper",
     "ConnectedHardware",
     "ConnectedTwistBase",
```
mustafab0: We will perhaps need a ConnectedEndEffector hardware interface for more complex ee, like dextrous hands, and such. 

But I imagine they would also work very similar to an abstract ConnectedHardware with Jointstate commands.

dimos/control/hardware_interface.py
mustafab0: Before creating new interface types we should first exploit these existing interfaces. 

The fewer interfaces we have the more consistent user/dev experience will be.

dimos/hardware/manipulators/xarm/adapter.py
```diff
@@ -62,6 +62,7 @@ def __init__(self, address: str, dof: int = 6, **_: object) -> None:
         self._dof = dof
         self._arm: XArmAPI | None = None
         self._control_mode: ControlMode = ControlMode.POSITION
+        self._gripper_enabled: bool = False
```
mustafab0, line 65: why is this required?

dimos/teleop/quest/quest_types.py
```diff
@@ -94,16 +94,19 @@ def from_joy(cls, joy: Joy, is_left: bool = True) -> "QuestControllerState":
...
+    as 7-bit integers in the upper 16 bits for proportional gripper control.
 
     Bit layout:
-        Left  (bits 0-6): trigger, grip, touchpad, thumbstick, primary, secondary, menu
-        Right (bits 8-14): trigger, grip, touchpad, thumbstick, primary, secondary, menu
+        Left  (bits 0-6):   trigger, grip, touchpad, thumbstick, primary, secondary, menu
+        Right (bits 8-14):  trigger, grip, touchpad, thumbstick, primary, secondary, menu
+        Bit 7, 15:          reserved
+        Bits 16-22:         left trigger analog (7-bit, 0=0.0 … 127=1.0)
+        Bits 23-29:         right trigger analog (7-bit, 0=0.0 … 127=1.0)
```
mustafab0, line 105: 👍

mustafab0 (changes requested):
Need to justify why a new hardware interface is necessary

dimos/control/blueprints.py
```diff
@@ -473,30 +475,39 @@
...
     joint_state_frame_id="coordinator",
     hardware=[
         HardwareComponent(
             hardware_id="arm",
             hardware_type=HardwareType.MANIPULATOR,
-            joints=make_joints("arm", 6),
+            joints=make_joints("arm", 7),
             adapter_type="xarm",
-            address="192.168.1.210",
+            address="192.168.2.235",
```
paul-nechifor, line 489: This is personal config, it doesn't belong in a blueprint. One idea would be to have `GlobalConfig.arm_ip` with a default of None. And you can put `ARM_IP=192.168.2.235` in your `.env` file.

dimos/control/blueprints.py
ruthwikdasyam: Right, blueprints are overflowing on the coordinator/manipulators end. We can reduce the list and have smaller definitions.

dimos/control/hardware_interface.py
ruthwikdasyam: ConnectedGripper needs ManipulatorAdapter because right now grippers does not have a seperate SDK/adapter and we reply on arm's adapter for gripper controls -- which means same adapter for both.
It reuses the parent arm's adapter to read and write gripper poses.

dimos/control/hardware_interface.py
```diff
@@ -287,7 +287,53 @@ def write_command(self, commands: dict[str, float], _mode: ControlMode) -> bool:
...
+        self,
+        adapter: ManipulatorAdapter,
+        component: HardwareComponent,
+    ) -> None:
+        super().__init__(adapter, component)
+        self._last_commanded = {name: 0.0 for name in self._joint_names}
+        self._initialized = True
+
+    def disconnect(self) -> None:
+        """No-op: lifecycle is owned by the parent ConnectedHardware."""
```
paul-nechifor: It's not clear to me why `ConnectedHardware.disconnect` calls `self._adapter.disconnect()` to begin with. If the life cycle of `self._adapter` is managed outside `ConnectedHardware` then it shouldn't call disconnect to begin with (there no connect method for example).

dimos/control/hardware_interface.py
ruthwikdasyam: Considered this method intially,

gripper can't be an 8th joint without special casing ConnectedHardware anyway - the XArm SDK uses completely separate API methods (`set_gripper_position` vs `set_servo_angle`), so `write_joint_positions()` would reject an 8th value. We'd need gripper-specific branching in `write_command()` and `read_state()` regardless

ConnectedGripper keeps that branching seperate rather than adding it to the generic joint path. It implements the same `read_state()/write_command()` interface, so the tick loop treats it uniformly - it just routes to `read_gripper_position()/write_gripper_position()` instead of the joint methods

The benefit of havig it as separate hardware is - in blueprint - we can add/remove the gripper declaratively without touching coordinator logic, and if we swap in a gripper with a different adapter/SDK, only the adapter changes - the coordinator and task code stay the same

dimos/control/tasks/teleop_task.py
```diff
@@ -302,6 +322,19 @@ def on_buttons(self, msg: Buttons) -> bool:
...
+        if self._config.gripper_joint:
+            if hand == "left":
+                self.on_gripper_trigger(msg.left_trigger_analog, 0.0)
+            elif hand == "right":
+                self.on_gripper_trigger(msg.right_trigger_analog, 0.0)
+            else:
+                logger.warning(
+                    f"TeleopIKTask {self._name}: gripper_joint is set but hand is not "
+                    f"'left' or 'right' — cannot determine which trigger to use"
+                )
```
paul-nechifor: We should rely more on the python type system.

TeleopIKTaskConfig.hand has a type of str with a default of `""` (which is invalid).

But if only `"left"` and `"right"` are valid, its type should be set to `Literal["left", "right"]`

dimos/control/hardware_interface.py
ruthwikdasyam: yes, dexterous hands and other complex end-effectors will eventually need this pattern too.
Having a separate hardware interface makes them easier to integrate without complicating the joint path

dimos/hardware/manipulators/xarm/adapter.py
ruthwikdasyam: I added `gripper_enabled` to run `arm.gripper_enable` method to activate gripper. But, i see that from manipulation_module, gripper is able to take commands without any `arm.gripper_enable` method. - which means gripper should always be active.

I can remove this check, if we get some issues later, we will add it.

dimos/control/blueprints.py
ruthwikdasyam: True. Currently all manipulation blueprints hardcode the arm IP the same way. Moving these to  `.env` is a good idea - I can do a seperate PR cleaning up across all blueprints

dimos/control/hardware_interface.py
ruthwikdasyam: Hm that's true. I see `connect` method called in coordinator.py line-244. perhaps, we can call disconnect from coordinator too. 
@mustafab0 should be able to reason this precisely

dimos/control/tasks/teleop_task.py
ruthwikdasyam: Agreed. Was using `""` as a catch-all (both left and right), but `Literal["left", "right"]` is cleaner. Will update both TaskConfig and TeleopIKTaskConfig

dimos/control/hardware_interface.py
mustafab0: Unsure what you mean by
> without complicating the joint path
______
> they would also work very similar to an abstract ConnectedHardware with Jointstate commands.

What I mean by this is that we may not even need a separate interface type they could just use the ConnectedHardware interface

ruthwikdasyam: Oh, I get what you mean. we can just initiate grippers thorugh ConnectedHardware Interface. 
I see that ConnectedHardware is tightly coupled with adapters - which mean having arm control methods -  `read_joint_pos/read_joint_vel`, `write_joint_pos/write_joint_vel` in its definition.
which means, if we wanna control grippers with this, we need to add check, probably with names of joints, and use gripper methods if its a `gripper_joint_value`

So, my idea is to have a parallel pipeline along Arms, for Grippers, and have their own(depends) adapter and, methods.
Thus ConnectedGrippers will be defined to access `read_gripper/ee_pos` and `write_gripper/ee_pos` methods. through which we can maintain method consistency across all 'gripper_adapters' or any arm_adapters with gripper methods (naming doesnt change)

But, i get it - your suggestion makes things easier with current robots/grippers. If you're specific about this, I can remove GrippersInterface for now. Let me know!

dimos/hardware/manipulators/xarm/adapter.py
mustafab0: I think okay to keep it here, considering more complicated EE such as pneumatic grippers might want to be enabled disabled to save power,etc.

dimos/control/hardware_interface.py
mustafab0: >  if we wanna control grippers with this, we need to add check, probably with names of joints, and use gripper methods if its a gripper_joint_value

Yeah exactly. We already have the separation of concerns so the adapters write_command can call a small helper such _set_gripper_pos and do it.

Another reason why I am a little unsure about the ConnectedGripper is that it still needs the ManipulatorAdapter. 

> If you're specific about this, I can remove GrippersInterface for now. Let me know!

Yes please! I am sure that in future we probably need to have more interfaces but we should move carefully here.

dimos/control/hardware_interface.py
mustafab0: `seld._adapter` is simply a wrapper for the robot sdk. The ConnectedHardware is the actual runtime instance and has a lifecycle. 

Every HardwareInterface needs a connect/disconnect which are generic for the ControlCoordinator to call. But the specific order of operations are encoded in the `adapter.connect()/disconnect()` methods. 

@ruthwikdasyam you are right about the ControlCoordinator calling connect, that is an oversight. The coordinator shouldn't directly call connect and should always call the hardware interface to do so. 

This likely is the case because when I was initially designing control coordinator a HardwareInterface did not exist at all.

dimos/control/hardware_interface.py
```diff
@@ -153,8 +167,8 @@ def write_command(self, commands: dict[str, float], mode: ControlMode) -> bool:
                 )
                 self._warned_unknown_joints.add(joint_name)
 
-        # Build ordered list for adapter
-        ordered = self._build_ordered_command()
+        # Build ordered list for arm joints only
+        arm_ordered = [self._last_commanded[name] for name in self._arm_joint_names]
```
mustafab0, line 171: A bit confused about the semantic change.

Is it because we want to make sure the gripper is added exactly after the specific ConnectedHardware manipulator instance?

Or is it to distinguish between TwistBase Hardware and Manipulator Hardware?

mustafab0 (dismissed):
Looks ok. 👍

paul-nechifor (dismissed):

dimos/control/hardware_interface.py
ruthwikdasyam: yes, its to make sure we write joint commands to this ordered list - which only has arm_joints, and no grippers.

The old `_build_ordered_command()` iterated all joints including gripper which would break here, 
howevr used by TwistBase where it works fine - coz they don't have gripper joints, so joint_names is just the velocity DOFs

paul-nechifor (approved):

PR #1376, opened 2026-02-27
lazy load legacy modules

jeff-hykin:
## Problem

Loading extra modules

## Solution

Lazy load

## Breaking Changes

None

## How to Test

`dimos --replay run unitree-go2-basic`


leshy (changes requested):
as far as i can tell this __init__.py is not neccessary for anything

PR #1365, opened 2026-02-25
feat(dask): remove dask

paul-nechifor:
## Problem

* Dask is slow to start.
* Causes issues with shutdown (you get ugly errors when hitting Ctrl+C).
* Required a lot of ugly patches in `core/__init__.py`
* Causes confusion because things are imported once per worker. (E.g.: see [Jeff's rerun issue](https://discord.com/channels/1341146487186391173/1344622457327321150/1447797852662403102), or [my actor confusion](https://discord.com/channels/1341146487186391173/1344622457327321150/1451149593382490132))

Closes DIM-609

## Solution

* Removed the dask package entirely.
* Moved imports out of `dimos.core`. Now you have to import from the original place.
* Removed everything in `dimos/core/__init__.py`.
* Made `dimos` shut down cleanly, without any exceptions, and right away.
* Inlined some of the imports which were causing slowdowns.
* Added a short doc about py-spy so others can investigate speed if they want to.
* Fixed mujoco errors on shutdown.

## Breaking Changes

* UtilizationModule is gone.
* Blueprint configurations cannot use lambdas anymore because `multiprocessing` requires params to be pickleable, and lambdas are not because they don't have names. This worked with Dask because Dask serialized the bytecode of the lambda and reconstructed it on the other side.

## How to Test

Test any blueprint. Example:

```python
uv run dimos run unitree-go2
```

## Contributor License Agreement

- [x] I have read and approved the [CLA](https://github.com/dimensionalOS/dimos/blob/main/CLA.md).

PR #1379, opened 2026-02-28
chore: remove unnecessary "prevent unwanted imports" comments

spomichter:
## How to Run/Test
```bash
cd /home/ubuntu/dimos
git fetch origin
git checkout chore/remove-llm-comments
grep -rn "Here to prevent unwanted imports" --include="*.py"
# Should return nothing
```

## Summary
Removes 8 instances of the LLM-generated comment `# Here to prevent unwanted imports in the file.` that were introduced in PR #1365. The lazy imports themselves are correct and intentional — only the redundant comments are removed.

Reference: https://github.com/dimensionalOS/dimos/pull/1365/changes#r2867298612

## Changes
- Removed comment from `dimos/agents/agent.py` (line 106)
- Removed comment from `dimos/agents/skills/navigation.py` (line 62)
- Removed comment from `dimos/agents/skills/person_follow.py` (line 179)
- Removed comment from `dimos/agents/web_human_input.py` (line 55)
- Removed comment from `dimos/agents_deprecated/memory/spatial_vector_db.py` (line 59)
- Removed 3 comments from `dimos/models/vl/base.py` (lines 72, 258, 319)

## Testing
- [x] Pre-commit hooks pass (all 18 checks)
- [x] No functional code changed — comment-only removal
- [x] Verified zero remaining instances via grep

### DIFF
--- a/dimos/agents/agent.py
+++ b/dimos/agents/agent.py
@@ -103,7 +103,6 @@ class Agent(Module[AgentConfig]):
             model = MockModel(json_path=self.config.model_fixture)
 
         with self._lock:
-            # Here to prevent unwanted imports in the file.
             from langchain.agents import create_agent
 
             self._state_graph = create_agent(
--- a/dimos/agents/skills/navigation.py
+++ b/dimos/agents/skills/navigation.py
@@ -59,7 +59,6 @@ class NavigationSkillContainer(Module):
         super().__init__()
         self._skill_started = False
 
-        # Here to prevent unwanted imports in the file.
         from dimos.models.vl.qwen import QwenVlModel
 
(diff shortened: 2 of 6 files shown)


paul-nechifor:
> Removes 8 instances of the LLM-generated comment # Here to prevent unwanted imports in the file.

They're not LLM generated. I specifically added them there to indicate that those imports should not be moved to the top because they cause imports which slow down the start of `dimos run`.

Without those comments, people might move them to the top. I often move unnecessary local imports because they're added by LLMs. LLMs often prefer to add local imports because they have the habit of producing the smallest diff possible, even when it doesn't make sense.

PR #1384, opened 2026-02-28
Tiny Fix: lcmspy crash

jeff-hykin:
## Problem

lcmspy will eventually crash with an error about mutating a value while iterating

## Solution

Copy the list before iterating

## Breaking Changes

None

## How to Test

`lcmspy` (run for a long time while a bunch of topics publish)

### DIFF
--- a/dimos/utils/cli/lcmspy/lcmspy.py
+++ b/dimos/utils/cli/lcmspy/lcmspy.py
@@ -51,7 +51,7 @@ class Topic:
         """Get messages within the specified time window"""
         current_time = time.time()
         cutoff_time = current_time - time_window
-        return [(ts, size) for ts, size in self.message_history if ts >= cutoff_time]
+        return [(ts, size) for ts, size in list(self.message_history) if ts >= cutoff_time]
 
     # avg msg freq in the last n seconds
     def freq(self, time_window: float) -> float:


dimos/utils/cli/lcmspy/lcmspy.py
```diff
@@ -51,7 +51,7 @@ def _get_messages_in_window(self, time_window: float):  # type: ignore[no-untype
         """Get messages within the specified time window"""
         current_time = time.time()
         cutoff_time = current_time - time_window
-        return [(ts, size) for ts, size in self.message_history if ts >= cutoff_time]
+        return [(ts, size) for ts, size in list(self.message_history) if ts >= cutoff_time]
```
paul-nechifor, line 54: Works as a quick fix, but I don't know if it make sense to duplicate the list every time you iterate over it. At a glance, I can't tell if this has a limit, and just copies more memory as you increase the size.

An alternative solution would be to add `with._lock` when accessing/mutating.

jeff-hykin: I actually think this is the ideal fix. AFAIK `_get_messages_in_window` wants a copy because its trying to take a snapshot at a particular time. I believe `list(deque)` is atomic, so it grabs the snapshot as intended. 

Blocking all modifications to the deque (self.message_history) until the list comprehension is done I think would be slower and more complex than grabbing a copy and then iterating over the copy.

I can change it to a tuple instead of a list (I need to run CI again to get past the flakey tests anyways)

paul-nechifor: `list(deque)` is atomic because it's implemented in C and holds the GIL. It's an implementation detail. Python is already moving towards removing the GIL. When that's mainstream, you'll have Python code adding a value and C code iterating over it to create the list.

I don't think it's an ideal fix. It only prevents mutations by implication, because the code for `list()` is holding the global interpreter lock.

If you use `with lock`, you're both using an explicit lock and using less memory.

---

But I don't think `with lock` is the ideal solution either. Taking another look at the code, I think the ideal solution would be to use two circular buffers (one for size, one for number of messages), each with 60 entries (a bucket for each second), and when a message comes, add it's size to one, and increment its corresponding timestamp bucket. That would use a fixed memory size.

The code only uses a maximum of 60 seconds and only counts the number of messages and the sum of sizes.

jeff-hykin: I guess I'll come back to this after the rosnav stuff. I just PR'd thinking a 30sec fix would be helpful

If we change the approach away from copies, I'm guessing we'd probably also want to redo these existing lines in `GraphLCMSpy`.

```py
        while not self.graph_log_stop_event.is_set():
            self.update_graphs(self.config.graph_log_window)  # type: ignore[attr-defined]
            # Copy to list to avoid RuntimeError: dictionary changed size during iteration
            for topic in list(self.topic.values()):
                topic.update_graphs(self.config.graph_log_window)  # type: ignore[attr-defined]
            time.sleep(self.config.graph_log_window)  # type: ignore[attr-defined]
```

mypy in that file could also use help

`jeff/fix/lcmspy2` has changes for a proper fix (except for GraphLCMSpy), but I don't want to get too invested (that branch is untested)

leshy: btw guys lcmspy is totally vibed and I didn't read the code even, this cli tooling is super ad hoc, your call if you want to take it seriously or not.

this needs a full rewrite (re-vibe?) to take into account pluggable GLOB transports now, so it can spy on ros,lcm etc

paul-nechifor: lcmspy is in the docs so it's an official tool. Users will be annoyed when it crashes.

jeff-hykin: @paul-nechifor can we just apply the tuple patch until we get an lcmspy rewrite? It fixes the crashes for sure

A transport-spy tool that uses the generic transport topic discovery (like rerun) seems like a good direction, on top of lcmspy probably needing a rewrite. So it seems like kind of a waste of time to polish something that's going to be rewritten entirely within 6months.

paul-nechifor: What's the tuple patch?

jeff-hykin: The current PR as-is (7 char change that prevents the crash from happening in practice)

(It was changed from `list()` to `tuple()` although I see now I must've not pushed the other day)

paul-nechifor: Added a lock in https://github.com/dimensionalOS/dimos/pull/1419

leshy: yeah this is the type of PR I don't read carefully, I expect Jeff to have thought about this enough given it's being submitted and I don't expect to need any more tokens from me, I get upset post-hoc if there are issues or if fix is partial, so I would have let Jeff PR in

if I was Jeff though, I'd just immediately follow what Paul asks for since it's very simple and more correct.

IDK if we'll have modern py user before lcmspy rewrite, but just because Paul cares about it, I would have implemented immediately and moved on

leshy: merged above locks thing, was 20 lines

PR #1395, opened 2026-03-02
feat: interactive install script (DIM-645)

spomichter:
## Problem
Closes DIM-645

DimOS has no guided installation experience. Users must manually follow README steps, install system dependencies, choose between library/dev mode, configure Python environments, and hope they get it right. This is a significant barrier to adoption.

## Solution

Added `scripts/install.sh` — a fully interactive installer powered by [gum](https://github.com/charmbracelet/gum) (charmbracelet's TUI toolkit for shell scripts). The script auto-downloads the gum binary (~4.5MB) and falls back to basic prompts if unavailable.

### Features
- **Interactive TUI**: arrow-key selection for install mode (library/developer), extras packages, setup method (system packages vs Nix)
- **Smart defaults**: detects OS/arch, checks disk space (warns <10GB), skips already-installed apt packages
- **Two setup paths**: system packages (apt + uv) or Nix with flake support
- **Safe by design**: asks before overwriting existing `.venv`, Ctrl+C works everywhere (no subshell traps), proper curl-pipe detection (re-execs from temp file for TTY)
- **Post-install verification**: optional smoke tests with `dimos --simulation` and targeted pytest
- **Non-interactive mode**: `--non-interactive --mode library --extras base,unitree` for CI/scripting
- **Cross-platform**: Linux x64/arm64, macOS x64/arm64

### Files changed
- `scripts/install.sh` — the installer (~920 lines)
- `README.md` — added Interactive Install section at top of Installation

## Breaking Changes
None — this is purely additive. Existing installation methods are unchanged.

## How to Test
```bash
# Test the installer directly (dry-run mode available)
cd /home/ubuntu/dimos
git fetch origin && git checkout feat/dim-645-interactive-install

# Dry run (shows what would happen without executing)
bash scripts/install.sh --dry-run

# Full interactive run (in a temp directory)
mkdir /tmp/test-install && cd /tmp/test-install
bash /home/ubuntu/dimos/scripts/install.sh

# Non-interactive mode
bash scripts/install.sh --non-interactive --mode library --extras base

# Curl-pipe style (as end users would run it)
curl -fsSL https://raw.githubusercontent.com/dimensionalOS/dimos/feat/dim-645-interactive-install/scripts/install.sh | bash
```

- [x] I have read and approved the [CLA](https://github.com/dimensionalOS/dimos/blob/main/CLA.md)


scripts/uninstall.sh
```diff
@@ -0,0 +1,170 @@
...
+        local_size="$(du -sh "$dir" 2>/dev/null | cut -f1 || echo "?")"
+        if should_remove "Remove ${dir}/ (${local_size})?"; then
+            do_rm "$dir"
+        fi
+    fi
+done
+
+# ─── System packages (apt) ───────────────────────────────────────────────────
+if command -v apt-get &>/dev/null; then
+    PKGS="portaudio19-dev git-lfs libturbojpeg python3-dev pre-commit"
```
paul-nechifor: There's no check that these packages were installed by us.  We might unintentionally remove packages that people were using before but weren't aware they were using.

Personally, I just wouldn't uninstall system packages.

Also, these are not all the packages which are installed.

scripts/install.sh
```diff
@@ -0,0 +1,900 @@
+#!/usr/bin/env bash
+# Copyright 2025-2026 Dimensional Inc.
+# Licensed under the Apache License, Version 2.0
+#
+# Interactive installer for DimOS — the agentive operating system for generalist robotics.
+#
+# Usage:
+#   curl -fsSL https://dimensional.ai/install.sh | bash
```
paul-nechifor: If this script is hosted there, I think it should be broken up in two:

- a script in the repo which does all the work (this will change as the project evolves)
- a script which just gets the script from the repo and executes it (this will stay the same and can be uploaded to https://dimensional.ai/install.sh)

If the full script is hosted on https://dimensional.ai/install.sh we will have to keep changing it as the project changes.

scripts/install.sh
```diff
@@ -0,0 +1,913 @@
...
+
+# ─── system dependencies ─────────────────────────────────────────────────────
+install_system_deps() {
+    info "checking system dependencies..."
+
+    case "$DETECTED_OS" in
+        ubuntu|wsl)
+            # Check if all packages are already installed
+            local needed=""
+            local all_pkgs="curl g++ portaudio19-dev git-lfs libturbojpeg python3-dev pre-commit libgl1 libegl1"
```
paul-nechifor: I think the script should be broken up into utilities and specifics. The actual important stuff like the package list here is buried deep into this script.

scripts/install.sh
spomichter: for now ill change to `curl -fsSL https://raw.githubusercontent.com/dimensionalOS/dimos/feat/dim-645-interactive-install/scripts/install.sh | bash`

scripts/install.sh
```diff
@@ -0,0 +1,918 @@
+#!/usr/bin/env bash
+# Copyright 2025-2026 Dimensional Inc.
+# Licensed under the Apache License, Version 2.0
+#
+# Interactive installer for DimOS — the agentive operating system for generalist robotics.
+#
+# Usage:
+#   curl -fsSL https://raw.githubusercontent.com/dimensionalOS/dimos/dev/scripts/install.sh | bash
```
paul-nechifor, line 8: Should it be dev or main?

spomichter: dev for now

scripts/install.sh
```diff
@@ -0,0 +1,918 @@
...
+        [[ "$VERBOSE" == "1" ]] && dim "$ $*"
+        info "$title"
+        eval "$@"
+    fi
+}
+
+# ─── ascii banner ─────────────────────────────────────────────────────────────
+show_banner() {
+    if [[ "$NON_INTERACTIVE" == "1" ]] && [[ -z "${DIMOS_SHOW_BANNER:-}" ]]; then return; fi
+    local banner='   ▇▇▇▇▇▇╗ ▇▇╗▇▇▇╗   ▇▇▇╗▇▇▇▇▇▇▇╗▇▇▇╗   ▇▇╗▇▇▇▇▇▇▇╗▇▇╗ ▇▇▇▇▇▇╗ ▇▇▇╗   ▇▇╗ ▇▇▇▇▇╗ ▇▇╗
```
paul-nechifor: Logo doesn't show up right on a Ubuntu 24.04 terminal with a default window size.

Image

scripts/install.sh
```diff
@@ -0,0 +1,918 @@
...
+    else
+        printf "    %s# activate the environment%s\n    cd %s && source .venv/bin/activate\n\n" "$DIM" "$RESET" "$dir"
+    fi
+
+    if [[ "$EXTRAS" == *"unitree"* ]] || [[ "$EXTRAS" == "all" ]] || [[ "$EXTRAS" == *"base"* ]]; then
+        printf "    %s# simulation%s\n    dimos --simulation run unitree-go2\n\n" "$DIM" "$RESET"
+        printf "    %s# real hardware%s\n    ROBOT_IP=192.168.1.100 dimos run unitree-go2\n\n" "$DIM" "$RESET"
+    fi
+    if [[ "$EXTRAS" == *"sim"* ]] || [[ "$EXTRAS" == "all" ]]; then
+        printf "    %s# MuJoCo + click-nav%s\n    dimos --simulation run unitree-go2-click-nav --viewer-backend rerun\n\n" "$DIM" "$RESET"
```
paul-nechifor: Won't `click-nav` be eventually moved into `unitree-go2`? Is it worth mentioning `unitree-go2-click-nav` here?

scripts/install.sh
```diff
@@ -0,0 +1,918 @@
...
+detect_os() {
+    DETECTED_ARCH="$(uname -m)"
+    local uname_s; uname_s="$(uname -s)"
+    if [[ "$uname_s" == "Darwin" ]]; then
+        DETECTED_OS="macos"
+        DETECTED_OS_VERSION="$(sw_vers -productVersion 2>/dev/null || echo "unknown")"
+    elif [[ "$uname_s" == "Linux" ]]; then
+        if grep -qi microsoft /proc/version 2>/dev/null; then DETECTED_OS="wsl"
+        elif [[ -f /etc/NIXOS ]] || has_cmd nixos-version; then DETECTED_OS="nixos"
+        else DETECTED_OS="ubuntu"; fi
```
paul-nechifor: I wouldn't assume that every os that's not Windows or NixOS is Ubuntu.

The best way to check for systems which use `apt-get` is `grep -Ei 'debian' /etc/os-release`. If it doesn't match we should also use `die "unsupported operating system"`

paul-nechifor (dismissed):

PR #1397, opened 2026-03-02
feat(testing): add coverage

paul-nechifor:
## Problem

Without test coverage it's more difficult to tell what areas need more attention in testing and whether your tests are actually hitting the code you expect to test.

Closes DIM-556

## Solution

* Added `coverage`.
* It works across processes, including when running `dimos run` tests, so the `-m mujoco` tests also contribute to coverage, though not in CI since they don't run in CI.
* Also added a `bin/coverage-by-author` command for fun.

## Breaking Changes

None

## How to Test

Run:

```bash
./bin/pytest-coverage
(cd htmlcov; python -m http.server 8000)
```

And see the coverage files at <http://localhost:8000>.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -211,7 +211,8 @@ jobs:
     uses: ./.github/workflows/tests.yml
     secrets: inherit
     with:
-      cmd: "pytest --durations=0 -m 'not (tool or mujoco)'"
+      cmd: "_DIMOS_COV=1 coverage run -m pytest --durations=0 -m 'not (tool or mujoco)' && coverage combine && coverage html && coverage report"
+      upload-coverage: true
       dev-image: ros-dev:${{ (needs.check-changes.outputs.python == 'true' || needs.check-changes.outputs.dev == 'true' || needs.check-changes.outputs.ros == 'true') && needs.ros-dev.result == 'success' && needs.check-changes.outputs.branch-tag || 'dev' }}
 
   run-mypy:
--- a/bin/pytest-coverage
+++ b/bin/pytest-coverage
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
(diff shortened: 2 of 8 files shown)


.github/workflows/docker.yml
```diff
@@ -211,7 +211,8 @@ jobs:
     uses: ./.github/workflows/tests.yml
     secrets: inherit
     with:
-      cmd: "pytest --durations=0 -m 'not (tool or mujoco)'"
+      cmd: "_DIMOS_COV=1 coverage run -m pytest --durations=0 -m 'not (tool or mujoco)' && coverage combine && coverage html && coverage report"
```
leshy, line 214: just to have a single place where this stuff is specified, should this be running ./bin/pytest-coverage ?

bin/pytest-coverage
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+rm -f .coverage .coverage.*
+
+export _DIMOS_COV=1
+
+uv run coverage run -m pytest "$@" -m 'not tool'
```
leshy, line 9: just to have a single place where this stuff is specified, should this be running ./bin/pytest-fast && ./bin/pytest-slow  ? maybe not possible just mentioning preference

paul-nechifor: But we don't want to run multiple command, right? And this is running `not tool` which includes `./bin/pytest-mujoco`.

leshy (approved):

leshy (commented):
I see you are uploading artifacts, 
I'm very keen on performance stats now with resource monitor so should go to the same place,

to not start a db I had an idea of just appending .json to some stats repo, so that we can build graphs through time. and compare to dev. IDK if this is better placed in assets instead of a repo.

then we can graph test coverage by author, repo performance through time, etc. even performance gain/loss by author

PR #1426, opened 2026-03-05
Fix: Escaped brackets in pip install to avoid zsh glob expansion

### LINKED ISSUE #1404: Update osx.md [Good first Issue]
Need Simple update for quotes when installing dimos because in pip, those brackets define optional dependency groups called extras, but zsh thinks you are trying to match files in your folder using a glob pattern. It tries to expand [misc,sim,visualization,agents,web,perception,unitree,manipulation,cpu,dev] as a filename pattern.

hgkdzbf6:
Fixed zsh glob pattern issue by escaping square brackets in pip install command. The brackets [misc,sim,...] were being treated as glob patterns by zsh, causing installation to fail. Fixes #1404

### DIFF
--- a/docs/installation/osx.md
+++ b/docs/installation/osx.md
@@ -1,6 +1,6 @@
 # macOS Install (12.6 or newer)
 
-```sh
+\`\`\`sh
 # install homebrew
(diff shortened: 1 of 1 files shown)


RadientBrain:
You should not put backslash for quotes:
\`\`\`

not required, other quote insertion is fine

PR #1431, opened 2026-03-06
Docker Restoration and Fixes

jeff-hykin:
## Problem(s)

- **This is a blocker for G1 rosnav stack parity with Go2, ref DIM-569**
- DockerModules stopped working on dev
- `configure_stream` does not support different transport types (Blocker for DIM-569) 
- A failed DockerModule build wouldn't stop the container (leak)
- A new container name was generated every run (huge build up of containers)
- We have no tests for docker (how it got broken on dev in the first place)
- No way to pull an image (rather than local dockerfile)
- No way to change build args (ex: can't add network flag)
- Image never rebuilds even if dockerfile edited or build args changes
- DockerModules didn't deploy in parallel
- There was a bug with error handling of WorkerManager 

Closes DIM-662

## Solution

- Add tests
- Minimal change to `ModuleCoordinator` to restore functionality: deploy docker modules in parallel
- remove `configure_stream` (not needed) in favor of using `set_transport` work as intended to allow different stream types
- stop/cleanup fixes

## Breaking Changes

None

## How to Test

`python ./examples/docker_hello_world/hello_docker.py`


dimos/core/docker_module.py
```diff
@@ -157,107 +161,152 @@ def _extract_module_config(cfg: DockerModuleConfig) -> dict[str, Any]:
...
+            raise TypeError(
+                f"{module_class.__name__}.default_config must be a DockerModuleConfig subclass, "
+                f"got {config_class.__name__}"
+            )
         config = config_class(**kwargs)
 
-        # Module info
         self._module_class = module_class
-        self._config = config
+        self.config = config
```
jeff-hykin, line 138: changed to match Module's more closely (how module's __init__ normally sets config)

dimos/core/docker_runner.py
```diff
@@ -266,17 +315,12 @@ def status(self) -> dict[str, Any]:
...
-        """Configure stream transport in container. Mirrors Module.set_transport() for autoconnect()."""
-        topic = getattr(transport, "topic", None)
-        if topic is None:
-            return False
-        if hasattr(topic, "topic"):
-            topic = topic.topic
+        """Forward to the container's Module.set_transport RPC."""
         result, _ = self.rpc.call_sync(
-            f"{self.remote_name}/configure_stream", ([stream_name, str(topic)], {})
+            f"{self.remote_name}/set_transport", ([stream_name, transport], {})
```
jeff-hykin: This is what allows setting transport type (and also simplifies DockerModule)

dimos/core/docker_module.py
```diff
@@ -400,16 +445,45 @@ def _build_container_command(self, cfg: DockerModuleConfig) -> list[str]:
...
+            # When run as `python script.py`, __module__ is "__main__".
+            # Resolve to the actual dotted module path so the container can import it.
+            import __main__
+
+            spec = getattr(__main__, "__spec__", None)
+            if spec and spec.name:
+                module_name = spec.name
+            else:
+                # Fallback: derive from file path relative to cwd
+                main_file = getattr(__main__, "__file__", None)
```
jeff-hykin: This fallback case was a claude side effect, probably doesn't need to be this complex but looked fine.

dimos/core/docker_module.py
```diff
@@ -421,13 +495,14 @@ def _wait_for_ready(self) -> None:
 
             try:
                 self.rpc.call_sync(
-                    f"{self.remote_name}/start", ([], {}), rpc_timeout=RPC_READY_TIMEOUT
```
jeff-hykin, line 423: start has side effects, so its not great to call it in a loop, which is why this was switched to get_rpc_method_names

dimos/core/module_coordinator.py
```diff
@@ -36,7 +36,7 @@ class ModuleCoordinator(Resource):  # type: ignore[misc]
     _global_config: GlobalConfig
     _n: int | None = None
     _memory_limit: str = "auto"
-    _deployed_modules: dict[type[Module], ModuleProxy]
+    _deployed_modules: dict[type[Module], ModuleProxyProtocol]
```
jeff-hykin: This type had to be broadened to support DockerModule, it more or less only shows its effects internally though

dimos/core/module_coordinator.py
```diff
@@ -76,37 +76,76 @@ def stop(self) -> None:
...
+            deployed_module = DockerModule(module_class, *args, **kwargs)
+        else:
+            deployed_module = self._client.deploy(module_class, *args, **kwargs)
+        self._deployed_modules[module_class] = deployed_module
+        return deployed_module  # type: ignore[return-value]
 
     def deploy_parallel(
         self, module_specs: list[tuple[type[ModuleT], tuple[Any, ...], dict[str, Any]]]
     ) -> list[ModuleProxy]:
+        from dimos.core.docker_runner import DockerModule, is_docker_module
```
JalajShuklaSS: Here It might be a personal preference to not to have this inline import and can feel that it can go to top level imports

jeff-hykin: its avoiding a circular import: 
module_coordinator → docker_runner → module → blueprints → module_coordinator 

I'll add a comment

dimos/core/module_coordinator.py
```diff
@@ -76,37 +76,76 @@ def stop(self) -> None:
...
     def start_all_modules(self) -> None:
         modules = list(self._deployed_modules.values())
-        if isinstance(self._client, WorkerManager):
-            with ThreadPoolExecutor(max_workers=len(modules)) as executor:
-                list(executor.map(lambda m: m.start(), modules))
-        else:
-            for module in modules:
-                module.start()
+        if not modules:
+            raise ValueError("No modules deployed. Call deploy() before start_all_modules().")
```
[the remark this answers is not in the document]
jeff-hykin: It would have raised anyways because `ThreadPoolExecutor` can't have a size of 0. It just would've raised with an unhelpful error: no regression

dimos/core/module_coordinator.py
```diff
@@ -76,37 +76,76 @@ def stop(self) -> None:
...
+                            lambda spec: DockerModule(spec[0], *spec[1], **spec[2]), docker_specs
+                        )
+                    )
+        finally:
+            # Reassemble results in original input order
+            results: list[Any] = [None] * len(module_specs)
+            for idx, mod in zip(worker_indices, worker_results, strict=False):
+                results[idx] = mod
+            for idx, mod in zip(docker_indices, docker_results, strict=False):
+                results[idx] = mod
```
JalajShuklaSS: I just was confused and was going though the this function and needed more clarity, here the``executor.map`` submits all tasks immediately but ``list()`` iterates results in order. If any task raises, ``list()`` re-raises and the assignment ``docker_results = list(...)`` never completes, so ``docker_results`` stays ``[]``. Containers launched by tasks that  succeeded before the failure are now orphaned: their DockerModule objects are garbage-collected, but the Docker containers keep running and are never registered in ``_deployed_modules`` for ``stop()`` to clean up.
    
So please walk with me for this example and see if it makes sense or please help me to get more clarity on this

deploy_parallel ``([ModA, ModB, ModC])`` where all are docker.
- executor.map submits all 3 to threads; all start launching containers
- ``ModA `` succeeds (container running), ``ModB`` fails, ``ModC`` succeeds
- ``list()`` gets ``ModA`` result, then hits ``ModB `` exception ->  raises
- ``docker_results`` stays`` [] ``-> zip produces nothing -> no registration
- ``ModA`` and ``ModC`` containers are running with no owner -> leaked forever

jeff-hykin: I think this behavior is for non-docker containers too, good call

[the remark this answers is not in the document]
jeff-hykin: is this a problem for `self._client.deploy_parallel` too? (look at its implementation) @greptile-apps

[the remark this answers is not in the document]
jeff-hykin: okay, check the new implementation(s) now

dimos/core/rpc_client.py
```diff
@@ -80,6 +80,17 @@ def __setstate__(self, state) -> None:  # type: ignore[no-untyped-def]
         self._stop_rpc_client = None
 
 
+class ModuleProxyProtocol(Protocol):
```
jeff-hykin, line 87: needed to make this to have an umbrella type for DockerModule and RPCClient. Ideally we'd just use ProxyModule itself but that doesn't work for a handful of reasons that I'm too tired to write out at this point (just follow the import traces and attr inheritance)

dimos/core/worker_manager.py
```diff
@@ -78,17 +82,19 @@ def deploy_parallel(
...
             assignments.append((worker, module_class, args, kwargs))
 
-        def _deploy(
-            item: tuple[Worker, type[ModuleT], tuple[Any, ...], dict[Any, Any]],
-        ) -> RPCClient:
-            worker, module_class, args, kwargs = item
-            actor = worker.deploy_module(module_class, args=args, kwargs=kwargs)
-            return RPCClient(actor, module_class)
-
-        with ThreadPoolExecutor(max_workers=len(assignments)) as pool:
```
jeff-hykin, line 88: given [ A-finished, B-Error, C-pending ], the C won't get properly cleaned up because B will throw immediately.  `safe_thread_map` fixes that

Also I added tests to cover this case so we don't get a regression later

examples/docker_hello_world/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+    libidn2-0 libgfortran5 libgomp1 \
+    cowsay \
+    && rm -rf /var/lib/apt/lists/*
+
+
+# Copy example module so it's importable inside the container
+COPY examples/docker_hello_world/hello_docker.py /dimos/source/examples/docker_hello_world/hello_docker.py
+RUN touch /dimos/source/examples/__init__.py /dimos/source/examples/docker_hello_world/__init__.py
+
+WORKDIR /app
```
jeff-hykin, line 15: Docker example

examples/docker_hello_world/hello_docker.py
```diff
@@ -0,0 +1,157 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Hello World Docker Module
+==========================
+
```
jeff-hykin, line 18: example for Input stream, Output stream, rpc call, and config value passing

dimos/core/docker_module.py
```diff
@@ -157,107 +162,152 @@ def _extract_module_config(cfg: DockerModuleConfig) -> dict[str, Any]:
...
 
     def stop(self) -> None:
         """Gracefully stop the Docker container and clean up resources."""
-        # Signal remote module, stop RPC, unsubscribe handlers (ignore failures)
+        if not self._running:
+            return
         with suppress(Exception):
-            if self._running:
-                self.rpc.call_nowait(f"{self.remote_name}/stop", ([], {}))
+            self.rpc.call_nowait(f"{self.remote_name}/stop", ([], {}))
```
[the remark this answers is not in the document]
jeff-hykin: recheck @greptile-apps

dimos/core/module_coordinator.py
jeff-hykin: alright, fixed 👍 made a `safe_thread_map` helper and used it in both places

dimos/core/tests/test_docker_deployment.py
```diff
@@ -0,0 +1,195 @@
+# Copyright 2026 Dimensional Inc.
```
jeff-hykin, line 1: Smoke tests

dimos/core/tests/test_parallel_deploy_cleanup.py
```diff
@@ -0,0 +1,219 @@
+# Copyright 2026 Dimensional Inc.
```
jeff-hykin, line 1: meaninful / rigorous tests

dimos/core/docker_worker_manager.py
```diff
@@ -0,0 +1,50 @@
...
+            _outcomes: list[Any], successes: list[DockerModule], errors: list[Exception]
+        ) -> None:
+            for mod in successes:
+                with suppress(Exception):
+                    mod.stop()
+            raise ExceptionGroup("docker deploy_parallel failed", errors)
+
+        return safe_thread_map(
+            specs, lambda spec: DockerModule(spec[0], *spec[1], **spec[2]), _on_errors
+        )
```
jeff-hykin: I don't like this class (I don't really like that WorkerManager is a class either) but I'll follow [Chestertons fence](https://theknowledge.io/chestertons-fence-explained/) in this case.

I think adding`self._docker_client` (next to `self._client` in ModuleCoordinator) is too far though

dimos/utils/safe_thread_map.py
```diff
@@ -0,0 +1,92 @@
...
+from typing import TYPE_CHECKING, Any, TypeVar
+
+if TYPE_CHECKING:
+    from collections.abc import Callable, Sequence
+
+T = TypeVar("T")
+R = TypeVar("R")
+
+
+def safe_thread_map(
```
jeff-hykin, line 40: a generic tool used by both DockerWorkerManager and WorkerManager: parallel execution with correct cleanup

.gitignore
```diff
@@ -42,6 +42,7 @@ package-lock.json
 # Ignore build artifacts
 dist/
 build/
+.Dockerfile.dimos
```
jeff-hykin: I don't like that this artifact is generated but I've already made enough changes to pre-existing DockerModule and I need to get rosnav done.

This just prevents build artifacts from building up

examples/docker_hello_world/Dockerfile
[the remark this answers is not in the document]
jeff-hykin: No, this gets generated automatically, read `_convert_dockerfile` then respond

dimos/core/docker_build.py
```diff
@@ -90,27 +70,77 @@ def _convert_dockerfile(dockerfile: Path) -> Path:
...
+def _compute_build_hash(cfg: DockerModuleConfig) -> str:
+    """Hash Dockerfile contents, build args, and SSH flag."""
+    assert cfg.docker_file is not None
+    digest = hashlib.sha256()
+    digest.update(cfg.docker_file.read_bytes())
+    for key, val in sorted(cfg.docker_build_args.items()):
+        digest.update(f"{key}={val}".encode())
+    for arg in cfg.docker_build_extra_args:
+        digest.update(arg.encode())
+    return digest.hexdigest()
```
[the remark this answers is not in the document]
jeff-hykin: not a problem

examples/docker_hello_world/hello_docker.py
```diff
@@ -0,0 +1,144 @@
...
+    default_config = HelloDockerConfig
+
+    prompt: In[str]
+    greeting: Out[str]
+
+    @rpc
+    def start(self) -> None:
+        super().start()
+        self.prompt.subscribe(self._on_prompt)
```
paul-nechifor: Example code should always follow best practices because people/LLMs are going to copy it. So in this case, you need to add to `_disposables`.

examples/docker_hello_world/hello_docker.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def _cowsay(self, text: str) -> str:
+        """Run cowsay inside the container and return the ASCII art."""
+        result = subprocess.run(
+            ["/usr/games/cowsay", text],
+            capture_output=True,
+            text=True,
+            check=True,
+        )
+        return result.stdout
```
paul-nechifor:
```suggestion
        return subprocess.check_output(["cowsay", text], text=True)
```

examples/docker_hello_world/hello_docker.py
```diff
@@ -0,0 +1,144 @@
...
+    def send(self, text: str) -> None:
+        """Publish a prompt message onto the stream."""
+        self.prompt.publish(text)
+
+    def _on_greeting(self, text: str) -> None:
+        print(f"[PromptModule] Received: {text}")
+
+
+if __name__ == "__main__":
+    from dimos.core.blueprints import autoconnect
```
paul-nechifor: Move to top.

pyproject.toml
```diff
@@ -326,8 +326,12 @@ docker = [
     "sortedcontainers",
     "PyTurboJPEG",
     "rerun-sdk",
+    "typing_extensions",
     "open3d-unofficial-arm; platform_system == 'Linux' and platform_machine == 'aarch64'",
     "open3d>=0.18.0; platform_system != 'Linux' or platform_machine != 'aarch64'",
+    # these below should be removed later, right now they are needed even for running `dimos --help` (seperate non-docker issue)
+    "langchain-core",
+    "matplotlib",
```
jeff-hykin: should get fixed by https://github.com/dimensionalOS/dimos/pull/1571

examples/docker_hello_world/hello_docker.py
jeff-hykin: My bad, thanks for catching this

dimos/core/docker_build.py
```diff
@@ -53,28 +55,6 @@
...
-    result = subprocess.run(cmd, text=True)
-    return result.returncode
-
-
-def _docker_bin(cfg: DockerModuleConfig) -> str:
-    """Get docker binary path."""
-    return cfg.docker_bin or "docker"
-
-
-def _image_exists(docker_bin: str, image_name: str) -> bool:
```
jeff-hykin, line 72: bloat removed

dimos/core/module.py
```diff
@@ -79,6 +79,8 @@ def get_loop() -> tuple[asyncio.AbstractEventLoop, threading.Thread | None]:
 
 class ModuleConfig(BaseConfig):
     rpc_transport: type[RPCSpec] = LCMRPC
+    default_rpc_timeout: float = DEFAULT_RPC_TIMEOUT
+    rpc_timeouts: dict[str, float] = dict(DEFAULT_RPC_TIMEOUTS)
```
jeff-hykin: `rpc_timeouts` is a new API cause a better timeout API was (at one point) needed for docker. While docker doesn't need it now, the work was already done and is related

dimos/core/module.py
```diff
@@ -113,7 +116,10 @@ def __init__(self, config_args: dict[str, Any]):
         self._loop, self._loop_thread = get_loop()
         self._disposables = CompositeDisposable()
         try:
-            self.rpc = self.config.rpc_transport()
+            self.rpc = self.config.rpc_transport(
+                rpc_timeouts=self.config.rpc_timeouts,
+                default_rpc_timeout=self.config.default_rpc_timeout,
```
jeff-hykin, line 131: all rpc_transports now get timeouts explicitly. The module (user level) controls the default timeout value

dimos/core/module.py
```diff
@@ -456,17 +462,6 @@ def set_transport(self, stream_name: str, transport: Transport) -> bool:  # type
         stream._transport = transport
         return True
 
-    @rpc
-    def configure_stream(self, stream_name: str, topic: str) -> bool:
-        """Configure a stream's transport by topic. Called by DockerModule for stream wiring."""
-        from dimos.core.transport import pLCMTransport
-
```
jeff-hykin, line 463: no longer needed (one of the main changes)

dimos/core/resource.py
```diff
@@ -15,7 +15,13 @@
...
 
 from abc import abstractmethod
-from typing import TYPE_CHECKING, Self
+import sys
+from typing import TYPE_CHECKING
+
+if sys.version_info >= (3, 11):
+    from typing import Self
+else:
+    from typing_extensions import Self
```
jeff-hykin, line 24: not sure how this got past CI

dimos/core/run_registry.py
```diff
@@ -21,6 +21,7 @@
 import os
 from pathlib import Path
 import re
+import signal
 import time
```
jeff-hykin, line 25: formatting (also not sure how this got past CI)

dimos/core/module_coordinator.py
```diff
@@ -113,43 +114,99 @@ def stop(self) -> None:
...
+
+        # Deploy worker and docker modules in parallel.
+        results: list[Any] = [None] * len(module_specs)
+
+        def _deploy_workers() -> None:
+            if not worker_specs:
+                return
+            assert self._client is not None
+            for index, module in zip(
+                worker_indices, self._client.deploy_parallel(worker_specs), strict=False
```
paul-nechifor: You're using `strict=False` a lot, but that's the default. I assume the linter added it. It would be better to change it to `strict=True`.

dimos/core/module.py
paul-nechifor: Can't use `dict` as a default, because it's mutable. Defaults for class attributes need to be immutable because they're shared between all instances. I think you just need to remove the default here and add

```python
self.rpc_timeouts = dict(DEFAULT_RPC_TIMEOUTS)
```

in `__init__`.

dimos/core/docker_runner.py
```diff
@@ -156,107 +162,172 @@ def _extract_module_config(cfg: DockerModuleConfig) -> dict[str, Any]:
...
-                self.rpc.call_nowait(f"{self.remote_name}/stop", ([], {}))
+            self.rpc.call_nowait(f"{self.remote_name}/stop", ([], {}))
+        self._cleanup()
+
+    def _cleanup(self) -> None:
+        """Release all resources. Idempotent — safe to call from partial init or after stop()."""
         with suppress(Exception):
             self.rpc.stop()
-        for unsub in self._unsub_fns:
+        for unsub in getattr(self, "_unsub_fns", []):
```
paul-nechifor: `_unsub_fns` is defined in `__init__`. It definitely exists. No need for `getattr`. Below, you're using it without `getattr`.

dimos/core/docker_runner.py
```diff
@@ -156,107 +162,172 @@ def _extract_module_config(cfg: DockerModuleConfig) -> dict[str, Any]:
...
+            with suppress(Exception):
+                self.stop()
+            raise
 
     def stop(self) -> None:
         """Gracefully stop the Docker container and clean up resources."""
-        # Signal remote module, stop RPC, unsubscribe handlers (ignore failures)
+        if not self._running:
+            return
+        self._running = False  # claim shutdown before any side-effects
```
paul-nechifor: It would be nicer to use an event as booleans without a lock are not threadsafe.

dimos/core/docker_runner.py
```diff
@@ -156,107 +162,172 @@ def _extract_module_config(cfg: DockerModuleConfig) -> dict[str, Any]:
 # Host-side Docker-backed Module handle
 
 
-class DockerModule:
+class DockerModule(ModuleProxyProtocol):
```
paul-nechifor: I thought this was renamed to `DockerRunner` at one point since it's not a `Module`.

dimos/core/docker_runner.py
```diff
@@ -156,107 +162,172 @@ def _extract_module_config(cfg: DockerModuleConfig) -> dict[str, Any]:
...
-        # Config
+        from dimos.core.docker_build import (
+            _compute_build_hash,
+            _get_image_build_hash,
+            build_image,
+            image_exists,
+        )
+
+        # global_config is passed by deploy pipeline but isn't a config field
+        kwargs.pop("global_config", None)
```
paul-nechifor: This has been renamed recently to `g` I think. (I don't like the name, but forgot to comment on it.)

dimos/core/docker_module.py
```diff
@@ -156,107 +162,172 @@ def _extract_module_config(cfg: DockerModuleConfig) -> dict[str, Any]:
...
         self.rpc_calls: list[str] = getattr(module_class, "rpc_calls", [])
         self._unsub_fns: list[Callable[[], None]] = []
         self._bound_rpc_calls: dict[str, RpcCall] = {}
 
-        # Build image if needed (but don't start - caller must call start() explicitly)
-        if not image_exists(config):
-            logger.info(f"Building {config.docker_image}")
-            build_image(config)
+        # Build or pull image, launch container, wait for RPC server
+        try:
```
paul-nechifor, line 171: It would be nice if all of this could be done in `start`.

`__init__` is generally for setting up the object, not doing work. If you have failures in `__init__` the object is left in an invalid state and it's hard to clean up. I assume that's why you have so many `with suppress(Exception):` in `_cleanup`.

dimos/core/docker_worker_manager.py
paul-nechifor: I was thinking that DockerWorkerManager would be like WorkerManager (i.e. it manages multiple workers). But this class doesn't manage anything, so it's not really a manager.

I think `self._docker_client` does make sense in order to keep track of the Docker workers and shut them down. Looks like I forgot to remove this code:

```
        for module_class, module in reversed(self._deployed_modules.items()):
            logger.info("Stopping module...", module=module_class.__name__)
            try:
                module.stop()
            except Exception:
                logger.error("Error stopping module", module=module_class.__name__, exc_info=True)
            logger.info("Module stopped.", module=module_class.__name__)
```

Shutdown should happen when `self._client.close_all` is called. That way all modules can be closed in parallel too. I overlooked this. 😢 

So ideally, you should be calling `self._docker_client.close_all()` and it would take care of closing the docker modules in parallel.

dimos/core/docker_module.py
```diff
@@ -156,107 +162,172 @@ def _extract_module_config(cfg: DockerModuleConfig) -> dict[str, Any]:
...
+            return
+        self._running = False  # claim shutdown before any side-effects
         with suppress(Exception):
-            if self._running:
-                self.rpc.call_nowait(f"{self.remote_name}/stop", ([], {}))
+            self.rpc.call_nowait(f"{self.remote_name}/stop", ([], {}))
+        self._cleanup()
+
+    def _cleanup(self) -> None:
+        """Release all resources. Idempotent — safe to call from partial init or after stop()."""
```
paul-nechifor, line 261: I wouldn't say that ignoring all errors is safe and idempotent. For example what error are you expecting `self._unsub_fns.clear()` to throw? It's always at least an empty list. That won't throw anything.

I understand that best effort cleanup is sometimes needed, but this is complicated because so much is done in the constructor. A cleaner constructor would make the teardown cleaner too.

dimos/core/tests/test_docker_deployment.py
```diff
@@ -0,0 +1,290 @@
...
+        from dimos.core.docker_runner import DockerModule
+
+        dm = DockerModule.__new__(DockerModule)
+        # Don't set rpcs, _module_class, or any instance attrs — simulates early __init__ failure
+        with pytest.raises(AttributeError):
+            _ = dm.some_method
+
+    def test_getattr_no_recursion_on_cleanup_attrs(self):
+        """Accessing cleanup-related attrs before they exist must raise, not recurse."""
+        from dimos.core.docker_runner import DockerModule
```
paul-nechifor: Tests should have imports at the top because you're going to import them anyway.

dimos/core/tests/test_docker_deployment.py
```diff
@@ -0,0 +1,290 @@
...
+    def test_deploy_docker_propagates_constructor_failure(
+        self, mock_worker_manager_cls, mock_docker_module_cls
+    ):
+        mock_worker_mgr = MagicMock()
+        mock_worker_manager_cls.return_value = mock_worker_mgr
+
+        # Container launch fails inside __init__; DockerModule handles its own cleanup
+        mock_docker_module_cls.side_effect = RuntimeError("launch failed")
+
+        coordinator = ModuleCoordinator()
```
paul-nechifor: There's a `dimos_cluster` fixture you might want to use, or create your own. You shouldn't initialize it in a test because if it fails `coordinator.stop()` won't be called.

paul-nechifor (dismissed):

dimos/core/docker_module.py
jeff-hykin: This is an important design discussion, and the timeout changes in this PR were basically part of me trying that route (everything in start). And yes the suppresses are because of this. We have two issues. Basically any interactive questions (stdin) must be before the module is deployed - not super relevant for docker but feels bleh to tell people to put stdin-using stuff in an `__init__`. The second though is that things like LFS downloads, docker build, docker pull, anything that can take a super massive amount of time causes start as an RPC to timeout. We kind of need an optional `build` step thats allowed to have a really long timeout. Ideally it would be an RPC so that build steps could be done in parallel.

Thoughts? I'd have no problem reworking this as I would rather get it right. 

Native modules have this same issue

dimos/core/docker_runner.py
jeff-hykin: self._running was carried over from Jalaj's PR but I agree it should be switched out.

dimos/core/docker_worker_manager.py
jeff-hykin: > deally, you should be calling self._docker_client.close_all()

But why? We are already calling stop on every module individually in module coordinator. Why is calling stop in module coordinator bad?

The docker manager isnt meaniful because, unlike WorkerManager, there's no docker worker pool or task allocation. Every docker module is completely standalone. Close all would be calling stop on all modules individually (redundant / double-stop with module coordinator)

jeff-hykin: Right now I think Managers in their current form are a dirty abstraction because its very leaky: only WorkerManager needs n_workers, which leaks upwards by needing changes the args of ModuleCoordinator, and module coordinator doesn't just work with the high-level manager, it needs to coordinate with the stop/start/close, and blend individual deployment with parallel deployment: touches modules both individually and through the workermanagers. We cant "just add another manager" we have to deeply think about WorkerManager and ModuleCooridnator.

Which is why I think they should be flattened into one until we see a pattern. Like maybe when we add cloud modules having multiple managers would make sense. But abstracting early IMO made it easier for bugs to hide and slower for me to add docker.

dimos/core/module.py
jeff-hykin: I did forget to change this back to an immutable dict / mapping proxy. 

When users pass config values its a full overwrite, and the module class should never mutate the dict. But it would be good for me to freeze it to enforce that. Is that acceptable?

I think its important that timeouts are in the config (self.config.rpc_timeouts rather than self.rpc_timeouts)

dimos/core/docker_module.py
```diff
@@ -72,9 +75,9 @@ class DockerModuleConfig(ModuleConfig):
     )  # (host, container, proto)
 
     # Runtime resources
-    docker_gpus: str | None = "all"
-    docker_shm_size: str = "2g"
-    docker_restart_policy: str = "on-failure:3"
+    docker_gpus: str | None = None
+    docker_shm_size: str = "4g"
+    docker_restart_policy: str = "no"
```
jeff-hykin, line 82: Important change to match defaults for docker CLI. 

shm_size is an exception, I debated being consistent with it but it is commonly needed to be high for ML stuff and the downside is just slightly more resource usage. There are lots of complaints about dockers default value and setting it high shouldn't cause a problem (unlike the other two defaults)

dimos/core/docker_build.py
```diff
@@ -1,120 +0,0 @@
-# Copyright 2025-2026 Dimensional Inc.
```
jeff-hykin, line 1: This was circularly dependent on docker_runner, didn't make sense to keep seperate
All functions moved as-is to bottom of docker_runner

dimos/core/docker_module.py
```diff
@@ -93,206 +103,240 @@ class DockerModuleConfig(ModuleConfig):
...
 def is_docker_module(module_class: type) -> bool:
     """Check if a module class should run in Docker based on its default_config."""
     default_config = getattr(module_class, "default_config", None)
-    return default_config is not None and issubclass(default_config, DockerModuleConfig)
-
-
-# Docker helpers
-
-
-def _run(cmd: list[str], *, timeout: float | None = None) -> subprocess.CompletedProcess[str]:
```
jeff-hykin, line 113: all helpers moved to below the class (otherwise untouched)

dimos/core/docker_module.py
```diff
@@ -473,7 +552,155 @@ def wait(self) -> None:
         self._shutdown.wait()
 
 
-def _install_signal_handlers(runner: StandaloneModuleRunner) -> None:
+def _run(cmd: list[str], *, timeout: float | None = None) -> subprocess.CompletedProcess[str]:
```
jeff-hykin, line 516: (these were just moved down from the top)

dimos/core/worker_manager.py
```diff
@@ -44,6 +50,12 @@ def start(self) -> None:
             self._workers.append(worker)
         logger.info("Worker pool started.", n_workers=self._n_workers)
 
+        if self._cfg.dtop:
+            from dimos.core.resource_monitor.monitor import StatsMonitor
+
+            self._stats_monitor = StatsMonitor(self)
+            self._stats_monitor.start()
```
jeff-hykin, line 58: stats monitor only works for python processes, so it was moved to the python worker manager (all coordinator did was give it worker manager anyways)

dimos/core/worker_manager.py
```diff
@@ -78,17 +92,34 @@ def deploy_parallel(self, module_specs: Iterable[ModuleSpec]) -> list[RPCClient]
...
+            assignments,
+            lambda item: RPCClient(item[0].deploy_module(item[1], item[2], item[3]), item[1]),
+            _on_errors,
+        )
+
+    def should_manage(self, module_class: type) -> bool:
+        """Catch-all — accepts any module not claimed by another manager."""
+        return True
+
+    def health_check(self) -> bool:
```
jeff-hykin, line 106: ModuleCoordinator was hardcoded to only doing the healthcheck for WorkerManager, so the health_check got pushed down into worker manager and is just called from ModuleCoordinator

dimos/core/worker_manager_docker.py
```diff
@@ -0,0 +1,93 @@
+# Copyright 2025-2026 Dimensional Inc.
```
jeff-hykin, line 1: Okay Paul, after our chat I was convinced to make the full WorkerManager for docker (and I feel good about it!)

dimos/hardware/sensors/camera/realsense/camera.py
```diff
@@ -445,7 +445,7 @@ def get_depth_scale(self) -> float:
 
 
 def main() -> None:
-    dimos = ModuleCoordinator(n=2)
+    dimos = ModuleCoordinator()
```
jeff-hykin, line 448: WorkerManager now gets the value directly from global_config (which could be set here but 2 is already the default)

This way ModuleCoordinator doesn't have worker-specific args. Every worker gets its config from global_config

dimos/protocol/rpc/pubsubrpc.py
```diff
@@ -290,12 +294,13 @@ def execute_and_respond() -> None:
 
 
 class LCMRPC(PubSubRPCMixin[Topic, Any], PickleLCM):
-    def __init__(self, **kwargs: Any) -> None:
-        # Need to ensure PickleLCM gets initialized properly
-        # This is due to the diamond inheritance pattern with multiple base classes
+    def __init__(
+        self, rpc_timeouts: dict[str, float], default_rpc_timeout: float, **kwargs: Any
```
jeff-hykin, line 298: just passing down rpc_timeouts

dimos/protocol/rpc/test_lcmrpc.py
```diff
@@ -18,11 +18,12 @@
 
 from dimos.constants import LCM_MAX_CHANNEL_NAME_LENGTH
 from dimos.protocol.rpc.pubsubrpc import LCMRPC
+from dimos.protocol.rpc.spec import DEFAULT_RPC_TIMEOUT
 
 
 @pytest.fixture
 def lcmrpc() -> Generator[LCMRPC, None, None]:
-    ret = LCMRPC()
+    ret = LCMRPC(rpc_timeouts={}, default_rpc_timeout=DEFAULT_RPC_TIMEOUT)
```
jeff-hykin, line 26: just making args explicit

dimos/core/docker_module.py
jeff-hykin: This should be fixed now

dimos/core/docker_module.py
jeff-hykin: Alright, build added

dimos/core/docker_runner.py
jeff-hykin: doesn't seem so. Is calling it DockerProxyModule okay?

dimos/core/worker_manager_docker.py
```diff
@@ -0,0 +1,93 @@
...
+    def deploy(
+        self,
+        module_class: type[ModuleBase],
+        global_config: GlobalConfig,
+        kwargs: dict[str, Any],
+    ) -> ModuleProxyProtocol:
+        # inlined to prevent circular dependency
+        from dimos.core.docker_module import DockerModuleProxy
+
+        mod = DockerModuleProxy(module_class, g=global_config, **kwargs)  # type: ignore[arg-type]
```
paul-nechifor: There are a few `# type: ignore`s in this file. Can they be avoided?

dimos/core/worker_manager_docker.py
paul-nechifor: Nice. 👍 You even backdated it to 2025.

dimos/core/worker_manager.py
paul-nechifor: > stats monitor only works for python processes

Does it, though? @leshy wrote StatsMonitor with more than just Python workers in mind. Docker containers also have stats like memory usage, number of children, cpu_percent, is_alive, running modules (just one), etc. Some might be hard to get, but you should still provide the ones that can be provided.

In the interest of time, it might make sense to do it in a second PR though.

dimos/core/worker.py
```diff
@@ -215,25 +215,27 @@ def deploy_module(
...
+
+            self._modules[module_id] = actor
+            logger.info(
+                "Deployed module.",
+                module=module_class.__name__,
+                worker_id=self._worker_id,
+                module_id=module_id,
+            )
+            return actor
+        finally:
```
paul-nechifor, line 237: This is for `raise RuntimeError (...`, right? You might want to just surround that statement with try-finally.

dimos/core/test_e2e_daemon.py
```diff
@@ -237,21 +235,19 @@ def test_status_shows_live_blueprint(self, live_blueprint):
...
+    def test_status_shows_live_entry_via_registry(self, live_blueprint):
+        _coord, entry = live_blueprint
 
         runs = list_runs(alive_only=True)
         matching = [r for r in runs if r.run_id == entry.run_id]
         assert len(matching) == 1
 
     def test_stop_kills_real_workers(self, live_blueprint):
-        coord, _entry = live_blueprint
+        from dimos.core.worker_manager import WorkerManager
```
paul-nechifor: Please move all these imports to the top.

dimos/core/module_coordinator.py
```diff
@@ -14,142 +14,156 @@
...
 
-    @property
-    def workers(self) -> list[Worker]:
-        """Active worker processes."""
-        if self._client is None:
-            return []
-        return self._client.workers
-
-    @property
-    def n_workers(self) -> int:
```
paul-nechifor, line 61: You can still have `n_workers`. You just have to sum of the workers from both managers.

dimos/core/module_coordinator.py
```diff
@@ -14,142 +14,156 @@
...
-        if not self._client:
+        if not self._managers:
             raise ValueError("Not started")
 
-        modules = self._client.deploy_parallel(module_specs)
-        for (module_class, _, _), module in zip(module_specs, modules, strict=True):
-            self._deployed_modules[module_class] = module  # type: ignore[assignment]
-        return modules  # type: ignore[return-value]
+        # Group specs by manager, tracking original indices for reassembly
+        groups: dict[int, WorkerManagerDocker | WorkerManager] = {}
```
paul-nechifor: You shouldn't use this type: `WorkerManagerDocker | WorkerManager`. You should define a `WorkerManagerProtocol`. Can be done in a second PR.

dimos/core/module_coordinator.py
```diff
@@ -14,142 +14,156 @@
...
 
     def stop(self) -> None:
-        if self._stats_monitor is not None:
-            self._stats_monitor.stop()
-            self._stats_monitor = None
-
         for module_class, module in reversed(self._deployed_modules.items()):
             logger.info("Stopping module...", module=module_class.__name__)
-            try:
+            with suppress(Exception):
```
paul-nechifor: Why not log to logger.error like before?

dimos/core/module_coordinator.py
```diff
@@ -14,142 +14,156 @@
...
+            WorkerManager(g=self._global_config),
+        ]
+        for m in self._managers:
+            m.start()
+
+    def _find_manager(
+        self, module_class: type[ModuleBase[Any]]
+    ) -> WorkerManagerDocker | WorkerManager:
+        for m in self._managers:
+            if m.should_manage(module_class):
```
paul-nechifor: I don't like the pattern here. What if two managers want it? First one wins, but I don't think the reason for the order is explained.

Also, the Python worker manager says it can manage anything, but I'm not sure that's true. I don't think it can run a module intended for Docker.

I'm not sure what the best future-proof solution is here. But I assume a better solution for now is to remove should_manage and just use is_docker_module directly here. That leaves the choice for how to decide between managers for the future to whoever extends this.

dimos/robot/cli/dimos.py
```diff
@@ -177,9 +177,8 @@ def run(
             coordinator.stop()
             raise typer.Exit(1)
 
-        n_workers = coordinator.n_workers
         n_modules = coordinator.n_modules
-        typer.echo(f"✓ All modules started ({n_modules} modules, {n_workers} workers)")
```
paul-nechifor, line 182: It doesn't hurt to leave this in.

dimos/core/rpc_client.py
```diff
@@ -67,23 +67,46 @@ def __call__(self, *args, **kwargs):  # type: ignore[no-untyped-def]
...
+        )
         self._unsub_fns.append(unsub_fn)
         return result
 
     def __getstate__(self):  # type: ignore[no-untyped-def]
         return (self._name, self._remote_name)
 
     def __setstate__(self, state) -> None:  # type: ignore[no-untyped-def]
-        self._name, self._remote_name = state
+        # Support both old 2-tuple and new 3-tuple (legacy) state for pickle compat.
```
paul-nechifor: What's the legacy case? I only see `__getstate__` returning 2-tuple

dimos/core/tests/test_parallel_deploy_cleanup.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+class TestWorkerManagerDockerPartialFailure:
+    """WorkerManagerDocker.deploy_parallel must stop successful containers when one fails."""
+
+    @patch("dimos.core.docker_module.DockerModuleProxy")
+    def test_middle_module_fails_stops_siblings(self, mock_docker_module_cls):
+        """Deploy 3 modules where the middle one fails. The other two must be stopped."""
+        from dimos.core.global_config import GlobalConfig
+        from dimos.core.worker_manager_docker import WorkerManagerDocker
```
paul-nechifor: These are duplicate imports imported into every test. Please move them to the top.

dimos/core/worker_manager_docker.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def stop(self) -> None:
+        for mod in reversed(self._deployed):
+            with suppress(Exception):
+                mod.stop()
+        self._deployed.clear()
+
+    def health_check(self) -> bool:
+        # TODO: in the future decide on what a meaninful health check would be
+        return True
```
paul-nechifor, line 84: Can't you check if the container is running?

dimos/core/module_coordinator.py
```diff
@@ -14,142 +14,156 @@
...
+            specs_by_manager.setdefault(mid, []).append(spec)
+
+        results: list[Any] = [None] * len(module_specs)
+
+        def _deploy_group(mid: int) -> None:
+            deployed = groups[mid].deploy_parallel(specs_by_manager[mid])
+            for index, module in zip(indices_by_manager[mid], deployed, strict=True):
+                results[index] = module
+
+        safe_thread_map(list(groups.keys()), _deploy_group)
```
paul-nechifor: This omits `on_errors`. I think the effect is that if you have an error in one manager, the modules for that worker are closed, but not the entire system.

This calls `deploy_parallel` on each manager which deploys each module in parallel also using `safe_thread_map`. But, in that case, safe_thread_map uses on_error and each manager handles the errors and stops the modules if a single one errors.

But if you have an error from a module of one manager, the modules of the other manager are not stopped.

dimos/core/worker_manager_docker.py
```diff
@@ -0,0 +1,93 @@
...
+        self,
+        module_class: type[ModuleBase],
+        global_config: GlobalConfig,
+        kwargs: dict[str, Any],
+    ) -> ModuleProxyProtocol:
+        # inlined to prevent circular dependency
+        from dimos.core.docker_module import DockerModuleProxy
+
+        mod = DockerModuleProxy(module_class, g=global_config, **kwargs)  # type: ignore[arg-type]
+        mod.build()
```
paul-nechifor: Why are you calling `build` here? Shouldn't the build happen at the build_all_modules stage? If you build during deployment by the time build_all_modules is call all modules will have been build already.

dimos/core/tests/test_docker_deployment.py
```diff
@@ -0,0 +1,229 @@
...
+        with patch.object(DockerModuleProxy, "__init__", lambda self: None):
+            dm = DockerModuleProxy.__new__(DockerModuleProxy)
+            dm._running = threading.Event()
+            dm._running.set()
+            dm._container_name = "test_container"
+            dm._unsub_fns = []
+            dm.rpc = MagicMock()
+            dm.remote_name = "TestModule"
+            dm._reconnected = True
+            dm.config = FakeDockerConfig(docker_reconnect_container=True)
```
paul-nechifor, line 219: This looks like the type of AI written test which recreates an entire object.

This uses `__new__` to avoid calling `__init__`, but it essentially duplicates `__init__` by manually setting properties on it. I'm okay with using mocks, but it shouldn't just duplicate what the `__init__` is doing because it's fragile and will break.

Also, this is duplicated in three tests. It should be a fixture at best.

dimos/core/docker_module.py
```diff
@@ -11,23 +11,23 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
```
paul-nechifor, line 14: Why was this file moved to `docker_module.py`. It gives the impression that this subclassed from `Module`.

dimos/core/worker_manager.py
jeff-hykin: yeah I mostly meant for the scope of this PR (restoring/fixing Docker) its out of scope (it'd be an additional feature). Stats Moinitor needs to be aware of the internals, because docker containers could have multiple processes inside, and we would want each to show up in dtop

dimos/core/docker_module.py
jeff-hykin: I kinda think we should expand our definition of a Module, whether its a protocol or base class or whatver, just something that is conceptually above PythonModule, NativeModule, DockerModule, CloudModule. Right now because we tell users we "have support for docker modules" I think its best that something is called/prefixed/suffixed with "docker module" so that what we tell people can be connected with whats in the code. I agree we should get these names straightend out at some point (especially for this blur with RpcClient/RpcServer), but lets do renaming after this big merge.

dimos/core/rpc_client.py
jeff-hykin: leftover from timeout changes, good call to remove

dimos/core/module_coordinator.py
jeff-hykin: thank you was a major problem. Originally full-stop was one of the fixes of this PR, somewhere along addressing the misc comments it got dropped. Changed now to a try-catch to make the behavior super clear

dimos/core/worker_manager_docker.py
jeff-hykin: good point, thats easy enough. Mostly was just trying not to add extra features ("restore functionality and fix")

dimos/core/module_coordinator.py
jeff-hykin: sounds good, added `DeploymentManager: TypeAlias = WorkerManagerDocker | WorkerManager` as a stop-gap before the second PR

dimos/core/module_coordinator.py
jeff-hykin: Maybe in a later PR, but a docker container could contain multiple python workers. "Do we count a docker container as one worker, or do we count all workers within it?" isn't a question I wanted to answer quite yet (this is similar to the StatsMonitor problem)

paul-nechifor (approved):

PR #1470, opened 2026-03-07
fix: command center socket.io auto-detect host instead of hardcoded localhost

aphexcx:
## Problem

The command center's Socket.IO connection is hardcoded to `ws://localhost:7779` in `Connection.ts`. When the command center is accessed remotely (e.g., via NAT from a different machine), the browser tries to connect back to `localhost` — which points to the browser's own machine, not the server. This silently breaks all command center functionality (costmap display, keyboard control, etc.).

## Fix

Changed `io("ws://localhost:7779")` → `io()` in `Connection.ts:22`.

`io()` with no arguments auto-detects the host from the page URL, which is the correct behavior for both:
- **Local dev**: page served from `localhost:7779`, connects to `localhost:7779` ✓
- **Remote access**: page served from `10.x.x.x:7779`, connects to `10.x.x.x:7779` ✓

Rebuilt `data/command_center.html` with the fix.

## Files changed
- `dimos/web/command-center-extension/src/Connection.ts` — one-line change
- `data/command_center.html` — rebuilt
- `data/.lfs/command_center.html.tar.gz` — rebuilt


data/command_center.html
```diff
@@ -0,0 +1,129 @@
+<!DOCTYPE html>
+<html lang="en">
```
spomichter, line 2: @aphexcx delete this then should be quick merge

spomichter (changes requested):

paul-nechifor:
We've removed command center so this is no longer relevant. Closed.

PR #1472, opened 2026-03-07
nix install for non debian based linux

leshy:
## Problem
reintroduces nix for other distors


leshy:
pushed greptile suggestion

> @leshy can you resolve greptile youre the only one that can rlly test

easy to test this by changing DETECTED_OS to "linux" - makes the installer not assume ubuntu apt-get is available

spomichter (approved):

PR #1476, opened 2026-03-07
feat(cli): dimos restart command (DIM-683)

spomichter:
## Summary

Adds `dimos restart` command that stops the running DimOS instance and re-invokes with the same arguments.

**Supersedes #1445** (which targeted the now-merged daemon branch and referenced removed `_stop_entry` function).

## Usage

```bash
dimos restart          # graceful stop + restart
dimos restart --force  # SIGKILL + restart
```

## How it works

1. Reads the most recent alive run from the registry
2. Saves its `original_argv` (exact `sys.argv` from the original run)
3. Calls `stop_entry()` (SIGTERM or SIGKILL with `--force`)
4. Waits for old process to fully exit (up to 2s poll)
5. Uses `os.execvp()` to replay the original command

No CLI flag reconstruction — just replays `sys.argv` verbatim. If the original run used `--daemon`, the restart inherits it automatically.

## Changes

- `dimos/core/run_registry.py`: Add `original_argv` field to `RunEntry` (+1 line)
- `dimos/robot/cli/dimos.py`: Store `sys.argv` in RunEntry, add `restart` command (+30 lines)

## Related

- Closes DIM-683
- Depends on daemon mode from #1436 (merged)
- Supersedes #1445

- [x] I have read and approved the [CLA](https://github.com/dimensionalOS/dimos/blob/main/CLA.md)


dimos/core/run_registry.py
```diff
@@ -52,6 +52,7 @@ class RunEntry:
     cli_args: list[str] = field(default_factory=list)
     config_overrides: dict[str, object] = field(default_factory=dict)
     grpc_port: int = 9877
+    original_argv: list[str] = field(default_factory=list)
```
Dreamsorcerer, line 55: Very minor suggestion, but it might be good to make these kind of things immutable in future generally. This also avoids needing the default_factory:
```suggestion
    original_argv: Sequence[str] = ()
```

PR #1488, opened 2026-03-08
feat(patrol): add patrolling module

paul-nechifor:
## Problem

We need a way to patrol explored areas.

Closes DIM-659

## Solution

* Wrote a few patrolling algorithms, and I'm using `coverage` by default because it looks the best.

* Wrote a test to visualize how well it covers the space (all tests are on the Hong Kong office scan from Ivan).:

<table>
  <tr>
    <td>coverage<br>patrolling_coverage_coverage</td>
    <td>frontier<br>patrolling_coverage_frontier</td>
    <td>random<br>patrolling_coverage_random</td>
  </tr>
</table>

* Tested the best values for `saturation_threshold` and `clearance_radius_m`.

patrol_router_optimization

* Tested the best values for `candidates_to_consider`.

candidates_optimization_equal_cells

* Verified that the robot patrols for a significant while.

patrol_path ignore_03

## Breaking Changes

None.

## How to Test

```bash
uv run dimos run --disable spatial-memory unitree-go2-agentic
```
and tell it to start patrolling.

### Issues

This doesn't always work because you get bad detections such as this (that's actually part of the chair):

debug-2026-03-12T00-26-49 829560+00-00

It messes up the entire flow. We need persistent 3D detections, and maybe a 360 camera.


leshy:
I'd maybe add some of those nice pictures + a short description on how it works, into docs somewhere? `docs/capabilities` ?

dimos/agents/skills/person_follow.py
```diff
@@ -107,14 +111,25 @@ def stop(self) -> None:
...
+        initial_image: str | None = None,
+    ) -> str:
         """Follow a person matching the given description using visual servoing.
 
         The robot will continuously track and follow the person, while keeping
         them centered in the camera view.
 
         Args:
             query: Description of the person to follow (e.g., "man with blue shirt")
+            initial_bbox: Optional pre-computed bounding box [x1, y1, x2, y2].
```
leshy, line 126: this is exactly what Detection2d/bbox model is, should we use those for detections to follow? 

I can imagine visual following making sense in many cases, would be good to support standard detection types?

dimos/navigation/patrolling/module.py
```diff
@@ -0,0 +1,141 @@
...
+from dimos.navigation.patrolling.create_patrol_router import create_patrol_router
+from dimos.navigation.patrolling.routers.patrol_router import PatrolRouter
+from dimos.navigation.replanning_a_star.module_spec import ReplanningAStarPlannerSpec
+from dimos.utils.logging_config import setup_logger
+
+logger = setup_logger()
+
+
+class PatrollingModule(Module):
+    odom: In[PoseStamped]
```
leshy, line 37: why not use self.tf?

dimos/navigation/patrolling/module.py
```diff
@@ -0,0 +1,141 @@
...
+    global_costmap: In[OccupancyGrid]
+    goal_reached: In[Bool]
```
leshy, line 39: this might be a good situation for jeffs RPC in which we feed nav module directly to this module as a dependancy, so it can call rpcs from it? we get nice type level dependance right? I haven't done this yet so slightly unsure

dimos/navigation/patrolling/module.py
```diff
@@ -0,0 +1,141 @@
...
+        self._stop_patrolling()
+        super().stop()
+
+    @skill
+    def start_patrol(self) -> str:
+        """Start patrolling the known area. The robot will continuously pick patrol goals from the router and navigate to them until `stop_patrol` is called."""
+        with self._patrol_lock:
+            if self._patrol_thread is not None and self._patrol_thread.is_alive():
+                return "Patrol is already running. Use `stop_patrol` to stop."
+            self._planner_spec.set_replanning_enabled(False)
```
leshy, line 84: How come? don't we have dynamic env issue if we disable replanning?

paul-nechifor: In normal navigation, when you set a goal, you need to get there. With patrolling, what's important is covering the patrol area, not necessarily getting to a specific point.

When setting a specific random target, there are issues that cause re-planning every once in a while (maybe someone got in the way, maybe some random noise blocked the path, maybe the goal was unreachable to begin with).

Instead of re-planning to get to the arbitrary target, it's a lot faster to just select another arbitrary target. If the goal is unreachable, the robot wastes a lot of time trying to get to a place that it cannot get to.

The router is in charge of making sure enough of the area is covered, so if the goal was actually important to get to, it will try to get there soon enough.

dimos/navigation/patrolling/module.py
paul-nechifor: Because `tf` does not have subscribe. You always have to request it.

dimos/navigation/patrolling/module.py
paul-nechifor: I've used Jeff's RPC in multiple places, including this PR, but I don't think it makes sense here. In this case, I want the navigation module to notify me when the goal has been reached (or declared a failure). I can't do that via RPC.

dimos/agents/skills/person_follow.py
paul-nechifor: `person_follow.py` does support navigating by Detection3D, but it's not the default. Following by EdgeTAM detections works well enough in practice. With Detection3D the distance isn't estimated in all situations, for example if a person is detected in the distance, but no voxels have been detected there yet. Visual servoing works quite nice from close-up (feet only detections) to far away detections.

But this is probably not related to this PR so maybe it should be discussed further on an issue.

leshy: aha no I didn't mean detection3D, I mean these arguments

```python
initial_bbox: list[float] | None = None,
initial_image: str | None = None,
```
are literally `Detection2d/bbox` type, just re-implemented, This is a type that carries image + bbox detection, knows how to draw itself, is returned by all our detectors (VLM, clip etc) so probably should use that as your I/O?

```
@dataclass
class Detection2DBBox(Detection2D):
    bbox: Bbox
    track_id: int
    class_id: int
    confidence: float
    name: str
    ts: float
    image: Image
```

I see now edgetam is a separate implementation of detectors, ideally it follows our standards and is just one of the 2d detectors in `detection/detectors`, we have also `dimos.perception.detection.reid` for reid models, (edgetam shoukd implement both? lmk if reid type spec is lacking tho)

so IMO we want to integrate into standard detection pipeline, so in the future I can do

(check `dimos/dimos/models/vl/test_vlm.py`)
```python
persontracker.track(any_vlm_model.query_detections("red ball"))
```

and once agents can access PY api they might do exactly that


for now ok to merge imo, but would recode edgetam at some point, what do you think?

leshy (dismissed):

leshy (approved):

PR #1512, opened 2026-03-10
fix TUI tools on MacOS/Linux when network not configured

jeff-hykin:
## Problem

Stuff like `lcmspy` doesn't work when the system isn't configured (Linux and MacOS)

Why?
- TUI takes over STDIN
- then `configure_system` runs, and tries to ask for yes/no input
- deadlock because user can't confirm yes/no 

## Solution

- patch so that system conf runs before starting the TUI system

## Breaking Changes

None

## How to Test

de-configure your multicast or network settings (maybe reboot) then run `lcmspy`

### DIFF
--- a/dimos/utils/cli/lcmspy/run_lcmspy.py
+++ b/dimos/utils/cli/lcmspy/run_lcmspy.py
@@ -80,7 +80,10 @@ class LCMSpyApp(App):  # type: ignore[type-arg]
 
     def __init__(self, *args, **kwargs) -> None:  # type: ignore[no-untyped-def]
         super().__init__(*args, **kwargs)
+        # start LCM before .run() takes over the terminal (raw mode),
+        # because autoconf uses typer.confirm() which deadlocks inside a TUI.
         self.spy = GraphLCMSpy(autoconf=True, graph_log_window=0.5)
+        self.spy.start()
         self.table: DataTable | None = None  # type: ignore[type-arg]
 
     def compose(self) -> ComposeResult:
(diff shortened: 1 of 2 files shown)


dimos/utils/cli/lcmspy/run_lcmspy.py
```diff
@@ -80,7 +80,10 @@ class LCMSpyApp(App):  # type: ignore[type-arg]
 
     def __init__(self, *args, **kwargs) -> None:  # type: ignore[no-untyped-def]
         super().__init__(*args, **kwargs)
+        # start LCM before .run() takes over the terminal (raw mode),
+        # because autoconf uses typer.confirm() which deadlocks inside a TUI.
         self.spy = GraphLCMSpy(autoconf=True, graph_log_window=0.5)
+        self.spy.start()
```
jeff-hykin, line 86: yeah calling start inside of init is cringe or whatever, but all other solutions are even more messy. We need LCM to start and ask the user about config stuff before the textual app starts, otherwise were going to get a deadlock

spomichter (approved):

dimos/utils/cli/lcmspy/run_lcmspy.py
paul-nechifor: I checked on an Ubuntu VM, and indeed lcmspy is broken without configuring on Linux as well.

But the fix could be a lot simpler. I think you just need to call autoconf. We're doing at the top of dimos.py. We should probably do it in every other command entrypoint.

Image

jeff-hykin: I think double-calling autoconf (once on the outside again on the inside) is more of a hack. Claude started with that and created a flag for turning autoconf off on the inside, but IMO thats even more of a mess.

After looking at it more, I think the better fix is for autoconf to be called in `__init__` instead of in start. I'll make a PR for that

PR #1515, opened 2026-03-10
fix: remove redundant autoconf from LCMService.start()

SUMMERxYANG:
## Problem

After PR #1503 (mac configurator) merged, users see repeated "Apply these changes now? [y/N]" prompts on startup — once per LCMService instance across all worker processes. The system configuration prompt loops indefinitely.

## Solution

Remove the `autoconf()` call from `LCMService.start()`. The blueprint system already runs all system configurators via `_run_configurators` before deploying any modules to workers, making the per-service call redundant. Standalone scripts (e.g. `drone.py`) also call `autoconf()` explicitly before creating any services.

Also removes the now-dead `LCMConfig.autoconf` field and cleans up all callsites that passed it, since the field was only ever read by the deleted code path.

## Breaking Changes

- `LCMConfig.autoconf` field removed. Any code passing `autoconf=True/False` to `LCMService`, `LCMPubSubBase`, `LCM`, `PickleLCM`, etc. will get a TypeError. Fix: remove the `autoconf=` kwarg.

## How to Test

1. Run a blueprint that spawns multiple workers (e.g. G1 agentic blueprint)
2. Verify the system configuration prompt appears only once on startup
3. Verify LCM communication still works correctly

### DIFF
--- a/dimos/robot/cli/topic.py
+++ b/dimos/robot/cli/topic.py
@@ -67,6 +67,11 @@ def topic_echo(topic: str, type_name: str | None) -> None:
             typer.echo("\nStopped.")
         return
 
+    # Warn about missing system config for standalone CLI usage.
+    from dimos.protocol.service.lcmservice import autoconf
+
+    autoconf(check_only=True)
+
     # Inferred typed mode: listen on /topic#pkg.Msg and decode from the msg_name suffix.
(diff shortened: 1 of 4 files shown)


SUMMERxYANG:
@greptile

PR #1523, opened 2026-03-11
Jeff/fix/autoconf2

jeff-hykin:
## Problem

I would get the "apply system changes?" every time on MacOS (not just every reboot but every execution)

## Solution

Run a command to correctly set the multicast value on systems that already have a value set

## Breaking Changes

Should be none, tested on summer's mac and mine.

## How to Test

```sh
dimos run demo-camera
``` 

(should only get asked questions once per reboot)

### DIFF
--- a/dimos/protocol/service/system_configurator/lcm.py
+++ b/dimos/protocol/service/system_configurator/lcm.py
@@ -168,6 +168,17 @@ class MulticastConfiguratorMacOS(SystemConfigurator):
         return f"Multicast: - sudo {' '.join(self.add_route_cmd)}"
 
     def fix(self) -> None:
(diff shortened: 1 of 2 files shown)


dimos/core/resource_monitor/stats.py
```diff
@@ -90,7 +90,7 @@ class IoStats(TypedDict):
 def _collect_io(proc: psutil.Process) -> IoStats:
     """Collect IO counters in bytes. Call inside oneshot()."""
     try:
-        io = proc.io_counters()
+        io = proc.io_counters()  # type: ignore[attr-defined]  # not available on macOS
```
jeff-hykin, line 93: my `mypy` wouldn't pass without this.

PR #1536, opened 2026-03-12
Feat/memory2

leshy:
## Summary

New memory subsystem (`memory2`) — a lazy, pull-based stream engine for storing, querying, and transforming robot observations with full type safety.

**Key pieces:**
- Generic `Stream[T]` / `Observation[T]` model with typed filters (`TimeRange`, `NearFilter`, spatial R*Tree index)
- Pluggable backends: in-memory `ListBackend` and full `SqliteBackend` (JSONB metadata, vec0 vector search, R*Tree spatial index, cursor pagination)
- Codec layer (Pickle, JPEG/TurboJPEG, LCM) with blob storage (file / SQLite)
- Embedding & vector search with `search_embedding(str | image)`
- Live reactive layer via `LiveChannel` with `.observable()` / `.subscribe()`
- Transform streams (projection, chaining, bare generator functions as transforms)

**Pluggable extension points:** `Store` (observation backend), `LiveChannel` (reactive pub/sub), `BlobStore` (binary data), `EmbeddingStore` (vector search) — each swappable independently.

**Documentation:**
- [Intro & usage](https://github.com/dimensionalOS/dimos/blob/feat/memory2/dimos/memory2/intro.md)
- [Streaming model](https://github.com/dimensionalOS/dimos/blob/feat/memory2/dimos/memory2/streaming.md)
- [Architecture overview](https://github.com/dimensionalOS/dimos/blob/feat/memory2/dimos/memory2/architecture.md)
- [Embeddings & vector search](https://github.com/dimensionalOS/dimos/blob/feat/memory2/dimos/memory2/embeddings.md)
- [Blob storage](https://github.com/dimensionalOS/dimos/blob/feat/memory2/dimos/memory2/blobstore/blobstore.md)
- [Codecs](https://github.com/dimensionalOS/dimos/blob/feat/memory2/dimos/memory2/codecs/README.md)
- [Backend implementations](https://github.com/dimensionalOS/dimos/blob/feat/memory2/dimos/memory2/impl/README.md)

## Breaking Changes

None — `memory2` is a new parallel module; existing `memory/` is untouched.

## How to Test

```bash
cd dimos/memory2 && python -m pytest
```


dimos/msgs/sensor_msgs/Image.py
```diff
@@ -377,15 +377,21 @@ def crop(self, x: int, y: int, width: int, height: int) -> Image:
...
-        sy = cv2.Sobel(gray.data, cv2.CV_32F, 0, 1, ksize=5)
-        magnitude = cv2.magnitude(sx, sy)
-        mean_mag = float(magnitude.mean())
-        if mean_mag <= 0:
+        """Return sharpness score.
+
+        Downsamples to ~160px wide before computing Laplacian variance
+        for fast evaluation (~10-20x cheaper than full-res Sobel).
+        """
+        gray = self.to_grayscale().data
```
leshy, line 384: just efficiency improvements for this algo

dimos/perception/detection/type/imageDetections.py
```diff
@@ -45,12 +45,13 @@ class ImageDetections(Generic[T], TableStr):
...
         return self.image.ts
 
-    def __init__(self, image: Image, detections: list[T] | None = None) -> None:
+    def __init__(self, image: Image | None = None, detections: list[T] | None = None) -> None:
         self.image = image
         self.detections = detections or []
-        for det in self.detections:
-            if not det.ts:
-                det.ts = image.ts
+        if image is not None:
```
leshy: decoupled detection2d from image, not super happy, but don't want to do deep interventions here in this PR

dimos/utils/threadpool.py
```diff
@@ -36,7 +36,7 @@ def get_max_workers() -> int:
         environment variable, defaulting to 4 times the CPU count.
     """
     env_value = os.getenv("DIMOS_MAX_WORKERS", "")
-    return int(env_value) if env_value.strip() else multiprocessing.cpu_count()
+    return int(env_value) if env_value.strip() else min(8, multiprocessing.cpu_count())
```
leshy, line 39: each process adds threads per core, so 32 threads (lagely unused) which wastes a lot of processing (gil context switching)

dimos/utils/docs/doclinks.py
```diff
@@ -360,12 +360,13 @@ def replace_code_match(match: re.Match[str]) -> str:
         resolved_path = resolve_candidates(candidates, file_ref)
 
         if resolved_path is None:
+            doc_rel = doc_path.relative_to(root) if doc_path.is_relative_to(root) else doc_path
```
leshy, line 363: needed doclinks to print a file that has an issue when error reporting

leshy:
@greptileai review pls

leshy:
@greptileai pls review

dimos/memory2/store.py
```diff
@@ -0,0 +1,164 @@
...
+
+        Per-stream ``overrides`` (e.g. ``live_channel=``) are merged on top of
+        the session-level defaults from :class:`SessionConfig`.
+        """
+        if name not in self._streams:
+            resolved = {k: v for k, v in vars(self.config).items() if v is not None}
+            resolved.update({k: v for k, v in overrides.items() if v is not None})
+            backend = self._create_backend(name, payload_type, **resolved)
+            self._streams[name] = Stream(source=backend)
+        return cast("Stream[T]", self._streams[name])
```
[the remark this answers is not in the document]
leshy: this is in-memory test store I don't care

dimos/memory2/impl/sqlite.py
```diff
@@ -0,0 +1,683 @@
...
+        return sorted(db_names | set(self._streams.keys()))
+
+    def delete_stream(self, name: str) -> None:
+        self._streams.pop(name, None)
+        self._registry_conn.execute(f'DROP TABLE IF EXISTS "{name}"')
+        self._registry_conn.execute(f'DROP TABLE IF EXISTS "{name}_blob"')
+        self._registry_conn.execute(f'DROP TABLE IF EXISTS "{name}_vec"')
+        self._registry_conn.execute(f'DROP TABLE IF EXISTS "{name}_rtree"')
+        self._registry_conn.execute("DELETE FROM _streams WHERE name = ?", (name,))
+        self._registry_conn.commit()
```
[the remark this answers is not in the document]
leshy: it's fine, multiprocess also won't know about the deletion, this is a very rare occurance

dimos/memory2/impl/sqlite.py
```diff
@@ -0,0 +1,683 @@
...
+            vs = self.config.vector_store
+            if vs is not None:
+                emb = getattr(obs, "embedding", None)
+                if emb is not None:
+                    vs.put(self._name, row_id, emb)
+
+            self._conn.commit()
+
+        obs.id = row_id
+        self._channel.notify(obs)
```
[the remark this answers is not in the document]
leshy: sorted

dimos/memory2/type.py
```diff
@@ -0,0 +1,114 @@
...
+    @property
+    def data(self) -> T:
+        val = self._data
+        if isinstance(val, _Unloaded):
+            if self._loader is None:
+                raise LookupError("No data and no loader set on this observation")
+            loaded = self._loader()
+            self._data = loaded
+            self._loader = None  # release closure
+            return loaded
```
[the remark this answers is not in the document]
leshy: fixed

dimos/mapping/occupancy/gradient.py
```diff
@@ -50,7 +50,7 @@ def gradient(
 
     # Compute distance transform (distance to nearest obstacle in cells)
     # Unknown cells are treated as if they don't exist for distance calculation
-    distance_cells = ndimage.distance_transform_edt(1 - obstacle_map)
+    distance_cells: np.ndarray = ndimage.distance_transform_edt(1 - obstacle_map)  # type: ignore[assignment]
```
paul-nechifor: Should use `NDArray[sometype]`

dimos/memory2/test_blobstore.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+# ── Tests ─────────────────────────────────────────────────────────
+
+
+class TestBlobStoreIntegration:
+    def test_append_stores_in_blobstore(self, tmp_path: Path) -> None:
+        bs = FileBlobStore(tmp_path / "blobs")
+        bs.start()
+        store = MemoryStore()
```
paul-nechifor: `store` should be a fixture. You need stop called. You also need to call stop on bs.

dimos/memory2/test_blobstore.py
```diff
@@ -0,0 +1,185 @@
...
+            assert len(results) == 1
+            # Data should be loaded (not _UNLOADED)
+            assert not isinstance(results[0]._data, type(_UNLOADED))
+            assert results[0].data == "payload"
+
+    def test_per_stream_eager_override(self, tmp_path: Path) -> None:
+        bs = FileBlobStore(tmp_path / "blobs")
+        bs.start()
+        store = MemoryStore()
+        with store.session(blob_store=bs) as session:
```
paul-nechifor: The entire session should be a fixture. That way you don't need to use `with` in the test.

dimos/memory2/test_e2e_import.py
```diff
@@ -0,0 +1,148 @@
...
+            return self._data[-1]
+        if ts - self._timestamps[idx - 1] <= self._timestamps[idx] - ts:
+            return self._data[idx - 1]
+        return self._data[idx]
+
+
+@pytest.fixture(scope="module")
+def store() -> Generator[SqliteStore, None, None]:
+    s = SqliteStore(path=str(DB_PATH))
+    yield s
```
paul-nechifor: ლ(¯ロ¯"ლ) For the love of God, you need to close things...

leshy: sorry at some point I actually deleted your conftest that checks for threads (some unrelated issue) and never reintroduced :face_holding_back_tears: so didn't catch these

dimos/mapping/occupancy/gradient.py
leshy: this is some mistake, removed

dimos/core/resource.py
```diff
@@ -43,3 +48,30 @@ def dispose(self) -> None:
...
+
+class CompositeResource(Resource):
+    """Resource that owns child disposables, disposed on stop()."""
+
+    _disposables: CompositeDisposable
+
+    def __init__(self) -> None:
+        self._disposables = CompositeDisposable()
+
+    def own(self, *disposables: DisposableBase) -> None:
```
leshy: proly there is a better name for this

paul-nechifor: register_disposables

dimos/core/worker.py
```diff
@@ -297,6 +297,15 @@ def _worker_entrypoint(
...
 ) -> None:
+    # Limit OpenCV internal threads to avoid idle thread contention.
+    # Modules that need parallel cv2 ops can call cv2.setNumThreads() in start().
+    try:
+        import cv2
+
+        cv2.setNumThreads(2)
+    except ImportError:
+        pass
+
```
paul-nechifor: I don't think opencv stuff belongs in worker.py. Maybe add a `bootstrap_worker` function in a different file and call it here.

dimos/memory2/backend.py
```diff
@@ -0,0 +1,165 @@
...
+    @abstractmethod
+    def notify(self, obs: Observation[T]) -> None:
+        """Fan out *obs* to all current subscribers."""
+        ...
+
+
+# ── Blob storage ──────────────────────────────────────────────────
+
+
+class BlobStore(Resource, ABC):
```
paul-nechifor: Resource is already an ABC, so you don't need to inherit from it again. That's only needed for Protocols.

dimos/core/resource.py
```diff
@@ -43,3 +48,30 @@ def dispose(self) -> None:
 
         """
         self.stop()
+
+    def __enter__(self) -> Resource:
+        self.start()
+        return self
+
+    def __exit__(self, *args: object) -> None:
+        self.stop()
```
paul-nechifor: By inheriting from DisposableBase you're limiting what `__enter__` can do. In this case, it has to return `Resource`, but that's not useful at all.

If you have `class MyClass(Resource)`, you can't do:

```
with MyClass() as my_cls:
    my_cls.my_method()
```

Because `my_method` is not declared on `DisposableBase`. The real return type should be `Self`.

dimos/core/resource.py
```diff
@@ -43,3 +48,30 @@ def dispose(self) -> None:
 
         """
         self.stop()
+
+    def __enter__(self) -> Resource:
+        self.start()
+        return self
+
+    def __exit__(self, *args: object) -> None:
```
paul-nechifor: `object` is almost never the right type. You probably want Any. But `DisposableBase` uses:

```
    def __exit__(
        self,
        exctype: Optional[Type[BaseException]],
        excinst: Optional[BaseException],
        exctb: Optional[TracebackType],
    ) -> None:
```

dimos/memory2/blobstore/blobstore.md
```diff
@@ -0,0 +1,86 @@
...
+## ABC (`backend.py`)
+
+```python
+class BlobStore(Resource, ABC):
+    def put(self, stream: str, key: int, data: bytes) -> None: ...
+    def get(self, stream: str, key: int) -> bytes: ...    # raises KeyError if missing
+    def delete(self, stream: str, key: int) -> None: ...  # silent if missing
+```
+
+- `stream` — stream name (used to organize storage: directories, tables)
```
paul-nechifor: Maybe it should be called `stream_name`.

dimos/memory2/backend.py
```diff
@@ -0,0 +1,165 @@
...
+    """Persistent storage for encoded payload blobs.
+
+    Separates payload data from metadata indexing so that large blobs
+    (images, point clouds) don't penalize metadata queries.
+
+    Extends Resource (start/stop) but does NOT manage its dependencies'
+    lifecycle — the caller owns the session / connection.
```
paul-nechifor: ~~I don't get it. Why does it not manage its own dependencies? I see Sqlite has blank start/stop and file blobstore has mkdir in start (which is not needed because the parents are created when `put()-ing` files.~~

If it doesn't use start/stop why does it need to be a resource?

**Edit:** I see. Sqlite for example takes a connection. In that case BlobStore should not be a Resource.

Also, these types of objects are usually called repositories: https://www.geeksforgeeks.org/system-design/repository-design-pattern/

So maybe rename it to `BlobRepository`?

dimos/memory2/type/backend.py
```diff
@@ -0,0 +1,165 @@
...
+        disposable that unsubscribes when disposed."""
+        ...
+
+    @abstractmethod
+    def notify(self, obs: Observation[T]) -> None:
+        """Fan out *obs* to all current subscribers."""
+        ...
+
+
+# ── Blob storage ──────────────────────────────────────────────────
```
paul-nechifor: Shouldn't this class be in `memory2/blobstore/`? Maybe `memory2/blobstore/blobstore.py`. It's documented in that dir.

dimos/memory2/blobstore/file.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def start(self) -> None:
+        self._root.mkdir(parents=True, exist_ok=True)
+
+    def stop(self) -> None:
+        pass
+
+    # ── BlobStore interface ───────────────────────────────────────
+
+    def put(self, stream: str, key: int, data: bytes) -> None:
```
paul-nechifor: Could be useful to sanitize `stream` to check it doesn't contain `/` or some other bad characters.

In the other class, stream names must work as tables, so maybe check for alphanumeric + underscore

dimos/memory2/blobstore/sqlite.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def stop(self) -> None:
+        pass
+
+    # ── BlobStore interface ───────────────────────────────────────
+
+    def put(self, stream: str, key: int, data: bytes) -> None:
+        self._ensure_table(stream)
+        self._conn.execute(
+            f'INSERT OR REPLACE INTO "{stream}_blob" (id, data) VALUES (?, ?)',
```
paul-nechifor: Dropping unescaped strings into SQL is bad.

dimos/memory2/blobstore/sqlite.py
```diff
@@ -0,0 +1,81 @@
...
+        self._ensure_table(stream)
+        row = self._conn.execute(
+            f'SELECT data FROM "{stream}_blob" WHERE id = ?', (key,)
+        ).fetchone()
+        if row is None:
+            raise KeyError(f"No blob for stream={stream!r}, key={key}")
+        result: bytes = row[0]
+        return result
+
+    def delete(self, stream: str, key: int) -> None:
```
paul-nechifor: I'm not sure it makes sense for get/delete (especially get since it shouldn't mutate) to call `_ensure_table`. If the table isn't there it can't delete the row or read the row.

dimos/memory2/blobstore/test_blobstore.py
```diff
@@ -0,0 +1,114 @@
...
+
+@pytest.fixture(params=["file", "sqlite"])
+def blob_store(
+    request: pytest.FixtureRequest,
+    file_store: FileBlobStore,
+    sqlite_store: SqliteBlobStore,
+) -> BlobStore:
+    if request.param == "file":
+        return file_store
+    return sqlite_store
```
paul-nechifor: I think this should be:

```suggestion
@pytest.fixture(params=["file_store", "sqlite_store"])
def blob_store(request: pytest.FixtureRequest) -> BlobStore:
    return request.getfixturevalue(request.param)
```

Otherwise, you're requesting both fixtures but only using one.

dimos/memory2/blobstore/sqlite.py
```diff
@@ -0,0 +1,81 @@
...
+
+from typing import TYPE_CHECKING
+
+from dimos.memory2.backend import BlobStore
+
+if TYPE_CHECKING:
+    import sqlite3
+
+
+class SqliteBlobStore(BlobStore):
```
paul-nechifor, line 40: Should this API be thread-safe? Looking at the code, it does look like it might be used from different threads at the same time?

dimos/memory2/codecs/jpeg.py
```diff
@@ -0,0 +1,63 @@
...
+    """Codec for Image types — stores as JPEG bytes (lossy, ~10-20x smaller).
+
+    Uses TurboJPEG (libjpeg-turbo) for 2-5x faster encode/decode vs OpenCV.
+    Preserves ``frame_id`` as a short header: ``<len_u16><frame_id_utf8><jpeg_bytes>``.
+    Pixel data is lossy-compressed; ``ts`` is NOT preserved (stored separately).
+    """
+
+    def __init__(self, quality: int = 50) -> None:
+        self._quality = quality
+        from turbojpeg import TurboJPEG  # type: ignore[import-untyped]
```
paul-nechifor: Please move to the top.

dimos/memory2/backend.py
leshy: sry that was a mistake in the comment

I expected normally that blobstores (and vectorstores) are fully fledged stores that manage their lifecycles (or have an option to, hence contextmanager stuff. since we might have some ultra-specialized systems for these things.

Sqliteblobstore + Sqlitestore is a special case in which sqlite store takes over blob/vector store lifecycle management (since they need to share a session) it then uses efficient JOINs to fetch data if .data for a stream is specified as non-lazy

this smart-coupling is a bit rough around the edges atm, this PR is mostly for us to agree on general approach and how we treat data. there will be a bunch of work and can split and organize IMO. we don't really have multiple databases yet to properly test coupling/decoupling yet etc, but this is the idea

reason for a decoupling between blob store and metadata stores is that actually we want medatata in relational database but likely will have a more appropriate storage system for actual large binary blobs

dimos/memory2/blobstore/test_blobstore.py
leshy: yeah completely rewrote this, test fixtures were very stupid overall

dimos/memory2/codecs/jpeg.py
leshy: rewrite this completely, jpeg codec should have used Image.lcm_jpeg_encode/decode to preserve the header

dimos/memory2/blobstore/sqlite.py
leshy: idea is mostly not, database `session` should be used in a single thread. and underlying database is responsible for efficient parallelization. (memory store is just a prototype for tests) 

there are a few contrived ways to escape this single thread restriction and cause problems (via .observable() or .subscribe()) but I think you'd really have to look to cause an issue there

dimos/core/worker.py
leshy: yeah me niether, 

this was actually consuming a bunch of resources so quickly prefixed dimos startup with this setting but obviously doesn't belong here.

should we have some "general lib config" file that's preloaded for all dimos runs? where should this be ran from?

dimos/memory2/blobstore/blobstore.md
leshy: also updated for VectorStore

dimos/memory2/type/backend.py
leshy: idk I felt like these fundamental types belong in the same place so didn't want to split around dirs and didn't want to put in type.py since it holds observations.

maybe can move all to type/

dimos/core/resource.py
leshy: all true, weird mypy didn't catch this

dimos/memory2/blobstore/sqlite.py
leshy: actually this is a convinience API for executing direct queries

dimos/memory2/type/backend.py
leshy: moved to type/

dimos/memory2/blobstore/file.py
leshy: sanitized

dimos/memory2/blobstore/sqlite.py
leshy: fixed

dimos/core/resource.py
leshy: fixed, sry about this, was vibes

dimos/memory2/backend.py
leshy: fixed

dimos/memory2/test_blobstore.py
leshy: rewrote tests/fixtures completely, they were dev crutches and really dumb

dimos/core/worker.py
paul-nechifor: I think just configuring this in a different file should be sufficient.

dimos/memory2/blobstore/sqlite.py
```diff
@@ -0,0 +1,87 @@
...
+            raise KeyError(f"No blob for stream={stream_name!r}, key={key}")
+        if row is None:
+            raise KeyError(f"No blob for stream={stream_name!r}, key={key}")
+        result: bytes = row[0]
+        return result
+
+    def delete(self, stream_name: str, key: int) -> None:
+        try:
+            self._conn.execute(f'DELETE FROM "{stream_name}_blob" WHERE id = ?', (key,))
+        except Exception:
```
paul-nechifor, line 105: Hm, I'm not sure we should be hiding this. It should be at least a warning.

dimos/memory2/impl/memory.py
```diff
@@ -0,0 +1,208 @@
...
+
+    default_config: type[BackendConfig] = BackendConfig
+
+    def __init__(
+        self, name: str = "<memory>", payload_type: type[Any] | None = None, **kwargs: Any
+    ) -> None:
+        super().__init__(**kwargs)
+        self._name = name
+        self._observations: list[Observation[T]] = []
+        self._next_id = 0
```
paul-nechifor: You can use this instead of `_next_id` (it's thread safe).

```python
from dimos.utils.sequential_ids import SequentialIds
_seq_ids = SequentialIds()
new_id = _seq_ids.next()
```

dimos/memory2/codecs/lz4.py
```diff
@@ -0,0 +1,44 @@
...
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any, TypeVar
+
+import lz4.frame  # type: ignore[import-untyped]
+
+if TYPE_CHECKING:
+    from dimos.memory2.codecs.base import Codec
+
+T = TypeVar("T")
```
paul-nechifor: It's not used.

```suggestion
```

dimos/memory2/codecs/base.py
```diff
@@ -0,0 +1,115 @@
...
+    """Auto-select codec based on payload type."""
+    from dimos.memory2.codecs.pickle import PickleCodec
+
+    if payload_type is not None:
+        from dimos.msgs.sensor_msgs.Image import Image
+
+        if issubclass(payload_type, Image):
+            from dimos.memory2.codecs.jpeg import JpegCodec
+
+            return JpegCodec()
```
paul-nechifor, line 41: I'm not sure JPEG is necessarily the best codec for an image because we have things like depth images with 16bits per channel.

But should we use a different format? I think we've used JPEG because we needed it in Foxglove, but if we're saving things for our own use we can pick our own format and not care about compatibility.

Here's a quick benchmark. I think JXL would be better, although TurboJPEG is by far the fastest.

```
Format           Library                    Time       Size  vs orig
--------------------------------------------------------------------
JPEG q50         TurboJPEG                0.003s   147.3 KB    20.3%
WebP q50         webp (pywebp)            0.078s   100.5 KB    13.9%
JPEG XL q50      pillow-jxl-plugin        0.021s    86.0 KB    11.9%
AVIF q50         pillow-avif-plugin       0.257s    87.6 KB    12.1%

```

---

Edit: JXL supports 16-bit channels so it would work for depth images too.

dimos/memory2/codecs/test_codecs.py
```diff
@@ -0,0 +1,197 @@
...
+    def test_lcm_type_returns_lcm(self) -> None:
+        from dimos.memory2.codecs.lcm import LcmCodec
+        from dimos.msgs.geometry_msgs.PoseStamped import PoseStamped
+
+        assert isinstance(codec_for(PoseStamped), LcmCodec)
+
+    def test_image_type_returns_jpeg(self) -> None:
+        pytest.importorskip("turbojpeg")
+        from dimos.memory2.codecs.jpeg import JpegCodec
+        from dimos.msgs.sensor_msgs.Image import Image
```
paul-nechifor: I think all imports in tests should be at the top because there is no speed concern since everything has to be imported anyway.

dimos/utils/threadpool.py
paul-nechifor: Yeah, this file is quite annoying as it creates the threadpool at import time whether it's needed or not.

dimos/memory2/impl/memory.py
```diff
@@ -0,0 +1,208 @@
...
+        with self._lock:
+            obs.id = self._next_id
+            self._next_id += 1
+            if encoded is not None:
+                assert bs is not None
+                bs.put(self._name, obs.id, encoded)
+                # Replace inline data with lazy loader
+                stream_name, key, codec = self._name, obs.id, self._codec
+                assert codec is not None
+                obs._data = _UNLOADED  # type: ignore[assignment]
```
paul-nechifor: Why is `# type: ignore[assignment]` needed?

dimos/memory2/impl/memory.py
```diff
@@ -0,0 +1,208 @@
...
+                if eager:
+                    _ = obs.data  # trigger lazy loader
+                yield obs
+        except (ClosedError, StopIteration):
+            pass
+        finally:
+            sub.dispose()
+
+    def count(self, query: StreamQuery) -> int:
+        return sum(1 for _ in self.iterate(query))
```
paul-nechifor: Using `iterate` triggers `_ = obs.data` which is not needed for knowing just the number of observations.

dimos/memory2/impl/sqlite.py
```diff
@@ -0,0 +1,669 @@
...
+    """
+
+    default_config: type[BackendConfig] = BackendConfig
+
+    def __init__(self, conn: sqlite3.Connection, name: str, **kwargs: Any) -> None:
+        Configurable.__init__(self, **kwargs)
+        CompositeResource.__init__(self)
+        self._conn = conn
+        self._name = name
+        self._codec: Codec[Any] = self.config.codec  # type: ignore[assignment]
```
paul-nechifor: Why `# type: ignore`

dimos/memory2/transform.py
```diff
@@ -0,0 +1,118 @@
...
+    def __call__(self, upstream: Iterator[Observation[T]]) -> Iterator[Observation[R]]: ...
+
+    def __str__(self) -> str:
+        parts: list[str] = []
+        for name in inspect.signature(self.__init__).parameters:  # type: ignore[misc]
+            val = getattr(self, name, _MISSING)
+            if val is _MISSING:
+                val = getattr(self, f"_{name}", _MISSING)
+            if val is _MISSING:
+                continue
```
paul-nechifor: I don't think you need _MISSING. You can just use `hasattr` instead of defaulting to _MISSING.

paul-nechifor (dismissed):

dimos/memory2/codecs/base.py
leshy: actually speed/cost was important for me, as memory module was storing lidar + camera@30fps + odom and I didn't want it to be close to overload of that single core, especially on light machines, that's why turbojpeg. we will have more resources for this once sqlite is not the backend (disk writing can go to psql or something) for now was too expensive

encoding individual images is a bit nuts, we should actually be storing actual videostreams, potentially passed straight through from the webrtc/whatever generates it, but required bigger rewrite

dimos/core/worker.py
leshy: added to dimos/__init__.py - also not a great place. we'll discuss/see

dimos/mapping/occupancy/gradient.py
leshy: actually this is some mypy issue in dev, not so familiar with numpy to fix atm but ignored to pass the tests

dimos/memory2/codecs/test_codecs.py
```diff
@@ -0,0 +1,181 @@
...
+def _jpeg_case() -> Case | None:
+    try:
+        from dimos.memory2.codecs.jpeg import JpegCodec
+        from dimos.utils.testing.replay import TimedSensorReplay
+
+        replay = TimedSensorReplay("unitree_go2_bigoffice/video")
+        frames = [replay.find_closest_seek(float(i)) for i in range(1, 4)]
+        codec = JpegCodec(quality=95)
+    except (ImportError, RuntimeError):
+        return None
```
Dreamsorcerer, line 130: This seems like a silent test failure. Why would we fail to import?

dimos/memory2/notifier/subject.py
```diff
@@ -0,0 +1,70 @@
...
+    from dimos.memory2.type.observation import Observation
+
+T = TypeVar("T")
+
+
+class SubjectNotifierConfig(NotifierConfig):
+    pass
+
+
+class SubjectNotifier(Notifier[T], Generic[T]):
```
Dreamsorcerer: I believe this is redundant:

```suggestion
class SubjectNotifier(Notifier[T]):
```

dimos/memory2/observationstore/sqlite.py
```diff
@@ -0,0 +1,444 @@
...
+    return (x, y or 0.0, z or 0.0, qx or 0.0, qy or 0.0, qz or 0.0, qw or 1.0)
+
+
+def _compile_filter(f: Filter, stream: str, prefix: str = "") -> tuple[str, list[Any]] | None:
+    """Compile a filter to SQL WHERE clause. Returns None for non-pushable filters.
+
+    ``stream`` is the raw stream name (for R*Tree table references).
+    ``prefix`` is a column qualifier (e.g. ``"meta."`` for JOIN queries).
+    """
+    if isinstance(f, AfterFilter):
```
Dreamsorcerer, line 87: Do we need to handle subclasses here? If not, then it might make sense to disallow them with `@final` and turn this into a mapping for better efficiency?
```
{AfterFilter: lambda f, p: f"...", ...}[type(f)](f, filter)
```

dimos/memory2/observationstore/sqlite.py
```diff
@@ -0,0 +1,444 @@
...
+        pose = f.pose
+        if pose is None:
+            return None
+        if hasattr(pose, "position"):
+            pose = pose.position
+        cx, cy, cz = _xyz(pose)
+        r = f.radius
+        # R*Tree bounding-box pre-filter + exact squared-distance check
+        rtree_sql = (
+            f'{prefix}id IN (SELECT id FROM "{stream}_rtree" '
```
Dreamsorcerer, line 114: Also, shouldn't we use sqlalchemy here? These strings are all highly vulnerable to SQL injection if there's any possibility of an attacker controlling the inputs.

dimos/memory2/observationstore/sqlite.py
```diff
@@ -0,0 +1,444 @@
...
+
+    def _ensure_tables(self) -> None:
+        """Create the metadata table and R*Tree index if they don't exist."""
+        self._conn.execute(
+            f'CREATE TABLE IF NOT EXISTS "{self._name}" ('
+            "    id      INTEGER PRIMARY KEY AUTOINCREMENT,"
+            "    ts      REAL    NOT NULL UNIQUE,"
+            "    pose_x  REAL, pose_y REAL, pose_z REAL,"
+            "    pose_qx REAL, pose_qy REAL, pose_qz REAL, pose_qw REAL,"
+            "    tags    BLOB    DEFAULT (jsonb('{}'))"
```
Dreamsorcerer, line 287: What if we need to change the table layout in future? Do we just throw away the data, or do we have DB migration in place? Again, if we use sqlalchemy, we can get this from alembic easily.

dimos/memory2/buffer.py
```diff
@@ -0,0 +1,248 @@
...
+        with self._cond:
+            self._closed = True
+            self._cond.notify_all()
+
+    def __len__(self) -> int:
+        with self._cond:
+            return len(self._buf)
+
+
+class Unbounded(BackpressureBuffer[T]):
```
Dreamsorcerer, line 212: Don't all these classes already exist in stdlib?
Certainly in asyncio, we have several of these type of queues:
https://docs.python.org/3/library/asyncio-queue.html#asyncio.Queue

dimos/memory2/conftest.py
```diff
@@ -0,0 +1,94 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Generator
+    from pathlib import Path
+
+    from dimos.memory2.blobstore.base import BlobStore
+    from dimos.memory2.store.base import Store
+
+
+@pytest.fixture
+def memory_store() -> Generator[MemoryStore, None, None]:
```
Dreamsorcerer: It's very rare to need Generator, all these can be changed:

```suggestion
def memory_store() -> Iterator[MemoryStore]:
```

dimos/memory2/conftest.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    return request.getfixturevalue(request.param)
+
+
+@pytest.fixture
+def file_blob_store(tmp_path: Path) -> Generator[FileBlobStore, None, None]:
+    store = FileBlobStore(root=str(tmp_path / "blobs"))
+    store.start()
+    yield store
+    store.stop()
```
Dreamsorcerer: Didn't these have context managers implemented?
```suggestion
    with FileBlobStore(root=str(tmp_path / "blobs")) as store:
        yield store
```

dimos/memory2/embed.py
```diff
@@ -0,0 +1,79 @@
...
+T = TypeVar("T")
+
+
+def _batched(it: Iterator[T], n: int) -> Iterator[list[T]]:
+    """Yield successive n-sized chunks from an iterator."""
+    while True:
+        batch = list(islice(it, n))
+        if not batch:
+            return
+        yield batch
```
Dreamsorcerer, line 37:
```suggestion
if sys.version_info >= (3, 12):
    from itertools import batched
else:
    def batched(it: Iterator[T], n: int) -> Iterator[list[T]]:
        """Yield successive n-sized chunks from an iterator."""
        while True:
            batch = list(islice(it, n))
            if not batch:
                return
            yield batch
```

dimos/memory2/registry.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class RegistryStore(Configurable[RegistryStoreConfig]):
+    """SQLite persistence for stream name -> config JSON."""
+
+    default_config: type[RegistryStoreConfig] = RegistryStoreConfig
+
+    def __init__(self, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        assert self.config.conn is not None, "conn is required"
```
Dreamsorcerer: Why allow None in the first place?

dimos/memory2/codecs/test_codecs.py
leshy: some bizzare CI issue with turbojpeg, needs resolving

leshy:
generally on @Dreamsorcerer comments - all very legit, but primarily in this PR I researched the correct structure and user/agent facing API and how we think about robot memory in general. I think each store, each implementation detail can use a person improving it. 

Likely this needs to be postgres and not sqlite at all, binary blobs might need a dedicated binary blob store, images should be an actual H.264 or something similar, etc etc

Codecs need to be shared between transports and memory! atm we have encoders for transports as separate classes.

I don't want to do this in the first pass as I want to start using this with actual robots and agents and visualizations, see how solid it is, how much needs changing, before comitting fully to making it rock solid in the backend.

Most feedback I'd focus on is actually high level, usage, theory and plans for this, will be much easier to give once docs are better and we are actually running on robots

dimos/memory2/registry.py
leshy: true

dimos/memory2/observationstore/sqlite.py
leshy: yeah I didn't want to deal with db migrations in the first pass, atm you reupload new lfs data blobs

dimos/memory2/observationstore/sqlite.py
leshy: yeah agreed, this is ridicilous

leshy: I hope u like php from 1995

dimos/memory2/buffer.py
leshy: I'd hope so, I agree this whole file I assume can be much more beautiful (external lib if not stdlib, idk)  PRs accepted :sweat_smile:

dimos/memory2/embed.py
leshy: idk how much I like the this, we could add dimos/utils/iterables or something that hides this?

Dreamsorcerer:
> Most feedback I'd focus on is actually high level, usage, theory and plans for this, will be much easier to give once docs are better and we are actually running on robots

I think the only high level thing for me, as mentioned privately, would be to consider making the APIs async now so we can leverage that in future.

leshy:
> > Most feedback I'd focus on is actually high level, usage, theory and plans for this, will be much easier to give once docs are better and we are actually running on robots
> 
> I think the only high level thing for me, as mentioned privately, would be to consider making the APIs async now so we can leverage that in future.

chatting in discord on this

leshy:
> > > Most feedback I'd focus on is actually high level, usage, theory and plans for this, will be much easier to give once docs are better and we are actually running on robots
> > 
> > 
> > I think the only high level thing for me, as mentioned privately, would be to consider making the APIs async now so we can leverage that in future.
> 
> chatting in discord on this

Sam Bull
Yeah, probably not an easy solution then. Probably need to make modules async first.

paul-nechifor (approved):

dimos/memory2/embed.py
Dreamsorcerer: I'd rather keep as per my suggestion. The main point being that we can remove the manual function (automatically via ruff UP) when we drop support for Python 3.11.

Issue #1401, opened 2026-03-03
Add Doom Control to drive robot


christiefhyang:
1. Quest-Side Application
Use Unity / Unreal / or native OpenXR to build a Quest app.
Obtain pose, button, joystick, and other input data from the headset and controllers.
Package these inputs into simple “control commands” (e.g., target velocity / relative pose / high-level instructions).

2. Network Protocol
Send the data from the Quest device to the PC via WebSocket / UDP / gRPC, etc.
On the PC / DimOS side, implement a small module to receive these messages.

3. “Quest Teleop” Module on the DimOS Side
Convert the received Quest control messages into:
	•	Twist → /cmd_vel (velocity), or
	•	PoseStamped → /goal_pose (pose),
Fully reuse the control interface I have already implemented under #1159.

4. Integration Testing and Simulation Validation
First connect the Quest control to a MuJoCo simulation to verify input/output consistency.
After validation in simulation, connect to the real robot.

PR #1554, opened 2026-03-14
feat: add Doom-style teleop control module (#1401)

christiefhyang:
This PR adds a Doom-style teleoperation module that lets both Unitree Go2 and G1 be driven with a shared WSAD + mouse control scheme, modeled after first-person shooter (FPS) controls. The DoomTeleop module is robot-agnostic and only uses standard navigation interfaces: continuous velocity via /cmd_vel and optional discrete pose goals via /goal_pose and /cancel_goal.

The module is then integrated into the updated Unitree Go2 and G1 blueprint architecture so that both robots can use the same keyboard plus mouse teleop without changing any existing transport topics.

## Motivation:
- Provide a more natural, FPS-like teleop experience that combines keyboard and mouse, instead of pure keyboard-only teleop.
- Unify teleop behavior across Go2 and G1 so that users can reuse the same mental model and input mapping on both platforms.
- Keep the implementation robot-agnostic, exposing only standard navigation interfaces so it can be reused in future blueprints without further changes.

## Implementation Details：
DoomTeleop Module:
- Keyboard control (continuous velocity)
•	Maps keyboard input to Twist messages:
•	W / S: move forward / backward
•	A / D: turn left / right
•	Space: emergency stop (publishes a zero Twist and clears internal input state)
•	Publishes cmd_vel: Out[Twist], which is transported on the existing /cmd_vel channel in the blueprints.
•	When the window loses focus, the module clears its input state and sends a zero Twist so the robot does not keep moving on a stuck key.

- Mouse control (yaw & optional pose)
	•	Horizontal mouse motion adjusts yaw by adding to angular.z, providing an FPS-like look and turn behavior on top of keyboard turning.
	•	Optional discrete pose helpers when navigation goal topics are wired:
	•	Right-click: sends a small forward step goal (PoseStamped) in front of the current pose.
	•	Middle-click: sends a small in-place rotation goal around the current position.
	•	Uses the existing navigation goal interfaces:
	•	goal_pose: Out[PoseStamped] mapped to /goal_pose
	•	cancel_goal: Out[Bool] mapped to /cancel_goal
	•	Space also cancels any active goal when cancel_goal is available.

- Robot-agnostic design
	•	DoomTeleop does not contain any Unitree- or robot-specific code.
	•	It only publishes Twist, PoseStamped and Bool on the standard navigation interfaces, and can be reused with any robot that exposes /cmd_vel, /goal_pose and /cancel_goal in its transports.

- Lifecycle and threading
	•	Uses a background thread plus pygame to handle keyboard and mouse events asynchronously so teleop input does not block the rest of the system.
	•	start and stop follow the existing Module lifecycle:
	•	start calls super().start and spawns the input thread.
	•	stop stops the thread, publishes a final zero Twist and cancels any active goal.

- Go2 Blueprint Integration
	•	Adds a go2_with_doom blueprint that composes the existing Unitree Go2 base navigation stack with doom_teleop.
	•	Reuses the existing Go2 transports for:
	•	/cmd_vel for velocity commands
	•	/odom and navigation goal topics where available for optional pose goals
	•	No changes are made to existing Go2 blueprints; go2_with_doom is an additive configuration.

- G1 Blueprint Integration
	•	Integrates doom_teleop into the updated modular G1 blueprint structure so it can be used alongside the existing joystick and keyboard teleop options.
	•	Wires DoomTeleop only through the established /cmd_vel and navigation goal interfaces (goal_pose, cancel_goal, odom), matching the existing G1 navigation stack.
	•	Existing G1 blueprints remain available and unchanged; Doom teleop is exposed as an additional configuration that can be selected when WSAD + mouse control is desired.

### Simulation (MuJoCo)

**Environment:** Ubuntu with MuJoCo support.

**Commands**

Install dependencies:

```bash
uv sync --extra dev --extra cpu --extra sim
```

**Go2**

```bash
uv run dimos --simulation run unitree-go2 --doom_teleop
```

**G1**

```bash
uv run dimos --simulation run unitree-g1-joystick --doom_teleop
```

### Verified

- **WSAD + mouse** control the robot as expected  
  - **W / S**: move forward / backward  
  - **A / D**: turn left / right  
  - **Mouse motion**: adds yaw to robot movement
- **Space** and the mouse-based emergency behavior immediately stop the robot and clear motion commands.
- When the teleop window **loses focus**, the module stops publishing motion commands so the robot does not continue moving on a stuck key.

---

### Robot
- Unitree **Go2** and/or **G1**

### Command

**Go2**

```bash
dimos run unitree-go2 --doom_teleop
```

**G1**

```bash
dimos run unitree-g1-joystick --doom_teleop
```

### Environment

Open indoor area with conservative speed limits on both the robot and the teleop module.

---

## Checkpoints

- Smooth **start, turning, and stopping**.
- **Emergency stop (Space / mouse)** works immediately in any state.
- **No stuck-key behavior**: releasing keys or losing window focus stops the robot instead of continuing to accelerate.

## Breaking Changes

None. This PR introduces a new module that integrates with existing interfaces without modifying or breaking any existing control paths.


dimos/robot/unitree/g1/blueprints/unitree_g1_blueprints.py
```diff
@@ -0,0 +1,269 @@
+#!/usr/bin/env python3
```
paul-nechifor: Both this and the other PR contain `unitree_g1_blueprints.py`, a file which we haven't used in a while. They also use `n_dask_workers` which has been renamed to `n_workers` so these are old files.

dimos/teleop/keyboard/doom_teleop.py
```diff
@@ -0,0 +1,323 @@
...
+                if pygame.K_w in self._keys_held:
+                    twist.linear.x += self._base_linear_speed
+                if pygame.K_s in self._keys_held:
+                    twist.linear.x -= self._base_linear_speed
+
+                # A/D = turn left/right
+                if pygame.K_a in self._keys_held:
+                    twist.angular.z += self._base_angular_speed
+                if pygame.K_d in self._keys_held:
+                    twist.angular.z -= self._base_angular_speed
```
Dreamsorcerer: I have 2 opposing suggestions here, due to limitations in pygame.

One is that we could use event.scancode, rather than event.key, assuming these produce the same result on different machines. This would keep the keys in the same place, even if the keyboard layout is different (e.g. my W is at the bottom-right on dvorak). Testing on my machine, the scancodes are 24, 4, 22 and 7.

If not, and we want to stick to the key constants, then we should probably use pygame.key.get_pressed() rather than trying to track the pressed keys ourselves. I recall several gotchas with this approach in the past (maybe to do with losing window focus etc.).

dimos/teleop/keyboard/doom_teleop.py
```diff
@@ -0,0 +1,214 @@
...
+                        stop_twist = Twist()
+                        stop_twist.linear = Vector3(0, 0, 0)
+                        stop_twist.angular = Vector3(0, 0, 0)
+                        self.cmd_vel.publish(stop_twist)
+                        print("EMERGENCY STOP!")
+                    elif event.key == pygame.K_ESCAPE:
+                        # ESC quits
+                        self._stop_event.set()
+
+            keys = pygame.key.get_pressed()
```
christiefhyang: @Dreamsorcerer I switched DoomTeleop over to use pygame.key.get_pressed() for WASD instead of maintaining our own key set, while still keeping the existing focus-loss handling. This should avoid the gotchas you mentioned around tracking pressed keys manually.

christiefhyang (commented):
Updated file path

dimos/robot/unitree/g1/blueprints/unitree_g1_blueprints.py
christiefhyang: i just updated both blueprint files

christiefhyang:
This update primarily fixes the interaction conflicts between “map click-to-goal” and “keyboard teleop” in the Command Center. The changes are focused on the actual web control pipeline, rather than the keyboard_teleop.py module itself.

Scope of Fixes
Updated frontend interaction logic in dimos/web/command-center-extension
Added backend support for cancel_goal events in dimos/web/websocket_vis/websocket_vis_module.py

Issues Resolved
Fixed the issue of “modifying the wrong file”: changes are now correctly applied to the Command Center / WebSocket control pipeline instead of mistakenly modifying keyboard_teleop.py
Fixed the issue where right-click and middle-click had no effect
Fixed the issue where users could still click to set a goal while teleop was active, causing conflicts between control modes

Behavior Changes
Left-click on map:
Only sends navigation goals when not in keyboard teleop mode
Right-click on map:
Sends cancel_goal to cancel the current navigation target
Middle-click on map:
Sends a stop teleop command, immediately zeroing velocity commands
Keyboard teleop state:
Frontend now synchronizes activation/deactivation state with upper layers
Automatically disables click-to-goal when teleop is active to prevent control conflicts

Main Modified Files
dimos/web/websocket_vis/websocket_vis_module.py
Added cancel_goal: Out[Bool]
Added cancel_goal socket event
dimos/web/command-center-extension/src/Connection.ts
Added cancelGoal()
dimos/web/command-center-extension/src/KeyboardControlPanel.tsx
Added onActiveChange
Syncs keyboard teleop state to upper layers
dimos/web/command-center-extension/src/components/VisualizerWrapper.tsx

Differentiates interactions by mouse button:
Left-click: set goal (only in goal mode)
Middle-click: stop teleop
Right-click: cancel goal
Disables browser default right-click menu
dimos/web/command-center-extension/src/App.tsx
Added isKeyboardTeleopActive
Disables map click-to-goal based on teleop state

dimos/teleop/keyboard/doom_teleop.py
Dreamsorcerer: Think you reverted it again?
https://github.com/dimensionalOS/dimos/pull/1554/changes/310939cfc600129e4c0d45e3c1ba7c87b7468290

christiefhyang: Yes I will update this soon. Encountered some issues with controls related to middle mouse button and right click.

christiefhyang:
Addressed the current review comments:

- Added the missing websocket_vis import in the Go2 blueprint file.
- Added the missing utilization import in the G1 blueprint file.
- Removed the stale dimos.robot.doom_teleop imports and kept the current dimos.teleop.keyboard.doom_teleop import.
- Replaced n_dask_workers with n_workers in the new G1 and Go2 blueprint global_config calls.

Validation:
python -m py_compile dimos/robot/unitree/go2/blueprints/unitree_go2_blueprints.py dimos/robot/unitree/g1/blueprints/unitree_g1_blueprints.py

Note: ruff was not available in the local Python environment.

paul-nechifor:

PR #1568, opened 2026-03-15
RosNav DockerModule

jeff-hykin:
DRAFT: Docker PR needs to be reviewed first

File change count is so high because g1 blueprints were redone

## Problem

G1 parity with Go2 


dimos/robot/unitree/g1/blueprints/agentic/_mujoco_skills.py
```diff
@@ -0,0 +1,160 @@
+#!/usr/bin/env python3
```
jeff-hykin, line 1: adaptation of existing code, renamed because we now have both Unity and Mujoco Sim

PR #1434, opened 2026-03-06
Booster K1 intergation

Kaweees:
## Problem

This merge request adds support for the [Booster K1](https://www.booster.tech/booster-k1/) robot, along with accompanying blueprints (`booster-k1-basic`, `booster-k1`, `booster-k1-spatial`, `booster-k1-agentic`).

> [!NOTE]
> This relies on [booster-rpc](https://pypi.org/project/booster-rpc/) ([GitHub source](https://github.com/Kaweees/booster-rpc)), a library I wrote that provides high-level control of the Booster K1 by interfacing with the same RPC and WebSocket protocols used by the [Booster App](https://www.booster.tech/open-source/). Through booster-rpc, Dimensional will be the first and only way to control the Booster wirelessly with no installation or SSH required. Additionally, the App and Dimensional can control the robot concurrently. In the future this will be generalized across all Booster robots 

Closes [DIM-583](https://linear.app/dimensional/issue/DIM-583/write-adapter-for-booster)

## Breaking Changes

None

## How to Test

`ROBOT_IP` can be found on the Booster App
Untitled drawing(3)

```sh
uv sync --extra booster
ROBOT_IP=10.0.0.15 uv run dimos --viewer rerun-web run booster-k1-agentic
```

## Contributor License Agreement

- [x] I have read and approved the [CLA](https://github.com/dimensionalOS/dimos/blob/main/CLA.md).

PR #1597, opened 2026-03-17
initial depth anything module implementation

leshy:


Kaweees:
This will be really useful for #1434, might be the nav stack needed for the Booster series

dimos/perception/depth/depth_anything3/module.py
```diff
@@ -0,0 +1,240 @@
...
+                            )
+
+            backpressure(aligned).subscribe(process_multi)
+
+    @rpc
+    def stop(self) -> None:
+        return super().stop()  # type: ignore[no-any-return]
+
+
+def deploy(
```
jeff-hykin, line 225: Probably delete

dimos/perception/depth/depth_anything3/module.py
```diff
@@ -0,0 +1,240 @@
...
+    publish_confidence: bool = False
+
+
+class DepthAnything3Module(Module[Config]):
+    default_config = Config
+
+    color_image: In[Image]
+
+    depth_image: Out[Image]
+    confidence: Out[Image]
```
jeff-hykin, line 78: I'd recommend rename to `depth_confidence` for the sake of other modules.

dimos/perception/depth/depth_anything3/module.py
```diff
@@ -0,0 +1,240 @@
...
+        from depth_anything_3.api import DepthAnything3
+
+        self._model = DepthAnything3.from_pretrained(self.config.model.value)
+        self._model = self._model.to(device=self.config.device)
+
+        # Create extra camera inputs/outputs for multi mode
+        if self.config.mode == DA3Mode.MULTI:
+            for i in range(1, self.config.num_cameras):
+                setattr(self, f"color_image_{i}", In(Image, f"color_image_{i}", self))
+                setattr(self, f"depth_image_{i}", Out(Image, f"depth_image_{i}", self))
```
jeff-hykin, line 92: I'd just leave this module to 1 camera and then lets create a general fix for multi-things rather than a module-specific multi hack. For drones we're going to need to solve the multi-agent thing pretty soon.

jeff-hykin: I understand it's better to instantiate the model once and then use batching, but the general problem of 'many modules/streams need to use < same object >' is a general problem: ex control coordinator modules sharing an event loop thread and/or network connection object

dimos/perception/depth/depth_anything3/module.py
```diff
@@ -0,0 +1,240 @@
...
+
+    # inference options
+    use_ray_pose: bool = False
+    ref_view_strategy: str = "saddle_balanced"
+
+    # outputs
+    publish_confidence: bool = False
+
+
+class DepthAnything3Module(Module[Config]):
```
jeff-hykin, line 72: A docstring explaining single/multi/temporal would be nice.

dimos/perception/depth/depth_anything3/module.py
```diff
@@ -0,0 +1,240 @@
...
+    @simple_mcache
+    def _sharp_stream(self) -> Observable[Image]:
+        return backpressure(
+            self.color_image.pure_observable().pipe(
+                sharpness_barrier(self.config.max_freq),
+            )
+        )
+
+    @simple_mcache
+    def _single_stream(self) -> Observable[tuple[Image, np.ndarray, np.ndarray]]:
```
jeff-hykin, line 126: Not sure why this is a method (compared to the much large multi-case being inlined) I don't see the cache benefit of a function that's only ever called once

jeff-hykin:
Would be nice if Paul could use this in his security demo, great reusable module.

jeff-hykin:
@greptile are the models auto-downloaded as needed?

paul-nechifor:

Issue #1596, opened 2026-03-17
Monocular SLAM experiment


leshy:
let's try

## odometry, loop closure
use orbslam3 for odometry + loop closure - write a native module, nix build -  implemented @ https://github.com/dimensionalOS/dimos/pull/1598

## dense pointclouds after odom is ok

use depthanything 3

not single image to depth! - this is trash noise, use temporal window, submit multiple frames with paralax - how do we deal with timestamp alignment? calculated depth needs to be ts aligned since it depends on odom

1. use sharpness_barrier to select sharp frames
2. calculate frames that are distinct enough (either optical flow or just pixel diff) 
3. submit blocks of frames to depth_anything in bulk
4. feed extrinsics from orbslam3 into depth_anything (https://github.com/ByteDance-Seed/depth-anything-3/blob/main/docs/API.md#-pose-conditioned-depth-estimation) 
5. convert depth to pointclouds, build a global map - write a module
depth anything 3 module implemented @ https://github.com/dimensionalOS/dimos/pull/1597

https://www.youtube.com/watch?v=EX5d3sjTI0w
https://www.youtube.com/watch?v=349oLODVVtI

## global map

use new memory2 to assemble frames based on global keypoint locations, as loop closure moves keypoints around, (each pointcloud frame placed in relation to it's nearest keypoint, then fused into global voxel grid)

global voxel grid then pruned with some statistical cutoff for hits, better sparse then wrong

PR #1598, opened 2026-03-17
feat(perception): add ORB-SLAM3 native module

leshy:
part of monocular nav experiment https://github.com/dimensionalOS/dimos/issues/1596 but ORB-SLAM3 is a good thing to have around in general

- moved orb-slam3 bindings to a separate repo due to GPL3 issues https://github.com/dimensionalOS/dimos-orb-slam3

otherwise extremely light native-module

TBH I don't understand rerun transforms (I keep forgetting and re-reading the docs) and not sure that our bridge API in this context is most convinient, so will review, this doesn't give you a nice moving camera frame, just odom and image messages, I will implement this later, it's a rerun config issue


jeff-hykin:
So we've got a problem (I'm having this with stuff in the nav stack too) ... ORB SLAM3 is copy left, GPLv3. We can't really include it in our codebase.

I think for a lot of our native modules we're going to need a way to keep it in a separate codebase and then pull it in at runtime if the user opts to use the module.

leshy:
> So we've got a problem (I'm having this with stuff in the nav stack too) ... ORB SLAM3 is copy left, GPLv3. We can't really include it in our codebase.
> 
> I think for a lot of our native modules we're going to need a way to keep it in a separate codebase and then pull it in at runtime if the user opts to use the module.

True, despite this not including orb-slam3 code, main.cpp links against orb-slam3. so we might want to put this as a separate GPL3 repo - edit - done

jeff-hykin:
if you add a comment about that weirdness (basically brokenness) you were showing me last night, then it looks good to me

leshy:
> if you add a comment about that weirdness (basically brokenness) you were showing me last night, then it looks good to me

done

jeff-hykin:
Right now I've got a voxel slam branch, a lt-mapper, a khronos-slam, better-fastlio2, etc. I'm kinda thinking after rosnav we add a handful of loop-closure benchmarks (lidar and visual) and dynamic map change benchmarks to DimOS, make sure stuff like ORB-SLAM3 is benchmarking how its supposed to, then let openclaw go crazy porting every major slam system to DimOS with validation

PR #1606, opened 2026-03-19
chore(blueprints): remove aliases

paul-nechifor:
## Problem

Blueprint aliases are annoying because there are two ways to do the same thing.

Closes DIM-538

## Solution

Removed `my_moduel()`. Now, you can only do `MyModule.blueprint()`.

## Breaking Changes

You can no longer import blueprint aliases. Just import the actual Module.


leshy (approved):
perfect yeah, I actually even think that autoconnect should be smart enough to detect it's received a module class, and call .blueprint() on it automatically?

PR #1619, opened 2026-03-20
feat(security): add security demo

paul-nechifor:
## Problem

* Agents aren't very good at coordinating skill calls.

Closes DIM-614

## Solution

* Add an additional security module which integrates everything.

## Breaking Changes

None

## How to Test

# VERY IMPORTANT: YOU MUST EXPLORE FIRST

* Move around the environment you want to patrol so that the global map fills in.
* Then start `uv run dimos --dtop --robot-ip <IP> run --disable spatial-memory unitree-go2-security`
* Run `humancli` and tell it `start the security patrol. Just call start_security_patrol. Do not ask me anything.`


dimos/robot/get_all_blueprints.py
```diff
@@ -45,9 +45,9 @@ def get_blueprint_by_name(name: str) -> Blueprint:
 def get_module_by_name(name: str) -> Blueprint:
     if name not in all_modules:
         _fail_unknown(name, list(all_modules.keys()))
-    attr_name = name.replace("-", "_")
-    python_module = __import__(all_modules[name], fromlist=[attr_name])
-    return getattr(python_module, attr_name)()  # type: ignore[no-any-return]
+    module_path, class_name = all_modules[name].rsplit(".", 1)
+    python_module = __import__(module_path, fromlist=[class_name])
+    return getattr(python_module, class_name).blueprint()  # type: ignore[no-any-return]
```
jeff-hykin, line 50: thanks for breaking this out into that other PR. It does make it easier to understand.

dimos/stream/audio/tts/node_openai.py
```diff
@@ -204,8 +204,12 @@ def dispose(self) -> None:
 
         self.is_running = False
 
+        # Clear pending items so the thread doesn't start new synthesis.
+        with self.queue_lock:
+            self.text_queue.clear()
+
         if self.processing_thread and self.processing_thread.is_alive():
-            self.processing_thread.join(timeout=5.0)
+            self.processing_thread.join(timeout=30.0)
```
jeff-hykin: why so big? this makes me worry about other timeouts

dimos/stream/audio/node_output.py
```diff
@@ -128,14 +130,15 @@ def stop(self) -> None:
...
+        with self._stream_lock:
+            if self._stream:
+                self._stream.stop()
+                self._stream.close()
+                self._stream = None
 
     def _play_audio_event(self, audio_event) -> None:  # type: ignore[no-untyped-def]
         """Play audio from an AudioEvent."""
-        if not self._running or not self._stream:
+        if not self._running:
```
jeff-hykin: why does self._stream need a lock but self._running doesn't?

dimos/experimental/security_demo/depth_estimator.py
```diff
@@ -0,0 +1,94 @@
+# Copyright 2026 Dimensional Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
```
jeff-hykin, line 3: I think it'd be kinda nice if this was a general tool as a module. (Follow up PR tho)

dimos/agents/skills/speak_skill.py
```diff
@@ -64,8 +70,32 @@ def speak(self, text: str) -> str:
...
+            return f"Speaking (non-blocking): {text}"
+
+        return self._speak_blocking(text)
+
+    def _speak_bg(self, text: str) -> None:
+        try:
+            self._speak_blocking(text)
+        finally:
+            with self._bg_threads_lock:
+                self._bg_threads = [
```
jeff-hykin, line 90: What's this for? (Or maybe why is it in the speak action rather than when a new thread is added)

dimos/agents/skills/speak_skill.py
```diff
@@ -41,6 +43,10 @@ def start(self) -> None:
 
     @rpc
     def stop(self) -> None:
+        with self._bg_threads_lock:
+            threads = list(self._bg_threads)
+        for t in threads:
+            t.join(timeout=10.0)
```
jeff-hykin: If you could, please put hardcoded numbers in a const or private class attr. It always makes me think "why 10? We use 5 and 3 in other places"

jeff-hykin: Hopefully with thread utils we can centralize the timeouts instead of hardcoded and different values everywhere

dimos/experimental/security_demo/depth_estimator.py
```diff
@@ -0,0 +1,94 @@
...
+from transformers import AutoImageProcessor, AutoModelForDepthEstimation
+
+from dimos.msgs.sensor_msgs.Image import Image, ImageFormat
+
+_DEPTH_MODEL_NAME = "depth-anything/Depth-Anything-V2-Small-hf"
+_DEPTH_MAX_WIDTH = 640
+
+
+class DepthEstimator:
+    """Runs depth estimation in a background thread, always processing only the latest image."""
```
jeff-hykin: Some details/warnings about how larger-than-max-width image sizes are handled would be nice in the docstring.

Could be very relevant for 360 cameras

dimos/experimental/security_demo/depth_estimator.py
```diff
@@ -0,0 +1,94 @@
...
+                self._process(image)
+
+    def _process(self, image: Image) -> None:
+        rgb = image.to_rgb()
+        pil_image = PILImage.fromarray(rgb.data)
+        if pil_image.width > _DEPTH_MAX_WIDTH:
+            scale = _DEPTH_MAX_WIDTH / pil_image.width
+            new_h = int(pil_image.height * scale)
+            pil_image = pil_image.resize((_DEPTH_MAX_WIDTH, new_h), PILImage.Resampling.BILINEAR)
+        inputs = self._processor(images=pil_image, return_tensors="pt").to("cuda")
```
jeff-hykin: Please make .to('cuda') configurable so this can run on cpu and macOS metal. The cuda-available check can be in the init or start and default to cuda if available

dimos/experimental/security_demo/security_module.py
```diff
@@ -0,0 +1,401 @@
...
+State = Literal["IDLE", "PATROLLING", "FOLLOWING"]
+
+
+class SecurityModuleConfig(ModuleConfig):
+    camera_info: CameraInfo
+    follow_frequency: float = 20.0
+
+
+def _create_router(global_config: GlobalConfig) -> PatrolRouter:
+    clearance_multiplier = 0.5
```
jeff-hykin, line 112: We probably want clearance to be user configurable (module config), which I think seems easier than trying to get the robot width.

dimos/experimental/security_demo/security_module.py
```diff
@@ -0,0 +1,401 @@
...
+        persons visually and then follow them automatically.
+        """
+        with self._lock:
+            if self._main_thread is not None and self._main_thread.is_alive():
+                return "Security patrol is already running. Use `stop_security_patrol` to stop."
+
+        self._router.reset()
+
+        self._planner_spec.set_replanning_enabled(False)
+        self._planner_spec.set_safe_goal_clearance(self.config.g.robot_rotation_diameter / 2 + 0.2)
```
jeff-hykin: Please put the + 0.2 into a var or config. The units or "when would I maybe change this" is not very obvious

dimos/experimental/security_demo/security_module.py
```diff
@@ -0,0 +1,401 @@
...
+            self._has_active_goal = True
+
+        if self._goal_reached_event.is_set():
+            self._goal_reached_event.clear()
+            self._has_active_goal = False
+
+        with self._lock:
+            image = self._latest_image
+        if image is None:
+            self._stop_event.wait(timeout=0.01)
```
jeff-hykin: Same as other timeout complaints: why 0.01 when others are orders of magnitude different

dimos/experimental/security_demo/security_module.py
```diff
@@ -0,0 +1,401 @@
...
+
+        best = max(detections.detections, key=lambda d: d.bbox_2d_volume())
+        twist = self._visual_servo.compute_twist(best.bbox, latest_image.width)
+        self.cmd_vel.publish(twist)
+
+        overlay = latest_image.data.copy()
+        if hasattr(best, "mask") and best.mask is not None:
+            mask_bool = best.mask > 0
+            green = np.zeros_like(overlay)
+            green[:, :] = (0, 255, 0) if latest_image.format == ImageFormat.BGR else (0, 128, 0)
```
jeff-hykin: I'm a bit confused by this. The 0,255,0 makes sense. I'm not sure why all other formats are 128, and not sure about the color being hardcoded. I don't want to block on this if it's just a quick hack that'll be cleaned up later though. Maybe later we can add image format agnostic methods for masking

dimos/experimental/security_demo/security_module.py
```diff
@@ -0,0 +1,401 @@
...
+
+        self._planner_spec.set_replanning_enabled(True)
+        self._planner_spec.reset_safe_goal_clearance()
+
+        self._cancel_current_goal()
+
+        with self._lock:
+            thread = self._main_thread
+        if thread is not None:
+            thread.join(timeout=5.0)
```
jeff-hykin: (See other timeouts comments)

dimos/robot/unitree/go2/blueprints/smart/unitree_go2_spatial.py
```diff
@@ -14,14 +14,17 @@
...
 from dimos.perception.perceive_loop_skill import PerceiveLoopSkill
 from dimos.perception.spatial_perception import SpatialMemory
 from dimos.robot.unitree.go2.blueprints.smart.unitree_go2 import unitree_go2
+from dimos.robot.unitree.go2.connection import GO2Connection
 
 unitree_go2_spatial = autoconnect(
     unitree_go2,
     SpatialMemory.blueprint(),
     PerceiveLoopSkill.blueprint(),
+    SecurityModule.blueprint(camera_info=GO2Connection.camera_info_static),
```
jeff-hykin, line 27: It's too bad we don't have a way to get this from a camera spec (kinda like rpc). Maybe I'll put this in the core 2.0 wishlist

dimos/robot/get_all_blueprints.py
jeff-hykin: Huh I guess that other PR still hasn't been merged

jeff-hykin (changes requested):
Minor changes: cuda, the distance parameter, and the self-added todo-fix

dimos/robot/get_all_blueprints.py
paul-nechifor: This one: https://github.com/dimensionalOS/dimos/pull/1689/changes ? It has.

dimos/stream/audio/tts/node_openai.py
paul-nechifor: Changed to 2.

dimos/stream/audio/node_output.py
paul-nechifor: True. Changed to Event.

dimos/experimental/security_demo/depth_estimator.py
paul-nechifor: Ivan has a PR for Depth Anything V3. I've used V2 here because it's easy to use. It's not the best model. Also, this whole thing is in experimental.

dimos/agents/skills/speak_skill.py
paul-nechifor: When a new non-block speak command comes, it's added to `_bg_threads`. So that on stop they can be stopped. This is the end of the speak command. This thread is exiting, so it's being removed from `_bg_threads`.

I've added a comment to explain this.

dimos/agents/skills/speak_skill.py
paul-nechifor: Changed it to 2. I'll do a separate PR to change all timeouts.

dimos/experimental/security_demo/depth_estimator.py
paul-nechifor: > Some details/warnings about how larger-than-max-width image sizes are handled would be nice in the docstring.

But I have no idea how that's handled... This is only used here to make visualization more interesting. It doesn't actually do anything.

dimos/experimental/security_demo/depth_estimator.py
paul-nechifor: I don't think that's worth while. If the model cannot run in CUDA, you shouldn't run the model at all. I'm using this:

```
        if torch.cuda.is_available():
            self._depth_estimator = DepthEstimator(self.depth_image.publish)
```

so the model won't run if CUDA is not available. I have CUDA and the model already cannot process everything in real time (skipping frames with real 720p video). I can't imagine it would be of any use without CUDA.

dimos/experimental/security_demo/security_module.py
paul-nechifor: I disagree. I think that when you make something user-configurable it should be done so with intention and with the knowledge that you have to support that.

`clearance_multiplier` is not something I want people to be able to configure. It affects how the algorithm works. There is an optimal value. It's not something the user can adjust based on preference. It's also not something I want to commit to using. Maybe we'll change how the algorithm works and it's not needed at all.

If someone has a need to configure it, then we should expose it, but I don't believe in exposing everything by default because it leaves us with a larger surface area which have to support.

---

If you want, I can make it an internal constant.

dimos/robot/get_all_blueprints.py
jeff-hykin: Then why would these show up as changes? Should be same as dev right?

dimos/experimental/security_demo/security_module.py
paul-nechifor: Changed to _ANTI_BUSY_LOOP_TIMEOUT though I'm not too sure it's needed.

dimos/experimental/security_demo/security_module.py
paul-nechifor: Moved.

dimos/experimental/security_demo/security_module.py
paul-nechifor: :sweat_smile: I didn't even look at the generated numpy code. Yeah, that doesn't make any sense. I've removed it.

dimos/experimental/security_demo/security_module.py
paul-nechifor: Changed to 2.

dimos/experimental/security_demo/depth_estimator.py
jeff-hykin: Sounds good, and yep I went ahead and reviewed it after I did this review 👍

dimos/experimental/security_demo/security_module.py
jeff-hykin: > There is an optimal value. It's not something the user can adjust based on preference

Ah this changes my understanding then. I don't yet understand why/how it would effect the algorithm beyond being a preference. 

How was the number calculated?

dimos/experimental/security_demo/depth_estimator.py
jeff-hykin: Sorry I should've said non-cuda rather than cpu, the model almost certainly runs fine on macOS metal and amd GPUs

jeff-hykin: In general I'm not a fan of hardware gatekeeping, especially on a specific brand/company. As a poor grad student I ran many many oversized models at 0.5fps on my cpu to great success for what I wanted (often requiring hours or days of codebase hacks). Sometimes all I wanted to do was printout a value to understand an algorithm, or see if the codebase had too much bit-rot. The definition of "fast enough" is inherently defined by the end user, not us.

dimos/experimental/security_demo/depth_estimator.py
jeff-hykin: Ah okay

jeff-hykin: Maybe put a note like: only intended for visualization (human consumption only)

dimos/experimental/security_demo/security_module.py
paul-nechifor: > How was the number calculated?

With SCIENCE!

0.5 is used since it halves the width and uses clearance_radius_m=0.2

https://github.com/dimensionalOS/dimos/pull/1488

560743723-67fd7f42-934f-4fd5-a9b1-09d04355270f

jeff-hykin: Okay lol. I'd say a comment referencing that is good for me!

dimos/experimental/security_demo/depth_estimator.py
paul-nechifor: This is supposed to run in real time. If it runs half a minute behind, the user doesn't have a legitimate objection. He's just wrong. None of this stops Mac users from running this. It doesn't run a model which their computer doesn't support.

jeff-hykin: When I run this on my mac, I have to change the .to("cuda") for depth to work. When I make that change it renders in real time no problem.

jeff-hykin: https://github.com/user-attachments/assets/6ae6ee03-a485-4811-abdc-3722d8782659

paul-nechifor: Sorry, didn't realize it could run without CUDA. Fixed.

dimos/experimental/security_demo/depth_estimator.py
paul-nechifor: Fixed.

jeff-hykin (dismissed):

dimos/agents/skills/speak_skill.py
```diff
@@ -64,8 +70,33 @@ def speak(self, text: str) -> str:
         if self._tts_node is None:
             return "Error: TTS not initialized"
```
leshy, line 72: this speak threading code is brutal, probably rxpy would make this much more readable, like you can feed stuff to say to a Subject, but won't block

dimos/navigation/patrolling/module.py
```diff
@@ -83,7 +84,7 @@ def start_patrol(self) -> str:
                 return "Patrol is already running. Use `stop_patrol` to stop."
             self._planner_spec.set_replanning_enabled(False)
             self._planner_spec.set_safe_goal_clearance(
-                self._global_config.robot_rotation_diameter / 2 + 0.2
+                self._global_config.robot_rotation_diameter / 2 + EXTRA_CLEARANCE
```
leshy, line 88: extra clearance seems like a navigation module configuration setting, why are we using constants for this? shouldn't this be provided as a default config for Nav module, and potentially overriden per robot base blueprint?

dimos/robot/unitree/go2/blueprints/agentic/unitree_go2_security.py
```diff
@@ -0,0 +1,96 @@
...
+                row_shares=[1, 2],
+            ),
+            column_shares=[1, 2],
+        ),
+        rrb.TimePanel(state="hidden"),
+        rrb.SelectionPanel(state="hidden"),
+    )
+
+
+rerun_config = {
```
leshy, line 78: just so you know, if you don't provide a rerun blueprint you are able to click click build your dash and save as a file, you can load later, put in assets/lfs etc, for these temp things

leshy (approved):

dimos/navigation/patrolling/module.py
paul-nechifor: No, I'm just adding more space here so that when selecting goals, it picks something further away from objects. The idea being that in patroling, you just want to move around the space, going through the most common parts rather than waste time going close to the walls.

dimos/agents/skills/speak_skill.py
paul-nechifor: I'll try this on a separate PR.

PR #1663, opened 2026-03-25
Threading Utils (and fix for Native Module flakey test)

jeff-hykin:
## Problem

Flakey test, and a history of Flakey tests surrounding threads in modules.

## Solution

Smart thread tooling with auto-cleanup to reduce the risk of bad cleanup and reduce bloat.

## Breaking Changes

None, did more testing than usual because this touches core.

## How to Test

```bash
# Thread utility tests (deadlocks, races, stress)
python -m pytest dimos/utils/test_thread_utils.py -v --noconftest

# NativeModule + MCP server tests
python -m pytest dimos/core/test_native_module.py dimos/agents/mcp/ -v --noconftest

# Full suite
python -m pytest --timeout=120 -q --ignore=dimos/perception/detection/type
```

### DIFF
--- a/dimos/agents/mcp/mcp_server.py
+++ b/dimos/agents/mcp/mcp_server.py
@@ -180,7 +180,6 @@ class McpServer(Module):
     def stop(self) -> None:
         if self._uvicorn_server:
             self._uvicorn_server.should_exit = True
-            loop = self._async_thread.loop
             if self._serve_future is not None:
                 self._serve_future.result(timeout=5.0)
             self._uvicorn_server = None
--- a/dimos/agents/mcp/mcp_server.py
+++ b/dimos/agents/mcp/mcp_server.py
@@ -181,7 +181,7 @@ class McpServer(Module):
         if self._uvicorn_server:
             self._uvicorn_server.should_exit = True
             loop = self._async_thread.loop
-            if loop is not None and self._serve_future is not None:
+            if self._serve_future is not None:
                 self._serve_future.result(timeout=5.0)
             self._uvicorn_server = None
             self._serve_future = None
--- a/dimos/agents/mcp/mcp_server.py
+++ b/dimos/agents/mcp/mcp_server.py
@@ -180,7 +180,7 @@ class McpServer(Module):
     def stop(self) -> None:
         if self._uvicorn_server:
             self._uvicorn_server.should_exit = True
-            loop = self._loop
+            loop = self._async_thread.loop
             if loop is not None and self._serve_future is not None:
                 self._serve_future.result(timeout=5.0)
             self._uvicorn_server = None
@@ -250,6 +250,5 @@ class McpServer(Module):
         config = uvicorn.Config(app, host=_host, port=_port, log_level="info")
         server = uvicorn.Server(config)
         self._uvicorn_server = server
-        loop = self._loop
-        assert loop is not None
+        loop = self._async_thread.loop
         self._serve_future = asyncio.run_coroutine_threadsafe(server.serve(), loop)
--- a/dimos/agents/mcp/test_mcp_server.py
+++ b/dimos/agents/mcp/test_mcp_server.py
@@ -16,9 +16,13 @@ from __future__ import annotations
 
 import asyncio
 import json
+import socket
+import time
 from unittest.mock import MagicMock
 
-from dimos.agents.mcp.mcp_server import handle_request
+import requests
+
+from dimos.agents.mcp.mcp_server import McpServer, handle_request
 from dimos.core.module import SkillInfo
 
 
@@ -111,3 +115,56 @@ def test_mcp_module_initialize_and_unknown() -> None:
 
     response = asyncio.run(handle_request({"method": "unknown/method", "id": 2}, [], {}))
     assert response["error"]["code"] == -32601
+
+
+def _free_port() -> int:
+    with socket.socket() as s:
+        s.bind(("", 0))
+        return s.getsockname()[1]
+
+
+def test_mcp_server_lifecycle() -> None:
+    """Start a real McpServer, hit the HTTP endpoint, then stop it.
+
+    This exercises the AsyncModuleThread event loop integration that the
+    unit tests above do not cover.
+    """
+    port = _free_port()
+
+    server = McpServer()
+    server._start_server(port=port)
+    url = f"http://127.0.0.1:{port}/mcp"
+
+    # Wait for the server to be ready
+    for _ in range(40):
+        try:
+            resp = requests.post(
+                url,
+                json={"jsonrpc": "2.0", "method": "initialize", "id": 1},
+                timeout=0.5,
+            )
+            if resp.status_code == 200:
+                break
+        except requests.ConnectionError:
+            time.sleep(0.1)
+    else:
+        server.stop()
+        raise AssertionError("McpServer did not become ready")
+
+    # Verify it responds
+    data = resp.json()
+    assert data["result"]["serverInfo"]["name"] == "dimensional"
+
+    # Stop and verify it shuts down
+    server.stop()
+    time.sleep(0.3)
+
+    with socket.socket() as s:
+        # Port should be released after stop
+        try:
+            s.connect(("127.0.0.1", port))
+            s.close()
+            # If we could connect, the server is still up — that's a bug
+            raise AssertionError("McpServer still listening after stop()")
+        except ConnectionRefusedError:
+            pass  # expected — server is down
--- a/dimos/agents/mcp/test_mcp_server.py
+++ b/dimos/agents/mcp/test_mcp_server.py
@@ -123,6 +123,25 @@ def _free_port() -> int:
         return s.getsockname()[1]
 
 
+def _wait_for_mcp(url: str, timeout: float = 4.0) -> requests.Response:
+    """Poll *url* until it answers an ``initialize`` RPC or *timeout* expires."""
+    deadline = time.monotonic() + timeout
+    while True:
+        try:
+            resp = requests.post(
+                url,
+                json={"jsonrpc": "2.0", "method": "initialize", "id": 1},
+                timeout=0.5,
+            )
+            if resp.status_code == 200:
+                return resp
+        except requests.ConnectionError:
+            pass
+        if time.monotonic() >= deadline:
+            raise AssertionError("McpServer did not become ready")
+        time.sleep(0.1)
+
+
 def test_mcp_server_lifecycle() -> None:
     """Start a real McpServer, hit the HTTP endpoint, then stop it.
 
@@ -133,30 +152,16 @@ def test_mcp_server_lifecycle() -> None:
 
     server = McpServer()
     server._start_server(port=port)
-    url = f"http://127.0.0.1:{port}/mcp"
+    try:
+        url = f"http://127.0.0.1:{port}/mcp"
+        resp = _wait_for_mcp(url)
 
-    # Wait for the server to be ready
-    for _ in range(40):
-        try:
-            resp = requests.post(
-                url,
-                json={"jsonrpc": "2.0", "method": "initialize", "id": 1},
-                timeout=0.5,
-            )
-            if resp.status_code == 200:
-                break
-        except requests.ConnectionError:
-            time.sleep(0.1)
-    else:
+        # Verify it responds
+        data = resp.json()
+        assert data["result"]["serverInfo"]["name"] == "dimensional"
+    finally:
         server.stop()
-        raise AssertionError("McpServer did not become ready")
 
-    # Verify it responds
-    data = resp.json()
-    assert data["result"]["serverInfo"]["name"] == "dimensional"
-
-    # Stop and verify it shuts down
-    server.stop()
     time.sleep(0.3)
 
     with socket.socket() as s:
@@ -164,7 +169,6 @@ def test_mcp_server_lifecycle() -> None:
         try:
             s.connect(("127.0.0.1", port))
             s.close()
-            # If we could connect, the server is still up — that's a bug
             raise AssertionError("McpServer still listening after stop()")
         except ConnectionRefusedError:
             pass  # expected — server is down
--- a/dimos/core/module.py
+++ b/dimos/core/module.py
@@ -95,9 +95,9 @@ class ModuleBase(Configurable[ModuleConfigT], Resource):
     rpc_calls: list[str] = []
 
     def __init__(self, config_args: dict[str, Any]):
+        self.mod_state = ThreadSafeVal[ModState]("init")
         super().__init__(**config_args)
         self._disposables = CompositeDisposable()
-        self.mod_state = ThreadSafeVal[ModState]("init")
         self._async_thread = (
             AsyncModuleThread(  # NEEDS to be created after self._disposables exists
                 module=self
--- a/dimos/core/module.py
+++ b/dimos/core/module.py
@@ -46,7 +46,7 @@ from dimos.utils import colors
 from dimos.utils.generic import classproperty
 from dimos.utils.thread_utils import AsyncModuleThread, ThreadSafeVal
 
-ModState = Literal["init", "started", "stopping", "stopped"]
+ModState = Literal["init", "started", "stopped"]
 
 if TYPE_CHECKING:
     from dimos.core.blueprints import Blueprint
@@ -130,9 +130,9 @@ class ModuleBase(Configurable[ModuleConfigT], Resource):
 
     def _stop(self) -> None:
         with self.mod_state as state:
-            if state in ("stopping", "stopped"):
+            if state == "stopped":
                 return
-            self.mod_state.set("stopping")
+            self.mod_state.set("stopped")
 
         if self.rpc:
             self.rpc.stop()  # type: ignore[attr-defined]
--- a/dimos/core/module.py
+++ b/dimos/core/module.py
@@ -11,14 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import asyncio
 from collections.abc import Callable
 from dataclasses import dataclass
 from functools import partial
 import inspect
 import json
 import sys
-import threading
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -100,8 +98,10 @@ class ModuleBase(Configurable[ModuleConfigT], Resource):
         super().__init__(**config_args)
         self._disposables = CompositeDisposable()
         self.mod_state = ThreadSafeVal[ModState]("init")
-        self._async_thread = AsyncModuleThread( # NEEDS to be created after self._disposables exists
-            module=self
+        self._async_thread = (
+            AsyncModuleThread(  # NEEDS to be created after self._disposables exists
+                module=self
+            )
         )
         try:
             self.rpc = self.config.rpc_transport()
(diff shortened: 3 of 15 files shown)


dimos/core/module.py
```diff
@@ -128,58 +119,43 @@ def frame_id(self) -> str:
 
     @rpc
     def start(self) -> None:
-        pass
+        with self.mod_state as state:
+            if state == "stopped":
+                raise RuntimeError(f"{type(self).__name__} cannot be restarted after stop")
+            self.mod_state.set("started")
```
jeff-hykin, line 125: I know lots of modules don't call super().start() but they also wouldn't be using mod_state cause its a new thing. 

Different/off-topic discussion, but I think core2 should have ModuleBase as class decorator instead of an inherited class (we can basically wrap methods instead of saying "please remember to call super").

dimos/core/module.py
```diff
@@ -128,58 +119,45 @@ def frame_id(self) -> str:
...
 
-        self._close_rpc()
-
-        # Save into local variables to avoid race when stopping concurrently
-        # (from RPC and worker shutdown)
-        loop_thread = getattr(self, "_loop_thread", None)
-        loop = getattr(self, "_loop", None)
+        # dispose of things BEFORE making aspects like rpc and _tf invalid
+        if hasattr(self, "_disposables"):
+            self._disposables.dispose()  # stops _async_thread via disposable
```
jeff-hykin, line 137: I think its important to move disposables up before the rpc stop and the tf stop

dimos/agents/mcp/mcp_server.py
```diff
@@ -180,8 +180,7 @@ def start(self) -> None:
     def stop(self) -> None:
         if self._uvicorn_server:
             self._uvicorn_server.should_exit = True
-            loop = self._loop
-            if loop is not None and self._serve_future is not None:
```
jeff-hykin, line 184: the loop is always there until super().stop() is called

dimos/agents/mcp/mcp_server.py
```diff
@@ -250,6 +249,5 @@ def _start_server(self, port: int | None = None) -> None:
         config = uvicorn.Config(app, host=_host, port=_port, log_level="info")
         server = uvicorn.Server(config)
         self._uvicorn_server = server
-        loop = self._loop
-        assert loop is not None
```
jeff-hykin, line 255: loop always there until stop is called

dimos/agents/mcp/test_mcp_server.py
```diff
@@ -111,3 +115,56 @@ def test_mcp_module_initialize_and_unknown() -> None:
...
     assert response["error"]["code"] == -32601
+
+
+def _free_port() -> int:
+    with socket.socket() as s:
+        s.bind(("", 0))
+        return s.getsockname()[1]
+
+
+def test_mcp_server_lifecycle() -> None:
```
jeff-hykin, line 145: new test

dimos/core/test_core.py
```diff
@@ -89,7 +89,7 @@ def test_classmethods() -> None:
     )
     assert hasattr(class_rpcs["start"], "__rpc__"), "start should have __rpc__ attribute"
 
-    nav._close_module()
+    nav._stop()
```
jeff-hykin: I'm trying to consolidate our naming to be "stop" instead of half "stop" half "close"

dimos/utils/thread_utils.py
```diff
@@ -0,0 +1,550 @@
...
+    from dimos.core.module import ModuleBase
+
+T = TypeVar("T")
+R = TypeVar("R")
+
+
+# ThreadSafeVal: a lock-protected value with context-manager support
+
+
+class ThreadSafeVal(Generic[T]):
```
jeff-hykin, line 44: this is my favorite util. I hate having _thing and _thing_lock and _thing2 and _thing2_lock, but I also hate seeing _thing being used in a method and thinking "hmm ... does _thing have a lock thats not being used?". This prevents ambiguity about what vals need locks and what vals don't

dimos/utils/thread_utils.py
```diff
@@ -0,0 +1,550 @@
...
+        """True after ``stop()`` has been called."""
+        return self._stop_event.is_set()
+
+    def start(self) -> None:
+        """Start the underlying thread."""
+        self._stop_event.clear()
+        self._thread.start()
+
+    def stop(self) -> None:
+        """Signal the thread to stop and join it.
```
jeff-hykin, line 148: this is probably the part that needs the most review

dimos/utils/thread_utils.py
```diff
@@ -0,0 +1,551 @@
...
+            stderr=subprocess.PIPE,
+            **self._popen_kwargs,
+        )
+        logger.info("Process started", pid=self._process.pid)
+
+        self._watchdog = ModuleThread(
+            module=self._module,
+            target=self._watch,
+            name=f"proc-{self._process.pid}-watchdog",
+        )
```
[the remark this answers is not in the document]
jeff-hykin: super().start() will throw if its called more than once. We can/should assume start isn't being called multiple times AFAIK. 

setstate getstate are different though, start could be called after setstate I believe

dimos/utils/thread_utils.py
```diff
@@ -0,0 +1,550 @@
...
+                    logger.warning("malformed JSON from process", raw=line)
+            proc = self._process
+            log_fn(line, pid=proc.pid if proc else None)
+        stream.close()
+
+
+# safe_thread_map: parallel map that collects all results before raising
+
+
+def safe_thread_map(
```
jeff-hykin, line 457: Not used in this PR, but is used by the docker branch so getting it in here a bit early cause this is the util file it belongs in

dimos/utils/typing_utils.py
```diff
@@ -0,0 +1,45 @@
...
+import sys
+
+if sys.version_info < (3, 13):
+    from typing_extensions import TypeVar
+else:
+    from typing import TypeVar
+
+if sys.version_info < (3, 11):
+
+    class ExceptionGroup(Exception):  # type: ignore[no-redef]  # noqa: N818
```
jeff-hykin, line 29: I didn't want to repeat all this cludge so I put it here. Let me know if there's a better spot

dimos/utils/thread_utils.py
```diff
@@ -0,0 +1,550 @@
...
+        with self._stop_lock:
+            if self._stopped:
+                return
+            self._stopped = True
+
+        self._stop_event.set()
+        if self._thread.is_alive() and self._thread is not threading.current_thread():
+            self._thread.join(timeout=self._close_timeout)
+
+    def join(self, timeout: float | None = None) -> None:
```
paul-nechifor, line 161: I don't think you need join since you're already join()-ing in stop.

dimos/utils/thread_utils.py
```diff
@@ -0,0 +1,550 @@
...
+        start: bool = True,
+        close_timeout: float = 2.0,
+        **thread_kwargs: Any,
+    ) -> None:
+        thread_kwargs.setdefault("daemon", True)
+        self._thread = threading.Thread(**thread_kwargs)
+        self._stop_event = threading.Event()
+        self._close_timeout = close_timeout
+        self._stopped = False
+        self._stop_lock = threading.Lock()
```
paul-nechifor: Why do you need _stopped and _stop_lock? You have _stop_event.

dimos/utils/thread_utils.py
```diff
@@ -0,0 +1,550 @@
...
+            self.start()
+
+    @property
+    def stopping(self) -> bool:
+        """True after ``stop()`` has been called."""
+        return self._stop_event.is_set()
+
+    def start(self) -> None:
+        """Start the underlying thread."""
+        self._stop_event.clear()
```
paul-nechifor: You don't need this. It's already off. If you want ModuleThread to be restartable, then you need to use another thread since threads aren't restartable.

dimos/utils/thread_utils.py
```diff
@@ -0,0 +1,550 @@
...
+    ) -> None:
+        thread_kwargs.setdefault("daemon", True)
+        self._thread = threading.Thread(**thread_kwargs)
+        self._stop_event = threading.Event()
+        self._close_timeout = close_timeout
+        self._stopped = False
+        self._stop_lock = threading.Lock()
+        module._disposables.add(Disposable(self.stop))
+        if start:
+            self.start()
```
paul-nechifor: Noooooo, don't autostart in the constructor. 😭

dimos/utils/thread_utils.py
```diff
@@ -0,0 +1,550 @@
...
+
+    Example::
+
+        class MyModule(Module):
+            @rpc
+            def start(self) -> None:
+                self._worker = ModuleThread(
+                    module=self,
+                    target=self._run_loop,
+                    name="my-worker",
```
paul-nechifor, line 118: It would be nice if ModuleThread used `self.module.__class__.__name__` as the prefix so we can just leave `name` blank most of the time and it still produces a useful name for debugging.

dimos/utils/thread_utils.py
```diff
@@ -0,0 +1,550 @@
...
+
+    def __setstate__(self, state: dict[str, Any]) -> None:
+        self._lock = threading.RLock()
+        self._value = state["_value"]
+
+    def __repr__(self) -> str:
+        return f"ThreadSafeVal({self._value!r})"
+
+
+# ModuleThread: a thread that auto-registers with a module's disposables
```
paul-nechifor: Why add this if there's a docstring below?

dimos/core/module.py
```diff
@@ -128,58 +119,45 @@ def frame_id(self) -> str:
...
 
     @rpc
     def stop(self) -> None:
-        self._close_module()
+        self._stop()
 
-    def _close_module(self) -> None:
-        with self._module_closed_lock:
-            if self._module_closed:
+    def _stop(self) -> None:
```
paul-nechifor: `_close_module` is a remnant from the the Module class hierarchy was more complicated. Some classes were skipping `Module.__init__` and didn't initialize `self._disposables` for example. That's why I'm using `hasattr(self, "_disposables")` or `hasattr(self, "_tf")`. We didn't even have `stop` then.

I think it's not needed at all anymore. This could be deleted if you want and moved into `def stop`.

jeff-hykin: happily! I though it was a rpc vs non-rpc thing

dimos/utils/thread_utils.py
jeff-hykin: cause AI loves redundancy
(I'll remove it, thanks for bringing attention)

dimos/utils/thread_utils.py
jeff-hykin, line 118:
```suggestion
                    name=self.module.__class__.__name__+"_my_worker",
```

dimos/utils/thread_utils.py
jeff-hykin: 😈 no boilerplate

But fr, how do you feel about ModuleThread().start()

dimos/agents/mcp/test_mcp_server.py
```diff
@@ -111,3 +115,56 @@ def test_mcp_module_initialize_and_unknown() -> None:
 
     response = asyncio.run(handle_request({"method": "unknown/method", "id": 2}, [], {}))
     assert response["error"]["code"] == -32601
+
+
+def _free_port() -> int:
```
paul-nechifor, line 120: This already exists as `_find_free_port` in the codebase. I've actually converted it to a fixture (find_free_port) in [my PR](https://github.com/dimensionalOS/dimos/pull/1665/changes#diff-30c2012ee6266bc37c71b0f71036ef9d29a4b787f0e691fb23a03b6473e64f17) but unlikely to be merged soon.

dimos/agents/mcp/test_mcp_server.py
```diff
@@ -111,3 +115,56 @@ def test_mcp_module_initialize_and_unknown() -> None:
...
+        try:
+            resp = requests.post(
+                url,
+                json={"jsonrpc": "2.0", "method": "initialize", "id": 1},
+                timeout=0.5,
+            )
+            if resp.status_code == 200:
+                break
+        except requests.ConnectionError:
+            time.sleep(0.1)
```
paul-nechifor: Stash already has 3 instances of this exact sequance (as functions called wait_for_mcp). Please don't add a 4th that is inlined.

dimos/agents/mcp/test_mcp_server.py
```diff
@@ -111,3 +115,56 @@ def test_mcp_module_initialize_and_unknown() -> None:
...
+    else:
+        server.stop()
+        raise AssertionError("McpServer did not become ready")
+
+    # Verify it responds
+    data = resp.json()
+    assert data["result"]["serverInfo"]["name"] == "dimensional"
+
+    # Stop and verify it shuts down
+    server.stop()
```
paul-nechifor: The server should be stopped even if the test fails. The best way to do it is to use a fixture for the mcp server.

dimos/core/module.py
```diff
@@ -128,58 +119,43 @@ def frame_id(self) -> str:
...
-
-        self._close_rpc()
+            self.mod_state.set("stopped")
 
-        # Save into local variables to avoid race when stopping concurrently
-        # (from RPC and worker shutdown)
-        loop_thread = getattr(self, "_loop_thread", None)
-        loop = getattr(self, "_loop", None)
+        # dispose of things BEFORE making aspects like rpc and _tf invalid
+        if hasattr(self, "_disposables"):
```
paul-nechifor, line 136: You'll have lots of conflicts here with what Ivan is doing for disposables in his PR: https://github.com/dimensionalOS/dimos/pull/1682 .

dimos/core/module.py
```diff
@@ -128,58 +119,43 @@ def frame_id(self) -> str:
...
-        if loop_thread:
-            if loop_thread.is_alive():
-                if loop:
-                    loop.call_soon_threadsafe(loop.stop)
-                loop_thread.join(timeout=2)
-            self._loop = None
-            self._loop_thread = None
+        if self.rpc:
+            self.rpc.stop()  # type: ignore[attr-defined]
+            self.rpc = None  # type: ignore[assignment]
```
paul-nechifor, line 141: Please avoid `type: ignore`

dimos/core/module.py
```diff
@@ -98,20 +87,22 @@ class ModuleBase(Configurable[ModuleConfigT], Resource):
...
     _rpc: RPCSpec | None = None
     _tf: TFSpec[Any] | None = None
-    _loop: asyncio.AbstractEventLoop | None = None
-    _loop_thread: threading.Thread | None
+    _async_thread: AsyncModuleThread
     _disposables: CompositeDisposable
     _bound_rpc_calls: dict[str, RpcCall] = {}
-    _module_closed: bool = False
-    _module_closed_lock: threading.Lock
+    mod_state: ThreadSafeVal[ModState]
```
paul-nechifor, line 93: I don't like `mod` for `module`. It's not much of an abbreviation (it makes more sense for modification).

dimos/core/module.py
paul-nechifor: That's one of the reasons I don't like inheritance. But can you explain what you mean by ModuleBase being a decorator? At first glace that seems more complicated.

dimos/utils/thread_utils.py
```diff
@@ -0,0 +1,559 @@
+# Copyright 2025-2026 Dimensional Inc.
```
paul-nechifor, line 1: I think these are separate enough that they could be their own files. It's often said that "utils" is a dumping ground for any odd code, but this could be put into a `dimos/core/threading/` directory.

dimos/utils/thread_utils.py
```diff
@@ -0,0 +1,559 @@
...
+        """
+        with self._stopped as stopped:
+            if stopped:
+                return
+            self._stopped.set(True)
+
+        if self._owns_loop and self._loop is not None and self._loop.is_running():
+            self._loop.call_soon_threadsafe(self._loop.stop)
+
+        if self._thread is not None and self._thread.is_alive():
```
paul-nechifor, line 255: It's good to also check that this call isn't from the current thread.

dimos/utils/thread_utils.py
```diff
@@ -0,0 +1,559 @@
...
+    """
+
+    def __init__(
+        self,
+        module: ModuleBase[Any],
+        *,
+        close_timeout: float = 2.0,
+    ) -> None:
+        self._close_timeout = close_timeout
+        self._stopped = ThreadSafeVal(False)
```
paul-nechifor, line 203: This is what Event is for.

dimos/utils/thread_utils.py
```diff
@@ -0,0 +1,559 @@
...
+        self._owns_loop = False
+        self._thread: threading.Thread | None = None
+        self._loop: asyncio.AbstractEventLoop | None = None
+        self._module_name = type(module).__name__
+
+        module._disposables.add(Disposable(self.stop))
```
paul-nechifor, line 209: ModuleBase has an AsyncModuleThread. It's rather odd for AsyncModuleThread to tell its owner what to do.

AsyncModuleThread shoud not take ModuleBase in its `__init__`. ModuleBase should register the shutdown of AsyncModuleThread

dimos/utils/thread_utils.py
```diff
@@ -0,0 +1,559 @@
...
+                target=self._loop.run_forever,
+                daemon=True,
+                name=f"{self._module_name}-event-loop",
+            )
+            self._thread.start()
+
+    @property
+    def loop(self) -> asyncio.AbstractEventLoop:
+        """The managed event loop."""
+        return self._loop
```
paul-nechifor, line 235: I'm surprised mypy doesn't complain that you're returning None when only asyncio.AbstractEventLoop is allowed. Greptile is right about the comment above.

dimos/utils/thread_utils.py
```diff
@@ -0,0 +1,559 @@
...
+        self._on_exit = on_exit
+        self._shutdown_timeout = shutdown_timeout
+        self._kill_timeout = kill_timeout
+        self._log_json = log_json
+        self._log_tail_lines = log_tail_lines
+        self._popen_kwargs = popen_kwargs
+        self._process: subprocess.Popen[bytes] | None = None
+        self._watchdog: ModuleThread | None = None
+        self._module = module
+        self._stopped = ThreadSafeVal(False)
```
paul-nechifor, line 316: Use Event.

dimos/utils/thread_utils.py
```diff
@@ -0,0 +1,559 @@
...
+        self.last_stdout: collections.deque[str] = collections.deque(maxlen=log_tail_lines)
+        self.last_stderr: collections.deque[str] = collections.deque(maxlen=log_tail_lines)
+
+        module._disposables.add(Disposable(self.stop))
```
paul-nechifor, line 320: Module should add ModuleProcess.stop to its disposables. ModuleProcess should not help itself to Module's private data like `_disposables`.

dimos/utils/thread_utils.py
```diff
@@ -0,0 +1,559 @@
...
+        else:
+            successes.append(v)
+
+    if errors:
+        if on_errors is not None:
+            zipped = [(items[i], outcomes[i]) for i in range(len(items))]
+            return on_errors(zipped, successes, errors)  # type: ignore[return-value, no-any-return]
+        raise ExceptionGroup("safe_thread_map failed", errors)
+
+    return [outcomes[i] for i in range(len(items))]  # type: ignore[misc]
```
paul-nechifor, line 523: Please don't ignore.

dimos/core/module.py
jeff-hykin: great! I'm actually a general a fan of full/long names for vars, I've just gotten used to other devs wanting things short.

PR #1682, opened 2026-03-26
memory2 StreamModules, cleanup

leshy:


dimos/core/resource.py
```diff
@@ -75,18 +77,17 @@ def __exit__(
...
-        self._disposables = CompositeDisposable()
+    _disposables: CompositeDisposable | None = None
 
-    def register_disposables(self, *disposables: DisposableBase) -> None:
-        """Register child disposables to be disposed when this resource stops."""
-        for d in disposables:
-            self._disposables.add(d)
+    def register_disposable(self, disposable: D) -> D:
+        """Register a child disposable to be disposed when this resource stops."""
+        if self._disposables is None:
```
paul-nechifor, line 84: Previous version was better. `self._disposables` should be set in `__init__`.

dimos/core/resource.py
```diff
@@ -75,18 +77,17 @@ def __exit__(
...
+            self._disposables = CompositeDisposable()
+        self._disposables.add(disposable)
+        return disposable
```
paul-nechifor, line 87: Why does this return the disposable?

dimos/mapping/voxels.py
```diff
@@ -12,133 +12,93 @@
...
+        self,
+        voxel_size: float = 0.05,
+        block_count: int = 2_000_000,
+        device: str = "CUDA:0",
+        carve_columns: bool = True,
+        frame_id: str = "world",
+    ) -> None:
+        self.voxel_size = voxel_size
+        self.carve_columns = carve_columns
+        self.frame_id = frame_id
```
paul-nechifor: These should probably start with an underscore.

dimos/mapping/voxels.py
```diff
@@ -176,36 +133,125 @@ def _carve_and_insert(self, new_keys: o3c.Tensor) -> None:
...
+
+    def dispose(self) -> None:
+        """Free GPU resources. The object is unusable after this call."""
+        if self._disposed:
+            return
+        self._disposed = True
+        self.get_global_pointcloud.invalidate_cache(self)  # type: ignore[attr-defined]
+        self.get_global_pointcloud2.invalidate_cache(self)  # type: ignore[attr-defined]
+        self.vbg = None  # type: ignore[assignment]
+        self._voxel_hashmap = None  # type: ignore[assignment]
```
paul-nechifor: Why so many ignores?

dimos/memory2/observationstore/memory.py
```diff
@@ -30,18 +31,28 @@
...
     default_config = ListObservationStoreConfig
     config: ListObservationStoreConfig
 
     def __init__(self, **kwargs: Any) -> None:
         super().__init__(**kwargs)
         self._name = self.config.name
-        self._observations: list[Observation[T]] = []
+        max_size = self.config.max_size
+        self._observations: deque[Observation[T]] = deque(
+            maxlen=max_size if max_size is not None else None
```
paul-nechifor: If `maxlen` supports None why add the `if-else`?

dimos/memory2/store/test_null.py
```diff
@@ -0,0 +1,61 @@
...
+from dimos.memory2.observationstore.memory import ListObservationStore
+from dimos.memory2.store.null import NullStore
+from dimos.memory2.type.filter import StreamQuery
+from dimos.memory2.type.observation import Observation
+
+
+def test_max_size_zero_monotonic_ids() -> None:
+    """ListObservationStore(max_size=0) assigns monotonically increasing IDs."""
+    store = ListObservationStore(name="test", max_size=0)
+    store.start()
```
paul-nechifor: You're not calling stop.

dimos/memory2/store/test_null.py
```diff
@@ -0,0 +1,61 @@
...
+    id2 = store.insert(Observation(id=-1, ts=3.0, _data="!"))
+
+    assert id0 == 0
+    assert id1 == 1
+    assert id2 == 2
+
+
+def test_max_size_zero_empty_query() -> None:
+    """ListObservationStore(max_size=0) query always returns empty."""
+    store = ListObservationStore(name="test", max_size=0)
```
paul-nechifor: Shouldn't tests in this file be for NullStore, not ListObservationStore?

dimos/memory2/test_voxel_map.py
```diff
@@ -0,0 +1,137 @@
...
+from dimos.mapping.voxels import VoxelMapTransformer
+from dimos.memory2.store.sqlite import SqliteStore
+from dimos.memory2.type.observation import Observation
+from dimos.msgs.sensor_msgs.PointCloud2 import PointCloud2
+from dimos.utils.data import get_data_dir
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+DB_PATH = get_data_dir() / "go2_bigoffice.db"
```
paul-nechifor: ~~Isn't this writing to the data dir? I think you should use the `tmp_path` pytest fixture.~~

---

Edit: Oh, you're only reading the file. In that case you need to use `get_data`, no?

dimos/memory2/test_visualizer.py
```diff
@@ -135,13 +136,71 @@ def test_search_reconstruct_full_path(self, store: SqliteStore) -> None:
...
+    def test_build_global_map(self, store: SqliteStore) -> None:
+        import pickle
+
+        global_map = pickle.loads(get_data("unitree_go2_bigoffice_map.pickle").read_bytes())
+        print(f"Global map: {len(global_map)}")
+
+    # we semantically search, then detect with a detection model
+    #
+    # VIS GOAL: draw 2d detections somehow, or project into 3d, draw 3d bounding boxes
+    def test_detect_objects_smart(self, store: SqliteStore, clip: CLIPModel) -> None:
```
paul-nechifor, line 163: There are a few things wrong with this test:

* it uses inline imports
* it doesn't assign anything
* it doesn't close vlm
* it prints (I assume those should be the assertions)

dimos/memory2/test_store.py
```diff
@@ -525,3 +526,51 @@ def test_accessor_dynamic(self, session: Store) -> None:
...
+        """Backend.stop() propagates to metadata_store, blob_store, vector_store."""
+        s = session.stream("z", int)
+        backend = s._source
+        assert isinstance(backend, Backend)
+        metadata_store = backend.metadata_store
+
+        session.stop()
+
+        # metadata_store should be stopped (CompositeResource._disposables disposed)
+        if hasattr(metadata_store, "_disposables") and metadata_store._disposables is not None:
```
paul-nechifor: In general, tests should not use conditional logic. Tests should be deterministic. If a test has to check if metadata_store has _disposables it means the test doesn't know what it's testing.

dimos/memory2/test_store.py
```diff
@@ -525,3 +526,51 @@ def test_accessor_dynamic(self, session: Store) -> None:
...
+
+        backend1 = s1._source
+        backend2 = s2._source
+        assert isinstance(backend1, Backend)
+        assert isinstance(backend2, Backend)
+
+        session.stop()
+
+        # Both backends' disposables are disposed
+        assert backend1._disposables is None or backend1._disposables.is_disposed
```
paul-nechifor: Shouldn't this be `assert backend1._disposables is not None` **and** `backend1._disposables.is_disposed`

dimos/memory2/test_module.py
```diff
@@ -0,0 +1,260 @@
...
+    class Double(Transformer[int, int]):
+        def __call__(self, upstream: Iterator[Observation[int]]) -> Iterator[Observation[int]]:
+            for obs in upstream:
+                yield obs.derive(data=obs.data * 2)
+
+    pipeline = Stream().transform(Double()).map(lambda obs: obs.derive(data=obs.data + 1))
+
+    # Should have a chain of transforms
+    assert pipeline._xf is not None
+    assert isinstance(pipeline._source, Stream)
```
paul-nechifor: I'm not sure how these assertions verify that "Unbound streams support .transform() and .map() chaining."

You should at least verify that the numbers are indeed doubled. But also, where are the numbers even coming from? Does `Stream()` generate incremental numbers by default?

dimos/memory2/test_module.py
```diff
@@ -0,0 +1,260 @@
...
+
+    class Double(Transformer[int, int]):
+        def __call__(self, upstream: Iterator[Observation[int]]) -> Iterator[Observation[int]]:
+            for obs in upstream:
+                yield obs.derive(data=obs.data * 2)
+
+    pipeline = Stream().transform(Double()).map(lambda obs: obs.derive(data=obs.data + 1))
+
+    # Should have a chain of transforms
+    assert pipeline._xf is not None
```
paul-nechifor: From this perspective, the meaning of `xf` is quite cryptic.

dimos/memory2/test_module.py
```diff
@@ -0,0 +1,260 @@
...
+    stream_names = {s.name for s in atom.streams}
+    assert "numbers" in stream_names
+    assert "result" in stream_names
+
+
+def test_stream_module_runtime_wiring() -> None:
+    """End-to-end: push data into In port, assert transformed data on Out port."""
+    import threading
+
+    from dimos.core.transport import pLCMTransport
```
paul-nechifor: Please move the imports to the top.

dimos/memory2/test_module.py
```diff
@@ -0,0 +1,260 @@
...
+
+    time.sleep(0.5)  # let live stream iterator spin up
+
+    # Push data through the In port's transport
+    module.numbers.transport.publish(42)
+
+    assert done.wait(timeout=5.0), f"Timed out, received={received}"
+
+    unsub()
+    module.stop()
```
paul-nechifor: Tests should clean up after themselves even if they fail.

dimos/memory2/test_module.py
```diff
@@ -0,0 +1,260 @@
...
+
+    # Shutdown the global RxPY thread pool so conftest thread-leak check passes
+    from dimos.utils.threadpool import get_scheduler
+
+    get_scheduler().executor.shutdown(wait=True)
```
paul-nechifor: This is what I mean when I say that having no root object owning things is problematic. Not only is it not clear where the thread pool is started, but it's not clear where this is used and why it needs to be closed. 

It just looks like this test reaches into a random part of codebase to perform some unrelated action.

I assume this is from `Stream.observable`.

There are two solution that I can see:

* Given that multiple Streams share the same thread pool, there needs to be a common objects which manages a set of streams. (I don't particularly like this)
* `Stream.observable` does not belong on `Stream`. There should be a function like `def stream_to_observable(stream, thread_pool):`. You could make it use the global pool if it's not given, but personally I don't like having a global pool. Global mutable state is a source of problems.

The API looks pure in general, so it should not use hidden global state to mask that `Stream.observable` is not in fact pure.

dimos/memory2/test_module.py
paul-nechifor: I see that you use `xf` as a abbreviation for `transformer`. I don't see this being commonly used. From search, I see some people use `xfmr` for electrical transformers, but I don't think that's common either.

dimos/memory2/test_e2e.py
```diff
@@ -126,6 +126,31 @@ def test_import_lidar(
         assert lidar.count() == count
         print(f"Imported {count} lidar frames")
 
+    def test_embed_and_save(self, session: SqliteStore, clip: CLIPModel) -> None:
+        """Embed video frames at 1Hz and persist to an embedded stream."""
+        video = session.stream("color_image", Image)
+
+        # Clear any prior run so the test is idempotent
+        if "color_image_embedded" in session.list_streams():
+            session.delete_stream("color_image_embedded")
```
paul-nechifor, line 135: This is a very incriminating comment. :)

The test which polluted a shared fixture should clean it up. Otherwise, the fixture shouldn't be shared.

dimos/memory2/test_e2e.py
```diff
@@ -126,6 +126,31 @@ def test_import_lidar(
...
+            .save(embedded)
+        )
+
+        count = 0
+        for obs in pipeline:
+            count += 1
+            print(f"  [{count}] ts={obs.ts:.2f} pose={obs.pose}")
+
+        assert count > 0
+        print(f"Embedded {count} frames (1Hz from {video.count()} total)")
```
paul-nechifor, line 152: I think you should delete these. Prints are only useful for debugging.

dimos/memory2/test_voxel_map.py
```diff
@@ -0,0 +1,137 @@
...
+
+@pytest.fixture(scope="module")
+def store() -> Iterator[SqliteStore]:
+    db = SqliteStore(path=str(DB_PATH))
+    with db:
+        yield db
+
+
+@pytest.mark.tool
+def test_build_global_map(self, store: SqliteStore) -> None:
```
paul-nechifor: Was this on a class? Why `self`?

dimos/core/resource.py
leshy: I added a bunch of classes to this (all modules) so I don't want to create overhead of creating a compositedisposable if not used

dimos/core/resource.py
leshy: convinience API, you often create an object, and want to register as disposable

bla = self.register_disposable(new Bla())

dimos/memory2/module.py
```diff
@@ -0,0 +1,109 @@
...
+
+        ((in_name, inp_port),) = self.inputs.items()
+        ((_, out_port),) = self.outputs.items()
+
+        store = self.register_disposable(NullStore())
+        store.start()
+        stream: Stream[Any] = store.stream(in_name, inp_port.type)
+
+        # we push input into the stream
+        inp_port.subscribe(lambda msg: stream.append(msg))
```
paul-nechifor, line 75: You're not keeping track of the subscription.

dimos/core/resource.py
paul-nechifor: Oh, I actually didn't realize this before, but ModuleBase already has `self._disposables`. This is why multiple inheritance is confusing. disposables are created in ModuleBase, but also in CompositeResource.

dimos/memory2/module.py
```diff
@@ -0,0 +1,109 @@
...
+
+        live = stream.live()
```
paul-nechifor, line 77: stream.live() creates a new Stream. It also needs to be closed.

dimos/core/resource.py
jeff-hykin: ~~kinda like the convenience of Thread().start() some might say 🤔~~

dimos/memory2/type/observation.py
```diff
@@ -50,6 +51,16 @@ class Observation(Generic[T]):
     _loader: Callable[[], T] | None = field(default=None, repr=False)
     _data_lock: threading.Lock = field(default_factory=threading.Lock, repr=False)
 
+    @property
+    def pose_stamped(self) -> PoseStamped:
+        """Return the pose as a PoseStamped with this observation's timestamp."""
```
jeff-hykin: this docstring is just fluff ("pose stamped is a method that returns a pose that is stamped")

dimos/core/resource.py
leshy: aha ok, can I remove from modules? on-demand CompositeResource is so convinient for Modules

dimos/memory2/stream.py
```diff
@@ -46,38 +46,46 @@
...
         self._source = source
         self._xf = xf
         self._query = query
 
-    def start(self) -> None:
-        pass
-
     def stop(self) -> None:
-        """Close the live buffer (if any), unblocking iteration."""
+        """Close live buffer and dispose subscriptions."""
```
jeff-hykin: docstring probably not helpful (given the comments below are actually helpful)

dimos/memory2/test_e2e.py
leshy: sorry about all this, tests were pretty vibed in this run, obviously super bad since agents read the tests as examples of good code

dimos/memory2/module.py
leshy: I am very smart and shutdown of module now shuts down transports which unsubscribes these things

https://github.com/dimensionalOS/dimos/pull/1682/changes#diff-2a78d9a0954c76ba1473eb7f8c2a22c05ea722416c4f15869e46435eed1dde54R166

https://github.com/dimensionalOS/dimos/pull/1682/changes#diff-106cdf9cd8240a8e0e920a92eb61398e93a5f2b33f85f8062403c3fbd5938cb6R138

I think this manual resource handling is annoying and fragile, stopping an owner object should propagade down the tree

dimos/memory2/module.py
leshy: same as above, shutdown of store propagades to it's streams

`store = self.register_disposable(NullStore())`

dimos/memory2/test_voxel_map.py
leshy: true wtf am I doing

dimos/memory2/test_visualizer.py
leshy: since this is a @tool inline imports of heavy modules were for on-demand exec, I assumed user will run a single specific test here.

prints are also here for a user, this is not an actual unit tests, but a sketch of data queries, for someone writing a visualizer

dimos/memory2/store/memory.py
```diff
@@ -12,10 +12,36 @@
...
 
-    pass
+    def _create_backend(
+        self, name: str, payload_type: type[Any] | None = None, **config: Any
+    ) -> Backend[Any]:
+        if "observation_store" not in config and self.config.observation_store is None:
+            obs: ListObservationStore[Any] = ListObservationStore(
+                name=name, max_size=self.config.max_size
+            )
+            obs.start()
```
paul-nechifor: I don't think you need start here. This is called as:

```
            backend = self.register_disposable(self._create_backend(name, payload_type, **resolved))
            backend.start()
```

`backend.start()` calls start on the `obs` again. I assume start is idempotent, but you never know how people will write their own stores.

dimos/memory2/stream.py
```diff
@@ -328,13 +345,82 @@ def subscribe(
...
+        Iteration runs on the dimos thread pool (via :meth:`subscribe`).
+        Returns a ``DisposableBase`` suitable for ``_disposables.add()``.
+
+        Example::
+
+            lidar.live().transform(VoxelMapTransformer()).publish(self.global_map)
+        """
+        import logging
+
+        log = logging.getLogger(__name__)
```
paul-nechifor: Shouldn't this use `logger = setup_logger()`?

dimos/memory2/stream.py
```diff
@@ -328,13 +342,82 @@ def subscribe(
...
+                result = result.transform(xf)
+            for f in query.filters:
+                result = result._with_filter(f)
+            if query.limit_val is not None:
+                result = result.limit(query.limit_val)
+            if query.offset_val is not None and query.offset_val != 0:
+                result = result.offset(query.offset_val)
+            if query.order_field is not None:
+                result = result.order_by(query.order_field, desc=query.order_desc)
+        return result  # type: ignore[return-value]
```
paul-nechifor: I see a lot of `# type: ignore` in new code which should probably be avoided.

dimos/memory2/todo.md
```diff
@@ -0,0 +1,57 @@
+# PR #1682 Review Issues
```
paul-nechifor: Did you mean to commit this. You can use `todo.ignore.md` to add any file that's ignored in Git.

dimos/memory2/test_module.py
leshy: fixed

dimos/memory2/test_module.py
leshy: true, renamed to _transformer

dimos/memory2/test_module.py
leshy: I agree, this global pool is bad and breaks containment and chain of custody.
I haven't written this scheduler and we shouldn't have it. 

This is how it is in dimos right now, (all rxpy streams share this global scheduler) this needs rewriting.

I can fix for streams (can I do it later? not pressed since this is a dimos wide issue)

dimos/memory2/test_e2e.py
leshy: these are @tool tests, more like tooling for users, I always run them vebose and check prints

paul-nechifor (dismissed):

paul-nechifor (dismissed):

paul-nechifor (dismissed):

spomichter (approved):

PR #1689, opened 2026-03-26
fix(blueprints): get module by name

paul-nechifor:
## Problem

In https://github.com/dimensionalOS/dimos/pull/1606 I broke the ability to get modules by name.

Closes DIM-XXX

## Solution

* Also specify the class name in `all_modules`.


jeff-hykin:
does this change mean we can do something like
```
dimos run CameraModule
```

instead of `dimos run camera-module` (which needed camera_module to exist)

paul-nechifor:
> does this change mean we can do something like
> 
> ```
> dimos run CameraModule
> ```
> 
> instead of `dimos run camera-module` (which needed camera_module to exist)

No. It just means `camera-module` is a blueprint for CameraModule.blueprint() like before.

dimos/robot/test_all_blueprints_generation.py
```diff
@@ -250,9 +250,9 @@ def _path_to_module_name(path: Path, root: Path) -> str:
 
 def _find_blueprints_in_file(
     file_path: Path, module_classes: set[str] | None = None
-) -> tuple[list[str], list[str]]:
+) -> tuple[list[str], list[tuple[str, str]]]:
```
mustafab0, line 253: We could use NamedTuple for code readability.

mustafab0 (approved):
Looks alright to me. Can't find any issues

dimos/robot/test_all_blueprints_generation.py
paul-nechifor: True, but this is a test and there's no typechecking done on tests. (I don't think they're all that useful on tests.)

PR #1694, opened 2026-03-27
feat: add MujocoCamera module for sim-based perception

ruthwikdasyam:
## Problem

Sim-based manipulation requires a camera for perception
This blocks testing and developing the full perception, planning, execution pipeline in simulation

Closes DIM-733
Closes DIM-404

## Solution

- Added `MujocoCamera` a drop-in replacement for `RealSenseCamera` that renders RGB + depth from a named camera in the MuJoCo sim loop.
- Shared engine registry: `get_or_create_engine()` keyed by resolved MJCF path lets the sim adapter and camera independently resolve the same engine without explicit wiring
- Gripper fix: Corrected inverted linear interpolation in `sim_manip_interface.py`
- PointCloud2 pickle fix: Strip `bounding_box_dimensions` cached property before pickling (was missing, could cause stale cache after unpickling)
- `Sequence[str]` → `list[str]` in `RobotModelConfig.tf_extra_links` for Pydantic v2 compatibility

## Breaking Changes

None

Important to Note: - `SimMujocoAdapter` now uses the engine registry - two adapters pointing to the same MJCF share state (previously each got an independent engine). This is intentional for the camera sharing use case.

## How to Test

```bash
# Unit tests
pytest dimos/simulation/manipulators/test_mujoco_camera.py -v -m mujoco

# Full sim-agent blueprint
dimos run xarm-perception-sim-agent
humancli
# Verify: RGB/depth frames render in rerun, able to command through humancli, observe and manipulate
```

## Note:

Added rerun in sim blueprint for a reason to view rgb and depth camera views
Can remove later, if not a need

Object detection is not great, should be improved later (if any optimzes needed for sim-cam)


dimos/hardware/manipulators/sim/adapter.py
mustafab0: This inherits? from SimManipInterface, which look exactly like adapter.py implementation

why do we need 2 of these files?
adapter.py and sim_manip_interface.py

ruthwikdasyam: True. 2 reasons
SimManipInterface is built upon generic `SimulationEngine` class, to mimic `adapter.py` implmentation.
The actual Mujoco adapter.py that inherits from `SimManipInterface` connects with `MujocoEngine`
Say we add a GenesisEngine, we can simply create another adapter that inherits from SimManipInterface and use it.

In addition, registry.py for adapters need adapters to be at the location `hardware/manipulators/xxxxx/adapter.py`, so had to place the simadapter here

dimos/simulation/engines/mujoco_sim_module.py
mustafab0, line 1: let's put this under simulation/sensors/

mustafab0 (dismissed):

dimos/msgs/sensor_msgs/PointCloud2.py
```diff
@@ -100,9 +100,13 @@ def __getstate__(self) -> dict[str, object]:
...
-        # Remove cached_property entries that hold unpicklable Open3D types
-        state.pop("oriented_bounding_box", None)
-        state.pop("axis_aligned_bounding_box", None)
+        # Remove cached_property entries that hold Open3D C++ objects
+        for key in (
+            "axis_aligned_bounding_box",
+            "oriented_bounding_box",
+            "bounding_box_dimensions",
+        ):
+            state.pop(key, None)
```
paul-nechifor: This will presumably break again in the future once another `cached_property` is added. How about:

```
for key in list(state):
    if isinstance(getattr(type(self), key, None), functools.cached_property):
        del state[key]
```

dimos/simulation/sensors/mujoco_camera.py
```diff
@@ -0,0 +1,411 @@
...
+
+    color_image: Out[Image]
+    depth_image: Out[Image]
+    pointcloud: Out[PointCloud2]
+    camera_info: Out[CameraInfo]
+    depth_camera_info: Out[CameraInfo]
+
+    default_config = MujocoCameraConfig
+
+    def __init__(self, *args, **kwargs) -> None:  # type: ignore[no-untyped-def]
```
paul-nechifor: `**kwargs: Any` to remove `type: ignore`

dimos/simulation/sensors/test_mujoco_camera.py
```diff
@@ -0,0 +1,285 @@
...
+        cam = MujocoCamera(camera_name="wrist_camera")
+        assert cam._camera_link == "wrist_camera_link"
+        assert cam._color_frame == "wrist_camera_color_frame"
+        assert cam._color_optical_frame == "wrist_camera_color_optical_frame"
+        assert cam._depth_frame == "wrist_camera_depth_frame"
+        assert cam._depth_optical_frame == "wrist_camera_depth_optical_frame"
+        cam.stop()
+
+
+@pytest.mark.mujoco
```
paul-nechifor: Why is this marked with `mujoco`? I don't see it starting mujoco.

dimos/simulation/sensors/test_mujoco_camera.py
```diff
@@ -0,0 +1,285 @@
...
+from dimos.simulation.engines.mujoco_engine import (
+    CameraConfig,
+    CameraFrame,
+    MujocoEngine,
+    _clear_registry,
+    _engine_registry,
+    get_or_create_engine,
+    unregister_engine,
+)
+from dimos.simulation.manipulators.test_sim_adapter import _patch_mujoco_engine
```
paul-nechifor: You should not import from tests.

If tests need to share code, it should be put into `conftest.py` as a fixture.

dimos/simulation/engines/mujoco_engine.py
```diff
@@ -16,25 +16,112 @@
...
+    depth_renderer: mujoco.Renderer
+    interval: float
+    last_render_time: float = 0.0
+
+
+_engine_registry: dict[str, MujocoEngine] = {}
+_engine_registry_lock = threading.Lock()
+
+
+def get_or_create_engine(
```
paul-nechifor: Mujoco is quite heavy. I don't understand the need to store MujocoEngine in a global variable.

Where is this intended to be used? Can't all places get it from the same initial place that creates it?

There is a danger with relying on global variables. We use a lot of processes. If the places which use `get_or_create_engine` do so from different processes, they will not see each other and and up creating duplicate mujoco instances.

dimos/simulation/sensors/mujoco_camera.py
```diff
@@ -0,0 +1,411 @@
...
+        super().__init__(*args, **kwargs)
+        self._engine: MujocoEngine | None = None
+        self._stop_event = threading.Event()
+        self._thread: threading.Thread | None = None
+        self._camera_info_base: CameraInfo | None = None
+
+    def set_engine(self, engine: MujocoEngine) -> None:
```
paul-nechifor: Why would you want to inject an engine into a Module? (I can only see this being used in tests.)

MujocoCamera is a Module which (probably) runs on its own process so I don't understand from where you could set the engine. If this is just for tests, you don't need `set_engine` you can patch it in tests without it.

dimos/simulation/sensors/mujoco_camera.py
```diff
@@ -0,0 +1,411 @@
...
+
+        self._camera_info_base = CameraInfo(
+            height=h,
+            width=w,
+            distortion_model="plumb_bob",
+            D=D,
+            K=K,
+            P=P,
+            frame_id=self._color_optical_frame,
+        )
```
paul-nechifor: We have this code for constructing fake camera infos in several places in the codebase (e.g. robot/drone/camera_module.py and robot/unitree/modular/detect.py and hardware/sensors/camera/zed/camera.py

Perhaps you can add a method on CameraInfo to construct a CameraInfo from camera 4-numbers intrinsics.

dimos/simulation/sensors/mujoco_camera.py
```diff
@@ -0,0 +1,411 @@
...
+
+    @rpc
+    def start(self) -> None:
+        if self._engine is None and self.config.address:
+            from pathlib import Path
```
paul-nechifor: Move to top.

dimos/simulation/sensors/mujoco_camera.py
```diff
@@ -0,0 +1,411 @@
...
+    def get_depth_scale(self) -> float:
+        return 1.0
+
+    def _build_camera_info(self) -> None:
+        """Compute camera intrinsics from the MuJoCo model (pinhole, no distortion)."""
+        if self._engine is None:
+            return
+        fovy_deg = self._engine.get_camera_fovy(self.config.camera_name)
+        if fovy_deg is None:
+            logger.error(f"Camera '{self.config.camera_name}' not found in MJCF")
```
paul-nechifor: In structlog it's more common to log as:

```suggestion
            logger.error(f"Camera not found in MJCF", camera_name=self.config.camera_name})
```

(Please do the same for the others.)

dimos/simulation/sensors/test_mujoco_camera.py
```diff
@@ -0,0 +1,285 @@
...
+        patches = _patch_mujoco_engine(ARM_DOF)
+        for p in patches:
+            p.start()
+        try:
+            engine = get_or_create_engine(config_path=Path("/fake/scene.xml"), headless=True)
+            assert engine is not None
+            assert len(_engine_registry) == 1
+        finally:
+            for p in patches:
+                p.stop()
```
paul-nechifor: There is a lot of duplicate cleanup code between tests. This indicates you should be using a fixture.

dimos/simulation/sensors/test_mujoco_camera.py
paul-nechifor: Yeah, it looks like none of these tests should be mujoco. `mark.mujoco` is for those few VERY slow tests which actually bring up mojoco to test things.

dimos/simulation/sensors/mujoco_camera.py
ruthwikdasyam: `set_engine()` was actually written for sharing the engine between adapter and camera, but with the registry `get_or_create_engine` handling that now, this is only used in tests. will remove it

dimos/msgs/sensor_msgs/PointCloud2.py
ruthwikdasyam: yes, this is great. changing...

dimos/simulation/sensors/test_mujoco_camera.py
ruthwikdasyam: Got it, yes not starting mujoco, not a slow test

dimos/simulation/sensors/test_mujoco_camera.py
ruthwikdasyam: Thankyou, yes, added fixture in conftest. removed dublicate code from both `test_sim_adapter` and `test_mujoco_camera` tests

dimos/simulation/engines/mujoco_engine.py
ruthwikdasyam: sim adapter and sim camera both need to share the MujocoEngine instance, both are called separately in the blueprint. As they mimic real arm adapter and realsense camera module, they lie in different places, and same instance cannot be internally shared
One way to do it is to create the instance in the blueprint and pass it to the modules, which I don't think is a good idea

ruthwikdasyam:
### sim v2 update:

The original design used a global engine registry (get_or_create_engine) where multiple modules shared the same MujocoEngine instance in-process. 
Fails - This broke under WorkerManager when the adapter and camera landed in different worker processes

### New design: 
One MujocoSimModule owns the engine exclusively and exposes joint state/commands via shared memory (SHM). The adapter `ShmMujocoAdapter` reads/writes SHM buffers instead of touching the engine directly. Camera streams are now published as Out ports from the same module, replacing the separate MujocoCamera.

Both sides discover each other via a deterministic key derived from the MJCF path - no registry, no RPC.

dimos/simulation/engines/mujoco_shm.py
```diff
@@ -0,0 +1,347 @@
...
+
+    def _array(self, buf: SharedMemory, n: int, dtype: Any) -> NDArray[Any]:
+        return np.ndarray((n,), dtype=dtype, buffer=buf.buf)
+
+    def _control(self) -> NDArray[np.int32]:
+        return np.ndarray((4,), dtype=np.int32, buffer=self.shm.control.buf)
+
+    def _increment_seq(self, index: int) -> None:
+        seq_arr = np.ndarray((8,), dtype=np.int64, buffer=self.shm.seq.buf)
+        seq_arr[index] += 1
```
mustafab0, line 261: Look suspiciously like magic numbers. Should be constants

mustafab0, line 261: This is a non atomic operation on a shared memory buffer. 

Should have a lock

dimos/simulation/engines/mujoco_shm.py
```diff
@@ -0,0 +1,347 @@
...
+                logger.warning("SHM close failed", name=shm.name, error=str(exc))
+
+    def _control(self) -> NDArray[np.int32]:
+        return np.ndarray((4,), dtype=np.int32, buffer=self.shm.control.buf)
+
+    def _set_command_mode(self, mode: int) -> None:
+        self._control()[CTRL_COMMAND_MODE] = mode
+
+    def _increment_seq(self, index: int) -> None:
+        seq_arr = np.ndarray((8,), dtype=np.int64, buffer=self.shm.seq.buf)
```
mustafab0: magic numbers here too

dimos/hardware/manipulators/sim/adapter.py
```diff
@@ -12,59 +12,232 @@
...
+                    )
+                    return False
+                time.sleep(_ATTACH_RETRY_POLL_S)
+
+        # Wait for sim module to signal ready.
+        deadline = time.monotonic() + _READY_WAIT_TIMEOUT_S
+        while not self._shm.is_ready():
+            if time.monotonic() > deadline:
+                logger.error("sim module not ready", timeout_s=_READY_WAIT_TIMEOUT_S)
+                return False
```
mustafab0, line 101: should clean up the `self._shm` reader after connect() timesout

dimos/simulation/engines/mujoco_shm.py
ruthwikdasyam: I think its okay. each sequence counter has a single writer. no two processes ever increment same index.
lock seems to be overhead

dimos/simulation/engines/mujoco_shm.py
ruthwikdasyam: fixed, thanks!

mustafab0 (approved):

PR #1696, opened 2026-03-28
chore(rpc): remove old rpc linking

paul-nechifor:
## Problem

Still had the two old RPC linking systems.

Closes DIM-527

## Solution

* Remove them.

## Breaking Changes

Old RPCs are gone.


dimos/core/test_core.py
```diff
@@ -77,7 +77,7 @@ def test_classmethods() -> None:
     # Check that we have the expected RPC methods
     assert "navigate_to" in class_rpcs, "navigate_to should be in rpcs"
     assert "start" in class_rpcs, "start should be in rpcs"
-    assert len(class_rpcs) == 9
+    assert len(class_rpcs) == 7
```
mustafab0, line 80: wouldn't number of class_rpcs change?

mustafab0 (approved):
Seems all good

dimos/core/test_core.py
paul-nechifor: The change is because I've removed get_rpc_method_names and set_rpc_method.

PR #1699, opened 2026-03-28
Feature: Unified RobotConfig for robot description import

mustafab0:
## Problem
Robot configuration is scattered across three manually-wired config objects (RobotModelConfig, HardwareComponent, TaskConfig). Every blueprint duplicates joint names 3x, task names 2x, and DOF counts 2x. Adding a new robot requires touching 8+ files and manually keeping values in sync across config layers.

## Solution
Introduce RobotConfig — a single source of truth where the URDF/MJCF model file is the ground truth. Joint names, DOF, limits, and base link are parsed from the model file automatically. The user provides the model path + a few semantic choices (end-effector link, hardware adapter), and everything else is derived.

## Breaking Changes
None — this is additive. Existing configs, helpers, and blueprints continue to work unchanged.

#### 1. Launch the blueprint
dimos run xarm7-planner-coordinator

#### 2. Open meshcat visualization
Go to http://localhost:7006 in your browser — should see xArm7 at home pose

#### 3. In a separate terminal, run the manipulation client
python -i -m dimos.manipulation.planning.examples.manipulation_client


dimos/robot/model_parser.py
```diff
@@ -0,0 +1,253 @@
...
+            pkg_path = str(Path(package_paths[pkg_name]).resolve())
+            return resolved.replace(f"$({a})", pkg_path)
+        return str(original_find(resolved, a, args, context))
+
+    substitution_args._find = custom_find
+    try:
+        doc = xacro.process_file(str(path), mappings=xacro_args)
+        return str(doc.toprettyxml(indent="  "))
+    finally:
+        substitution_args._find = original_find
```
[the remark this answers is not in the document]
mustafab0: added a threading lock for xacro files processing

dimos/robot/model_parser.py
```diff
@@ -0,0 +1,253 @@
...
+            )
+        )
+
+    # Root link = parent that is never a child
+    root_link = ""
+    for link_name in links:
+        if link_name not in child_links:
+            root_link = link_name
+            break
+
```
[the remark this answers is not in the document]
mustafab0: added root link edge case checks

dimos/robot/catalog/panda.py
```diff
@@ -0,0 +1,67 @@
...
+    address: str | None = None,
+    **overrides: Any,
+) -> RobotConfig:
+    """Create a Franka Emika Panda robot configuration (7 DOF).
+
+    Args:
+        name: Robot identifier (must contain 'panda' for VAMP auto-detection).
+        joint_prefix: Prefix for coordinator joint namespace.
+        adapter_type: Hardware adapter ("mock").
+        address: Connection address.
```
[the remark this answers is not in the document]
mustafab0: removed stale parameter

dimos/robot/catalog/ufactory.py
```diff
@@ -0,0 +1,125 @@
...
+        "home_joints": [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
+        "base_pose": [0, y_offset, z_offset, 0, 0, 0, 1],
+        "package_paths": {"xarm_description": get_data("xarm_description")},
+        "xacro_args": xacro_args,
+        "auto_convert_meshes": True,
+        "collision_exclusion_pairs": XARM_GRIPPER_COLLISION_EXCLUSIONS if add_gripper else [],
+        "tf_extra_links": tf_extra_links or [],
+    }
+    defaults.update(overrides)
+    return RobotConfig(**defaults)
```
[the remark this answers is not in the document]
mustafab0: attach_xyz positions link_base relative to a world link within the URDF via a fixed joint. Then Drake welds link_base (not world) to the world frame using base_pose. So:

attach_xyz shifts link_base relative to world link in the URDF
Drake welds link_base to world frame at base_pose
These are on different links — attach_xyz offsets within the URDF's own world→link_base joint, while base_pose welds link_base to Drake's world. So if the URDF already has a world→link_base weld with the offset baked in (line 270-275 shows it skips the weld if one exists), there's no double-counting. The base_pose weld gets skipped.

dimos/robot/catalog/__init__.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2025-2026 Dimensional Inc.
```
paul-nechifor: Could you please remove this file? I've deleted all the previous `__init__.py` files and replaced them with just importing directly. It avoids having to import unrelated things when you import from an `__init__.py`.

dimos/robot/catalog/piper.py
```diff
@@ -0,0 +1,73 @@
...
+    Args:
+        name: Robot identifier.
+        adapter_type: Hardware adapter ("mock", "piper").
+        address: CAN port (e.g., "can0").
+        y_offset: Y-axis offset for base pose (multi-arm setups).
+        **overrides: Override any RobotConfig field.
+    """
+    defaults: dict[str, Any] = {
+        "name": name,
+        "model_path": get_data("piper_description") / "urdf/piper_description.xacro",
```
paul-nechifor: Is this function meant to be used at import time in some file? If so you'd have to use LfsPath instead of get_data to avoid doing the download at import time.

dimos/robot/model_parser.py
```diff
@@ -0,0 +1,262 @@
...
+    path: Path,
+    package_paths: dict[str, Path],
+    xacro_args: dict[str, str],
+) -> str:
+    """Expand a xacro file to URDF XML string."""
+    try:
+        import xacro  # type: ignore[import-not-found,import-untyped]
+    except ImportError:
+        raise ImportError(
+            "xacro is required for processing .xacro files. Install with: pip install xacro"
```
paul-nechifor: We might want something more relevant to dimos, like "install the manipulation extra (where xacro is)".

dimos/robot/model_parser.py
```diff
@@ -0,0 +1,262 @@
...
+        original_find = substitution_args._find
+
+        def custom_find(resolved: str, a: str, args: list[str], context: dict[str, str]) -> str:
+            pkg_name = args[0] if args else ""
+            if pkg_name in package_paths:
+                pkg_path = str(Path(package_paths[pkg_name]).resolve())
+                return resolved.replace(f"$({a})", pkg_path)
+            return str(original_find(resolved, a, args, context))
+
+        substitution_args._find = custom_find
```
paul-nechifor: This is not very robust. Why do we have to patch private xacro functions. It's bound to break eventually.

dimos/robot/catalog/ufactory.py
```diff
@@ -0,0 +1,125 @@
...
+from __future__ import annotations
+
+from typing import Any
+
+from dimos.robot.config import RobotConfig
+from dimos.utils.data import get_data
+
+# XArm gripper collision exclusions (parallel linkage mechanism)
+# The gripper uses mimic joints where non-adjacent links can overlap legitimately
+XARM_GRIPPER_COLLISION_EXCLUSIONS: list[tuple[str, str]] = [
```
paul-nechifor, line 26: There's one in blueprints.py (seems identical):

```python
XARM_GRIPPER_COLLISION_EXCLUSIONS: list[tuple[str, str]] = [
    # Inner knuckle <-> outer knuckle (parallel linkage)
    ("right_inner_knuckle", "right_outer_knuckle"),
    ("left_inner_knuckle", "left_outer_knuckle"),
    # Inner knuckle <-> finger (parallel linkage)
    ("right_inner_knuckle", "right_finger"),
    ("left_inner_knuckle", "left_finger"),
    # Cross-finger pairs (mimic joint symmetry)
    ("left_finger", "right_finger"),
    ("left_outer_knuckle", "right_outer_knuckle"),
    ("left_inner_knuckle", "right_inner_knuckle"),
    # Outer knuckle <-> opposite finger
    ("left_outer_knuckle", "right_finger"),
    ("right_outer_knuckle", "left_finger"),
    # Gripper base <-> all moving parts (can touch at limits)
    ("xarm_gripper_base_link", "left_inner_knuckle"),
    ("xarm_gripper_base_link", "right_inner_knuckle"),
    ("xarm_gripper_base_link", "left_finger"),
    ("xarm_gripper_base_link", "right_finger"),
    # Arm link6 <-> gripper (attached via fixed joint, can touch)
    ("link6", "xarm_gripper_base_link"),
    ("link6", "left_outer_knuckle"),
    ("link6", "right_outer_knuckle"),
]
```

dimos/robot/config.py
```diff
@@ -0,0 +1,248 @@
...
+        return f"traj_{self.name}"
+
+    # -- Converter methods ----------------------------------------------------
+
+    def to_robot_model_config(self) -> RobotModelConfig:
+        """Generate RobotModelConfig for ManipulationModule."""
+        from dimos.manipulation.planning.spec.config import RobotModelConfig
+        from dimos.msgs.geometry_msgs.PoseStamped import PoseStamped
+        from dimos.msgs.geometry_msgs.Quaternion import Quaternion
+        from dimos.msgs.geometry_msgs.Vector3 import Vector3
```
paul-nechifor: Do these have to be inline?

dimos/robot/catalog/__init__.py
paul-nechifor: I actually have a test which fails the build when it encounters `__init__.py` files:

```
FAILED dimos/test_no_init_files.py::test_no_init_files - AssertionError: Found __init__.py files in dimos/:
  - robot/catalog/__init__.py

__init__.py files are not allowed because they lead to unnecessary extraneous imports. Everything should be imported straight from the source module.
```

dimos/robot/catalog/piper.py
paul-nechifor: Yep, looks like these need to be changed to not download at import time.

mustafab0: changed to LfsPath.

dimos/robot/catalog/__init__.py
mustafab0: Removed

dimos/robot/model_parser.py
mustafab0: Updated the error message

dimos/robot/config.py
```diff
@@ -0,0 +1,248 @@
...
+
+    # Collision exclusion pairs (gripper-specific, cannot be parsed from model)
+    collision_exclusion_pairs: list[tuple[str, str]] = Field(default_factory=list)
+
+    _parsed: ModelDescription = PrivateAttr()
+
+    @model_validator(mode="after")
+    def _parse_model_and_fill_defaults(self) -> RobotConfig:
+        """Parse model file eagerly and fill defaults."""
+        parsed = parse_model(self.model_path, self.package_paths, self.xacro_args)
```
paul-nechifor: RobotConfig cannot parse the model at constructor time because it's created at import time so it downloads the file at import time.

I think you have to change the way it works so that the Module creates the RobotConfig at `def start` or `def build` (added recently by Jeff to work before `def start`).

dimos/robot/catalog/ufactory.py
```diff
@@ -0,0 +1,125 @@
...
+    defaults: dict[str, Any] = {
+        "name": name,
+        "model_path": get_data("xarm_description") / "urdf/xarm_device.urdf.xacro",
+        "end_effector_link": "link_tcp" if add_gripper else "link7",
+        "adapter_type": adapter_type,
+        "address": address,
+        "joint_names": [f"joint{i}" for i in range(1, 8)],
+        "base_link": "link_base",
+        "home_joints": [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
+        "base_pose": [0, y_offset, z_offset, 0, 0, 0, 1],
```
paul-nechifor: I probably don't have enough context, but why are the other values not settable? E.g. should I be able to set `x_offset`? I see you have `pitch`, but it's not used. It is used in attach_rpy above.

dimos/robot/config.py
```diff
@@ -0,0 +1,248 @@
...
+            home_joints=self.home_joints,
+            pre_grasp_offset=self.pre_grasp_offset,
+        )
+
+    def to_hardware_component(self) -> HardwareComponent:
+        """Generate HardwareComponent for ControlCoordinator."""
+        from dimos.control.components import (
+            HardwareComponent,
+            HardwareType,
+        )
```
paul-nechifor: You might want to configure claude to always move imports to the top.

dimos/robot/model_parser.py
```diff
@@ -0,0 +1,262 @@
...
+    def get_joint(self, name: str) -> JointDescription | None:
+        for j in self.joints:
+            if j.name == name:
+                return j
+        return None
+
+
+logger = setup_logger()
+
+_xacro_patch_lock = threading.Lock()
```
paul-nechifor: Nit: it's probably better to have variable definitions before class definitions.

dimos/robot/model_parser.py
mustafab0: By default, xacro resolves this via ROS's ament_index_python / rospack, which requires a full ROS workspace with packages installed

we download robot description packages via LFS to custom paths. So we need to tell xacro "when you see $(find xarm_description), use this path instead."

mustafab0: Now temporarily override the command dispatch table

dimos/robot/catalog/ufactory.py
mustafab0: This will  be addressed in the follow up PR comming soon

dimos/robot/config.py
mustafab0: updated

dimos/robot/config.py
mustafab0: Moved them to top

dimos/robot/config.py
mustafab0: I added a lazy parsing workaround, such that it does not download at import time and is only used when trying to access derived properties.

the other option is to do a major refactor. which I ma happy to do if you think that is the better alternative.

dimos/robot/model_parser.py
mustafab0: Done

dimos/robot/catalog/ufactory.py
mustafab0: Updated to be more consistent.

dimos/robot/config.py
mustafab0: Let me know for this @paul-nechifor

paul-nechifor (dismissed):

dimos/robot/config.py
paul-nechifor: :+1:

paul-nechifor (approved):

PR #1707, opened 2026-03-29
fix(modules): allow disabling modules again

paul-nechifor:
## Problem

My recent commit broke disabling modules.

Closes DIM-XXX


leshy (approved):
thanks I hate it

PR #1708, opened 2026-03-29
Record replay

### LINKED ISSUE #1575: RecordReplay
# intro

for now we record a robot via hardcoding RPC call to record() @

`dimos/dimos/robot/unitree/go2/connection.py:193`

and replay via injecting custom fake connection class in place of actual connection

`dimos/dimos/robot/unitree/go2/connection.py:77`

```python
class ReplayConnection(UnitreeWebRTCConnection):
    dir_name = "unitree_go2_bigoffice"
```

This is pretty dumb. we should be able to record any module or groups of modules, and easily replace their outputs with recorded ones in a blueprint. We want a recorder class.

This is a vague sketch

Recorder produces a Recording that's a memory2 store.

StoreConfig is @ `/dimos/memory2/store/base.py:65`
Store is @ `/dimos/memory2/store/base.py:80`

configures backend basically.

```python

class RecordReplay(ABC):
   topics: list[Topic]

    def __init__(self, store: Store | None, filename: str | None):
       ...
       # should init default store (sqlite) if not set use filename? thinking of ergonomics here..

   # idk maybe we want to make topics attribute on a recorder some magical class with topics as attributes? 
   # idk think about ergonomics
   # store for example has `store.streams.some_stream` magic namespace

   # I'll provide most naive possible methods here but you can design whatever:
  
   def add_topic(topic: Topic):
      ...
      
   def remove_topic(topic: Topic):
      ...

   # inspects module Out topics, records all
   def add_module(module: Module):
      ...
      
   def remove_module(module: Module):
      ...

   def start_recording():
      ...
      
   def stop_recording():
      ...

   # some timeline and playback tooling?
   
   def play(playback_speed: float = None):
      ...

   def seek(seconds: float = None)

   def get_seek() -> float:
       ... # idk I'm vibing

   # some timeline editing tooling like:
   def delete_section(time_range):
       ... 
   # idea is you'd record, then you can replay (viewing in rerun) and choose to trim some stuff, then save
   # idk if actual memory2.store supports convinient point deletion? or range deletions? should we add?
   # should we keep triming as an extra/todo later?
   
   # maybe remove_topic method here also DELETES topic from a store, if recorder made a recording?

   def save():
      ...
   # idk, do we call save() after trimming or is it auto-saved?

```

"a recording" this produces is whatever is required to re-init the RecordReplay, I guess since we use sqlite that's just an sqlite file, 

idk if we want to keep my proposal with `__init__` taking pluggable alternative stores. we can choose to keep it simple.. so `__init__` can also be just take this sqlite filename

`memory2/test_e2e.py` shows how to record sensor data into the Store

Above is basically "video editor ui" but in API

## UI

this is the "video editor UI" but in cli

Image

So ideally you start something that looks like lcmspy, you select modules (TBD!) or topics you want, and press record. you can also press stop, play, rewind etc. cool thing is that dimos pubsub Topic type is something Recorder can emit into

if you run `rerun-bridge` in the side terminal, it listens to LCM so you can watch your recording.

allow triming the recording, deleting, show timeline basically nice simple terminal VLC :) - this can be vibed super easily, I don't code my cli tooling at all.

you can have diff colors or use braille (like `dtop`) for streams, depending on message counts, bytes, idk whatever seems cool, don't bother too much with details, I vibe these things, if RecordReplay API is good this is easy to re-vibe etc

## Usage

blueprint.replay(RecordReplay) or something like this? should kill modules whose OUTs are in the recording itself, but run the rest of the blueprint (talk to Paul if needed - this is the biggest unknown for me but should be easy)

`blueprint.replay(RecordReplay.seek(10).map(every_other_msg).map(rate_limiter))` <- in theory (for fun, idk if useful) we can offer all memory2 query/filter methods on this? just pass down to individual streams

We also want to hook --replay into this right? so blueprint could potentially have a config seting for initing it's "appropriate" recording, so if --replay is used this recording is replayed, 

but --replay should support loading files also right? (cli expects standard sqlite store)

## Misc

we store recordings in lfs system, dimos/data/.lfs - dimos/utils/data.py helps with this - there are docs also, find them in docs/

### Fancy blueprint level default recording control
you could allow blueprint to specify default modules to record? (via blueprint config? bonus points think about higher order blueprints appending not rewriting?) so we can get `dimos --record` (module OUT topics are easily inspectable) 

so like

```python

unitree_go2_basic = (
    autoconnect(
        with_vis,
        go2_connection(),
        websocket_vis(),
    )
    .global_config(n_workers=4, robot_model="unitree_go2")
    .configurators(ClockSyncConfigurator())
    .default_record_modules(go2_connection) # or part of .global_config? idk
)

# so when I feed into a recorder (deployed blueprint? undeployed? idk)

recorder = Recorder("mydb.db") # relative path so places it in dimos/data via `get_data`
recorder.record_blueprint(unitree_go2_basic) # it knows the topics

# later (will disable go2_connection since it will discover that recorder offers those topics)
unitree_go_basic.replay_recoriding(Recorder("mydb.db"))


```
### actually always record

we will have a rolling db for all runs, robots record all of their sensors by default. If you like your run, you can go into cli, run the "video editing / player" cli tool, and `rerun-bridge`, trim, commit to LFS

(memory2 and fancy memory functionality will use this db as well, in realtime for historical queries etc)

### external module level record

we don't have a system to OUTSIDE of dimos know which modules emit which topics actually, nice to add, then in your UI you can see modules and select modules and not just topics (can have a tree of depth 1 right? module->topics)

if you use `dimos --dtop` and type `dtop` in another terminal, this is some system that reports modules and worker states (via LCM broadcast), we could have a similar (or same idk) system that reports a blueprint being ran for example (I think it's seriazable) so you can externally inspect

Dreamsorcerer:
## Problem

We want a recorder tool with the ability to record a selection of streams for modules using the new memory2 module, and the ability to replay those files.

Closes #1575 

## Solution

Added a RecordReplay class to handle the main functionality, a separate recorder UI, and appropriate adjustments to link it all together.

Also made the run() entrypoint async, so it is possible to migrate gradually to asyncio (probably needs some discussion).

## How to Test

Run simulation or similar, while also running `dimos recorder`. A selection of streams should appear in the recorder. Select some streams with space and then use r to start/stop a recording.
Also possible to record an entire session using `--record-path`.

Replay with `--replay --replay-file=...`.


dimos/core/module_coordinator.py
```diff
@@ -162,10 +162,10 @@ def start_all_modules(self) -> None:
     def get_instance(self, module: type[ModuleBase]) -> ModuleProxy:
         return self._deployed_modules.get(module)  # type: ignore[return-value, no-any-return]
 
-    def loop(self) -> None:
-        stop = threading.Event()
+    async def loop(self) -> None:
```
jeff-hykin: Maybe not this specific edit but I think async loop on module coordinator is probably is a good long term change but definitely will need to talk to Paul and Ivan about it

.gitignore
```diff
@@ -77,3 +77,6 @@ CLAUDE.MD
 htmlcov/
 .coverage
 .coverage.*
+
+# Created from simulation
+MUJOCO_LOG.TXT
```
jeff-hykin, line 82: Yeah we've been missing this for a long time

flake.nix
```diff
@@ -42,6 +42,7 @@
           { vals.pkg=pkgs.opensshWithKerberos;flags={}; }
           { vals.pkg=pkgs.unixtools.ifconfig; flags={}; }
           { vals.pkg=pkgs.unixtools.netstat;  flags={}; }
+          { vals.pkg=pkgs.uv;                 flags={}; }
```
jeff-hykin: This will break on many (but not all) Linux systems. It is a task on the backlog to get uv in the nix env but don't push as part of the record/replay change.

Dreamsorcerer: No problem, was debugging with Ivan. Will revert.

dimos/record/record_replay.py
```diff
@@ -0,0 +1,391 @@
...
+    def is_playing(self) -> bool:
+        return self._play_task is not None and not self._play_task.done()
+
+    @property
+    def is_paused(self) -> bool:
+        return self.is_playing and not self._resume.is_set()
+
+    def start_recording(
+        self,
+        pubsubs: Collection[LCMPubSubBase],
```
leshy: needs to be generic, idk what's the best interface to specify topics, but we should be able to take a subscription from any transport at this layer.

at this stage doesn't have to be a glob supporting pubsub, this is for discovery, all topics you want to record are known at this stage, so ros transport, shm should work here.

dimos/record/record_replay.py
```diff
@@ -0,0 +1,391 @@
...
+        self._recording = False
+        for unsub in self._unsubscribes:
+            unsub()
+        self._unsubscribes.clear()
+        logger.info("Recording stopped")
+
+    def _on_message(self, msg: bytes, topic: Topic) -> None:
+        stream_name = topic_to_stream_name(topic.pattern)
+
+        if self._topic_filter is not None and stream_name not in self._topic_filter:
```
leshy: you are doing subscribe_all then filtering in python which is too costly, at this stage python code already decoded LCM message and built a class for you just to throw it away, recorder will be used in actual robot modules with predefined topics, so we need efficient compute here..

this also makes me realize we might want to extend our pubsub base definition to support just watching for topics without subscribing/parsing so actual CLI recorder can be efficient, but can do it in a follow up

dimos/record/record_replay.py
Dreamsorcerer: Yes, was just looking at updating this annotation. Complication is that something like AllPubSub doesn't have .start() method.

leshy:
btw sharing if helpful, current module I use to hardcode implicit recording for specific robots
```python
class Recorder(Module[RecorderConfig]):
    """Records all ``In`` ports to a memory2 SQLite database.

    Subclass with the topics you want to record::

        class MyRecorder(Recorder):
            color_image: In[Image]
            lidar: In[PointCloud2]

        blueprint.add(MyRecorder, db_path="session.db")
    """

    default_config = RecorderConfig

        super().__init__(**kwargs)
        self._store: SqliteStore | None = None

    @rpc
    def start(self) -> None:
        super().start()

        self._store = self.register_disposable(SqliteStore(path=self.config.db_path))
        self._store.start()

        if not self.inputs:
            logger.warning("Recorder has no In ports — nothing to record, subclass the Recorder")
            return

        for name, port in self.inputs.items():
            stream: Stream[Any] = self._store.stream(name, port.type)
            unsub = port.subscribe(lambda msg, s=stream: s.append(msg))
            self.register_disposable(Disposable(unsub))
            logger.info("Recording %s (%s)", name, port.type.__name__)

    @rpc
    def stop(self) -> None:
        super().stop()
```

dimos/protocol/pubsub/impl/lcmpubsub.py
```diff
@@ -92,7 +91,7 @@ def publish(self, topic: Topic | str, message: bytes) -> None:
         topic_str = str(topic) if isinstance(topic, Topic) else topic
         self.l.publish(topic_str, message)
 
-    def subscribe_all(self, callback: Callable[[bytes, Topic], Any]) -> Callable[[], None]:
+    def subscribe_all(self, callback: Callable[[bytes, Topic], None]) -> Callable[[], None]:
```
leshy, line 93: suspicious, this should return unsubscribe() or Disposable

Dreamsorcerer: That's the return type, right? Not the callback. The return value is not read from the callback, so should be None.

jeff-hykin: I think Sam is right here

dimos/core/module.py
```diff
@@ -371,6 +372,57 @@ def blueprint(self) -> _BlueprintPartial:
 
         return partial(Blueprint.create, self)  # type: ignore[arg-type]
 
+    @rpc
+    def start_recording(self, db_path: str) -> None:
```
leshy: I'm not sure if we need modules to self-record, this will introduce GIL and local load, and we shouldn't depend on this for recording other language modules etc. Record system is external

leshy:
Can we get docs on how this is used? how do I test full functionality? I see recording to record can be set via CLI?

leshy:
@copilot resolve the merge conflicts in this pull request

dimos/core/module.py
```diff
@@ -376,6 +383,36 @@ def blueprint(self) -> _BlueprintPartial:
...
```
paul-nechifor, line 387: `start_recording` is a bit too generic/unprecise because it's available on all modules. How about `record_outputs` or `start_recording_outputs`.

dimos/core/module.py
```diff
@@ -376,6 +383,36 @@ def blueprint(self) -> _BlueprintPartial:
...
+        from dimos.memory2.store.sqlite import SqliteStore
+
+        self._rec_store = SqliteStore(path=db_path)
```
paul-nechifor, line 390: You might want to check that `start_recording` hasn't been called already... Also, technically, `@rpc` functions are called from different threads so `self._rec_store` needs a lock.

dimos/core/module.py
```diff
@@ -376,6 +383,36 @@ def blueprint(self) -> _BlueprintPartial:
...
+        for name, out in self.outputs.items():
+            stream = self._rec_store.stream(name, out.type)
+            reg = self._rec_store._registry.get(name)
```
paul-nechifor, line 393: You're accessing `._registry`, a private variable. You should either:

* rename it to `registry` and make it public.
* or add helper methods to the store like `get_stream_meta` and `update_stream_meta`.

dimos/memory2/observationstore/base.py
```diff
@@ -69,5 +69,9 @@ def fetch_by_ids(self, ids: list[int]) -> list[Observation[T]]:
         """Batch fetch by id (for vector search results)."""
         ...
 
+    def delete_range(self, t1: float, t2: float) -> list[int]:
+        """Delete observations with ts in [t1, t2]. Returns deleted IDs."""
+        raise NotImplementedError
```
paul-nechifor, line 74: Why not make it an `@abstractmethod`?

dimos/memory2/observationstore/sqlite.py
```diff
@@ -439,5 +439,21 @@ def fetch_by_ids(self, ids: list[int]) -> list[Observation[T]]:
...
+                f'SELECT id FROM "{self._name}" WHERE ts >= ? AND ts <= ?', (t1, t2)
+            ).fetchall()
+            ids = [r[0] for r in rows]
+            if ids:
+                placeholders = ",".join("?" * len(ids))
+                self._conn.execute(f'DELETE FROM "{self._name}" WHERE id IN ({placeholders})', ids)
+                self._conn.execute(
+                    f'DELETE FROM "{self._name}_rtree" WHERE id IN ({placeholders})', ids
+                )
+                self._conn.commit()
```
paul-nechifor, line 455: You can use DELETE-RETURNING and use 2 executes instead of 3 (i.e. merge the SELECT and DELETE into a single statement).

```python
rows = self._conn.execute(
    f'DELETE FROM "{self._name}" WHERE ts >= ? AND ts <= ? RETURNING id',
    (t1, t2),
).fetchall()
```

dimos/record/record_replay.py
```diff
@@ -0,0 +1,394 @@
...
+from dimos.protocol.service.spec import Service
+from dimos.types.timestamped import Timestamped
+from dimos.visualization.rerun.bridge import RerunConvertible, is_rerun_multi
+
+if sys.version_info >= (3, 11):
+    from typing import Self
+else:
+    from typing import Any as Self
+
+logger = logging.getLogger(__name__)
```
paul-nechifor, line 65: Use setup_logger

dimos/robot/cli/dimos.py
```diff
@@ -325,11 +333,28 @@ def run(
...
+    daemon: bool = typer.Option(False, "--daemon", "-d", help="Run in background"),
+    disable: list[str] = typer.Option([], "--disable", help="Module names to disable"),
+    blueprint_args: list[str] = typer.Option((), "--option", "-o"),
+    config_path: Path = typer.Option(
+        CONFIG_DIR / "dimos", "--config", "-c", help="Path to config file"
+    ),
+    show_help: bool = typer.Option(False, "--help"),
+) -> None:
+    """Start a robot blueprint"""
+    logger.info("Starting DimOS")
```
paul-nechifor, line 354: You haven't removed `logger.info("Starting DimOS")` from `_run`. Doesn't this just print the same message twice?

Dreamsorcerer:
#1976

PR #1713, opened 2026-03-30
feat(tool-streams): add tool streams

paul-nechifor:
## Problem

* Agents don't have a coherent idea of what response is coming from what tool.

Closes DIM-633

## Solution

* Added MCP streaming. Now agents receive updates from one tool in the same tool context.

## Breaking Changes

None

### DIFF
--- a/dimos/agents/mcp/tool_stream.py
+++ b/dimos/agents/mcp/tool_stream.py
@@ -1,0 +1,177 @@
+# Copyright 2026 Dimensional Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Server-initiated tool-stream notifications.
+
+A skill uses `ToolStream` to push text updates out to any connected MCP client
+(Claude Code, our own `McpClient`, curl, ...) while the skill's background work is
+still running.
+
+Transport: each `ToolStream.send` publishes a ready-made JSON-RPC
+`notifications/message` frame on the shared `/tool_streams` LCM topic. Skill
+workers and the `McpServer` process typically live in different workers, so we
+lean on LCM's local-multicast bus to cross that boundary. `McpServer` subscribes
+to the topic once, forwards each frame to every connected `GET /mcp` SSE client,
+and drops frames when nobody is listening.
+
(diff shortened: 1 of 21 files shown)


mustafab0:
Went through the PR, it seems good. Did not identify anything that would be a concern. 👍

docs/usage/tool_streams.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Quick example: inline updates
+
+```python
+import time
+
+from dimos.agents.annotation import skill
+from dimos.core.module import Module
+
+
```
leshy, line 32: I'd hope to just do:

```python
class Counter(Module):
    @skill
    def count_to(self, n: int) -> str:
        for i in range(n):
            yield f"At {i + 1} of {n}")
            time.sleep(0.1)
        return f"Counted to {n}."
```

leshy: spoke in private, will merge this, implement generators/async generator RPCs and tool streams will be natural like this later. tool streams are rare atm and will be easy to convert

dimos/agents/mcp/tool_stream.py
```diff
@@ -0,0 +1,177 @@
...
+    if message:
+        params["message"] = message
+    if tool_name is not None:
+        params["_meta"] = {"tool_name": tool_name}
+    return {"jsonrpc": "2.0", "method": NOTIFICATIONS_PROGRESS_METHOD, "params": params}
+
+
+def subscribe(callback: ToolStreamCallback) -> Callable[[], None]:
+    """Subscribe to the tool-stream LCM topic and return a cleanup callable."""
+    transport: pLCMTransport[dict[str, Any]] = pLCMTransport(TOOL_STREAM_TOPIC)
```
leshy, line 86: so we'll switch to zenoh, RPC, TF, transports are part of a module config so we can easily transition, this isn't changable on the outside, not a big deal but mentioning

docs/usage/tool_streams.md
paul-nechifor: Yes, I think this would work nicely with async. But it's probably not a good idea for sync because calls are limited by the RPC timeout period.

Need to implement RPC generators first though.

leshy (dismissed):
I think it's ok, clean, but would prefer RPCs and (async) generators then a new protocol, we spoke in private, this isn't heavily used atm and will be easy to switch over once ready

dimos/agents/mcp/tool_stream.py
paul-nechifor: How should I handle this? Should I switch it to Zenoh when that's done. Or should I allow this to be configurable and set it to LCM for now but Zenoh later?

I.e. is the switch to Zenoh permanent or just an option?

leshy: we have moduleconfig which lets you switch out transports on a module level,
I'd put the config for this there, and set to pLCM for now

https://github.com/dimensionalOS/dimos/blob/dev/dimos/core/module.py#L91

leshy (approved):

PR #1716, opened 2026-03-30
Have rerun render shared memory topics (fix dev for MacOS)

jeff-hykin:
Draft b/c `ShmSubset` probably shouldn't have `subscribe_all` since its basically a shim. Better fix would be changing rerun bridge to support topics without subscribe all. 

## Problem

MacOS doesn't render any camera images to rerun, cause rerun doesn't support shared memory transport


paul-nechifor:

PR #1543, opened 2026-03-13
Config options

Dreamsorcerer:
## Problem

Users need to be able to pass configuration options for blueprints. This adds basic support for CLI arguments, envvars and config files.

Closes #1082.

## Solution

Typer is not compatible with this kind of dynamic configuration, therefore we basically tell it to give us arbitrary strings via `-o` and then do all the processing ourselves, along with providing our own help output.

## How to Test

Should be able to add `--help` to dimos run to see blueprint options and then configure them with `-o`, env vars and JSON config files. For example:
`RERUNBRIDGEMODULE__MEMORY_LIMIT="50%" dimos --simulation run unitree-go2 -o voxelgridmapper.voxel_size=1 --config=foo.json`

PR #1727, opened 2026-04-01
Import time

Dreamsorcerer:
## Problem

This improves import time for some dimos.core modules, which are needed for #1543. Separating the import improvements for easier review.

## Solution

To avoid having too many repetitive inline imports harming readability, JpegEncoderMixin is moved to a separate module to avoid the heavy import of Image when importing the other classes in that module,

### DIFF
--- a/dimos/core/blueprints.py
+++ b/dimos/core/blueprints.py
@@ -28,8 +28,7 @@ if TYPE_CHECKING:
 from dimos.core.global_config import GlobalConfig, global_config
 from dimos.core.module import ModuleBase, ModuleSpec, is_module_type
 from dimos.core.module_coordinator import ModuleCoordinator
-from dimos.core.stream import In, Out
-from dimos.core.transport import LCMTransport, PubSubTransport, pLCMTransport
+from dimos.core.stream import In, Out, PubSubTransport
 from dimos.spec.utils import Spec, is_spec, spec_annotation_compliance, spec_structural_compliance
 from dimos.utils.generic import short_id
 from dimos.utils.logging_config import setup_logger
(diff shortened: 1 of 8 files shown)


bri-tong:
## Code Review

### Summary

Clean refactoring that defers heavy imports (`scipy`, `langchain_core`, `Image`/`cv2`/`rerun` via JPEG codecs) to improve module load time. The approach is sound — splitting `JpegEncoderMixin`/`JpegLCM` into `jpeg_lcm.py` and converting top-level imports to deferred inline imports is a well-targeted strategy. However, there is one regression that will cause a runtime crash.

---

### 🔴 Blocking Issue

**`NameError` at runtime in `dimos/core/transport.py`**

The `from dimos.utils import colors` import was removed ([line 26](https://github.com/dimensionalOS/dimos/pull/1727/files#diff-transport-L26)), but `colors.green()` and `colors.blue()` are still used in `PubSubTransport.__str__()`:

```python
def __str__(self) -> str:
    return (
        colors.green(f"{self.__class__.__name__}(")  # ❌ NameError
        + colors.blue(self.topic)
        + colors.green(")")
    )
```

Any call to `str()` on a transport instance (e.g. during logging in `Blueprint._connect_streams`) will crash.

**Fix**: Either restore the import, or defer it into `__str__()` if it has undesirable transitive deps:
```python
def __str__(self) -> str:
    from dimos.utils import colors
    return (
        colors.green(f"{self.__class__.__name__}(")
        + colors.blue(self.topic)
        + colors.green(")")
    )
```

---

### Non-blocking Suggestions

1. **`Quaternion.py` — duplicate deferred import**: `from scipy.spatial.transform import Rotation` appears identically in both `from_rotation_matrix()` and `to_euler()`. Python caches module imports so there's no performance penalty, but a shared private helper like `_scipy_rotation()` would keep the deferral in one place and reduce visual noise. Minor readability preference.

2. **Backwards-compat re-export**: `JpegEncoderMixin` was removed from `encoders.py` and `JpegLCM` from `lcmpubsub.py`. I verified all consumers on `dev` import via `dimos.core.transport.JpegLcmTransport`, so nothing is broken today. But if external users or downstream forks import from the old module paths, they'll get `ImportError`. Consider adding a deprecation re-export in `lcmpubsub.py` if backwards compat matters for these internal symbols.

3. **New module `jpeg_lcm.py`**: Well-structured — good docstring explaining why it exists, correct MRO composition. No issues.

---

### Questions / Clarifications

None — the intent is clear from the PR description and code.

paul-nechifor (approved):

PR #1731, opened 2026-04-02
fix: voxelized check before snapping in callback

ruthwikdasyam:
## Problem

Closes DIM-XXX

### DIFF
--- a/dimos/msgs/sensor_msgs/PointCloud2.py
+++ b/dimos/msgs/sensor_msgs/PointCloud2.py
@@ -56,11 +56,9 @@ class PointCloud2(Timestamped):
         pointcloud: o3d.geometry.PointCloud | o3d.t.geometry.PointCloud | None = None,
         frame_id: str = "world",
         ts: float | None = None,
-        voxelized: bool = False,
     ) -> None:
         self.ts = ts  # type: ignore[assignment]
         self.frame_id = frame_id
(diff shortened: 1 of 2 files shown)


dimos/msgs/sensor_msgs/PointCloud2.py
```diff
@@ -660,11 +662,12 @@ def to_rerun(
         elif mode == "boxes":
             box_size = size if size is not None else voxel_size
             half = box_size / 2
-            # Snap points to voxel grid centers so boxes tile properly
-            points = np.floor(points / box_size) * box_size + half
-            points, unique_idx = np.unique(points, axis=0, return_index=True)
-            if point_colors is not None and isinstance(point_colors, np.ndarray):
-                point_colors = point_colors[unique_idx]
+            if not self.voxelized:
```
leshy, line 663: I think should completely remove this self.voxelized thing and never do any snapping. also we shouldn't need any numpy operations ideally.

also idk if `self.as_numpy()` is needed and if it's expensive, I'm assuming rerun might then convert BACK into some other format, can we feed o3d pct directly to rerun or with less data transformation?

40ms is still pretty bad, the whole global mapper takes 4ms (on my machine) but sending to rerun takes 40

ruthwikdasyam:
was created just to share changes

PR #1734, opened 2026-04-03
fix: reduce rerun pipeline latency

leshy:
## Problem

to_rerun() methods for PointCloud2 and OccupancyGrid had heavy per-frame computations. 
The bottlenecks were all on the DimOS side before data even reached Rerun

Closes DIM-727

## Solution

- class_id colormaps (PointCloud2.py) - one-time AnnotationContext registration - viewer resolves colors instead of Python
- LUT costmap rendering OccupancyGrid
- Remove redundant floor+unique as Points from VoxelGridMapper are already on-grid and unique
- points_f32 - Read directly from tensor as float32, skip legacy float64 conversion
- Move voxel coordinates to CPU immediately after extraction so GPU memory is reserved for the hashmap
- Per-entity throttle - heavy types with max_hz dict per entity

Pointcloud2 `to_rerun` - before: ~350ms, Now: ~5ms
Costmap `to_rerun` - before: ~40ms, Now: ~5ms

## Breaking Changes

None

## How to Test

run `dimos --replay --replay-dir=unitree_go2_bigoffice run unitree-go2`
Should run 5+ min with no lag, Great at 0.1 voxel size, Good at 0.05 voxel size
will be improved in upcoming PR

- Follow-up is required to remove throttling and implement dropping msgs wrt rerun backpressure


dimos/robot/unitree/go2/blueprints/basic/unitree_go2_basic.py
```diff
@@ -110,6 +108,11 @@ def _go2_rerun_blueprint() -> Any:
         "world/global_map": _convert_global_map,
         "world/navigation_costmap": _convert_navigation_costmap,
     },
+    "max_hz": {
+        "world/global_map": 5,  # publishes at ~7.8 Hz
+        "world/color_image": 10,  # publishes at ~14 Hz
```
leshy, line 115: this is a bit funny, why does it publish at ~14 hz and not 10

ruthwikdasyam: thats the actual publish rate, without any cap.
mentioned there, So, we should set max_hz to some number less than that

dimos/msgs/nav_msgs/OccupancyGrid.py
```diff
@@ -548,9 +496,27 @@ def to_rerun(
...
-        # Generate RGBA texture and flip to match world coordinates
-        # Grid row 0 is at world y=origin (bottom), but texture row 0 is at UV v=0 (top)
-        rgba = np.flipud(self._generate_rgba_texture(colormap, opacity, cost_range, background))
+        # Downsample grid before LUT to avoid allocating huge RGBA array
+        max_tex = 256
+        grid = self.grid[::-1]  # flip for world coords (view, no copy)
+        if grid.shape[0] > max_tex or grid.shape[1] > max_tex:
+            step_h = max(1, grid.shape[0] // max_tex)
+            step_w = max(1, grid.shape[1] // max_tex)
+            grid = grid[::step_h, ::step_w]
```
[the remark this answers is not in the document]
ruthwikdasyam: the goal is to avoid huge arrays (1000x1000+), not to hit exactly 256, so this is fine

dimos/mapping/voxels.py
```diff
@@ -201,8 +201,10 @@ def get_global_pointcloud2(self) -> PointCloud2:
     # @timed()
     def get_global_pointcloud(self) -> o3d.t.geometry.PointCloud:
         voxel_coords, _ = self.vbg.voxel_coordinates_and_flattened_indices()
-        pts = voxel_coords + (self.config.voxel_size * 0.5)
-        out = o3d.t.geometry.PointCloud(device=self._dev)
+        # Move to CPU immediately to avoid holding a large duplicate on GPU.
+        cpu = o3c.Device("CPU:0")
```
leshy, line 205: this is not great becasue there might be global pointcloud manipulations we want to do on a GPU still,
not all get_global_pointcloud calls are for CPU

I've written this a while ago so no sure - but isn't this slowing down our most expensive operation by forcing a CPU execution on GPU machines? 

Please be careful with this, you should know the timing changes for both CPU and GPU exec for each line changed here in voxels.py

dimos/robot/unitree/go2/blueprints/basic/unitree_go2_basic.py
leshy: isn't go2 camera fps 30? this seems like it's already capped at input here by something?

ruthwikdasyam: they come in at ~14/15 fps with replay and hardware. so, seems not

dimos/mapping/voxels.py
ruthwikdasyam: sometimes we are getting CUDA OOM issue when map is super huge.. at the end of running with 400k voxels.
from what i understood - along with voxel hashmap, ‘get_global_pointcloud’ kept the published point cloud on GPU duplicating all the voxel coordinates in VRAM just to eventually send them over LCM (CPU). As the map grew, this duplicate exhausted GPU memory - CUDA OOM

In voxels.py, after operations with hashmap, this is how it flows to publish
‘GPU hashmap → extract coords (GPU) → arithmetic (GPU) → cache as PointCloud (GPU) → to_legacy (GPU → CPU copy) → LCM’

This is creating a duplicate copy, and temp allocations during conversion. As the map grew, this competed with hashmap memory. so..

‘GPU hashmap → extract coords (GPU) → copy to CPU → arithmetic (CPU) → cache as PointCloud (CPU) → to_legacy (CPU) → LCM’

this brings read path to CPU and eliminates several of those from GPU at once. allocations no longer overlap with write path. giving some room to breathe


I see this changes with pc to pc. But felt like this is a good optimization. Its ~0.02 ms with GPU vs ~0.2 with CPU at 400k voxels

dimos/robot/unitree/go2/blueprints/basic/unitree_go2_basic.py
leshy: are you sure? that seems broken to me, can create an issue to check, out of scope for this I guess

dimos/mapping/voxels.py
leshy: yes it's totally ok if these are actual numbers
>  Its ~0.02 ms with GPU vs ~0.2 with CPU at 400k voxels

ruthwikdasyam:
redirected - https://github.com/dimensionalOS/dimos/pull/1747

PR #1735, opened 2026-04-03
feat(sim): dimsim as connection

paul-nechifor:
## Problem

We need a light-er weight simulator.

Closes DIM-xxx

## Solution

* Add the DimSim simulator.
* Add tests which use it.

## Breaking Changes

The `--simulation` flag now takes a parameter: either `mujoco` or `dimsim`.

## How to Test

Run the apt scene:

```bash
uv run dimos --simulation=dimsim run unitree-go2-agentic
```

Run the empty scene:

```bash
uv run dimos --dimsim-scene=empty --simulation=dimsim run unitree-go2-agentic
```

Run the dimsim tests:
```bash
uv run pytest -m dimsim
```

### DIFF
--- a/dimos/e2e_tests/test_dimsim_spatial_memory.py
+++ b/dimos/e2e_tests/test_dimsim_spatial_memory.py
@@ -1,0 +1,32 @@
+# Copyright 2025-2026 Dimensional Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+
+
+@pytest.mark.dimsim
+def test_go_to_the_bed(lcm_spy, start_blueprint, human_input, dim_sim, explore_house) -> None:
+    start_blueprint(
+        "run",
+        "unitree-go2-agentic",
+        simulator="dimsim",
+    )
+    lcm_spy.save_topic("/rpc/McpClient/on_system_modules/res")
+    lcm_spy.wait_for_saved_topic("/rpc/McpClient/on_system_modules/res", timeout=1200.0)
+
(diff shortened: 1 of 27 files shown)


dimos/e2e_tests/test_dimsim_spatial_memory.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+@pytest.mark.dimsim
+def test_go_to_the_bed(lcm_spy, start_blueprint, human_input, dim_sim, explore_house) -> None:
+    start_blueprint(
+        "run",
+        "unitree-go2-agentic",
+        simulator="dimsim",
+    )
+    lcm_spy.save_topic("/rpc/McpClient/on_system_modules/res")
```
leshy, line 25: a bit concerning since this hardcodes LCM dep, lcm shoiuld be an irrelvant internal transport detail, we carefully designed dimos to make LCM replacable with anything, with 3 config changes on a module - RPC, TF and Pubsub rebind, this couples us too tightly

probably this should be some RPC on a module?

leshy (dismissed):
I have no notes on this apart from LCM, can fix in the followup, amazing!

dimos/e2e_tests/test_dimsim_spatial_memory.py
paul-nechifor: I can switch this to zenoh when we switch RPC to it. But currently I see this as the only option to observe that the system is ready from the outside.

I'm listening for the RPC response to `on_system_modules` because that's the RPC call which indicates that the Agent is ready to accept requests. I could technically replace this with polling for the HTTP MCP server, but that would only help in knowing when the system is ready. We still need to know the specific transportation protocol to listen for when the robot arrives at a certain odom position.

Are you suggesting that we should make an RPC call from outside the system? Even if we made that, we would still need to know the specific protocol being used.

mustafab0 (dismissed):

dimos/e2e_tests/test_dimsim_spatial_memory.py
leshy: discussed in person, defering for later, external PY api needs to support this, we need to know system state, initial connection protocol to dimos is predefined but per topic transport, rpc transport can be fed then as globalconfig to a connecting script? we will discuss soon, merging now
```python
dimos.connect() 
dimos.ready()
```

leshy (dismissed):

leshy (dismissed):

leshy (approved):

PR #1744, opened 2026-04-04
feat(blueprints): support starting new blueprints after startup

paul-nechifor:
## Problem

* Module coordinator doesn't allow starting multiple blueprints.

Closes DIM-XXX

## Solution

* Blueprints are now fully immutable. They cannot be built. You have to use a ModuleCoordinator to build a blueprint.
* Restructured the files and moved related code into the `dimos/core/coordinator` directory.

## Breaking Changes

* The import locations have changed.
* `Blueprint.build` was removed. You have to use a `ModuleCoordinator` to build.


dimos/core/coordination/blueprints.py
```diff
@@ -0,0 +1,232 @@
...
+
+if sys.version_info >= (3, 11):
+    from typing import Self
+else:
+    from typing_extensions import Self
+
+logger = setup_logger()
+
+
+class DisabledModuleProxy:
```
leshy, line 41: actually will be used for record/replay probably!

paul-nechifor: Sorry, I'm not sure what you mean...

leshy (approved):

dimos/core/coordination/blueprints.py
leshy: Replay of a recording will auto-supress modules it's replaying, probably using above functionality

PR #1749, opened 2026-04-05
config refactor

leshy:
we now infer default config from Configurable[ConfigT] type, mypy auto-understands self.config, no need for default_config attribute


leshy:
actually not sure

PR #1756, opened 2026-04-07
feat(ci): single step

paul-nechifor:
## Problem

We have a docker image build step as part of the build (even if it is cached). Would be better to just `uv sync, pytest, mypy`.

Closes DIM-783

## Solution

* Split CI into two workflows: ci.yml (tests + mypy) and docker-build.yml (image builds only)
* Only run coverage on merges (to avoid slow down from instrumentation).

## Breaking Changes

None.

## How to Test

None.


leshy (dismissed):
honestly I have no idea how this works, but as is tradition, no one but latest person that worked on CI knows how CI works.

so I'll just ask user level questions:

1. where is the local cache now?
2. can I add a docker image that depends on some other docker image we build (g1 from ros-python let's say) - will it intelligently build it only when it has to, rebuild if it's parent is rebuilt, and can I run tests on it? - this is where most complexity of CI came from.. 

if you say no to 2 - can still approve, we will need this eventually, but not yet - can handle this on a case-by-case basis, until someone writes a custom dimos CI action

.github/workflows/ci.yml
```diff
@@ -0,0 +1,70 @@
...
+
+      - name: Run tests
+        if: github.event_name != 'push'
+        run: |
+          /entrypoint.sh bash -c "source .venv/bin/activate && pytest --durations=0 -m 'not (tool or mujoco)'"
+
+      - name: Run tests with coverage
+        if: github.event_name == 'push'
+        run: |
+          /entrypoint.sh bash -c "source .venv/bin/activate && _DIMOS_COV=1 coverage run -m pytest --durations=0 -m 'not (tool or mujoco)' && coverage combine && coverage html && coverage report"
```
Dreamsorcerer, line 53: Why are we running the tests twice?

Dreamsorcerer (dismissed):
I'll probably make further suggestions in followup PRs, but overall looks reasonable.

.github/workflows/ci.yml
paul-nechifor: We're not running them twice. The coverage tests only run on merges. The non-coverage ones run in PRs.

paul-nechifor:
> honestly I have no idea how this works, but as is tradition, no one but latest person that worked on CI knows how CI works.
> 
> so I'll just ask user level questions:
> 
>     1. where is the local cache now?
> 
>     2. can I add a docker image that depends on some other docker image we build (g1 from ros-python let's say) - will it intelligently build it only when it has to, rebuild if it's parent is rebuilt, and can I run tests on it? - this is where most complexity of CI came from..
> 
> 
> if you say no to 2 - can still approve, we will need this eventually, but not yet - can handle this on a case-by-case basis, until someone writes a custom dimos CI action

It's stored locally.  This checks out new branches without cleaning up previous ones.

```
      - uses: actions/checkout@v5
        with:
          clean: false
 ```

I know what you're thinking: "but that messes up different runs". True, but i'm also running this:

```
git clean -ffdx -e .venv
```

That removes every non-tracked file which is not in `.venv`.

So the end effect is that only the `.venv` dir is preserved between runs. `uv sync` automatically updates it if it's slightly off.

leshy:
> I know what you're thinking: "but that messes up different runs". True, but i'm also running this:

is there an issue with parallelization then? I assumed we have parallel checkouts if running multiple branches tests?
also do you need to clean data/* files?

paul-nechifor:
> > I know what you're thinking: "but that messes up different runs". True, but i'm also running this:
> 
> is there an issue with parallelization then? I assumed we have parallel checkouts if running multiple branches tests? also do you need to clean data/* files?

I don't know if there are multiple checkouts per worker, but that shouldn't be an issue in either case. It will just have as many "caches" as there are checkouts.

I think data needs to be cleared because the pre-commit script doesn't allow files that are not commited and other branches could polute that.

Essentially what is preserved is only what is in .venv and git files because git and uv are capable of bringing the checkout to a clean state. If I tried to preserve data, I would have to add some logic to handle different cases.

leshy (approved):
I'm unsure how this works wrt multiple parallel CI runs but one mind in the torment nexus of CI is enough

paul-nechifor:
> can I add a docker image that depends on some other docker image we build (g1 from ros-python let's say) - will it intelligently build it only when it has to, rebuild if it's parent is rebuilt, and can I run tests on it? - this is where most complexity of CI came from..

Sorry, forgot to answer this.

I've kept docker-build.yml the same. The only difference is that the images get built on merges to dev, main, or every Monday morning.

If you add a forth docker image for just g1, then it will be built under those same conditions.

leshy:
> If you add a forth docker image for just g1, then it will be built under those same conditions.

but docker images introduce system changes that might be required for running actual tests on that build (imagine DDS for example) so this will cause it to fail tests no?

we can shelf this, this is solving an issue we don't have yet, by sacrificing actual irl performance.

in a month someone might want us to CI build custom docker + run pytest on it that passes only within that system, so should keep that in mind

paul-nechifor:
> > If you add a forth docker image for just g1, then it will be built under those same conditions.
> 
> but docker images introduce system changes that might be required for running actual tests on that build (imagine DDS for example) so this will cause it to fail tests no?

The solution is to have multiple smaller commits. If you need to add a new system dependency, add it a as a separate PR. That will built a new image.

leshy:
> The solution is to have multiple smaller commits. If you need to add a new system dependency, add it a as a separate PR. That will built a new image.

yeah that's perfect! we simplify 99.9% usecases and make 0.1% slightly slower

PR #1765, opened 2026-04-09
Task: Add meaningful manipulation tests

mustafab0:
## Problem

The manipulation module used a single flat joint namespace which broke when multiple robots were connected — joint names collided and state updates routed to the wrong robot. Blueprint configs also had redundant `or None` patterns, and `drake_world.py` had dead code left over from a removed model cache.

---

## Solution

Added unit tests for the two most critical pieces of new logic:
- `_on_joint_state`: routing, splitting, missing-joint handling, init capture, multi-robot correctness
- `PickAndPlaceModule` heuristics: object detection lookup, occlusion offset math, grasp orientation, place_back guard

---

## Breaking Changes

`joint_name_mapping` on `RobotModelConfig` now uses `/` as the prefix separator (e.g. `left/joint1` instead of `left_joint1`). Existing blueprints in this repo are already updated. External code that constructs `RobotModelConfig` with custom `joint_name_mapping` dicts needs to adopt the new convention.

---

## How to Test

1. Run the unit tests: `pytest dimos/manipulation/test_manipulation_unit.py dimos/manipulation/test_pick_and_place_unit.py -v`
2. Run the full manipulation integration suite: `pytest dimos/manipulation/ -v`


dimos/manipulation/test_pick_and_place_unit.py
```diff
@@ -0,0 +1,153 @@
...
+        # At origin, no shift should occur (division-by-zero guard)
+        assert abs(ox) < 1e-3
+        assert abs(oy) < 1e-3
+
+    def test_grasp_orientation_near_is_top_down(self):
+        q = PickAndPlaceModule._grasp_orientation(gx=0.3, gy=0.0, xy_dist=0.3)
+        # Near object: pitch = 180 degrees (top-down), tilt = 0
+        # Quaternion should be valid (norm ~1)
+        norm = (q.x**2 + q.y**2 + q.z**2 + q.w**2) ** 0.5
+        assert abs(norm - 1.0) < 0.01
```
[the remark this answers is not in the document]
mustafab0: fixed

dimos/manipulation/test_pick_and_place_unit.py
```diff
@@ -0,0 +1,155 @@
...
+    obj.name = name
+    obj.object_id = object_id
+    obj.center = Vector3(x=center[0], y=center[1], z=center[2])
+    obj.size = Vector3(x=size[0], y=size[1], z=size[2])
+    return obj
+
+
+def _make_module() -> PickAndPlaceModule:
+    """Create a PickAndPlaceModule with mocked __init__."""
+    with patch.object(PickAndPlaceModule, "__init__", lambda self: None):
```
paul-nechifor: I'm not sure why you're patching `__init__` given that you're calling `__new__` which I don't think calls `__init__`.

But this function should be a fixture, and it's not a good fixture. It pretty much duplicates what `PickAndPlaceModule.__init__` does. This makes it fragile.

It would be better to actually call `PickAndPlaceModule()`. If `PickAndPlaceModule.__init__` does something you don't want, you can patch that specific thing.

dimos/manipulation/test_pick_and_place_unit.py
```diff
@@ -0,0 +1,155 @@
...
+from dimos.msgs.geometry_msgs.Vector3 import Vector3
+
+
+def _make_det_object(
+    name: str = "cup",
+    object_id: str = "abc12345",
+    center: tuple[float, float, float] = (0.5, 0.0, 0.3),
+    size: tuple[float, float, float] = (0.05, 0.05, 0.10),
+) -> MagicMock:
+    """Create a lightweight mock DetObject with the attributes used by PickAndPlaceModule."""
```
paul-nechifor: DetObject is Object which is a dataclass. Why can't you just instantiate it? It just holds values, so there's nothing to mock.

---

(BTW, it would be nice to actually rename Object to DetObject or something else.)

dimos/manipulation/test_pick_and_place_unit.py
```diff
@@ -0,0 +1,155 @@
...
+        result = module._find_object_in_detections("cup")
+        assert result is det
+
+    def test_find_by_object_id(self):
+        module = _make_module()
+        det = _make_det_object(object_id="abc12345")
+        module._detection_snapshot = [det]
+
+        # Truncated prefix match
+        result = module._find_object_in_detections("anything", object_id="abc1")
```
paul-nechifor, line 85: I thought you said you got rid of this functionality. Looks like if you have two objects abc12345 abc19999, this method just returns the first match. If matching on truncated prefix is just to avoid typing long ids, it should at least throw an error if there are multiple matches.

mustafab0: Updated to give error on multiple matches

dimos/manipulation/test_pick_and_place_unit.py
mustafab0: Created a DetObject data class instead of mock 👍

dimos/manipulation/test_pick_and_place_unit.py
mustafab0: Fixed, made a `module` fixture. Real init chain runs from `PickandPlaceModule` to `Module`

paul-nechifor (approved):

PR #1767, opened 2026-04-09
feat(workers): use types for worker messages

paul-nechifor:
## Problem

Closes DIM-XXX


leshy (approved):
great, this stuff is cool for logging also, as you can write __str__ on these etc

PR #1769, opened 2026-04-10
Feat/memory2 - plotting, examples, recorder module, semantic search

leshy:
- adds data_type attribute to observations
- adds raw storage of strings or numbers directly into metadata store (doesn't require separate fetch and blob store)
- adds basic memory module (module that inits a db for you) and a Recorder and Query/Embedding module as example
- adds 3d projections for a whole set of detections (ImageDetections2D -> ImageDetections3DPC) as opposed to previously singular ImageDetection2D -> ImageDetection3D
- implements a new TimedSensorReplay shim that uses memory2 (our full reply file is down to 200Mb compressed vs 2Gb!)
- switches over legacy tests to LegacyPickleStore (will delete dimos/memory soon entirely and deal with pickle files)
- adds many examples to docs for memory
- removes a test that ensures all modules have start and stop (was annoying to add boilerplate, especially now with `CompositeResource` and automatic detection of thread leaks we have)


dimos/memory/test_embedding.py
```diff
@@ -16,17 +16,16 @@
...
 
-dir_name = "unitree_go2_bigoffice"
-
 
 @pytest.mark.skip
 def test_embed_frame() -> None:
     """Test embedding a single frame."""
     # Load a frame from recorded data
-    video = TimedSensorReplay(get_data(dir_name) / "video")
+    from dimos.msgs.sensor_msgs.Image import Image
```
paul-nechifor, line 26: You don't need the import. It's just for type hinting, and that's not checked in test files.

dimos/memory2/store/base.py
```diff
@@ -61,6 +61,10 @@ def __repr__(self) -> str:
         names = object.__getattribute__(self, "_store").list_streams()
         return f"StreamAccessor({names})"
 
+    def items(self) -> list[tuple[str, Stream[Any]]]:
+        store: Store = object.__getattribute__(self, "_store")
+        return [(name, store.stream(name)) for name in store.list_streams()]
```
paul-nechifor: Why not just `self._store`? What is `__getattribute__` needed for?

```suggestion
        return [(name, self._store.stream(name)) for name in self._store.list_streams()]
```

docs/capabilities/memory/demo_rerun.py
```diff
@@ -0,0 +1,88 @@
+# Copyright 2026 Dimensional Inc.
```
paul-nechifor, line 1: What is this file for?

dimos/memory2/module.py
```diff
@@ -108,3 +127,133 @@ def _apply_pipeline(self, stream: Stream[Any]) -> Stream[Any]:
...
+    @rpc
+    def start(self) -> None:
+        super().start()
+
+        self.model = self.register_disposable(self.config.embedding_model())
+        self.model.start()
+
+        self.embeddings = self.store.stream("color_image_embedded", Image)
+
+        # fmt: off
```
paul-nechifor, line 218: I keep seeing these, but ruff aligns them nicely:

```python
        self.register_disposable(
            self.store.streams.color_image.live()
            .filter(lambda obs: obs.data.brightness > 0.1)
            .transform(QualityWindow(lambda img: img.sharpness, window=0.5))
            .transform(EmbedImages(self.model, batch_size=2))
            .save(self.embeddings)
            .drain_thread()
        )
```

dimos/memory2/stream.py
```diff
@@ -319,7 +396,11 @@ def drain(self) -> int:
             n += 1
         return n
 
-    def observable(self) -> reactivex.Observable[Observation[T]]:
+    def drain_thread(self) -> DisposableBase:
+        """Drain this stream on the dimos thread pool; returns a disposable."""
+        return self.subscribe(on_next=lambda _: None)
```
paul-nechifor: You don't need `on_next`.

```suggestion
        return self.subscribe()
```

paul-nechifor: Although, is drain_thread even needed if it's just equivalent to calling `subscribe`?

dimos/memory2/module.py
```diff
@@ -108,3 +127,133 @@ def _apply_pipeline(self, stream: Stream[Any]) -> Stream[Any]:
...
+        self.register_disposable(
```
paul-nechifor: The implementation of `drain_thread` calls `subscribe`. It's implementation already registers the disposable:

```python
    def subscribe(
        self,
        on_next: Callable[[O], None] | ObserverBase[O] | None = None,
        on_error: Callable[[Exception], None] | None = None,
        on_completed: Callable[[], None] | None = None,
    ) -> DisposableBase:
        """Subscribe to this stream as an RxPY Observable.

        The subscription is tracked and disposed when this stream is stopped.
        """
        return self.register_disposable(
            self.observable().subscribe(  # type: ignore[call-overload]
                on_next=on_next,
                on_error=on_error,
                on_completed=on_completed,
            )
        )
```
Aren't you effectively disposing twice? (I assume there's some protection against doing it twice, but it still should be clear where it's done).

dimos/memory2/module.py
leshy: oh, not for me, I'll check why

dimos/memory/test_embedding.py
leshy: I type check my test files

dimos/memory2/backend.py
```diff
@@ -87,22 +89,32 @@ def loader() -> Any:
...
             obs.id = row_id
 
-            # Store blob
+            # Store blob (non-scalar data only)
             if encoded is not None:
                 assert self.blob_store is not None
                 self.blob_store.put(self.name, row_id, encoded)
                 # Replace inline data with lazy loader
-                obs._data = _UNLOADED
+                obs._data = _UNLOADED  # type: ignore[assignment]
```
paul-nechifor: I see you've added some `# type: ignore`s. But this wasn't needed before. Why is it needed now?

dimos/memory2/backend.py
```diff
@@ -171,7 +186,7 @@ def _vector_search(self, query: StreamQuery) -> Iterator[Observation[T]]:
         vs = self.vector_store
         assert vs is not None and query.search_vec is not None
 
-        hits = vs.search(self.name, query.search_vec, query.search_k or 10)
+        hits = vs.search(self.name, query.search_vec, query.search_k or 4096)
```
paul-nechifor: Maybe move this to a config option somewhere. Seems very magical. I wouldn't expect it to be a power of 2.

dimos/memory2/module.py
```diff
@@ -108,3 +127,133 @@ def _apply_pipeline(self, stream: Stream[Any]) -> Stream[Any]:
...
+
+class MemoryModuleConfig(ModuleConfig):
+    db_path: str | Path = "recording.db"
+
+    @field_validator("db_path", mode="before")
+    @classmethod
+    def _resolve_path(cls, v: str | Path) -> Path:
+        p = Path(os.fspath(v))
+        if not p.is_absolute():
+            from dimos.utils.data import get_project_root
```
paul-nechifor: Imports should be at the top, but this is unrelated to data, so you should probably use `DIMOS_PROJECT_ROOT` instead of `get_project_root`

dimos/memory2/module.py
```diff
@@ -108,3 +127,133 @@ def _apply_pipeline(self, stream: Stream[Any]) -> Stream[Any]:
...
+    Provides a config with a db_path for the module's MemoryStore, and common start/stop logic.
+
+    If changing the backend globally in dimos, this class will be replaced
+    """
+
+    config: MemoryModuleConfig
+    _store: SqliteStore | None = None
+
+    @property
+    def store(self) -> SqliteStore:
```
paul-nechifor, line 189: This works in practice now, but it's not a good direction. If `self._store` is initialized on the first access, there could be a future subclass of MemoryModule which accesses it from two different threads so it could race.

It's annoying to use locks all over the place, so we will probably want to eventually moved to asyncio Modules.

dimos/memory2/module.py
```diff
@@ -108,3 +127,133 @@ def _apply_pipeline(self, stream: Stream[Any]) -> Stream[Any]:
...
+               .filter(lambda obs: obs.data.brightness > 0.1)
+               .transform(QualityWindow(lambda img: img.sharpness, window=0.5))
+               .transform(EmbedImages(self.model, batch_size=2))
+               .save(self.embeddings)
+               .drain_thread()
+        )
+        # fmt: on
+
+    @skill
+    def search(self, query: str) -> Stream[Any]:
```
paul-nechifor: Do agents understand `Stream[Any]`?

paul-nechifor: I don't think `@skill`s works without a docstring.

Can you add a test for this? I'm not sure what its meant to do.

dimos/memory2/module.py
leshy: Yes - ish, stream/store has an internal register_disposable that's there to shutdown everything given stream.stop() or store.stop() is called.

but also users can take care to subscribe/unsubscribe in their own context according to the modules needs.

In my case store shutdown and stopping this drain is the same event (module shutdown) so I can remove register_disposable

dimos/memory2/stream.py
leshy: yeah just a naming thing, I didn't like users calling subscribe() to drain.

I might actually remove .subscribe() entirely and keep .observable() for clarity that you are switching into rxpy context at that point

dimos/memory2/module.py
```diff
@@ -108,3 +127,133 @@ def _apply_pipeline(self, stream: Stream[Any]) -> Stream[Any]:
...
+            else:
+                raise FileExistsError(f"Recording already exists: {db_path}")
+
+        if not self.inputs:
+            logger.warning("Recorder has no In ports — nothing to record, subclass the Recorder")
+            return
+
+        for name, port in self.inputs.items():
+            stream: Stream[Any] = self.store.stream(name, port.type)
+            self.register_disposable(Disposable(port.subscribe(partial(Stream.append, stream))))
```
paul-nechifor: `partial` with the class is a bit confusing. Isn't this just:

```suggestion
            self.register_disposable(Disposable(port.subscribe(stream.append)))
```

docs/capabilities/memory/demo_rerun.py
leshy: renamed to rerun_demo.py - it's some random (only one) memory2 + rerun demo, 
It's messy but useful and I assume people prefer to see it.

docs are WIP and will do a second pass asap

dimos/memory2/module.py
leshy: True, let me investigate what actually I want here wrt data/ absolute paths etc. (import is inline on purpose since I didn't want to load if not used but don't mind moving)

## on constants

ugh I'm very unhappy about this global `constants.py`  why are we introducing this? would much rather people store their constants in ther respectative config files or worst case subdirs

Constants for shared memory should be in shared memory (as kwarg? why is this even a separate var?)
LCM_MAX_CHANNEL_NAME_LENGTH belongs in LCM, why is this a separate var? this is a hard protocol limit

dimos/memory2/module.py
leshy: true, created an issue to adress this https://github.com/dimensionalOS/dimos/issues/1910

dimos/memory2/module.py
leshy: no, this is a bit of a stub, also mistyped - fixing, it returns `PoseStamped` we can discuss how to wire this, this module replaces full old semantic search

dimos/memory2/backend.py
leshy: this is a hard limit of the vector db search, it should probably be in the backend hardcoded setting
it doesn't have to be power of 2 if you specify `k` yourself

dimos/memory2/backend.py
leshy: this one isn't, sry, reviewing

dimos/memory2/stream.py
```diff
@@ -262,15 +322,19 @@ def fetch(self) -> list[Observation[T]]:
             )
         return list(self)
 
-    def first(self) -> Observation[T]:
+    def to_list(self) -> list[O]:
```
paul-nechifor: Maybe delete `fetch`? Aliases are for when a library is so mature that it needs backwards compatibility.

dimos/memory2/stream.py
```diff
@@ -308,6 +372,19 @@ def summary(self) -> str:
         dur = t1 - t0
         return f"{self}: {n} items, {dt0} — {dt1} ({dur:.1f}s)"
 
+    def cache(self) -> Stream[T, O]:
```
paul-nechifor: How about `materialize`? `cache` implies `cache lookups`.

dimos/memory2/transform.py
```diff
@@ -105,17 +105,273 @@ def __call__(self, upstream: Iterator[Observation[T]]) -> Iterator[Observation[R
...
+            if last_ts is None or obs.ts - last_ts >= interval:
+                last_ts = obs.ts
+                yield obs
+
+    return FnIterTransformer(_throttle)
+
+
+def speed() -> FnIterTransformer[Any, float]:
+    """Compute speed (m/s) between consecutive observations from their poses."""
+    import math
```
paul-nechifor: Please move all these imports to the top.

dimos/memory2/store/base.py
leshy: was some kind of recursion guard, not needed now

dimos/memory2/backend.py
leshy: cleaned up all ignores introduced

dimos/memory2/module.py
paul-nechifor: > why are we introducing this?

`constants.py` has been there since last year. It's global, because some things are used throughout the project. For example `DIMOS_PROJECT_ROOT`, so it makes sense to be there.

The constants for shared memory used to be in `dimos/robot/unitree_webrtc/unitree_go2.py`. It obviously didn't belong there. I moved it to `dimos/constants.py` because I couldn't find a better place.

> LCM_MAX_CHANNEL_NAME_LENGTH belongs in LCM, why is this a separate var?

It is in LCM, but they don't expose it. So I had to copy it somewhere.

dimos/memory2/module.py
leshy: yes no idea what I was thinking late last night

dimos/memory2/transform.py
leshy: yeah these are on purpose, but I guess light enough to be on top

leshy: scipy I will leave inline

dimos/memory2/stream.py
leshy: yes that's much better

dimos/memory2/stream.py
leshy: yeah here I'm just using them as aliases not backwards compat, but I think to_list is cleaner and will just remove fetch yeah

dimos/memory2/transform.py
```diff
@@ -105,17 +105,273 @@ def __call__(self, upstream: Iterator[Observation[T]]) -> Iterator[Observation[R
...
+    return median + k * scale, median, scale
+
+
+def _otsu_threshold(values: list[float]) -> float:
+    """1D Otsu threshold: maximizes between-class variance over the value list."""
+    sorted_vals = sorted(values)
+    n = len(sorted_vals)
+    total = sum(sorted_vals)
+    best_var, best_thresh = -1.0, sorted_vals[-1]
+    cum = 0.0
```
paul-nechifor, line 253: Wot?

leshy: idk bro, some of these, these tresholds, similarity are quickly vibed while experimenting with data analysis - I  will fix this but heavily rewrite/review in other passes

dimos/msgs/nav_msgs/OccupancyGrid.py
```diff
@@ -478,10 +478,110 @@ def cell_value(self, world_position: Vector3) -> int:
 
         return int(self.grid[y, x])
 
+    def _generate_rgba_texture(
```
paul-nechifor: I don't think this should be private. It's used in a different class.

But I think it belongs in `mapping/occupancy/visualizations.py`. It's just a different way to visualize, right? I think `dimos/msgs` should be for messages and not algorithms which process them.

dimos/msgs/sensor_msgs/Image.py
```diff
@@ -374,6 +374,11 @@ def crop(self, x: int, y: int, width: int, height: int) -> Image:
 
         return Image(data=cropped_data, format=self.format, frame_id=self.frame_id, ts=self.ts)
 
+    @property
+    def brightness(self) -> float:
+        """Return mean brightness in [0, 1]."""
+        return float(self.data.mean() / 255.0)
```
paul-nechifor: We support 16 bit images, so you should derive the maximum value (255) based on `ImageFormat`.

dimos/memory2/stream.py
leshy: kept subscribe() as it does register_disposable and handles cleanup for users

dimos/robot/unitree/go2/blueprints/smart/unitree_go2.py
```diff
@@ -30,6 +36,22 @@
     ReplanningAStarPlanner.blueprint(),
     WavefrontFrontierExplorer.blueprint(),
     PatrollingModule.blueprint(),
-).global_config(n_workers=7, robot_model="unitree_go2")
+).global_config(n_workers=9, robot_model="unitree_go2")
+
+
+class Go2MemoryConfig(RecorderConfig):
+    db_path: str | Path = "recording_go2.db"
```
paul-nechifor, line 43: Is this saved in the root of the project? We probably need a directory for storing state at this point. For example, spatial memory uses `assets/output/memory/`. For most other stuff we use `~/.local/state/dimos`. But maybe we need a local directory we can dump all sorts of stuff? That way you wouldn't have to specifically ignore this in `.gitignore`.

dimos/perception/detection/type/detection2d/imageDetections2D.py
```diff
@@ -34,6 +36,22 @@
...
+
+        Args:
+            *predicates: Functions that take a detection and return True to keep it
+
+        Returns:
+            A new ImageDetections2D instance with filtered detections
+        """
+        filtered = list(self.detections)
+        for predicate in predicates:
+            filtered = [det for det in filtered if predicate(det)]
```
paul-nechifor: Isn't it better to loop over the detections first? If most predicates eliminate nothing you're recreating `filtered` over and over. (I don't have an intuition for what the practical length values for the lists are.)

dimos/robot/unitree/go2/blueprints/smart/unitree_go2.py
leshy: yes, this is kind of unresolved and we need to decide, dumping in project root is pretty bad

docs/capabilities/memory/index.py
```diff
@@ -0,0 +1,144 @@
+# Copyright 2026 Dimensional Inc.
```
paul-nechifor: Can you please rename all of these scripts with the `demo_` prefix?

dimos/perception/detection/type/detection2d/imageDetections2D.py
leshy: inconsequential since it's like 4-5 detections max but yeah this is weird and ugly, better to loop over detections, fixing

generally this is all bad, like `timestamped` class and operators based on collections of timed things, we should have a `Collection` protocol/base class for any collection. so native functools, py stuff should work on top of imageDetections2D

dimos/memory2/transform.py
paul-nechifor: No, I just found the variable name amusing. :sweat_smile:

dimos/memory2/module.py
leshy: let's start moving those to respectative dirs and ideally under pydantic configs as defaults if you don't mind? 

a bit out of scope here but just for future reviews - worry is in 2 months we will have these types of constants files with 200 entries, and we won't know which module owns, who introduced and why, and which modules depend on it. I understand that some constants are truly global but those are rare. 

LCM_MAX_CHANNEL_NAME_LENGTH can be in service/lcm.py ? 
SHM stuff in shm transport etc

dimos/utils/testing/replay.py
```diff
@@ -12,11 +12,292 @@
...
+                    if is_disposed:
+                        return None
+                    observer.on_next(data)
+                    if next_message is not None:
+                        schedule_emission(next_message)
+                    else:
+                        observer.on_completed()
+                    return None
+
+                sched.schedule_relative(delay, emit)
```
paul-nechifor: schedule_relative returns a CompositeDisposable which must be disposed of. It contains a Timer thread, so that must be shut down.

I've noticed that schedule_relative is used in 2 other place, none of which dispose of the result. But they're all very similar so maybe it should be abstracted. This place is at least very similar to `def stream` in `memory/timeseries/base.py`.

dimos/msgs/sensor_msgs/Image.py
leshy: unfortunately once again you are correct

leshy: fixed

dimos/utils/testing/replay.py
leshy: refactored, though all of this will be deleted once memory2 takes root in tests also, this is temp code now

dimos/msgs/nav_msgs/OccupancyGrid.py
leshy: Yes you are right, I don't know where to put this stuff atm, but will move there

dimos/memory2/backend.py
leshy: moved to sqlite impl

dimos/memory2/module.py
leshy: updated this to use constants.py

dimos/robot/unitree/go2/blueprints/smart/unitree_go2.py
leshy: https://github.com/dimensionalOS/dimos/issues/1914

leshy: to be solved

paul-nechifor (approved):

PR #1779, opened 2026-04-12
feat(api): add porcelain api with connect

paul-nechifor:
## Problem

Closes DIM-730

## Solution

See [the docs](https://github.com/dimensionalOS/dimos/blob/paul/feat/porcelain-api-with-connect/docs/usage/python-api.md) for how it works.

## How to Test

Included in the docs.

### DIFF
--- a/dimos/core/coordination/module_coordinator.py
+++ b/dimos/core/coordination/module_coordinator.py
@@ -22,6 +22,7 @@ import sys
 import threading
 from typing import TYPE_CHECKING, Any, cast
 
+from dimos.core.coordination.rpyc_server import RpycServer
 from dimos.core.coordination.worker_manager import WorkerManager
 from dimos.core.coordination.worker_manager_docker import WorkerManagerDocker
 from dimos.core.coordination.worker_manager_python import WorkerManagerPython
@@ -52,9 +53,7 @@ class ModuleCoordinator(Resource):
     ) -> None:
         self._global_config = g
         manager_types: list[type[WorkerManager]] = [WorkerManagerDocker, WorkerManagerPython]
-        self._managers: dict[str, WorkerManager] = {
-            cls.deployment_identifier: cls(g=g) for cls in manager_types
-        }
+        self._managers = {cls.deployment_identifier: cls(g=g) for cls in manager_types}
         self._deployed_modules = {}
         self._deployed_atoms: dict[type[ModuleBase], BlueprintAtom] = {}
         self._resolved_module_refs: dict[tuple[type[ModuleBase], str], type[ModuleBase]] = {}
@@ -62,6 +61,8 @@ class ModuleCoordinator(Resource):
         self._class_aliases: dict[type[ModuleBase], type[ModuleBase]] = {}
         self._module_transports: dict[type[ModuleBase], dict[str, PubSubTransport[Any]]] = {}
         self._started = False
+        self._modules_lock = threading.RLock()
+        self._rpyc = RpycServer(self)
 
     def start(self) -> None:
         from dimos.core.o3dpickle import register_picklers
@@ -72,6 +73,8 @@ class ModuleCoordinator(Resource):
         self._started = True
 
     def stop(self) -> None:
+        self._rpyc.stop()
+
         for module_class, module in reversed(self._deployed_modules.items()):
             logger.info("Stopping module...", module=module_class.__name__)
             try:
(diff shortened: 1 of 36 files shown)


dimos/core/coordination/python_worker.py
```diff
@@ -353,7 +371,62 @@ def _worker_entrypoint(conn: Connection, worker_id: int) -> None:
...
+
+        case SuppressConsoleRequest():
+            _suppress_console_output()
+            return WorkerResponse(result=True)
+
+        case StartRpycRequest():
+            if state.rpyc_server is not None:
+                return WorkerResponse(result=state.rpyc_server.port)
+            WorkerRpycService._instances = state.instances
+            state.rpyc_server = ThreadedServer(
```
leshy, line 405: you listen on 0.0.0.0 which I don't actually mind (we assume Physical layer is our boundary) but we keep getting these security PRs

But then also worries me having a host setting per *thing* since exposing dimos to network becomes super complicated with 50 host settings.

I'd propose let's define a global setting for "dimos listen host" which is 127.0.0.1 by default.

and then your listen address is that global setting unless explicilty set as a config setting for your feature?

dimos/core/coordination/module_coordinator.py
```diff
@@ -52,16 +53,16 @@ def __init__(
...
+        self._managers = {cls.deployment_identifier: cls(g=g) for cls in manager_types}
         self._deployed_modules = {}
         self._deployed_atoms: dict[type[ModuleBase], BlueprintAtom] = {}
         self._resolved_module_refs: dict[tuple[type[ModuleBase], str], type[ModuleBase]] = {}
         self._transport_registry: dict[tuple[str, type], PubSubTransport[Any]] = {}
         self._class_aliases: dict[type[ModuleBase], type[ModuleBase]] = {}
         self._module_transports: dict[type[ModuleBase], dict[str, PubSubTransport[Any]]] = {}
         self._started = False
+        self._modules_lock = threading.RLock()
+        self._rpyc = RpycServer(self)
```
leshy, line 65: this creates competing communciation channels through which to call skills etc. we already have `@rpc` and that rpc client shim, I think this is an ok boundary to what you can access on remote modules..  and this will work as we transition to zenoh, switch to remote etc.

now we have a totally separate new protocol API for calling skills/modules, maybe RpyServer should only be on the coordinator which holds those RPC module shims?

leshy (dismissed):
big comment on actual protocol to talk to workers, but I'm aproving as we iterate. protocol replacable later also.

Another 
```
run() and restart() are only available in local mode. On a connected instance they raise NotImplementedError.
```

IMO we especially want those in remote mode:
- you start dimos, then claude iterates on the code on individual module, it wants to restart that module
- openclaw controls dimos, wants to deploy a new module

I'm not totally sure why those features are not "basically for free" and require extra work

dimos/core/coordination/module_coordinator.py
paul-nechifor: The reason I added this is to have deep access to private data for debugging. With `@rpc` you can only access what you mark. RPyC gives you access to everything.

This is why I was saying that reloading/etc on `dimos run ...` is probably not needed. I'm not looking to make `RPyC` a competing communication channel. I see `dimos run ...` as more for production and running from the repl as more for debugging where you want access to everything.

If we use this for agents, I assume they want access to more things too and not just things marked with `@rpc`.

paul-nechifor:
> big comment on actual protocol to talk to workers, but I'm aproving as we iterate. protocol replacable later also.
> 
> Another
> 
> ```
> run() and restart() are only available in local mode. On a connected instance they raise NotImplementedError.
> ```
> 
> IMO we especially want those in remote mode:
> 
>     * you start dimos, then claude iterates on the code on individual module, it wants to restart that module
> 
>     * openclaw controls dimos, wants to deploy a new module
> 
> 
> I'm not totally sure why those features are not "basically for free" and require extra work

Implemented this. Required pickling classes.

dimos/core/coordination/python_worker.py
paul-nechifor: Fixed.

dimos/core/coordination/module_coordinator.py
leshy: Aha the way I see it (fight me on this if you think the other way)

We don't really know how agents will like to interact with running dimos.
is it via `dimos run` session and .runs and .restarts (I actually assume yes) or full session ownership via "local" runs?

In case of a local run - agent has to interact with something like a repl via something like a jupyter notebook - I tested this for memory - claude actually is trained explicitly for this and it has jupyter specific tool calls and we should investigate how this actually works 

I asked for the same API in both cases for a few reasons:

- we actually don't know what agents will prefer
- I see both "repl" and "local" usage as just "API clients" towards dimos, apart from light adapter layer, code being executed should be the same in both cases
- we actually don't know what WE will prefer. I can imagine this being utilized by some web ui etc to restart modules in a few months, who knows.
```python
router.get("/restart_module", lambda req, res:
   dimos = Dimos.connect(req.user.cluster_name)
   req.write("connected to cluster")
   dimos.navigation.stop()
   req.write("nav stopped")
   dimos.navigation.start()
   req.end("nav started")
)
```
makes a lot of sense

- it's simpler for users understanding, the whole beauty of this .connect vs dimos() is that you get the same behaving object in both cases
- I assume (plz correct me if missing something) that if our internal structure if set well we should support both usecases naturally - my thinking is .connect provides interface to modulecoordinator, but from that point onwards both Dimos and Dimos.connect() are communicating with remote processes and no changes between these 2 cases.
 

in terms of RPyC vs RPC

since I see both .connect and local run as just API clients, I think dask is a great example, you can do `dask = Dask()` or `dask = Dask("remote_cluster")`

https://docs.dask.org/en/stable/deploying.html

I actually like a hard boundary and single protocol so that agents (and users) don't get confused about what's running where and start doing weird things expecting there isn't a process boundary so I don't mind @rpc only access, and agents/users adding @rpc access to attributes they want to debug, I'm not hard set on this and won't block RPyC in workers, but presenting my case in case you agree

paul-nechifor: So what should I change? Remove RPyC and use @rpc for ModuleCoordinator?

leshy: we had a chat, decided to merge early and experiment

leshy (approved):
So generally yes, but it feels like we might want a refactor.

Deploying multiples of modules/blueprints,
To potentially use RPC and not RPyC for everything

so we can sketch all this out later, but let's kick this off :D

PR #1780, opened 2026-04-12
native rebuild changes

jeff-hykin:
## Problem

Editing source code doesnt cause native modules to rebuild. Very easy to forget and not friendly for ai edits. 

## Solution

Some generic utils:
- **`dimos/utils/change_detect.py`**: Content-hash-based file change detection using `xxhash`. 
- **`NativeModuleConfig.rebuild_on_change`**: Optional `list[str|Path|Glob]` 

## Breaking Changes

None

## How to Test

```bash
# Change detection tests
pytest dimos/utils/test_change_detect.py -v

# Native module rebuild tests
pytest dimos/core/test_native_rebuild.py -v

# Native module crash/thread leak test
pytest dimos/core/test_native_module.py::test_process_crash_triggers_stop -v

# LCM isolation tests
pytest dimos/protocol/pubsub/test_pattern_sub.py -v
pytest dimos/protocol/pubsub/impl/test_lcmpubsub.py -v

# Full fast suite
pytest -m 'not (tool or slow or mujoco)' dimos/
```


dimos/core/native_module.py
```diff
@@ -161,44 +184,76 @@ def start(self) -> None:
...
-            logger.info("Stopping native process", pid=self._process.pid)
+            logger.info(
+                "Stopping native process",
+                module=self._mod_label,
+                pid=self._process.pid,
+            )
             self._process.send_signal(signal.SIGTERM)
             try:
-                self._process.wait(timeout=self.config.shutdown_timeout)
+                self._process.wait(timeout=DEFAULT_THREAD_JOIN_TIMEOUT)
```
leshy: we do want configurable shutdown timeouts via config, since some process can have a legitimate longer cleanup time, let's try and utilize the configuration system, even if default is some global hardcoded value

dimos/utils/change_detect.py
```diff
@@ -0,0 +1,377 @@
...
+
+
+def _hash_files(files: list[Path]) -> str:
+    """Compute an aggregate xxhash digest over the sorted file list."""
+    h = xxhash.xxh64()
+    for fpath in files:
+        try:
+            # Include the path so additions/deletions/renames are detected
+            h.update(str(fpath).encode())
+            h.update(fpath.read_bytes())
```
leshy: this will load a full file into memory, ok for source files, but will for sure cause a weird issue for someone down the line. maybe easy hack for this is to check the file size and simply not hash if larger then 500k-1Mb or something

leshy:
so not sure, haven't looked into details, but is there a reason we aren't relying nix build before each module start, since nix essentially does the above by itself?

dimos/core/native_module.py
```diff
@@ -83,15 +84,21 @@ class NativeModuleConfig(ModuleConfig):
     extra_env: dict[str, str] = Field(default_factory=dict)
     shutdown_timeout: float = 10.0
     log_format: LogFormat = LogFormat.TEXT
+    rebuild_on_change: list[PathEntry] | None = None
 
     # Override in subclasses to exclude fields from CLI arg generation
-    cli_exclude: frozenset[str] = frozenset()
+    cli_exclude: frozenset[str] = frozenset({"rebuild_on_change"})
```
leshy: this breaks subclass overriding right? since now all nativemodules need to explicitly include this value

dimos/core/native_module.py
```diff
@@ -135,17 +143,32 @@ class NativeModule(Module):
...
     _watchdog: threading.Thread | None = None
     _stopping: bool = False
-    _last_stderr_lines: collections.deque[str]
+    _stderr_tail: collections.deque[str]
+    _stdout_tail: collections.deque[str]
+    _tail_lock: threading.Lock
+    _tail_size = 50
+
+    @property
+    def _mod_label(self) -> str:
```
leshy, line 161: @cachedproperty better here

dimos/core/native_module.py
```diff
@@ -135,17 +143,32 @@ class NativeModule(Module):
     _process: subprocess.Popen[bytes] | None = None
     _watchdog: threading.Thread | None = None
     _stopping: bool = False
-    _last_stderr_lines: collections.deque[str]
+    _stderr_tail: collections.deque[str]
+    _stdout_tail: collections.deque[str]
```
leshy: every line is already being passed to the logger.

So the tail buffers are keeping a duplicate copy of data that's already been logged just to re-log it again - why? I see also `_last_stderr_lines` was added previously.

IMO if we want more ergonomics with logging should fix the actual logger to allow you to filter, have per module log files you can tail in another term etc

dimos/core/native_module.py
```diff
@@ -161,44 +184,76 @@ def start(self) -> None:
...
             cmd=" ".join(cmd),
             cwd=cwd,
         )
+
+        # fix bad-close and leaked process issues.
+        # start_new_session=True is the thread-safe way to isolate the child
+        # from terminal signals (SIGINT from the tty).  preexec_fn is unsafe
+        # in the presence of threads (subprocess docs), so we only use it on
+        # Linux where prctl(PR_SET_PDEATHSIG) has no alternative.
+        def _child_preexec_linux() -> None:
```
leshy: nice

dimos/core/native_module.py
```diff
@@ -83,15 +84,21 @@ class NativeModuleConfig(ModuleConfig):
     extra_env: dict[str, str] = Field(default_factory=dict)
     shutdown_timeout: float = 10.0
     log_format: LogFormat = LogFormat.TEXT
+    rebuild_on_change: list[PathEntry] | None = None
```
leshy: btw this is the first actual setting that's different between prod and dev, we def don't want this on deployments, so probably if toggled True should be done not in code but the local config file level or via local env vars.

dimos/core/native_module.py
jeff-hykin: `self.config.shutdown_timeout` is only in native module (it was a change introduced by me in the unity sim PR I think) so I was debating it.

Thanks for the clarification though, I was debating if we want consistency or configurability. (fixed)

jeff-hykin:
> so not sure, haven't looked into details, but is there a reason we aren't relying nix build before each module start, since nix essentially does the above by itself?

Unfortunately yeah. I don't care much about speed, but rebuild_on_change is 1980x faster than nix's check on reboot, and merely 280x faster once nix has a "warm" cache. And that's for the mid360, which is absolutely tiny. For a native module like PathFollower being checked onboard the G1, nix takes 1037.89ms just to confirm nothing changed (`rebuild_on_change` runs in 5ms on the G1).

I'm not thrilled about `rebuild_on_change`, but it was important for development on the G1.

Now that I think about it though, we should probably have it always rebuild *unless* there is a `rebuild_on_change`, since its more or less an optimization

dimos/core/native_module.py
jeff-hykin: Whats the difference between prod and dev? I don't follow

dimos/utils/change_detect.py
jeff-hykin: Switched to streaming to reduce ram overhead, still runs fast (still 100x or 1000x nix's change checking)

Added a `max_file_size` arg, files greater than max_file_size will switch to using the last modified date instead of actually hashing the file. For rebuild_on_change the user is always expected to only mention source code so probably a non-problem but max_file_size should help this tool be more general.

dimos/core/native_module.py
jeff-hykin: Good catch, removed

dimos/core/native_module.py
jeff-hykin: Agreed, I think this was a side effect of two different logggng approaches that got merged. Went ahead pushed simplified logging

dimos/core/native_module.py
paul-nechifor: On production the code doesn't change so there's no need to rebuild. (I assume this is what Ivan meant.)

leshy: yeah this hashes all the files in the dir which can be quite costly for larger files, def don't want to run in prod

leshy: yes hashing all files is computationally expensive and we don't want prod deployments to do this during the startup

leshy:
>  I don't care much about speed, but rebuild_on_change is 1980x faster than nix's check on reboot, and merely 280x faster once nix has a "warm" cache. And that's for the mid360, which is absolutely tiny. For a native module like PathFollower being checked onboard the G1, nix takes 1037.89ms just to confirm nothing changed (rebuild_on_change runs in 5ms on the G1).
> 
> I'm not thrilled about rebuild_on_change, but it was important for development on the G1.

I feel like I'm missing something since 1 second doesn't actually sound like an issue to me. This runs only in dev envs right? Build systems often take way longer then this. Why was this important for g1 dev? 

Change detection in the first place is a fancy feature and `ssh g1:cd dimos/bla && make` seems ok to me as well

I don't mind this code in - since it's a generic file change detector that we might want elsewhere as well (I'd probably rely on modification times though) I'm not pushing back against this, just trying to understand this work.

btw you can also use linux inotify api and autobuild
```sh
while inotifywait -r -e modify,create,delete src; do make; done
```

or to piggyback on nativemodule internal rebuild step, just clean up the build artifacts
```sh
while inotifywait -r -e modify,create,delete src; do echo "rebuild scheduled"; rm bin/result; done
```

> Now that I think about it though, we should probably have it always rebuild unless there is a rebuild_on_change, since its more or less an optimization

Builds can get quite expensive so I wouldn't do this by default but I guess there are cases in which you want this yeah, in my mid360 dev I was doing `rm ./bin/result && build` by hand after cpp changes (to see compiler output), would do `inotifywait` if I wanted autobuilds

paul-nechifor:
> I don't mind this code in - since it's a generic file change detector that we might want elsewhere as well

Personally, I don't like the idea of adding code under the justification that it could be useful later. Code always has maintenance costs, and less code is always better.

I've said this before, but I don't understand the need for all this change detecting code. Build systems should already do this. It could be justifiable if it was intolerably slow, but nix taking 1037.89ms doesn't seem bad.

leshy:
> > I don't mind this code in - since it's a generic file change detector that we might want elsewhere as well
> 
> Personally, I don't like the idea of adding code under the justification that it could be useful later. Code always has maintenance costs, and less code is always better.

yeah I agree actually

> I've said this before, but I don't understand the need for all this change detecting code. Build systems should already do this. It could be justifiable if it was intolerably slow, but nix taking 1037.89ms doesn't seem bad.

yeah I'm also confused, seems like 1500 lines to shave off one second in dev, introduces dev/prod env config split etc

jeff-hykin:
> yes hashing all files is computationally expensive and we don't want prod deployments to do this during the startup

> I feel like I'm missing something since 1 second doesn't actually sound like an issue to me. 

how can these two things be true at the same time?

leshy:
> > yes hashing all files is computationally expensive and we don't want prod deployments to do this during the startup
> 
> > I feel like I'm missing something since 1 second doesn't actually sound like an issue to me.
> 
> how can these two things be true at the same time?

it's prod vs dev 

in dev - if rebuild in dev takes 1 second to start is a non issue

in prod deployments we are not rebuilding stuff, it's important that we don't re-compute a hash of every file in every native module for no reason (I'm assuming native modules can get pretty large and we'll lose startup time and compute every single time dimos runs)

dimos/core/native_module.py
jeff-hykin: > all the files in the dir

No, it doesn't hash all the files in the dir

> quite costly for larger files

It has the "if large use last edit time" feature, so no its not costly on large files (and why would people point it at large files?)

On my mac `rebuild_on_change` runs in under 1 millisecond for the largest nav native module. Absolutely unnoticeable. 

I feel like you guys are just going off theory

jeff-hykin:
> it could be useful later

This PR already has another place where the tooling is used`mesh_utils.py`

jeff-hykin:
At the time this PR was created (and still an open question) not every build command is guareteed to use nix. So I figured it'd be nice to both: not call the build command (whatever it might be) merely never or always -- but rather when something changed. 

My timeline:
1. manual rebuild is painful on G1 nav stack
2. I add attempt rebuild every time
3. (just for PR) I expect you guys would say rebuild every time is too slow, so I add `rebuild_on_change`
4. (just for PR) I expect you guys to say "file hashing is not specific to native modules please put it in a util" so I put it in a util
5. (just for PR) I expect the util to not be used, so I change mesh-whatever to use it
6. (just for PR) I want to make this easier to review so I break it out into a separate PR
7. (just for PR) wait a week between feedback

I'm just trying to get nav merged, I'm doing everything I can to be as generic/efficient/modular/small-pr's as possible

we can revert this in 20min a week from now, I don't care about the feature itself.


If you guys want a PROD=true env var that bypasses everything, that sounds fine to me.

dimos/core/coordination/python_worker.py
```diff
@@ -319,12 +320,15 @@ def _suppress_console_output() -> None:
 
 def _worker_entrypoint(conn: Connection, worker_id: int) -> None:
     apply_library_config()
+    # Ignore SIGINT so the coordinator can orchestrate shutdown via the pipe.
+    # Without this, workers race with the coordinator: they start tearing down
+    # modules locally while the coordinator tries to send stop() RPCs, causing
+    # BrokenPipeErrors.
+    signal.signal(signal.SIGINT, signal.SIG_IGN)
     instances: dict[int, Any] = {}
```
jeff-hykin, line 328: this is a problem that showed up more in native modules, but it is technically not related to native module rebuild. Let me know if this should be a one file PR

dimos/core/native_module.py
```diff
@@ -60,6 +59,27 @@ class MyCppModule(NativeModule):
 from dimos.core.module import Module, ModuleConfig
 from dimos.utils.logging_config import setup_logger
 
+# ctypes is only needed for the Linux child-preexec helper below.  Hoisting
+# the import out of the inner function avoids re-importing on every start().
+if sys.platform.startswith("linux"):
```
jeff-hykin, line 64: this is for better native module cleanup on shutdown

dimos/hardware/sensors/lidar/common/livox_sdk_config.hpp
```diff
@@ -40,16 +41,42 @@ struct SdkPorts {
...
     int fd = memfd_create("livox_sdk_config", 0);
     if (fd < 0) {
         perror("memfd_create");
         return {-1, ""};
     }
+#else
+    // mkstemp replaces the 6 X's in place — e.g. livox_sdk_config.aB3xY9.
+    // Honor $TMPDIR when set (sandboxed macOS apps and CI runners point
+    // it at a per-process scratch dir); fall back to /tmp.
+    const char* tmpdir = std::getenv("TMPDIR");
```
jeff-hykin, line 66: bad transfer, this should be in the rosnav merge, will remove (its macos support for livox)

PR #1783, opened 2026-04-13
fix: rework navigation map into structural/live/planning layers with confidence-based structural updates

christiefhyang:
## Summary 
This PR upgrades the navigation map pipeline from a single static binary map to a layered map model for better handling world changes.  

The new model includes structural_map, live_map, and planning_map.  

Global planner and local planner both consume planning_map so behavior stays consistent between global path generation and local obstacle handling.  

Map fusion and clearing logic are moved into occupancy operations as pure functions, keeping planner modules focused on planning logic only.

## Background and Motivation  

- Current map behavior is mostly static and does not robustly represent environment structure changes over time, such as moved desks or furniture updates.  
- The system needs a mechanism to distinguish short term perception changes from long term structural updates.  
- This PR introduces a confidence based structural update flow and a time bounded live layer, so the planner can react quickly to recent changes while preserving long term map stability.

## Design Goals  

- Keep responsibilities clean  
- Navigation map manages layered map state  
- Operations module owns fusion and clearing logic  
- Planner modules only consume planning map and do not perform map fusion  
- Keep changes minimal and targeted  
- Only touch required files and avoid broad refactor  
- Support incremental rollout  
- Add config switches and thresholds to tune behavior safely

## Implementation Details  
1) navigation_map.py  
Replaced single binary map storage with layered state  
structural_map stores long lived structure  
live_map stores most recent observation snapshot  
planning_map is composed from structural plus live layers  
Added per cell signed confirmation counts  
positive counts increase occupied confidence  
negative counts increase free confidence  
unknown observations do not change confidence  
Update flow  
on first update initialize structural map and counters  
on next updates optionally update structural map using confidence thresholds  
always refresh live map  
always refresh planning map through fusion  
Retention logic  
live layer retention is evaluated using receive time based age to avoid timestamp base mismatch between sources  
Compatibility  
binary_costmap is kept as an alias to planning_map to avoid broad call site changes

2) operations.py  
Added pure functions for map update and fusion  
update_confirmation_counts updates signed confidence counters  
update_structural_map applies write and clear thresholds to structural layer  
fuse_planning_map combines structural and live layers for planner consumption  
These functions isolate data transformation from planner logic and make behavior easier to reason about and test.

3) global_planner.py  
Safe goal check now reads planning_map instead of old binary map source  
No map fusion logic introduced here.

4) local_planner.py  
Path forward obstacle check now reads planning_map  
This ensures local stop and replan triggers use the same map semantics as global planner decisions.

5) global_config.py  
Added new configuration parameters  
structural_write_threshold  
structural_clear_threshold  
live_map_retention_sec  
enable_structural_update
 
## Behavioral Changes  
Planner map source is now unified as planning_map  
Recent obstacles can affect planning immediately through live layer  
Persistent observations can gradually update structural layer through confidence thresholds  
If live layer expires by retention time, planning map falls back to structural layer only  
Structural updates can be disabled by config for staged rollout

## Why This Is Safe  

- Scope is limited to five files only  
- Existing planner entry points and module boundaries remain intact  
- Compatibility alias reduces migration risk for existing binary_costmap usage  
- Fusion logic is centralized in pure functions with explicit inputs and outputs  
- Config gates allow conservative deployment and easy rollback of structural updates

## Validation  

- Syntax validation completed for all changed files using Python 3.11 py_compile  
- Full pytest run is currently blocked in this environment due to existing interpreter dependency mismatch unrelated to this PR  
- Code path audit confirms  
- global planner consumes planning_map  
- local planner clearance check consumes planning_map  
- map fusion and clearing logic stays outside planner modules

## Risks and Mitigations  
1) Structural update thresholds may be too aggressive or too conservative for different environments  
Mitigation; Use structural_write_threshold and structural_clear_threshold for tuning, and enable_structural_update for controlled rollout

2) Live layer may expire too quickly or too slowly depending on sensor cadence  
Mitigation; Tune live_map_retention_sec per deployment profile

3) Time base mismatch between map message timestamp and process clock  
Mitigation; Live retention now uses receive time based aging in navigation_map

## Rollout Plan  
Phase 1  
Enable layered planning map with structural updates enabled and conservative thresholds  
Phase 2  
Tune thresholds based on observed false occupied and false free transitions  
Phase 3  
Lock deployment defaults once behavior stabilizes in representative scenes

## Rollback Plan  
Set enable_structural_update to false to freeze structural layer updates  
Increase or decrease live_map_retention_sec to adjust live influence quickly  
If needed, planners still work with planning_map fallback behavior through existing interfaces without additional refactor

## Files Changed  
dimos/navigation/replanning_a_star/navigation_map.py  
dimos/navigation/replanning_a_star/global_planner.py  
dimos/navigation/replanning_a_star/local_planner.py  
dimos/mapping/occupancy/operations.py  
dimos/core/global_config.py

- [ x ] I have read and approved the [CLA](https://github.com/dimensionalOS/dimos/blob/main/CLA.md).


dimos/core/global_config.py
```diff
@@ -28,16 +30,12 @@ def _get_all_numbers(s: str) -> list[float]:
...
     simulation: bool = False
     replay: bool = False
-    replay_dir: str = "go2_sf_office"
-    new_memory: bool = False
-    viewer: ViewerBackend = "rerun"
-    n_workers: int = 2
+    rerun_enabled: bool = True
+    rerun_server_addr: str | None = None
+    viewer_backend: ViewerBackend = "rerun-web"
+    n_dask_workers: int = 2
```
jeff-hykin: Why were these changed?

dimos/navigation/replanning_a_star/global_planner.py
```diff
@@ -114,7 +104,7 @@ def stop(self) -> None:
         self._replan_event.set()
 
         if self._thread is not None and self._thread is not current_thread():
-            self._thread.join(DEFAULT_THREAD_JOIN_TIMEOUT)
+            self._thread.join(2)
```
jeff-hykin: Please revert

dimos/navigation/replanning_a_star/global_planner.py
```diff
@@ -138,13 +127,6 @@ def handle_goal_request(self, goal: PoseStamped) -> None:
         self._replan_limiter.reset()
         self._plan_path()
 
-    def set_safe_goal_clearance(self, clearance: float) -> None:
```
jeff-hykin, line 141: This kinda seems important. Is it still active but under a different name?

jeff-hykin:
I'm always happy to have help on nav. I would've hoped to have a bit more of a design discussion before work was started. Its true dynamic updates are a problem, but theyre mostly a problem for the 3d case. The 2d case is fine. Planning could be more advanced, but I'm concerned that a multi-layer approach is going to make things complicated. Ivan and myself are trying to make it where there are fewer maps (one source of truth) rather than more maps. 

Could you explain how the system can tell a difference between something structural vs nonstructural?

christiefhyang:
> rather

Reverting this PR back to draft. The core idea is to provide structural map updates and additions are mostly made in operations.py. I will take a look again on the revert changes you and Paul mentioned. Will circle back.

christiefhyang:
Addressed the review comments in the latest update:

- Changed structural threshold validation to require clear_threshold < write_threshold.
- Updated the structural map docstring to clarify that it returns a new OccupancyGrid instead of mutating in place.
- Added config-time validation for structural_clear_threshold < structural_write_threshold.
- Updated NavigationMap properties to return copies instead of exposing internal references.
- Preserved the planning map timestamp on property refresh.
- Reverted the thread join timeout back to DEFAULT_THREAD_JOIN_TIMEOUT.
- Restored set_safe_goal_clearance/reset_safe_goal_clearance and wired safe-goal validation back to the configurable clearance value.

Validation:
python -m py_compile dimos/mapping/occupancy/operations.py dimos/core/global_config.py dimos/navigation/replanning_a_star/navigation_map.py dimos/navigation/replanning_a_star/global_planner.py

paul-nechifor:
Closed. Can be reopened later.

PR #1784, opened 2026-04-13
Jeff/fix/rconnect2

jeff-hykin:
NOTE: this will become ready-for-review once this [dimos-viewer](https://github.com/dimensionalOS/dimos-viewer/pull/20) PR is merged

## Problem(s)

1. dimos-viewer with `--connect` doesn't work for remote connections. 
2. there's no way to remap what dimos-viewer publishes on. It publishes to a hardcoded LCM path of `cmd_vel` which ends up making stream-renaming a pain since everything else must be renamed to know if the cmd_vel is coming from the viewer or from a planner/module.
3. We do the "if global_config == rerun" in multiple places

## Solution

Use websockets instead of LCM for the viewer.

- A dimos module starts a websocket server listening for clicked_point and tele_cmd_vel
- The dimos-viewer tries to connect to a websocket and publishes clicked_point and tele_cmd_vel on there.
- Theres a consolidated `vis_module` to de-dup the visualization logic

## Breaking Changes

None

## How to Test

You'll need the `jeff/fix/connect` branch of `dimos-viewer` compiled and python installed into dimos:

```bash
# Build and install dimos-viewer (requires pixi)
cd <path-to-dimos-viewer>
git checkout jeff/fix/connect
pixi run build
uv pip install target/wheels/dimos_viewer-*.whl --force-reinstall
```

```bash
# Unit + E2E tests
uv run pytest dimos/visualization/rerun/test_websocket_server.py dimos/visualization/rerun/test_viewer_ws_e2e.py -v --timeout=30 -k "not test_viewer_ws_client_connects"
```

Alternatively

```bash
# Terminal 1: start the websocket server
python -c "
from dimos.visualization.rerun.websocket_server import RerunWebSocketServer
import threading
server = RerunWebSocketServer(port=3030)
server.clicked_point.subscribe(lambda pt: print(f'[CLICK] {pt.x:.3f},{pt.y:.3f},{pt.z:.3f}'))
server.tele_cmd_vel.subscribe(lambda tw: print(f'[TWIST] {tw}'))
server.start()
threading.Event().wait()
"

# Terminal 2: start the viewer
dimos-viewer --connect rerun+http://localhost:9877/proxy --ws-url ws://127.0.0.1:3030/ws
```

Click in the 3D viewport and use WASD keys — should see `[CLICK]` and `[TWIST]` in terminal 1.


dimos/visualization/rerun/constants.py
```diff
@@ -0,0 +1,23 @@
...
+
+from typing import Literal, TypeAlias
+
+ViewerBackend: TypeAlias = Literal["rerun", "foxglove", "none"]
+RerunOpenOption: TypeAlias = Literal["none", "web", "native", "both"]
+
+RERUN_OPEN_DEFAULT: RerunOpenOption = "native"
+RERUN_ENABLE_WEB = True
+RERUN_GRPC_PORT = 9876
+RERUN_WEB_PORT = 9090
```
leshy: don't bind to 9090, this is an extremely common port (our china VPN already uses for example, TOR is on 9090)

dimos/visualization/rerun/constants.py
```diff
@@ -0,0 +1,23 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Literal, TypeAlias
+
+ViewerBackend: TypeAlias = Literal["rerun", "foxglove", "none"]
+RerunOpenOption: TypeAlias = Literal["none", "web", "native", "both"]
+
+RERUN_OPEN_DEFAULT: RerunOpenOption = "native"
```
leshy, line 28: shouldn't rerun defaults belong to visualization/rerun in a standard config dict?

dimos/visualization/rerun/constants.py
jeff-hykin: - pre-existing (I just moved the variable)
- this doesn't set the port (its just for docker to know about the existing Rerun default port)

I think changing the code to set the port to a non-default port (and then testing everything again) is outside the scope of this PR, but I agree we should do it

dimos/visualization/rerun/constants.py
jeff-hykin: No because they need to be imported by global_config and we don't want to import rerun just to run `dimos --help`

jeff-hykin:
Sorry I forgot to port the project.toml change from rosnav8 back to this branch. Ready now though

dimos/visualization/rerun/websocket_server.py
```diff
@@ -0,0 +1,275 @@
...
+    z: float = 1.0
+    roll: float = 1.0
+    pitch: float = 1.0
+    yaw: float = 1.0
+
+
+class Config(ModuleConfig):
+    # Intentionally binds 0.0.0.0 by default so the viewer can connect from
+    # any machine on the network (the typical robot deployment scenario).
+    host: str = "0.0.0.0"
```
paul-nechifor: Please use `global_config.listen_host` which is `127.0.0.1` now. People can manually change that to `0.0.0.0` if they want to connect from another machine.

dimos/visualization/rerun/websocket_server.py
```diff
@@ -0,0 +1,275 @@
...
+
+
+def _handshake_noise_filter(record: logging.LogRecord) -> bool:
+    """Drop noisy "opening handshake failed" records from port scanners etc."""
+    msg = record.getMessage()
+    return not ("opening handshake failed" in msg or "did not receive a valid HTTP request" in msg)
+
+
+class CmdVelScaling(BaseModel):
+    """Per-dimension multipliers applied to outgoing teleop cmd_vel twists.
```
paul-nechifor: What is the use-case for this?

dimos/visualization/rerun/websocket_server.py
```diff
@@ -0,0 +1,275 @@
...
+        """Log full dimos-viewer commands that viewers can use to connect."""
+        import socket
+
+        from dimos.utils.generic import get_local_ips
+        from dimos.visualization.constants import RERUN_GRPC_PORT
+
+        local_ips = get_local_ips()
+        hostname = socket.gethostname()
+        ws_url = f"ws://127.0.0.1:{self.config.port}/ws"
+        grpc_url = f"rerun+http://127.0.0.1:{RERUN_GRPC_PORT}/proxy"
```
paul-nechifor: Use `global_config.listen_host` instead of 127.0.0.1

dimos/visualization/rerun/websocket_server.py
```diff
@@ -0,0 +1,275 @@
...
+            lines.append("")
+        lines.append(f"  hostname: {hostname}")
+        lines.append("=" * 60)
+        lines.append("")
+
+        logger.info("\n".join(lines))
+
+    def _run_server(self) -> None:
+        """Entry point for the background server thread."""
+        self._ws_loop = asyncio.new_event_loop()
```
paul-nechifor: You don't need to create a new loop, there's `self._loop`.

dimos/visualization/rerun/websocket_server.py
```diff
@@ -0,0 +1,275 @@
...
+        try:
+            self._ws_loop.run_until_complete(self._serve())
+        except Exception:
+            logger.error("RerunWebSocketServer failed to start", exc_info=True)
+        finally:
+            self._server_ready.set()  # unblock stop() even on failure
+            self._ws_loop.close()
+
+    async def _serve(self) -> None:
+        import websockets.asyncio.server as ws_server
```
paul-nechifor: Please instruct claude to move imports to the top.

dimos/visualization/rerun/websocket_server.py
```diff
@@ -0,0 +1,275 @@
...
+                    float(msg.get("linear_y", 0)) * s.y,
+                    float(msg.get("linear_z", 0)) * s.z,
+                ),
+                angular=Vector3(
+                    float(msg.get("angular_x", 0)) * s.roll,
+                    float(msg.get("angular_y", 0)) * s.pitch,
+                    float(msg.get("angular_z", 0)) * s.yaw,
+                ),
+            )
+            logger.debug(f"RerunWebSocketServer: twist → {twist}")
```
paul-nechifor: Personally, I'd remove all the debug logging. If someone wants to debug, he can add temporary prints himself. This just adds noise.

dimos/visualization/rerun/test_websocket_server.py
```diff
@@ -0,0 +1,480 @@
...
+        return self
+
+    def __exit__(self, *_: Any) -> None:
+        if self._ws is not None and self._loop is not None:
+            self._loop.run_until_complete(self._ws.close())
+        if self._loop is not None:
+            self._loop.close()
+
+    async def _connect(self) -> Any:
+        import websockets.asyncio.client as ws_client
```
paul-nechifor: You import this several times. Move it to the top.

dimos/visualization/rerun/test_websocket_server.py
```diff
@@ -0,0 +1,480 @@
...
+        with MockViewerPublisher(f"ws://127.0.0.1:{_TEST_PORT}/ws") as pub:
+            pub.send_click(1.5, 2.5, 0.0, "/world", timestamp_ms=1000)
+            pub.flush()
+
+        done.wait(timeout=2.0)
+        mod.stop()
+
+        assert len(received) == 1
+        pt = received[0]
+        assert abs(pt.x - 1.5) < 1e-9
```
paul-nechifor: Use pytest.approx

dimos/visualization/rerun/test_websocket_server.py
```diff
@@ -0,0 +1,480 @@
...
+        received: list[Any] = []
+        done = threading.Event()
+        mod.clicked_point.subscribe(_collect(received, done))
+
+        with MockViewerPublisher(f"ws://127.0.0.1:{_TEST_PORT}/ws") as pub:
+            pub.send_click(0.0, 0.0, 0.0, "/robot/base", timestamp_ms=2000)
+            pub.flush()
+
+        done.wait(timeout=2.0)
+        mod.stop()
```
paul-nechifor: You're stopping without try-finally. This is why I think fixtures are much better for cleanup.

dimos/visualization/rerun/test_websocket_server.py
```diff
@@ -0,0 +1,480 @@
...
+        done = threading.Event()
+        mod.clicked_point.subscribe(_collect(received, done))
+
+        with MockViewerPublisher(f"ws://127.0.0.1:{_TEST_PORT}/ws") as pub:
+            pub.send_click(0.0, 0.0, 0.0, "", timestamp_ms=5000)
+            pub.flush()
+
+        done.wait(timeout=2.0)
+        mod.stop()
+        assert received and abs(received[0].ts - 5.0) < 1e-6
```
paul-nechifor: Why two asserts in the same line? If I'm not mistaken, it complicates pytest diff when the assertion fails.

dimos/visualization/rerun/test_websocket_server.py
```diff
@@ -0,0 +1,480 @@
...
+
+        received: list[Any] = []
+        all_arrived = threading.Event()
+
+        def _cb(pt: Any) -> None:
+            received.append(pt)
+            if len(received) >= 3:
+                all_arrived.set()
+
+        mod.clicked_point.subscribe(_cb)
```
paul-nechifor: Don't forget the unsubscribe.

dimos/visualization/rerun/test_websocket_server.py
```diff
@@ -0,0 +1,480 @@
...
+        """Heartbeat messages must not trigger a clicked_point publish."""
+        mod = _make_module()
+        mod.start()
+        _wait_for_server(_TEST_PORT)
+
+        clicks: list[Any] = []
+        twists: list[Any] = []
+        twist_done = threading.Event()
+        mod.clicked_point.subscribe(clicks.append)
+        mod.tele_cmd_vel.subscribe(_collect(twists, twist_done))
```
paul-nechifor: Please, please, please, always clean up in tests.

dimos/visualization/rerun/test_websocket_server.py
```diff
@@ -0,0 +1,480 @@
...
+
+    def test_click_sets_frame_id_from_entity_path(self) -> None:
+        """entity_path is stored as frame_id on the published PointStamped."""
+        mod = _make_module()
+        mod.start()
+        _wait_for_server(_TEST_PORT)
+
+        received: list[Any] = []
+        done = threading.Event()
+        mod.clicked_point.subscribe(_collect(received, done))
```
paul-nechifor: I thought you said you preferred module over mod.

dimos/visualization/rerun/test_websocket_server.py
```diff
@@ -0,0 +1,480 @@
...
+
+        with MockViewerPublisher(f"ws://127.0.0.1:{_TEST_PORT}/ws") as pub:
+            pub.send_twist(0.5, 0.0, 0.0, 0.0, 0.0, 0.8)
+            pub.flush()
+
+        twist_done.wait(timeout=2.0)
+        mod.stop()
+        assert clicks == []
+
+    def test_stop_does_not_publish_clicked_point(self) -> None:
```
paul-nechifor: I don't quite understand what these tests are for. You're checking that stopping doesn't send click... but why would it? It's like trying to prove a negative.

If you do it like this, shouldn't you also check that clicking doesn't send stop, that stops don't send heartbeats, etc.

dimos/visualization/rerun/test_viewer_ws_e2e.py
```diff
@@ -0,0 +1,328 @@
...
+        server = _make_server()
+        server.start()
+        _wait_for_server(_E2E_PORT)
+
+        received: list[Any] = []
+
+        def _on_pt(pt: Any) -> None:
+            received.append(pt)
+
+        server.clicked_point.subscribe(_on_pt)
```
paul-nechifor: unsub

dimos/visualization/rerun/test_viewer_ws_e2e.py
```diff
@@ -0,0 +1,328 @@
...
+
+        # Start dimos-viewer in --connect mode, pointing it at a non-existent gRPC
+        # proxy (it will fail to stream data, but that's fine) and at our WS server.
+        # Use DISPLAY="" to prevent it from opening a window (it will exit quickly
+        # without a display, but the WebSocket connection happens before the GUI loop).
+        proc = subprocess.Popen(
```
paul-nechifor, line 163: please always put subprocesses in fixtures so you always shut them down and not leak them.

dimos/visualization/rerun/websocket_server.py
```diff
@@ -0,0 +1,275 @@
...
+            msg = json.loads(raw)
+        except json.JSONDecodeError:
+            logger.warning(f"RerunWebSocketServer: ignoring non-JSON message: {raw!r}")
+            return
+
+        if not isinstance(msg, dict):
+            logger.warning(f"RerunWebSocketServer: expected JSON object, got {type(msg).__name__}")
+            return
+
+        msg_type = msg.get("type")
```
paul-nechifor, line 214: It would be nice to have a discriminated union of TypedDicts so we can type messages we send. Something like:

```python
class ClickMsg(TypedDict):
    type: Literal["click"]
    x: float
    ...

class TwistMsg(TypedDict):
     type: Literal["twist"]
    ...

ClientMsg = Union[ClickMsg, TwistMsg, ...]

File: dimos/navigation/cmd_vel_mux.py [hunk 20]
```diff
@@ -0,0 +1,166 @@
...
+        self._timer = None
+        self._timer_gen = 0
+
+    def __del__(self) -> None:
+        # Cancel any pending cooldown timer so the daemon thread doesn't
+        # outlive the mux and trip pytest's thread-leak detector.
+        timer = getattr(self, "_timer", None)
+        if timer is not None:
+            timer.cancel()
+            timer.join(timeout=DEFAULT_THREAD_JOIN_TIMEOUT)
```
paul-nechifor: Why is this needed?

dimos/visualization/rerun/test_websocket_server.py
jeff-hykin: I do, but apologies for when I miss when claude does

dimos/navigation/cmd_vel_mux.py
```diff
@@ -0,0 +1,166 @@
...
+                # rather than accumulating under rapid teleop (50 Hz) and
+                # tripping pytest's thread-leak detector.
+                self._timer.cancel()
+                self._timer.join(timeout=DEFAULT_THREAD_JOIN_TIMEOUT)
+            self._timer_gen += 1
+            my_gen = self._timer_gen
+            # weakref prevents the Timer thread from keeping the mux alive
+            # via a bound-method reference — otherwise mux.__del__ can't
+            # run at test scope exit.
+            self_ref = weakref.ref(self)
```
paul-nechifor: But why do you need `__del__`? This is a module, it will be cleaned up by the module coordinator with `.stop()`. You don't need `__del__` and therefore you don't need `weakref`.

dimos/navigation/cmd_vel_mux.py
```diff
@@ -0,0 +1,166 @@
...
+
+            def _end() -> None:
+                obj = self_ref()
+                if obj is not None:
+                    obj._end_teleop(my_gen)
+
+            self._timer = threading.Timer(self.config.tele_cooldown_sec, _end)
+            self._timer.daemon = True
+            self._timer.start()
```
paul-nechifor: Timer is a subclass of Thread. You're spawning a new thread for every message. That does not seem like a good idea. Teleop is at what, 50Hz? That does not seem reasonable. We really need to focus on allow easy use of async code on modules... I'll try to take this soon.

dimos/navigation/test_cmd_vel_mux.py
```diff
@@ -0,0 +1,130 @@
...
+from dimos.constants import DEFAULT_THREAD_JOIN_TIMEOUT
+from dimos.msgs.geometry_msgs.Twist import Twist
+from dimos.msgs.geometry_msgs.Vector3 import Vector3
+from dimos.navigation.cmd_vel_mux import CmdVelMux, CmdVelMuxConfig
+
+
+def _make_mux(cooldown: float = 0.1) -> Any:
+    """Build a CmdVelMux with mocked output streams. __del__ cleans up the timer."""
+    with patch.object(CmdVelMux, "__init__", lambda self: None):
+        mux = cast("Any", CmdVelMux.__new__(CmdVelMux))
```
paul-nechifor: This is a very bad Claude pattern:

```
with ... X ... __init__:
    X.__new__
```

dimos/navigation/test_cmd_vel_mux.py
```diff
@@ -0,0 +1,130 @@
...
+    mux.config = CmdVelMuxConfig(tele_cooldown_sec=cooldown)
+    mux._teleop_active = False
+    mux._lock = threading.Lock()
+    mux._timer = None
+    mux._timer_gen = 0
```
paul-nechifor: Please create an actual module, do not duplicate in test code what __init__ is doing.

dimos/visualization/rerun/test_websocket_server.py
jeff-hykin: I do, but I don't always catch it when claude uses it. My bad

dimos/navigation/test_cmd_vel_mux.py
```diff
@@ -0,0 +1,130 @@
...
+
+        mux._end_teleop(stale_gen)
+        assert mux._teleop_active  # still active
+        assert mux._timer is current_timer  # current timer untouched
+
+
+class TestConfigDefaults:
+    def test_cooldown_default(self) -> None:
+        config = CmdVelMuxConfig()
+        assert config.tele_cooldown_sec == 1.0
```
paul-nechifor:
```suggestion
```

dimos/robot/cli/dimos.py
```diff
@@ -660,17 +664,47 @@ def send(
...
-    """Launch the Rerun visualization bridge."""
-    from dimos.visualization.rerun.bridge import run_bridge
+    """Launch the Rerun visualization bridge.
+
+    Standalone utility: runs the bridge directly in the main process (no
+    blueprint / worker pool) so users can attach a viewer to existing LCM
+    traffic without building a full module graph.
+    """
+    import signal
+    from typing import cast, get_args
```
paul-nechifor: Move to top.

dimos/robot/cli/dimos.py
```diff
@@ -660,17 +664,47 @@ def send(
...
+
+    from dimos.protocol.pubsub.impl.lcmpubsub import LCM
+    from dimos.protocol.service.lcmservice import autoconf
+    from dimos.visualization.constants import RerunOpenOption
+    from dimos.visualization.rerun.bridge import RerunBridgeModule
```
paul-nechifor: I think some of these should be at the top too. E.g. constants.

dimos/utils/generic.py
```diff
@@ -20,6 +20,24 @@
...
+
+def get_local_ips() -> list[tuple[str, str]]:
+    """Return ``(ip, interface_name)`` for every non-loopback IPv4 address.
+
+    Picks up physical, virtual, and VPN interfaces (including Tailscale).
+    """
+    import socket
+
+    import psutil
+
```
paul-nechifor: Top

dimos/visualization/rerun/bridge.py
```diff
@@ -290,43 +295,97 @@ def _on_message(self, msg: Any, topic: Any) -> None:
 
     @rpc
     def start(self) -> None:
+        import socket
+        from urllib.parse import urlparse
```
paul-nechifor: top

dimos/robot/unitree/go2/blueprints/basic/unitree_go2_webrtc_keyboard_teleop.py
```diff
@@ -31,6 +31,10 @@
 unitree_go2_webrtc_keyboard_teleop = autoconnect(
     unitree_go2_coordinator,
     KeyboardTeleop.blueprint(),
+).remappings(
+    [
+        (KeyboardTeleop, "tele_cmd_vel", "cmd_vel"),
```
paul-nechifor, line 36: This isn't needed, right? CmdVelMux should do the translation I think. That should be moved to the base.

dimos/visualization/rerun/test_viewer_ws_e2e.py
```diff
@@ -0,0 +1,328 @@
...
+        _wait_for_server(_E2E_PORT)
+
+        received: list[Any] = []
+        done = threading.Event()
+
+        def _on_pt(pt: Any) -> None:
+            received.append(pt)
+            done.set()
+
+        server.clicked_point.subscribe(_on_pt)
```
paul-nechifor: unsub

dimos/visualization/rerun/test_viewer_ws_e2e.py
```diff
@@ -0,0 +1,328 @@
...
+from dimos.visualization.rerun.websocket_server import RerunWebSocketServer
+
+_E2E_PORT = 13032
+
+
+def _make_server(port: int = _E2E_PORT) -> RerunWebSocketServer:
+    return RerunWebSocketServer(port=port)
+
+
+def _wait_for_server(port: int, timeout: float = 5.0) -> None:
```
paul-nechifor: You already added this in dimos/visualization/rerun/test_websocket_server.py and I'm pretty sure I've seen it in the codebase before under other names.

dimos/visualization/rerun/test_viewer_ws_e2e.py
```diff
@@ -0,0 +1,328 @@
...
+        try:
+            asyncio.run(_probe())
+            return
+        except Exception:
+            time.sleep(0.05)
+    raise TimeoutError(f"Server on port {port} did not become ready within {timeout}s")
+
+
+def _send_messages(port: int, messages: list[dict[str, Any]], *, delay: float = 0.05) -> None:
+    import websockets.asyncio.client as ws_client
```
paul-nechifor: Top

dimos/visualization/rerun/test_viewer_ws_e2e.py
```diff
@@ -0,0 +1,328 @@
...
+                    "y": 20.0,
+                    "z": 0.5,
+                    "entity_path": "/world/robot",
+                    "timestamp_ms": 42000,
+                }
+            ],
+        )
+
+        done.wait(timeout=3.0)
+        server.stop()
```
paul-nechifor: Need try-finally or fixture

dimos/visualization/rerun/constants.py
leshy: rerun bridge.py actually doesn't import rerun until start() is called so this can be imported freely

(at least it isn't supposed to, I see Sam added two imports at the top but those can be guarded by a if TYPE_CHECKING since they are just for typing)

generally if extracting (which is not needed here) I'd rather extract the config into an external file rerun/config.py then start creating constants.py files with dilluted ownership

dimos/visualization/rerun/constants.py
leshy: https://github.com/user-attachments/assets/ccd5f151-603f-4873-a6ac-b27cddf9d2cc

so IDK changes this PR introduces but dimos run doesn't work for me on this branch while dev works fine. This looks like your clicking server listen port, not rerun?



dev doesn't listen on 9090 for me, is there a way to rebind your listen port via some config?

dimos/navigation/cmd_vel_mux.py
jeff-hykin: (Resolving cause I'll pull in MovementManager which replaces this whole class)

dimos/navigation/cmd_vel_mux.py
jeff-hykin: (Resolving cause I'll pull in MovementManager which replaces this whole class)

dimos/visualization/rerun/constants.py
jeff-hykin: If you're saying move `class RerunConfig()` to config.py it makes a circular import:
global_config.py → rerun/config.py → core/module.py (ModuleConfig) → global_config.py

jeff-hykin: I improved the imports, made RERUN_WEB_PORT, but I think I'm gonna leave this this as-is since this was more or less pre-existing and this PR is really starting to expand.
We already need to clean up the terrible rerun_config later so we can clean it up with that.

dimos/robot/unitree/go2/blueprints/basic/unitree_go2_webrtc_keyboard_teleop.py
jeff-hykin: normally MovementManager takes tele_cmd_vel and nav_cmd_vel and outputs cmd_vel, but in this case there's no nav or MovementManager. So we pipe tele_cmd_vel gets direct control over the robot

dimos/visualization/rerun/websocket_server.py
jeff-hykin: the robot moves really slowly in sim but only with tele-op. Sometimes we want to change the speed. Right now its hardcoded on the rerun side. I've moved this scaling to MovementManager though since its job will be to make sure nav_cmd_vel and tele_cmd_vel are at basically the same scale.

dimos/visualization/rerun/websocket_server.py
jeff-hykin: thats a wonderful global_config value! One day hopefully we'll have something global for port mapping/preveting port conflicts too.

dimos/robot/unitree/mujoco_connection.py
```diff
@@ -126,12 +129,23 @@ def start(self) -> None:
...
         # Launch the subprocess
         try:
-            # mjpython must be used macOS (because of launch_passive inside mujoco_process.py)
+            # mjpython must be used on macOS (because of launch_passive inside mujoco_process.py).
+            # It needs libpython on the dylib search path; uv-installed Pythons
+            # use @rpath which doesn't always resolve inside venvs, so we
+            # point DYLD_LIBRARY_PATH at the real libpython directory.
             executable = sys.executable if sys.platform != "darwin" else "mjpython"
+            env = os.environ.copy()
+            if sys.platform == "darwin":
```
jeff-hykin, line 138: problem specific to newer mac's. (this got in the way of me testing this branch)

pyproject.toml
```diff
@@ -82,7 +82,7 @@ dependencies = [
     # TODO: rerun shouldn't be required but rn its in core (there is NO WAY to use dimos without rerun rn)
     # remove this once rerun is optional in core
     "rerun-sdk>=0.20.0",
-    "dimos-viewer>=0.30.0a2",
+    "dimos-viewer==0.30.0a6.dev99",
```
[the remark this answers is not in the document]
jeff-hykin: non-issue, dev99 is on pypi

dimos/visualization/rerun/constants.py
jeff-hykin: I realized whats happening, this would probably be a problem if you try to use the rerun web on dev. The web server was started in the background when native wasn't explicitly set - which I'm not going to debate right now so I just changed that to not start it.

Also the port should be properly configurable now.

dimos/core/global_config.py
```diff
@@ -37,6 +40,8 @@ class GlobalConfig(BaseSettings):
     replay_dir: str = "go2_sf_office"
     new_memory: bool = False
     viewer: ViewerBackend = "rerun"
+    rerun_open: RerunOpenOption = RERUN_OPEN_DEFAULT
+    rerun_web: bool = RERUN_ENABLE_WEB
```
paul-nechifor, line 44: It's kinda odd to have both a constant and a global config value? Which one are people supposed to use?

But what is this for? Isn't it controlled by passing `--viewer=rerun-web`?

dimos/robot/cli/dimos.py
```diff
@@ -660,17 +665,45 @@ def send(
...
 ) -> None:
-    """Launch the Rerun visualization bridge."""
-    from dimos.visualization.rerun.bridge import run_bridge
+    """Launch the Rerun visualization bridge.
+
+    Standalone utility: runs the bridge directly in the main process (no
+    blueprint / worker pool) so users can attach a viewer to existing LCM
+    traffic without building a full module graph.
+    """
+    import signal
```
paul-nechifor: Should be moved to the top.

dimos/visualization/rerun/bridge.py
```diff
@@ -156,47 +165,33 @@ def _default_blueprint() -> Blueprint:
...
 
 
+def _rebuild_config() -> None:
+    from rerun._baseclasses import Archetype
+    from rerun.blueprint import Blueprint
+
+    Config.model_rebuild(_types_namespace={"Archetype": Archetype, "Blueprint": Blueprint})
+
+
+_rebuild_config()
```
paul-nechifor: If this is called at import level it doesn't make sense for the imports to be inside the function.

dimos/visualization/rerun/bridge.py
```diff
@@ -237,10 +241,13 @@ def _visual_override_for_entity_path(
 
         # None means "suppress this topic entirely"
         if any(fn is None for fn in matches):
-            return lambda msg: None
+            result: Callable[[Any], RerunData | None] = lambda msg: None  # noqa: E731
+            self._override_cache[entity_path] = result
```
paul-nechifor: The linter is telling you you should use:

```suggestion
            def result(msg: Any) -> RerunData | None:
                return None
            self._override_cache[entity_path] = result
```

It's actually shorter than using Callable. Same for the other `# noqa`s.

dimos/visualization/rerun/bridge.py
```diff
@@ -290,65 +296,135 @@ def _on_message(self, msg: Any, topic: Any) -> None:
...
-        # Add pubsub stop as disposable
         for pubsub in self.config.pubsubs:
             if hasattr(pubsub, "stop"):
                 self.register_disposable(Disposable(pubsub.stop))  # type: ignore[union-attr]
 
         self._log_static()
 
+    def _log_connect_hints(self, grpc_port: int) -> None:
+        """Log CLI commands for connecting a viewer to this bridge."""
+        from dimos.utils.generic import get_local_ips
```
paul-nechifor: Why not add to conventions: "Don't use imports inside functions unless they're needed for performance reasons or to resolve circular imports." Then you can ask Claude to move all of these.

dimos/core/global_config.py
jeff-hykin: Please I just need to get nav merged 😭, a constant + global-config already existed before this PR

dimos/visualization/rerun/bridge.py
jeff-hykin: I actually did in a pass yesterday (I also added to my local claude config) it just didn't catch all of them.

Can we make these code-cosmetics changes after its merged? There's files on dev that aren't even formatted correctly.

dimos/navigation/smart_nav/modules/movement_manager/movement_manager.py
```diff
@@ -0,0 +1,132 @@
...
+        self._lock = threading.Lock()
+        self._teleop_active = False
+        self._last_teleop_time = 0.0
+
+    @rpc
+    def start(self) -> None:
+        super().start()
+        self.clicked_point.subscribe(self._on_click)
+        self.nav_cmd_vel.subscribe(self._on_nav)
+        self.tele_cmd_vel.subscribe(self._on_teleop)
```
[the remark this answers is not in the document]
jeff-hykin: good find

dimos/visualization/rerun/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+import asyncio
+import time
+
+
+def wait_for_server(port: int, timeout: float = 5.0) -> None:
```
paul-nechifor: You're not supposed to import from conftest.py. I don't remember the reason now, but it breaks pytest in subtle ways. You're only supposed to use it for fixtures (and other configs). So, define a fixture called `wait_for_server` and return this function from it.

dimos/visualization/rerun/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+    """Block until the WebSocket server on *port* accepts a connection."""
+    import websockets.asyncio.client as ws_client
```
paul-nechifor: Top.

dimos/visualization/rerun/test_viewer_ws_e2e.py
```diff
@@ -0,0 +1,207 @@
...
+
+_E2E_PORT = 13032
+
+
+@pytest.fixture()
+def server() -> RerunWebSocketServer:
+    module = RerunWebSocketServer(port=_E2E_PORT)
+    module.start()
+    wait_for_server(_E2E_PORT)
+    yield module  # type: ignore[misc]
```
paul-nechifor, line 40: No need for ignore, we don't type check tests.

dimos/visualization/rerun/test_viewer_ws_e2e.py
paul-nechifor: You need to move it to to the top.

dimos/visualization/rerun/test_viewer_ws_e2e.py
```diff
@@ -0,0 +1,207 @@
...
+        try:
+            proc.wait(timeout=3)
+        except subprocess.TimeoutExpired:
+            proc.kill()
+
+    @pytest.mark.skipif(
+        shutil.which("dimos-viewer") is None
+        or "--connect"
+        not in subprocess.run(["dimos-viewer", "--help"], capture_output=True, text=True).stdout,
+        reason="dimos-viewer binary not installed or does not support --connect",
```
paul-nechifor: Why skip? If `dimos-viewer` is not present or doesn't support `--connect` that sounds to me like we should know that. I.e., the test should fail.

dimos/visualization/rerun/test_websocket_server.py
paul-nechifor: This is not resolved.

dimos/visualization/rerun/test_websocket_server.py
```diff
@@ -0,0 +1,213 @@
...
+    module.stop()
+
+
+@pytest.fixture()
+def publisher(server: RerunWebSocketServer) -> MockViewerPublisher:
+    with MockViewerPublisher(f"ws://127.0.0.1:{_TEST_PORT}/ws") as publisher:
+        yield publisher  # type: ignore[misc]
+
+
+# ── Tests ────────────────────────────────────────────────────────────────
```
paul-nechifor, line 116:
```suggestion
```

dimos/visualization/rerun/test_websocket_server.py
```diff
@@ -0,0 +1,213 @@
...
+    done.wait(timeout=2.0)
+    unsub()
+
+    assert len(received) == 1
+    assert received[0].is_zero()
+
+
+def test_invalid_json_does_not_crash(server: RerunWebSocketServer) -> None:
+    """Malformed JSON is silently dropped; server stays alive for the next message."""
+    import websockets.asyncio.client as ws_client
```
paul-nechifor: Please move to the top.

dimos/navigation/smart_nav/modules/movement_manager/test_movement_manager.py
```diff
@@ -0,0 +1,117 @@
...
+    return PointStamped(ts=time.time(), frame_id="map", x=x, y=y, z=z)
+
+
+def test_teleop_suppresses_nav_and_cancels_goal(manager: MovementManager) -> None:
+    """Teleop arriving should suppress nav, publish stop_movement, and cancel the goal with NaN."""
+    manager.config.tele_cooldown_sec = 10.0
+    manager._on_teleop(_twist(lx=0.3))
+
+    # Nav is suppressed
+    manager.cmd_vel.publish.reset_mock()  # type: ignore[union-attr]
```
paul-nechifor, line 59: None of these type ignores are needed. We don't type check tests.

dimos/navigation/smart_nav/modules/movement_manager/movement_manager.py
```diff
@@ -0,0 +1,133 @@
...
+"""MovementManager: click-to-goal relay + teleop/nav velocity mux."""
+
+from __future__ import annotations
+
+import math
+import threading
+import time
+from typing import Any
+
+from dimos_lcm.std_msgs import Bool  # type: ignore[import-untyped]
```
paul-nechifor, line 24: Why is this ignore needed? I can see the import in other files without any ignore.

dimos/navigation/smart_nav/modules/movement_manager/movement_manager.py
```diff
@@ -0,0 +1,133 @@
...
+        self.stop_movement.publish(Bool(data=True))
+        # NOTE: this NaN goal is more of a safety fallback.
+        # It can be REALLY bad if a robot is supposed to stop moving but wont
+        # we should probably think a more robust/strict requirement on planners
+        cancel = PointStamped(
+            ts=time.time(), frame_id="map", x=float("nan"), y=float("nan"), z=float("nan")
+        )
+        self.way_point.publish(cancel)
+        self.goal.publish(cancel)
+        logger.debug("Navigation cancelled — waiting for new goal")
```
paul-nechifor, line 98:
```suggestion
```

dimos/visualization/rerun/websocket_server.py
paul-nechifor: Changing the speed makes sense, but why per axis? Wouldn't `linear_multiplier` and `angular_multiplier` make more sense. I don't understand why you'd need need faster movement in the X axis than the Y axis for example.

dimos/navigation/smart_nav/modules/movement_manager/movement_manager.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def _on_nav(self, msg: Twist) -> None:
+        with self._lock:
+            if self._teleop_active:
+                # check if cooldown has expired
+                elapsed = time.monotonic() - self._last_teleop_time
+                if elapsed < self.config.tele_cooldown_sec:
+                    return
+                self._teleop_active = False
+        self.cmd_vel.publish(msg)
```
paul-nechifor, line 108: Since you write to cmd_vel without a lock, order is not guaranteed for nav messages and teleop messages. For example, this function can run up till this line, determines that it should publish, but pauses before it does so. The second thread runs _on_teleop, determines to publish the teleop message to cmd_vel and does so. This first thread wakes up and publishes the nav message to cmd_vel even though `self._teleop_active` is True.

* One way to deal with it is to put everything behind `with self._lock`.
* Another way to untangle the threading issue is to publish messages to a queue and only have one thread process all these disparate sources.
* Another would be to use `self._loop.call_soon_threadsafe(self._handle_nav, msg)`. You publish all messages onto the asyncio loop as calls to async methods. They'll be on the same thread, so no need for locks.

paul-nechifor (dismissed):

dimos/visualization/rerun/websocket_server.py
jeff-hykin: I started with a single scalar value only applied to x-y then my ivan-bot-review flagged that and said it should be generic an cover all the aspects that could be scaled. So then I made it just a twist that scales.

dimos/navigation/smart_nav/modules/movement_manager/movement_manager.py
jeff-hykin: Movement manager is gonna get nuked when mustafa gets control coordinator merged in. So the code just needs to work for like 2 weeks.

paul-nechifor (approved):

PR #1786, opened 2026-04-13
ci(macos): add macos_bug marker and skip known-crashing worker tests

### LINKED ISSUE #1752: Ignore broken macos tests
looks a bit confusing to have fails in CI for all PRs

let's create a new pytest tag macos_bug or something, and just skip those tests in CI on mac, will make it easy to see where issues on mac are as well

ruthwikdasyam:
## Problem

macOS CI fails on nearly every PR because `dimos/core/coordination/test_worker.py` hits a native segfault (exit 139) partway through execution - the crash surfaces during `test_load_balancing_distributes_modules` but is cumulative state from prior tests in the same file. 
Since the process dies, pytest reports no `FAILED` tests, just a red job, which trains the team to ignore macOS CI entirely

Closes #1752 

## Solution

- Introduces a `macos_bug` pytest marker convention for tests known to fail or crash on the macOS runner
- Updates the macOS workflow's pytest filter to exclude `macos_bug` alongside the existing `tool` and `mujoco` exclusions.

Linux CI is unaffected.

None

## How to Test

- macOS check should now get past `test_worker.py` instead of segfaulting
- Linux CI (`ci.yml`) should continue to run the full suite including `test_worker.py`, since the marker is only excluded in `macos.yml`
- To verify locally on macOS
   ```bash
   python -m pytest --durations=10 -m 'not (tool or mujoco or macos_bug)' --timeout=120 dimos/

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -98,7 +98,7 @@ jobs:
           CI: "1"
         run: |
           source .venv/bin/activate
-          python -m pytest --durations=10 -m 'not (tool or mujoco or macos_bug)' --timeout=120 dimos/
+          python -m pytest --durations=10 -m 'not (tool or mujoco)' --timeout=120 dimos/
 
       - name: Check disk usage (post-test)
         if: always()
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -74,6 +74,7 @@ jobs:
       - name: Build C++ extensions
         run: |
           source .venv/bin/activate
+          uv pip install pybind11
           python setup.py build_ext --inplace
 
       - name: Check disk usage
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -93,7 +93,7 @@ jobs:
           CI: "1"
         run: |
           source .venv/bin/activate
-          python -m pytest --durations=10 -m 'not (tool or mujoco)' --timeout=120 dimos/
+          python -m pytest --durations=10 -m 'not (tool or mujoco or macos_bug)' --timeout=120 dimos/
 
       - name: Check disk usage (post-test)
         if: always()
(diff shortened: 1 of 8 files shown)


.github/workflows/macos.yml
```diff
@@ -93,7 +93,7 @@ jobs:
           CI: "1"
         run: |
           source .venv/bin/activate
-          python -m pytest --durations=10 -m 'not (tool or mujoco)' --timeout=120 dimos/
+          python -m pytest --durations=10 -m 'not (tool or mujoco or macos_bug)' --timeout=120 dimos/
```
paul-nechifor: This skips the tests in CI, but I think it would be better to add a code marker to skip for all Mac OS runs. I also think it's nicer because pytest always lists tests that are skipped this way at the end of the run.

Add this to `dimos/conftest.py`:

```python
        "skipif_macos_bug": (_is_macos(), "Some tests are buggy on Mac OS"),
```

and mark with `skipif_macos_bug`.

dimos/navigation/replanning_a_star/min_cost_astar.py
```diff
@@ -154,7 +156,10 @@ def min_cost_astar(
                 return None
             return _reconstruct_path_from_coords(path_coords, costmap)
         else:
-            logger.warning("C++ A* module could not be imported. Using Python.")
+            logger.warning(
+                "C++ A* module could not be imported (%s). Using Python.",
+                _CPP_IMPORT_ERROR,
```
paul-nechifor, line 161: If the C++ module doesn't work on the mac os build, it means it wasn't set up correctly.

ruthwikdasyam: yeah, added this to display the issue. sorted the installation, works now.

.github/workflows/macos.yml
ruthwikdasyam: Got it, now skips tests both in mac pc and CI

paul-nechifor (approved):

PR #1787, opened 2026-04-14
Feat/integrate zenoh

leshy:
## Problem

Closes DIM-XXX


dimos/protocol/pubsub/impl/test_zenohpubsub.py
```diff
@@ -0,0 +1,232 @@
...
+    for session in _sessions.values():
+        session.close()
+    _sessions.clear()
+
+
+def wait_for_delivery() -> None:
+    """Wait for published messages to be delivered to subscribers."""
+    import time
+
+    time.sleep(0.1)
```
leshy: don't use sleeps, use thread locks, otherwise this is either too slow or too flakey in CI

leshy:
running
`dimos --transport=zenoh --dtop --replay --replay-dir=unitree_go2_bigoffice run unitree-go2`

it crashes on import error (not below RuntimeError, I found this in code)

```python
            raise RuntimeError(
                "transport='zenoh' but eclipse-zenoh is not installed. "
                "Install with: uv sync --extra zenoh"
            )
```

but executing `uv sync --extra zenoh` uninstalls everything else from my uv env

vrinek:
@leshy I am not very familiar with `uv` and Python's package management yet. To unblock this, and simplify, I can make Zenoh a required dependency. What do you think?

vrinek:
# Benchmarks

I'm excluding the raw bytes transports to compare apples to apples.

```
uv run pytest -svm tool -k "not bytes" dimos/protocol/pubsub/benchmark/test_benchmark.py
```
image

dimos/protocol/pubsub/impl/test_zenohpubsub.py
vrinek: I agree. Zenoh does not provide any tooling to make this easy. I'll rewrite this using `threading.Event`.

vrinek: See https://github.com/dimensionalOS/dimos/pull/1787/changes/e699d04d3d060c8506f6a0272c2bc09e7491e876 for the fix.

vrinek:
Running the benchmarks for longer (50k msgs, 10sec) prints more reliable results. QoS at default of DROP + RELIABLE + DATA.

image

leshy:
> @leshy I am not very familiar with `uv` and Python's package management yet. To unblock this, and simplify, I can make Zenoh a required dependency. What do you think?

let's learn how to use UV, it's our main package manager, this is not the last feature of this type etc @paul-nechifor will know

leshy:
superseded by https://github.com/dimensionalOS/dimos/pull/1906

PR #1789, opened 2026-04-15
Fix mac UDP

Dreamsorcerer:
## Problem

Mac UDP limits are too low. It appears newer releases now support 32 MiB.

## Solution

To continue supporting higher limits in future releases, this code tries to set the actual target threshold and then keeps halving and retrying until it works, thus allowing us to use the maximum available on each system. (The max is a compiled-in constant to the kernel and is not exposed anywhere).

## How to Test

Run sim/replay on Mac OS X.


dimos/protocol/service/test_system_configurator.py
```diff
@@ -352,27 +354,78 @@ def test_check_returns_false_when_values_low(self) -> None:
             assert configurator.check() is False
             assert len(configurator.needs) == 3
 
+    def test_check_uses_saved_config_as_target(self, tmp_path) -> None:
+        conf = tmp_path / "sysctl.json"
+        conf.write_text('{"kern.ipc.maxsockbuf": 32000000}')
```
jeff-hykin: plz put 32000000 in a var

dimos/protocol/service/system_configurator/lcm.py
```diff
@@ -14,17 +14,34 @@
 
 from __future__ import annotations
 
+import json
 import re
 import resource
 import subprocess
 
+from dimos.constants import STATE_DIR
```
jeff-hykin, line 22: cool I didn't know we already had a STATE_DIR

jeff-hykin (changes requested):
32000000 in a var, then it looks good to me

leshy (dismissed):

jeff-hykin (approved):

leshy (approved):

PR #1793, opened 2026-04-16
perf(rerun): render voxel maps as Points3D spheres instead of Boxes3D

### LINKED ISSUE #1631: Rerun throttling is a hack, fix
there is a rate limiter in rerun bridge now

`/dimos/visualization/rerun/bridge.py`

```python

        # Rate-limit heavy data types to prevent viewer memory exhaustion.
        # High-bandwidth streams (e.g. 30fps camera, lidar) would otherwise
        # flood the viewer faster than it can evict, causing OOM.  Light
        # messages (Path, PointStamped, TF, etc.) pass through unthrottled.
        if self.config.min_interval_sec > 0 and isinstance(msg, _HEAVY_MSG_TYPES):
            now = time.monotonic()
            last = self._last_log.get(entity_path, 0.0)
            if now - last < self.config.min_interval_sec:
                return
            self._last_log[entity_path] = now
```

we should absolutely be able to render 30fps camera images so there should be no reason for this, investigate how rerun deals with backpressure, talk to rerun team, read docs, what exactly is the bottleneck here and why?

---
*Synced from* [DIM-725](https://linear.app/dimensional/issue/DIM-725) *by summer*

ruthwikdasyam:
## Problem

endering each voxel as an `rr.Boxes3D` instance is expensive - each box is 12 line segments in the
Rerun line renderer, whereas `rr.Points3D` with a world-space radius is a single billboarded quad per point 

Closes #1631 

## Solution

Flip all `PointCloud2.to_rerun(mode="boxes")` callers to `mode="spheres"`. Same world-space voxel footprint, renders as  spheres instead of cubes, ~10x cheaper at high point counts

- option to switch to `mode=boxes` is still present
- voxel sizes are reduced to 0.05 (from 0.1), as the rerun maintains good fps with spherical voxels
- toggling default set to `0 hz`. Can be adjusted when needed (if, boxes are used)

## Breaking Changes

None

## How to Test

- Any blueprint should launch rerun - with spherical voxels
quick test -> `dimos --replay --replay-dir=unitree_go2_bigoffice run unitree-go2'

spherevoxels

Looks good. Able to maintain good fps (~20) with 400k voxels at 0.05 radius

### DIFF
--- a/dimos/hardware/sensors/lidar/fastlio2/fastlio_blueprints.py
+++ b/dimos/hardware/sensors/lidar/fastlio2/fastlio_blueprints.py
@@ -23,7 +23,7 @@ mid360_fastlio = autoconnect(
     FastLio2.blueprint(voxel_size=voxel_size, map_voxel_size=voxel_size, map_freq=-1),
     RerunBridgeModule.blueprint(
         visual_override={
-            "world/lidar": lambda grid: grid.to_rerun(voxel_size=voxel_size, mode="boxes"),
+            "world/lidar": lambda grid: grid.to_rerun(voxel_size=voxel_size, mode="spheres"),
         }
     ),
 ).global_config(n_workers=2, robot_model="mid360_fastlio2")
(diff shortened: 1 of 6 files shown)


leshy (changes requested):
Hey we can probably just change PointCloud2 to_rerun() function to do this by default? and we can just delete explicit overrides from all blueprints

leshy (approved):

PR #1794, opened 2026-04-16
feat:rust native modules

aclauer:
## Problem

We need support for Rust native modules to offload performance critical modules.

Closes DIM-XXX

## Solution

See native/rust/examples for a Rust native ping pong example

Added NativeModule struct to allow easy publishing and subscribing to topics. NativeModules can use any transport that implements the Transport trait (currently only LCM).

Rust NativeModules also support configuration through stdin instead of cli args. Cli args are kept for backwards compatibility.

## Breaking Changes

None

## How to Test

For unit tests
cd dimos/native/rust
cargo test

For end to end test with two Rust modules talking over LCM
python3 rust_ping_pong.py 


dimos/core/native_module.py
```diff
@@ -168,13 +174,20 @@ def start(self) -> None:
...
         self._process = subprocess.Popen(
             cmd,
             env=env,
             cwd=cwd,
+            stdin=subprocess.PIPE,
             stdout=subprocess.PIPE,
             stderr=subprocess.PIPE,
         )
+        assert self._process.stdin is not None
+        self._process.stdin.write(stdin_blob)
```
leshy: nice, clean, will just mess with existing native modules, can we have a toggle? maybe a class itself can have an attribute on nativemodule version, new version expects stdin blob, old one, args

examples/native-modules/rust/src/native_ping.rs
```diff
@@ -0,0 +1,40 @@
+// NativeModule ping example.
+//
```
leshy, line 2: Can we move this to dimos/examples/ dir? so we that don't create more example dirs

slight issue - dimos/examples/language-interop has examples of totally external scripts talking to dimos, this is doing actual native modules (I assume proper way going forward), so should reformat the examples/ dir somehow to make sense

direct dimos vs native-module dimos interop examples? idk, you can decide. we'll point new devs / users to whatever you write here - so in your interest to make it nice otherwise they'll talk to you :)

leshy (changes requested):
small tweaks

dimos/core/native_module.py
aclauer: Sure i can add a field to the native module config to tell it if it should read from cli args or stdin for the configs. (default to cli args for now, but we should move any existing modules to stdin asap, should be a pretty minor change)

examples/native-modules/rust/src/native_ping.rs
aclauer: Makes sense, I'll move it over. My vision is to standardize the native module sdk across all the languages to match the Rust implementation, and then examples/native-modules can have an example module of each one and show them being coordinated by dimos.

leshy: yeah that's perfect

native/rust/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "dimos-native-module"
+version = "0.1.0"
```
leshy, line 3: I don't actually know if this should be in our main repo and how rust manages packages, but you can own these decisions

leshy (approved):

PR #1875, opened 2026-04-17
native module refinement

jeff-hykin:
Native module refinement (rebuild logic and logging) and keyboard interrupt improvement


dimos/core/native_module.py
```diff
@@ -68,11 +88,6 @@ class MyCppModule(NativeModule):
 logger = setup_logger()
 
 
-class LogFormat(enum.Enum):
```
leshy, line 71: why was this killed?

dimos/core/native_module.py
```diff
@@ -161,132 +197,185 @@ def start(self) -> None:
...
-        """Run ``build_command`` if the executable does not exist."""
+        """Run ``build_command`` when not in PROD mode, or if the executable is missing.
+
+        When ``PROD`` env var is set, skip rebuilding entirely — the executable
+        must already exist.  Otherwise, always invoke ``build_command`` and let
+        nix handle caching/cache-busting.
+        """
         exe = Path(self.config.executable)
-        if exe.exists():
+        is_prod = os.environ.get("PROD")
```
leshy: you are introducing magical env var, please use config system and global config

dimos/core/native_module.py
```diff
@@ -161,132 +197,185 @@ def start(self) -> None:
...
+
+        if is_prod:
+            if not exe.exists():
```
leshy: we do want autobuild to work in prod on the first run, in case of nix this will let us distribute binaries but intelligently locally build when needed (just like pip does for py libs)

dimos/core/native_module.py
```diff
@@ -161,132 +197,185 @@ def start(self) -> None:
...
+                raise FileNotFoundError(
+                    f"[{self._mod_label}] Executable not found: {exe}. "
+                    "Set build_command in config to auto-build, or build it manually."
+                )
+            return
+
         logger.info(
-            "Executable not found, running build",
+            "Building native module",
```
leshy, line 357: We shouldn't by default run build system change detection (even to validate file changes) for all native modules all of the time in dev, this is expensive, toggle this per module if you are developing it. if you have 5 native modules, we don't want to run 5 nix instances every time someone is running your blueprint in dev

dimos/core/native_module.py
leshy, line 71: why was this killed?

dimos/core/native_module.py
```diff
@@ -60,6 +59,27 @@ class MyCppModule(NativeModule):
 from dimos.core.module import Module, ModuleConfig
 from dimos.utils.logging_config import setup_logger
 
+# ctypes is only needed for the Linux child-preexec helper below.  Hoisting
+# the import out of the inner function avoids re-importing on every start().
+if sys.platform.startswith("linux"):
+    import ctypes
+
+    _LIBC = ctypes.CDLL("libc.so.6", use_errno=True)
```
paul-nechifor:
```suggestion
    from ctypes.util import find_library
    _LIBC = ctypes.CDLL(find_library("c"), use_errno=True)
```

leshy (changes requested):
some changes

dimos/core/native_module.py
```diff
@@ -60,6 +59,27 @@ class MyCppModule(NativeModule):
 from dimos.core.module import Module, ModuleConfig
 from dimos.utils.logging_config import setup_logger
 
+# ctypes is only needed for the Linux child-preexec helper below.  Hoisting
```
paul-nechifor: I assume this is LLM generate code because it adds a lot of unnecessary explanations which obscures that it's actually needed for. This could just as easily be:

```python
from ctypes.util import find_library
import ctypes

def set_process_to_die_when_parent_dies() -> None:
    if not sys.platform.startswith("linux"):
        return
    PR_SET_PDEATHSIG = 1
    libc = ctypes.CDLL(find_library("c"), use_errno=True)
    if libc.prctl(PR_SET_PDEATHSIG, signal.SIGTERM) != 0:
        err = ctypes.get_errno()
        raise OSError(err, f"prctl(PR_SET_PDEATHSIG) failed: {os.strerror(err)}")


...

self._process = subprocess.Popen(
            ...
            preexec_fn=set_process_to_die_when_parent_dies,
        )
```

dimos/core/native_module.py
jeff-hykin: The comments on the PR I thought were saying to simplify the stdout/err and AFAIK it wasn't used. 

If we're going to have JSON vs normal logging shouldn't that be for all modules? It doesn't make sense to me that that would be native module specific

leshy: it's not native module specific, our py logging system supports rich metadata (dictionaries) attached to each log line, for simplicity for native modules we support just STDOUT/STDERR parsing into strings, but rich json log is our dimos wide default

dimos/core/native_module.py
jeff-hykin: bruh I feel like I can't win here lol. on the last pr you said "1000ms isn't a big deal on dev" and I asked in #core if this logic was okay, and paul said yes and you saw it and didn't respond (AFAIK). 

all these builds/checks run in parallel, basically nobody is going to notice if its 1 module or 5.

paul-nechifor: This is my understanding as well (this implementation is fine).

dimos/core/native_module.py
```diff
@@ -161,132 +197,185 @@ def start(self) -> None:
...
+                    pid=proc.pid,
                 )
-                self._process.kill()
-                self._process.wait(timeout=5)
-        if self._watchdog is not None and self._watchdog is not threading.current_thread():
-            self._watchdog.join(timeout=DEFAULT_THREAD_JOIN_TIMEOUT)
-        self._watchdog = None
-        self._process = None
+                proc.kill()
+                proc.wait(timeout=self.config.shutdown_timeout)
```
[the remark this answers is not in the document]
jeff-hykin: fixed

dimos/core/native_module.py
```diff
@@ -161,132 +197,185 @@ def start(self) -> None:
...
         logger.info(
-            "Executable not found, running build",
+            "Building native module",
             executable=str(exe),
             build_command=self.config.build_command,
         )
+        build_start = time.perf_counter()
         proc = subprocess.Popen(
             self.config.build_command,
             shell=True,
```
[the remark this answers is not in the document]
jeff-hykin: non-issue

dimos/core/native_module.py
jeff-hykin: Restored and added https://github.com/dimensionalOS/dimos/issues/1904 since normal modules don't have a toggle, and ideally native modules wouldn't have a toggle too.

dimos/core/native_module.py
jeff-hykin: Edit: make sure to bulid on prod if missing

dimos/core/native_module.py
jeff-hykin: Yeah, we're on the same page with comments 👍  although, for the nav PR, I'll pre-apologize as I have not purged the junk comments/docstrings from it yet

paul-nechifor: BTW this should no longer be needed as I've added a watchdog process to kill orphans: https://github.com/dimensionalOS/dimos/pull/1886

dimos/core/global_config.py
```diff
@@ -52,6 +52,8 @@ class GlobalConfig(BaseSettings):
     nerf_speed: float = 1.0
     planner_robot_speed: float | None = None
     mcp_port: int = 9990
+    mcp_host: str = "127.0.0.1"
```
jeff-hykin: `mcp_host` needs to be removed

PR #1877, opened 2026-04-17
Make whisper an optional extra with faster-whisper by default

Dreamsorcerer:
## Problem

Whisper requires downloading a 150MB model and depends on torch (with GBs of CUDA downloads).

## Solution

Provide faster-whisper by default (2MB) and use as a fallback when whisper is not available.
This avoids the 150MB download, and means we are one step closer to not depending on torch for a base install.

## Breaking Changes

Users need to request `dimos[whisper]` now for full whisper feature.

## Test

```
python -c "
from dimos.stream.audio.pipelines import stt
node = stt()
node.emit_text().subscribe(on_next=lambda t: print(f"USER: {t}"))
from dimos.stream.audio.utils import keepalive
keepalive()
"
```


Dreamsorcerer:
TTS seems to work pretty well with faster-whisper anyway.

pyproject.toml
```diff
@@ -169,8 +169,13 @@ agents = [
 
     # Audio
     "openai",
-    "openai-whisper",
     "sounddevice",
+    "faster-whisper>=1.0.0",
+]
+
+whisper = [
+    "dimos[agents]",
```
paul-nechifor: Why include dimos[agents] here?

Dreamsorcerer: Does it make sense to install without agents? I was just assuming that we'd want all the other dependencies here.

pyproject.toml
```diff
@@ -169,8 +169,13 @@ agents = [
...
+    "openai-whisper",
 ]
```
Dreamsorcerer:
```suggestion
```

dimos/stream/audio/stt/node_whisper.py
```diff
@@ -27,10 +27,32 @@
...
+
+    _USE_FASTER_WHISPER = False
+except ImportError:
+    try:
+        from faster_whisper import WhisperModel  # type: ignore[import-untyped]
+
+        logger.warn(
+            "openai-whisper not installed, falling back to faster-whisper. "
+            "Install openai-whisper for the full backend: pip install openai-whisper",
+            stacklevel=2,
```
Dreamsorcerer:
```suggestion
```

dimos/stream/audio/stt/node_whisper.py
```diff
@@ -14,9 +14,9 @@
 # limitations under the License.
 
 from typing import Any
+import warnings
```
Dreamsorcerer:
```suggestion
```

paul-nechifor (approved):

PR #1879, opened 2026-04-18
feat(arduino): add ArduinoModule with full arduino sim (and real arduino) support

jeff-hykin:
# EXPERIMENTAL / TOY / FUN

## Problem

We don't have arduino support :( 

## Solution

Arduino support. Just like a native module, there's an Arduino module, but you give it an .ino file instead of a C++ file. 

- Includes: Full Arduino simulator (all via nix of course)
- ### How does it work?
    - put the entire arduino toolchain in a nix flake
    - 1. we build the .ino file for an Arduino based on python config values (which one to target)
    - 2. we try to push the payload to the Arduino (auto scan ports and use python config)
    - 3. we (hidden) build a C++ bridge module (very very very similar to a native module)
    - 4. we start the C++ native module (as normal)
    - 5. messages passing
        - So ... LCM has C support ... but it uses malloc, which the arduino doesn't (really) support
        - solution: we make a crappier version of most LCM messages, but they are fixed size and dont use malloc for encode/decode
        - So: an LCM message arrives at the C++ module (normal native module stuff)
        - it encodes the message into this crappier LCM version (it strips the 64 bit hash, which is sketchy I know but this is experimental. We should add a checksum later.)
        - that data is sent over serial to the arduino
        - the arduino code uses a dimos-provided header
            - the header patches the serial.print (because we overloaded the serial line for data and printing)
        - the message arrives on the arduino program through serial, it gets parsed, float64's are read as float32 (because Arduino doesn't support float64)
        - the Arduino does whatever it wants
        - the Arduino lcm_crappy encodes a message and publishes it on serial
        - the C++ reads the serial message, creates the 64bit header for the message, then publishes that as-is to the rest of the dimos system
- Tested on real Arduino hardware

```mermaid
%%{init: {'theme':'base', 'themeVariables': {'fontSize': '40px', 'fontFamily': 'Inter, sans-serif'}}}%%
flowchart LR
    subgraph Dimos["🖥️ dimos"]
        direction TB
        SrcModule["Some dimos module"]
        DstModule["Some other dimos module"]
        Terminal["📟 Terminalprintouts"]
    end

    subgraph Host["C++ Native Module"]
        direction TB
        Encode["Strip 64-bit LCM checksum"]
        Decode["Recreate 64-bit LCM checksum"]
        PrintHandler["Print handlerforward to terminal"]
    end

    subgraph Arduino["Arduino"]
        direction TB
        Parse["read serial, lcm_crappy decode (fixed size only)"]
        Logic["Arduino does its thing"]
        Emit["lcm_crappy encode and publish"]
        Printf["Serial.printpatched by dimos header"]
        Parse --> Logic --> Emit
        Logic -.-> Printf
    end

    SrcModule -->|"LCM message"| Encode
    Encode -->|"serial (data)"| Parse
    Emit -->|"serial (data)"| Decode
    Decode -->|"LCM message"| DstModule

    Printf -->|"serial (printout)"| PrintHandler
    PrintHandler -->|"stdout"| Terminal

    classDef dimos fill:#ede9fe,stroke:#6d28d9,stroke-width:2px,color:#1e1b4b
    classDef host fill:#d1fae5,stroke:#047857,stroke-width:2px,color:#064e3b
    classDef arduino fill:#1e3a8a,stroke:#1e40af,stroke-width:2px,color:#ffffff
    classDef terminal fill:#fef3c7,stroke:#b45309,stroke-width:2px,color:#78350f

    class Dimos,SrcModule,DstModule dimos
    class Terminal terminal
    class Host,Encode,Decode,PrintHandler host
    class Arduino,Parse,Logic,Emit,Printf arduino
```

## ~~Breaking Changes~~

# Caveats
- A lot
- We should test if float64 round trip doesn't mangle the data (is 64->32 is handled as a special case, but need to make sure its handled correctly)
- There isn't a way to get config data from the python side to arduino
- all the normal caveats of native modules (rpc calls and whatnot)

## How to Test

```bash
# On real hardware
uv run examples/arduino_led_echo/demo_led_blink.py

# Unit tests
uv run pytest dimos/core/tests/test_arduino_module.py -v

# Wire-compat C++ tests (requires nix)
cd dimos/hardware/arduino/test && nix develop --command bash -c "cmake -B build && cmake --build build && ./build/test_wire_compat"

# Virtual end-to-end (requires nix for QEMU + avr toolchain)
uv run pytest dimos/hardware/arduino/examples/arduino_twist_echo/test_e2e_virtual.py -v

# Or run the virtual blueprint directly
dimos run arduino-twist-echo-virtual
```


leshy:
pasting comment from Discord here:

Will chat on this in person first, I think it's really cool to auto-build and deploy onto a microcontroller when your blueprint runs, we can even have a small version control protocol to not write firmware every time.

I'd start with just establishing good comms, then do autobuilds etc separately.

I see a bunch of messages are redefined here by hand due to malloc, but I think we should just write a small arduino compatible encoder/decoder on our dimos-lcm cpp side and not merge 6000 lines of mirrored messages into main repo. LCM proto also gives you pubsub out of the box (topic encoding, packets etc, super helpful for serial) we can disable hashing or change the algo if this is an issue - actually pretty important for serial comms

on arduino side ideally we have the exact same cpp API, why can't I do something like (I'm completely vibing)
```
dimos = dimos_start(serial_port)
# msg parsing needs to be tied to a hardware interrupt not a stupid loop
# due to high CPU constraints - partial decoding of a topic name first and not decoding the rest if no active sub
dimos.subscribe('/twist' ...)
```

dimos/protocol/rpc/pubsubrpc.py
```diff
@@ -89,11 +89,13 @@ def __init__(
...
-        state: dict[str, Any]
-        if hasattr(super(), "__getstate__"):
-            state = super().__getstate__()  # type: ignore[assignment]
-        else:
-            state = self.__dict__.copy()
+        # ``object.__getstate__`` exists on Python 3.11+ but not 3.10, and
+        # mypy 3.10 reports ``__getstate__ undefined in superclass``.  Use
+        # ``getattr`` with a default to bypass the static lookup; the
+        # result is typed ``Any`` which lets the rebind to ``state`` pass
+        # both mypy versions without a ``# type: ignore``.
```
jeff-hykin: this shouldnt be here

dimos/robot/all_blueprints.py
```diff
@@ -167,6 +168,7 @@
     "spatial-memory": "dimos.perception.spatial_perception.SpatialMemory",
     "speak-skill": "dimos.agents.skills.speak_skill.SpeakSkill",
     "temporal-memory": "dimos.perception.experimental.temporal_memory.temporal_memory.TemporalMemory",
+    "twist-echo": "dimos.hardware.arduino.examples.arduino_twist_echo.module.TwistEcho",
```
jeff-hykin: should be named different

dimos/simulation/engines/mujoco_shm.py
```diff
@@ -203,15 +207,21 @@ def read_position_command(self, num_joints: int) -> NDArray[np.float64] | None:
...
         self._last_pos_cmd_seq = seq
         arr = self._array(self.shm.pos_t, MAX_JOINTS, np.float64)
-        return arr[:num_joints].copy()
+        # `.astype(np.float64)` is typed on both mypy 3.10 and 3.12 as
+        # returning ``NDArray[np.float64]`` — using it here (instead of
+        # ``.copy()`` on a slice, which mypy 3.10 loses dtype on) gives
+        # us a dtype-parameterized return without a cast.  `.astype`
+        # also always returns a fresh array, so the copy semantics are
+        # preserved.
+        return arr[:num_joints].astype(np.float64)
```
jeff-hykin: shouldnt be here

dimos/experimental/arduino/examples/arduino_multi_echo/tool_multi_echo_hardware.py
```diff
@@ -0,0 +1,355 @@
+# Copyright 2026 Dimensional Inc.
```
paul-nechifor, line 1: This is not a test. Please rename it to `tool_` maybe?

dimos/experimental/arduino/examples/arduino_twist_echo/test_e2e_virtual.py
```diff
@@ -0,0 +1,137 @@
...
+
+import shutil
+import time
+from types import MappingProxyType
+
+import pytest
+
+# Heavy imports are deferred past the skip check so a machine without the
+# Arduino toolchain can still collect this file cheaply.
+pytestmark = [pytest.mark.slow, pytest.mark.tool]
```
paul-nechifor: We no longer have `slow` and tool tests are now not marked anymore (just rename to `tool_`)

dimos/experimental/arduino/test_arduino_module.py
```diff
@@ -0,0 +1,525 @@
...
+
+    return msgs_dir
+
+
+def _main_cpp_path() -> Path:
+    return _ARDUINO_HW_DIR / "cpp" / "main.cpp"
+
+
+@pytest.mark.slow
+@pytest.mark.tool
```
paul-nechifor: slow has been renamed and tool no longer exists. Please move the tool test to a separate `tool_` file. (Same for the other one.)

paul-nechifor (approved):

Issue #1759, opened 2026-04-08
Execute documentation codeblocks in CI


leshy:
I bet a bunch are broken already, if fix isn't obvious feel free to mark those as `skip` and we need to fix them

read `docs/agents/docs/codeblocks.md` on how code block exec works.

to not be annoying for devs (fail tests or expect them to do this locally) do what 
`.github/workflows/code-cleanup.yml` action does - CI can commit the change into the branch

PR #1884, opened 2026-04-20
Feat/1759 doc codeblocks ci

bogwi:
## Problem
Hi, everyone.

This closes Execute documentation codeblocks in CI #1759 opened by @leshy .

## Solution

Two strategies. First introduced the mechanism, workflow and a script. The second - improves present doc codeblocks for: missing imports, modules resolution, some blocks are wrapped to be run-able, those that did not mean to be run are skipped like illustrative one-liners or very heavy blocks, etc. This workflow is disabled for macos.  This is a version 1 of the #1759 solution.

## Breaking Changes

None


bin/run-doc-codeblocks
```diff
@@ -0,0 +1,105 @@
...
+      exit 1
+    fi
+  done
+else
+  if [[ -d ./docs ]]; then
+    run_one_target ./docs
+  fi
+  if [[ -f README.md ]]; then
+    run_one_target README.md
+  fi
```
paul-nechifor: Why the `if`s? `docs` and `README.md` should always exist.

bogwi: `if` here is a harmless and defensive measure, but I can see it can look redundant.

.github/workflows/doc-codeblocks.yml
```diff
@@ -0,0 +1,66 @@
...
+        with:
+          python-version: '3.12'
+
+      - uses: astral-sh/setup-uv@v8
+        with:
+          enable-cache: true
+
+      - uses: actions/setup-node@v6
+        with:
+          node-version: '22'
```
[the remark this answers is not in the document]
bogwi: Aligning `doc-codeblocks.yml` with `code-cleanup.yml` and the rest of the repo: checkout@v5, setup-python@v5, setup-uv@v4. But no other workflow uses setup-node; will use setup-node@v4 as the peer to checkout@v5. Also aligning git-auto-commit-action with `code-cleanup.yml` (v5).

bogwi:
The last commit, 3be2ad5, addresses two previous minor ci run failures: 
(a) trailing-whitespace, and 
(b) astral's new "No more major and minor tags" policy from 8.0.0, use only immutable tags as versions.

The full ci should be green now and is safe to re-run.

If @paul-nechifor insists on "if" change - please say a word.

leshy:
https://github.com/dimensionalOS/dimos/actions/runs/24695981801/job/72277385519?pr=1884

didn't investigate in detail, approved workflow run, currently fails in some weird way, is this related to it being an external PR?

bogwi:
> Tracing this: https://github.com/dimensionalOS/dimos/actions/runs/24711607201/job/72278865724?pr=1884.

## What is actually failing
The job name is md-babel but the failure happens before `./bin/run-doc-codeblocks` even runs. It fails on:

`uv sync --extra dev --frozen --no-extra dds`

That installs the project in editable mode which runs setuptools -> build_ext for the pybind11 extensions in setup.py. Pybind11’s auto_cpp_level() compiles a tiny probe (flagcheck.cpp) with c++ and -std=c++17 / 14 / 11. When all of those probes fail it gets:

`RuntimeError: Unsupported compiler -- at least C++11 support is needed!`

So this looks as a native C++ build environment problem on the runner not md-babel or markdown execution at all.

Two plausible ways to fix it:
1) use an environment known to compile dimos (as in ci.yml), or 
2) to fix c++ / build tools on the self-hosted host.


The above last commit uses the same ci as in ci.yml. Also updated some docs to have no manipulation imports in CI.

bogwi:
Updated workflow compiles dimos OK.
 
New commit addresses these that erred:
a) Marked the YOLO snippet as ```python skip.

b) Marked all replay and asset-dependent session=qb blocks in both quality_filter.md files as skip session=qb (left the two synthetic rx / quality_barrier examples runnable).

c) Marked all session=align Python blocks in both temporal_alignment.md files as skip session=align (embedded  stays as documentation).

bin/run-doc-codeblocks
bogwi: Done in [0707499](https://github.com/dimensionalOS/dimos/pull/1884/commits/0707499d651452b350ce21d1dda3792079ab918c)

.github/workflows/doc-codeblocks.yml
```diff
@@ -0,0 +1,73 @@
...
+
+permissions:
+  contents: write
+  packages: read
+  pull-requests: read
+
+jobs:
+  md-babel:
+    if: github.event_name != 'pull_request' || github.event.pull_request.draft == false
+    timeout-minutes: 60
```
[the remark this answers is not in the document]
bogwi: Done in [0707499](https://github.com/dimensionalOS/dimos/pull/1884/commits/0707499d651452b350ce21d1dda3792079ab918c)

bogwi:
md-babel-py is 1.1.4 now, supports --execution-timeout flag which I authored exactly for the purpose: https://github.com/leshy/md-babel-py/releases/tag/v1.1.4. Giving the recent > 60s (md-babe-py default) runs, timeout was increased to 120s.  Done in [0707499](https://github.com/dimensionalOS/dimos/pull/1884/commits/0707499d651452b350ce21d1dda3792079ab918c)

On the local machine, the current state of the docs fluctuates between 8 and 33 seconds, and not every block runs. The Golden Rule in codeblocks.md, as stated now, is: "All code blocks must be executable. Never write illustrative/pseudo code blocks. If you're showing an API usage pattern, create a minimal working example that actually runs. This ensures the documentation stays correct as the codebase evolves." However, this might not be possible to achieve.

Update in: [30a12a3](https://github.com/dimensionalOS/dimos/pull/1884/commits/30a12a39129e97fc21974706ea1eb82ce08fe829)

Greptile might flag this as  Arbitrary code execution from fork PRs on self-hosted runner: github.event_name != 'pull_request' || github.event.pull_request.draft == false. To silence the bot - [0707499](https://github.com/dimensionalOS/dimos/pull/1884/commits/0707499d651452b350ce21d1dda3792079ab918c) was created, but it will not run on forks. @leshy, this needs your clarification as per #1759.

.github/workflows/doc-codeblocks.yml
```diff
@@ -0,0 +1,76 @@
...
+#   until then, this job fails on execution errors so drift is visible.
+# - For now this job only verifies that covered code blocks execute in CI.
+#   A follow-up can re-enable auto-committing refreshed <!--Result:--> /
+#   generated assets back onto the PR branch.
+name: doc-codeblocks
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+    paths:
+      - 'docs/**'
```
paul-nechifor, line 18: This is set to run only if docs are changed, right? But that's not correct.

The code examples use the actual `dimos/` package. If incompatible changes are made to the docs, but the .md files are not updated, then this task will not fail as it should.

docs/usage/transports/index.md
```diff
@@ -110,7 +110,7 @@ ros = nav.transports(
 
 Each **stream** on a module can use a different transport. Set `.transport` on the stream **before starting** modules.
 
-```python ansi=false
+```python skip ansi=false
```
paul-nechifor: `from dimos.core.module_coordinator import ModuleCoordinator` is no longer a valid import. It's `from dimos.core.coordination.module_coordinator import ModuleCoordinator` now.

Is this the reason this was skipped?

I think it would be better if the code blocks were fixed instead of skipped. Although I assume this is not easy.

docs/agents/style.md
```diff
@@ -38,7 +38,7 @@ something/
 
 Never add imports to `__init__.py` files. Re-exporting from `__init__.py` makes imports too wide and slow — importing one symbol pulls in the entire package tree.
 
-```python
+```python skip
```
paul-nechifor, line 41: I think the solution here is to break this into two examples, and only skip the bad one.

docs/agents/testing.md
```diff
@@ -59,7 +59,7 @@ def test_wiring() -> None:
 
 When a resource is shared across multiple tests, use a pytest fixture with `yield` instead of repeating context managers in each test:
 
-```python
+```python skip
```
paul-nechifor: This breaks because the import for `pytest` is missing. Maybe if you add it you don't have to skip.

.github/workflows/doc-codeblocks.yml
bogwi: Thanx, nice catch. 

We can do this,

```
  pull_request:
    types: [opened, synchronize, reopened, ready_for_review]
    paths:
      - 'dimos/**    # add this
      - 'docs/**'
      ...
```

so any PR that changes library code also runs documentation code block execution.

If we later find examples that import from other trees (for example scripts/ or small helpers outside dimos/), we add those globs too, but for the usual "examples import dimos" setup, adding dimos/** closes the hole described.

bin/run-doc-codeblocks
```diff
@@ -0,0 +1,124 @@
...
+if [[ ${#USER_PATHS[@]} -gt 0 ]]; then
+  for p in "${USER_PATHS[@]}"; do
+    if [[ -d "$p" ]] || [[ -f "$p" ]]; then
+      run_one_target "$p"
+    else
+      echo "Error: not a file or directory: $p" >&2
+      exit 1
+    fi
+  done
+else
```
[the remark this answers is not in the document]
bogwi: The README guard belongs **only** after `done`, in the `else` that pairs with `if [[ ${#USER_PATHS[@]} -gt 0 ]]`.

Will be implemented in the next cmt.

docs/usage/transports/index.md
bogwi: The `skip` matched the policy in `codeblocks.md` (multi-process coordinator + timing/hardware), but yes it also hid an incorrect snippet details. 

Make it run was relatively straightforward. If @leshy  agrees on this, because it is his block, I will make a change I just tested to work. If not, the example will be a skip even though we will change the correct import naturally.

Here is the proposed version with commentary:

The runnable example below uses a tiny synthetic image publisher instead of `CameraModule` so it works without a webcam and in CI; the wiring is the same as with a real camera.

```python ansi=false
import time

import numpy as np
import reactivex as rx

from dimos.core.core import rpc
from dimos.core.coordination.module_coordinator import ModuleCoordinator
from dimos.core.stream import In, Out
from dimos.core.transport import LCMTransport
from dimos.msgs.sensor_msgs.Image import Image, ImageFormat


class TickerCameraConfig(ModuleConfig):
    frequency_hz: float = 2.0


class TickerCameraModule(Module):
    """Publish synthetic frames so this example runs without a webcam."""

    config: TickerCameraConfig
    color_image: Out[Image]

    @rpc
    def start(self) -> None:
        super().start()

        def emit(_: int) -> None:
            img = Image.from_numpy(
                np.zeros((480, 640, 3), dtype=np.uint8),
                format=ImageFormat.RGB,
                frame_id="synthetic",
            )
            self.color_image.publish(img)

        period = 1.0 / max(self.config.frequency_hz, 0.1)
        self.register_disposable(rx.interval(period).subscribe(emit))


class ImageListener(Module):
    image: In[Image]

    async def handle_image(self, img: Image) -> None:
        print(f"Received: {img.shape}")


if __name__ == "__main__":
    # Start local cluster and deploy modules to separate processes
    dimos = ModuleCoordinator()
    dimos.start()

    camera = dimos.deploy(TickerCameraModule, frequency_hz=2.0)
    listener = dimos.deploy(ImageListener)

    # Choose a transport for the stream (example: LCM typed channel)
    camera.color_image.transport = LCMTransport("/camera/rgb", Image)

    # Connect listener input to camera output
    listener.image.connect(camera.color_image)

    dimos.start_all_modules()

    time.sleep(2)
    dimos.stop()
```


In general what we are trying to achieve with [Feat/1759 doc codeblocks ci](https://github.com/dimensionalOS/dimos/pull/1884#top)#1884 is to:
- incrementally improve the quality of our docs
- not make every line to run. * If to make a line run requires x10 of wrapping it defeats the purpose of the doc snippet and sways away the attention of the reader.

The best rule is to have codeblocks that are (a) ready to copy-pasting and (b) illustrative that show the intent but needed not to be running.

docs/agents/style.md
bogwi: Agree.

docs/agents/testing.md
bogwi: "just add pytest and drop skip" will not do here. The snippet also relies on DB_PATH without defining it. That only blows up if something runs the fixture body under pytest; md-babel never invoked it so it looked fine while skipped.
The assertions (count() > 0, len(results) == 5) assumed a DB already filled with "video" Image rows. An empty or temp DB would fail if someone actually ran these tests.

Fix below if the goal is to make it 100% run. But it's like +21 -7, net 14 lines of code. @leshy intent seems a bit diluted here.


```python
from collections.abc import Iterator
from pathlib import Path
import tempfile

import pytest

from dimos.msgs.sensor_msgs.Image import Image


@pytest.fixture(scope="module")
def store() -> Iterator[SqliteStore]:
    # Example uses an empty temp DB so it runs standalone; point `path` at shared test data when needed.
    with tempfile.TemporaryDirectory() as d:
        db_path = Path(d) / "memory.db"
        db = SqliteStore(path=str(db_path))
        with db:
            yield db


def test_query(store: SqliteStore) -> None:
    assert store.stream("video", Image).count() == 0


def test_search(store: SqliteStore) -> None:
    results = store.stream("video", Image).limit(5).to_list()
    assert results == []
```

.github/workflows/doc-codeblocks.yml
leshy: agreed, @paul I think most of the time this will catch problems with code changing and breaking docs

docs/usage/transports/index.md
leshy: yeah this looks good, whatever you guys decide

docs/agents/testing.md
leshy: yes I personally don't care too much about this being executable, 

main idea behind executable docs is that we catch when examples are broken due to code changes later. 

these `agents/testing.md` are sketches for how an agent should write tests so not very important imo

.github/workflows/doc-codeblocks.yml
```diff
@@ -0,0 +1,76 @@
...
+          git lfs pull
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '22'
+
+      - name: Install Python dependencies
+        run: |
+          uv sync --all-extras --no-extra dds --no-extra unitree-dds --frozen
+          uv pip install matplotlib httpx
```
leshy, line 69: let's create an issue to figure out why this is missing from our dev image

.github/workflows/doc-codeblocks.yml
```diff
@@ -0,0 +1,76 @@
...
+          git config --global --add safe.directory '*'
+          git clean -ffdx -e .venv
+
+      # Docs decode JPEG from SQLite via PyTurboJPEG; it needs the system lib (see docs/installation/ubuntu.md).
+      # macOS dev machines often have libjpeg-turbo from Homebrew; the ros-dev container does not by default.
+      # Examples also use get_data() which resolves Git LFS blobs under tests/data.
+      - name: Install TurboJPEG and pull Git LFS for doc examples
+        run: |
+          apt-get update
+          DEBIAN_FRONTEND=noninteractive apt-get install -y git-lfs libturbojpeg
```
leshy, line 58: let's create an issue to figure out why this is missing from our dev image

leshy (dismissed):

leshy (approved):

PR #1886, opened 2026-04-20
feat: never leave child processes alive after the parent

paul-nechifor:
## Problem

Certain child processes might not get shut down correctly, especially processes started by modules.

Closes DIM-XXX

## Solution

* Mark all processes by adding a `DIMOS_RUN_ID` env var for them.
* Start a watchdog process which watches for the main process to die. When it dies, all children (and children of children, etc) are killed.
* Also kill when a new process starts.

## Breaking Changes

None.


leshy (approved):
looks good, is it hard to give watchdog voice in the logs?
I'd like to get a warning when it's killing some runaway children, what do you think?

PR #1888, opened 2026-04-21
Drop perception. Update docs.

Dreamsorcerer:
## Problem

Dropping heavyweight perception from base install. With other PRs, this removes torch from base install.

Also dropping bitsandbytes which doesn't appear to be used anywhere at all?

Also updating the docs now to reflect all those changes.

## Breaking Changes

Users wanting perception now need to request it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -165,7 +165,6 @@ agents = [
     "langchain-text-splitters>=1,<2",
     "langchain-huggingface>=1,<2",
     "langchain-ollama>=1,<2",
-    "bitsandbytes>=0.48.2,<1.0; sys_platform == 'linux'",
     "ollama>=0.6.0",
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -164,7 +164,6 @@ agents = [
     "langchain-text-splitters>=1,<2",
     "langchain-huggingface>=1,<2",
     "langchain-ollama>=1,<2",
-    "bitsandbytes>=0.48.2,<1.0; sys_platform == 'linux'",
```
paul-nechifor, line 168: bitsandbytes is needed by langchain-ollama or langchain-huggingface. I don't remember which one, but I added it intentionally (should have left a comment at the time).

paul-nechifor (approved):

PR #1901, opened 2026-04-22
Split tests off of self hosted runner

Dreamsorcerer:
This splits linting and tests into separate jobs on standard GH runners.
Only the tests that need ROS, LFS or something similar are now run on self-hosted runner.

This speeds up the CI significantly (GH runner tests also get to use xdist to parallelise the tests).
This also allows external PRs to be tested and merged normally, with just the self-hosted tests being skipped.
Also some security improvements and improving maintainability of the code.


.github/workflows/ci.yml
```diff
@@ -14,7 +14,68 @@ permissions:
...
+      - name: Setup Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: 3.12
+
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libturbojpeg0-dev portaudio19-dev build-essential \
+            libgl1 libgl1-mesa-dri python3-dev libcyclonedds-dev
```
```suggestion
            libgl1 libgl1-mesa-dri python3-dev cyclonedds-dev
```

dimos/models/vl/test_base.py
```diff
@@ -24,6 +24,8 @@
 from dimos.perception.detection.type.detection2d.imageDetections2D import ImageDetections2D
 from dimos.utils.data import get_data
 
+pytestmark = pytest.mark.slow
+
```
Dreamsorcerer:
```suggestion
```

dimos/agents/mcp/mcp_client.py
```diff
@@ -156,7 +156,12 @@ def _try_fetch_tools(self, timeout: float, interval: float) -> dict[str, Any] |
             try:
                 self._mcp_request("initialize")
                 break
-            except (httpx.ConnectError, httpx.RemoteProtocolError):
+            # During cold start uvicorn may briefly answer with 404 (route
+            # table not yet built), close the connection mid-read, or have
+            # the JSON-RPC handler return an error before tools are
+            # registered. Retry on anything httpx raises plus our own
+            # `RuntimeError` from `_mcp_request`'s error-payload check.
+            except (httpx.HTTPError, RuntimeError):
```
Dreamsorcerer:
```suggestion
            except (httpx.ConnectError, httpx.RemoteProtocolError):
```

pyproject.toml
```diff
@@ -347,6 +297,108 @@ base = [
...
+    "python-socketio>=5.16.1",
+    "sounddevice>=0.5.5",
+    "trimesh>=4.12",
+    "watchdog>=3.0.0",
+
+    "pandas-stubs>=2.3.2.250926,<3",
+    "types-pyaudio",
+    "types-psycopg2>=2.9.21.20251012",
+    "types-PyYAML>=6.0.12.20250915,<7",  # yaml (msgs/sensor_msgs/CameraInfo.py, perception/common/utils.py)
+    "types-requests>=2.32.4.20260107,<3",  # requests (mcp_adapter, osm, cli, skills/rest, …)
```
Dreamsorcerer:
```suggestion
    "types-PyYAML>=6.0.12.20250915,<7",
    "types-requests>=2.32.4.20260107,<3",
```

pyproject.toml
```diff
@@ -461,16 +520,13 @@ env = [
   "GOOGLE_MAPS_API_KEY=AIzafake_google_key",
   "PYTHONWARNINGS=ignore:cupyx.jit.rawkernel is experimental:FutureWarning",
 ]
-addopts = "-s -v -r a -p no:warnings -p no:launch_testing -p no:launch_ros --import-mode=importlib --color=yes -m 'not (tool or slow or mujoco)'"
+addopts = "--dist=loadfile -v -r a -p no:warnings -p no:launch_testing -p no:launch_ros --import-mode=importlib --color=yes -m 'not (tool or self_hosted or mujoco)'"
```
Dreamsorcerer:
```suggestion
addopts = "--dist=loadfile --durations=10 -v -ra --showlocals -p no:launch_testing -p no:launch_ros --import-mode=importlib --color=yes -m 'not (tool or self_hosted or mujoco)'"
```

Dreamsorcerer:
@copilot resolve the merge conflicts in this pull request

.github/workflows/ci.yml
```diff
@@ -5,68 +5,277 @@ on:
...
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: gh run cancel ${{ github.run_id }} --repo ${{ github.repository }}
+
+  ci-complete:  # This is used for branch protection.
+    if: always()
+
+    needs:
+    - lint
+    - tests
```
Dreamsorcerer, line 264:
```suggestion
    - md-babel
    - tests
```

dimos/core/coordination/python_worker.py
```diff
@@ -412,6 +424,11 @@ def _handle_request(request: Any, state: _WorkerState) -> WorkerResponse:
                     "allow_pickle": True,
                 },
             )
+            # `ThreadedServer.__init__` binds the socket but does not call
+            # `listen()` — that happens in `start()` on the thread. Call
+            # `_listen()` synchronously first so the port we return is
+            # actually accepting connections.
+            state.rpyc_server._listen()
```
paul-nechifor: I've had the same issue in some other place. The solution is to wait for the connection to be open. It's not good practice to call private methods because we don't want to depend on something which they might change.

dimos/core/test_mcp_integration.py
```diff
@@ -37,6 +36,7 @@
 
 from dimos.agents.mcp.mcp_adapter import McpAdapter
 from dimos.agents.mcp.mcp_server import McpServer
+from dimos.conftest import _BUCKET
```
paul-nechifor: It's bad practice to import from conftest files. Just define a fixture, or move the constant outside.

dimos/protocol/service/system_configurator/base.py
```diff
@@ -77,8 +77,12 @@ def fix(self) -> None:
 
 
 def configure_system(checks: list[SystemConfigurator], check_only: bool = False) -> None:
-    if os.environ.get("CI"):
-        logger.info("CI environment detected: skipping system configuration.")
+    # Skip in test runs — we'd otherwise prompt for sudo from a non-
+    # interactive subprocess and kill the worker. The self-hosted CI
+    # runner has its host config baked in via the container image, so
+    # it doesn't need configurators to fix anything either.
+    if os.environ.get("PYTEST_VERSION"):
```
paul-nechifor, line 84: I think the idea of using CI is that we want to see the password prompt locally, but not when running in CI because there's nothing to be done there.

If you change it to check for PYTEST_VERSION then the system is never adjusted locally, right? So tests would only work if you ran `uv run dimos ...` and then `uv run pytest` after a fresh boot, but if you run `uv run pytest` as the first thing then presumably it wouldn't work? Is that so?

.pre-commit-config.yaml
```diff
@@ -5,8 +5,6 @@ repos:
   - repo: https://github.com/Lucas-C/pre-commit-hooks
     rev: v1.5.5
     hooks:
-      - id: forbid-crlf
-      - id: remove-crlf
```
paul-nechifor, line 10: Why would we want Windows line endings. This causes trouble in certain cases. If I remember correctly, certain shell scripts don't work with windows line endings. But even so, it's better to have uniform files.

pyproject.toml
```diff
@@ -350,9 +300,114 @@ base = [
...
+    # Needed to compile the in-tree extensions.
+    "pybind11>=2.12",
+]
+
+[tool.uv]
+# `unitree-dds` is a superset of `unitree` (adds DDS bridge libs that need
+# system cyclonedds). Mark them conflicting so uv's universal resolver
+# doesn't try to satisfy both at once when `unitree` is requested.
+conflicts = [
+    [{ extra = "unitree" }, { extra = "unitree-dds" }],
```
paul-nechifor: I can no longer install all the packages:

```bash
$ CYCLONEDDS_HOME=/opt/cyclonedds uv sync --all-extras
Resolved 448 packages in 1ms
error: Extras `unitree` and `unitree-dds` are incompatible with the declared conflicts: {`dimos[unitree]`, `dimos[unitree-dds]`}
```

.pre-commit-config.yaml
Dreamsorcerer: I set the config option in ruff format, so this step is redundant and means we now don't need this extra tool in the auto-fix workflow.

paul-nechifor: @Dreamsorcerer Ruff is for python files. It doesn't make these changes for other files.

leshy:
```
      - name: Ruff format
        run: uv run ruff format
      - name: uv lock
        run: uv lock
      - name: autofix.ci
        uses: autofix-ci/action@v1
```


some of this is (was?) done by pre-commit hooks which were ran by the CI - if pre-commit hooks change files in any way, it would auto-commit. 

feels here we are re-implementing part of this flow (ruff, uv lock can be added) - I liked the fact that dev machines and CI run exactly the same code path for code formatting, it's easier to maintain. can we rely on pre-commit hooks for autofix?

leshy:
can merge this early then iterate not to diverge

.pre-commit-config.yaml
Dreamsorcerer: >         files: \.py$

Neither does this.

pyproject.toml
Dreamsorcerer: I'm not familiar enough with uv to know if there is some way we should make this work.

Basically, unitree and unitree-dds conflict with each other, and this just makes it explicit, instead of getting really confused with what's actually running and wasting time debugging something that shouldn't have occurred.

Dreamsorcerer: Nevermind, I think I got confused with something else where there was a package shadowing another package. I've removed this and it seems to be fine so far.

dimos/protocol/service/system_configurator/base.py
Dreamsorcerer: So, this fails locally too for the same reason, which is why I left it like this. I don't think the password prompts are going to work correctly through xdist.

Atleast the default set of tests all work fine for me with the code here that just raises to the soft limit without needing sudo.

For the full self_hosted tests, then yes, you may need to run dimos first. I've not thought of a better way to do it.

dimos/core/test_mcp_integration.py
Dreamsorcerer: There was some complication that made that look difficult to do, but seems to have been resolved at some point.

dimos/protocol/service/system_configurator/base.py
Dreamsorcerer: If we've got a better proposal, let's add the improvement in a followup PR.

.pre-commit-config.yaml
paul-nechifor: Ah, true. I was confusing this for the editorconfig stuff. 😅

paul-nechifor (dismissed):

paul-nechifor (approved):

PR #1906, opened 2026-04-23
Default to Zenoh transport on macOS and document replay workflow

bogwi:
Supersedes #1787.

This PR carries forward the Zenoh transport integration from #1787 and wraps it into a merge-ready branch that fixes the remaining macOS Big Office replay gap.

## Validation

Typical **replay on macOS** when Zenoh is installed (default is already Zenoh, so no transport flag is required):

```bash
dimos --dtop --replay --replay-db=go2_bigoffice run unitree-go2
```

The same workload on **Linux** (default remains `lcm` until you opt in):

```bash
dimos --transport=zenoh --dtop --replay --replay-db=go2_bigoffice run unitree-go2
```

## Notes

- this PR is intended as the wrapped successor to #1787, not a separate redesign
- Linux behavior remains unchanged by default: explicit Zenoh still works, and the default transport remains LCM


dimos/core/test_zenoh_transport.py
```diff
@@ -0,0 +1,259 @@
...
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from dimos.core.coordination.blueprints import autoconnect
+from dimos.core.coordination.module_coordinator import _get_transport_for, _run_configurators
+from dimos.core.global_config import GlobalConfig, global_config
+from dimos.core.module import Module
```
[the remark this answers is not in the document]
bogwi: With e82b5dc, I moved the retry helper into import-safe shared test utilities and test_zenohpubsub now skips cleanly with pytest.importorskip("zenoh") so non-Zenoh environments can still collect the platform-agnostic suites.

dimos/visualization/rerun/bridge.py
```diff
@@ -172,8 +173,39 @@ def _resolve_viewer_mode() -> ViewerMode:
...
 class Config(ModuleConfig):
-    """Configuration for RerunBridgeModule."""
+    """Configuration for RerunBridgeModule.
+
+    The pubsubs field is accepted for backwards compatibility (existing blueprints
+    pass it), but ignored at start() — the actual pubsub backend is resolved
+    lazily from global_config.transport.
+    """
 
     pubsubs: list[SubscribeAllCapable[Any, Any]] = field(default_factory=lambda: [LCM()])
```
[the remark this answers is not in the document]
bogwi: In [e82b5dc](https://github.com/dimensionalOS/dimos/pull/1906/commits/e82b5dc31073d0c27ba1566656d921e58146048f) the bridge now preserves explicit non-default pubsubs overrides treating the legacy [LCM()] value as backward-compatible default behavior and so transport-driven resolution still works for existing blueprints.

dimos/protocol/service/test_zenohservice.py
```diff
@@ -0,0 +1,82 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tests for ZenohService — singleton session management."""
+
+from __future__ import annotations
+
+import pytest
+
+from dimos.protocol.service.zenohservice import ZenohConfig, ZenohService, _sessions
```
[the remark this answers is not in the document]
bogwi: Committed as: 89702c148 - fix(zenoh): skip zenohservice tests when dependency is missing

bogwi:
> ### Greptile Summary

> * **P1 — wrong env-var name in docs**: `docs/usage/transports/index.md` (and `osx.md`) document `DIMOS_TRANSPORT=zenoh` as the env-var override, but `GlobalConfig` uses `pydantic_settings.BaseSettings` with no `env_prefix`, so the actual variable pydantic-settings reads is `TRANSPORT`. Setting `DIMOS_TRANSPORT` will be silently ignored.
> 
> ### Confidence Score: 4/5
> Safe to merge with the env-var name corrected in docs; runtime behavior is unaffected by the docs bug.
> 
> One P1 finding: the documented env-var `DIMOS_TRANSPORT` will silently have no effect because `GlobalConfig` has no `env_prefix`. The code itself is correct and well-tested; the bug is isolated to documentation. P1 ceiling is 4/5.
> 
> docs/usage/transports/index.md and docs/installation/osx.md — both reference the incorrect `DIMOS_TRANSPORT` env-var name.
> 

This is not the purpose of this branch to add these changes to `global_config.py`:
```
model_config = SettingsConfigDict(
        env_prefix="DIMOS_",
        ...
```

It will be breaking cause anyone relying on unprefixed env vars for GlobalConfig will no longer get that effect; they need DIMOS_ for those settings to load from the environment / .env even though the cli.md in line 53 says, - "Environment variables and `.env` values must be prefixed with `DIMOS_`" . That is not yet done, but the docs for this PR have to be aligned so to not tackle them later in yet a new PR.

bogwi:
1) The branch defaults to zenoh on Mac yet needs a flag on the rest.

From `dimos/docs/usage/transports/index.md`

Typical **replay on macOS** when Zenoh is installed (default is already Zenoh, so no transport flag is required):

```bash
dimos --dtop --replay --replay-db=go2_bigoffice run unitree-go2
```

The same workload on **Linux** (default remains `lcm` until you opt in):

```bash
dimos --transport=zenoh --dtop --replay --replay-db=go2_bigoffice run unitree-go2
```

2) All issues mentioned in the opening PR #1787 and recent Finish Zenoh integration #1941 were addressed.

docs/usage/transports/index.md
```diff
@@ -29,6 +29,73 @@ So: treat the API as uniform, but pick a backend whose semantics match the task.
...
+
+| Situation | Default |
+|-----------|---------|
+| macOS and the `eclipse-zenoh` package is importable | `zenoh` |
+| Any other platform, or Zenoh not installed | `lcm` |
+
+**Two ways to override for one run or for your shell:**
+
+1. **CLI:** `dimos --transport=zenoh ...` or `dimos --transport=lcm ...` (see [CLI](/docs/usage/cli.md) for precedence with `.env` and blueprints).
+2. **Environment:** `DIMOS_TRANSPORT=zenoh` or `DIMOS_TRANSPORT=lcm`.
```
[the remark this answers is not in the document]
bogwi: https://github.com/dimensionalOS/dimos/pull/1906#issuecomment-4372456306

dimos/core/coordination/module_coordinator.py
```diff
@@ -613,7 +634,8 @@ def _run_configurators(blueprint: Blueprint) -> None:
     from dimos.protocol.service.system_configurator.base import configure_system
     from dimos.protocol.service.system_configurator.lcm_config import lcm_configurators
 
-    configurators = [*lcm_configurators(), *blueprint.configurator_checks]
+    lcm_checks = lcm_configurators() if global_config.transport == "lcm" else []
```
paul-nechifor: Even if the global transport is zenoh, we still need to run LCM configurators so that LCM works for individual LCM transports. LCM configurators can only be skipped if we remove LCM entirely, no?

dimos/protocol/pubsub/impl/zenohpubsub.py
```diff
@@ -0,0 +1,205 @@
...
+
+        sub = self.session.declare_subscriber(key_expr, on_sample)
+        with self._subscriber_lock:
+            self._subscribers.append(sub)
+
+        undeclared = False
+
+        def unsubscribe() -> None:
+            nonlocal undeclared
+            if undeclared:
```
paul-nechifor: This has to be under the lock too.

docs/development/testing.md
```diff
@@ -48,16 +48,20 @@ pytest dimos
...
 
+For non-interactive runs, especially on macOS, prefer `CI=1`. This skips system configurator prompts and avoids test runs trying to change local multicast or sysctl state.
+
 ### Slow tests
 
 Run the slow tests:
 
 ```bash
-./bin/pytest-slow
+CI=1 ./bin/pytest-slow
```
paul-nechifor: Why was this added? CI is for CI.

```suggestion
./bin/pytest-slow
```

dimos/visualization/rerun/test_viewer_integration.py
```diff
@@ -120,6 +122,33 @@ def test_bridge_has_fallback(self):
...
+        from dimos.visualization.rerun.bridge import Config, _resolve_pubsubs
+
+        config = Config(pubsubs=[LCM()], g=GlobalConfig(transport="lcm"))
+        pubsubs = _resolve_pubsubs(config)
+
+        assert len(pubsubs) == 1
+        assert isinstance(pubsubs[0], LCM)
+
+    def test_explicit_custom_pubsubs_override_is_honored(self):
+        from dimos.core.global_config import GlobalConfig
```
paul-nechifor: In tests, all imports should be at the top.

dimos/protocol/service/zenohservice.py
```diff
@@ -0,0 +1,86 @@
...
+import zenoh
+
+from dimos.protocol.service.spec import BaseConfig, Service
+
+zenoh.init_log_from_env_or("warn")
+from dimos.utils.logging_config import setup_logger
+
+logger = setup_logger()
+
+_sessions: dict[str, zenoh.Session] = {}
```
paul-nechifor: What is this for?

As far as I can see, this is only used for tests. It's bad practice to tailor source code to what tests need. Tests should keep track of the sessions they use. Production code should not keep state around just in case tests need it.

dimos/core/test_zenoh_transport.py
```diff
@@ -0,0 +1,271 @@
...
+    def _clean_sessions(self):
+        from dimos.protocol.service.zenohservice import _sessions
+
+        yield
+        for s in _sessions.values():
+            s.close()
+        _sessions.clear()
+
+    def test_zenoh_transport_broadcast_and_subscribe(self) -> None:
+        import threading
```
paul-nechifor: Please move all imports in all tests to the top.

dimos/core/test_zenoh_transport.py
```diff
@@ -0,0 +1,271 @@
...
+        config = GlobalConfig()
+        with pytest.raises(ValidationError, match="transport"):
+            config.update(transport=cast("object", "invalid"))
+
+
+class TestZenohAvailableGuard:
+    def test_zenoh_available_is_bool(self) -> None:
+        assert isinstance(ZENOH_AVAILABLE, bool)
+
+    @pytest.mark.skipif(not ZENOH_AVAILABLE, reason="zenoh not installed")
```
paul-nechifor: This may be controversial, but I think all tests should fail if a developer failed to install some dependency.

The point of tests is to make sure that the branch is in a valid state. If a developer doesn't install the dependencies and a bad change is made, the tests won't say so simply because the dependencies are skipped.

For users, this is different. Users may have a reason they want to not install certain dependencies. But developers should always install in order to run all tests.

docs/usage/cli.md
paul-nechifor, line 3: (Line unrelated). Does `dimos topic echo` work with Zenoh?

dimos/core/coordination/module_coordinator.py
```diff
@@ -543,7 +549,22 @@ def _get_transport_for(blueprint: Blueprint, name: str, stream_type: type) -> Pu
...
-    transport = pLCMTransport(topic) if use_pickled else LCMTransport(topic, stream_type)
+
+    if global_config.transport == "zenoh":
+        if not ZENOH_AVAILABLE:
+            raise RuntimeError(
+                "transport='zenoh' but eclipse-zenoh is not installed. " + ZENOH_INSTALL_HINT
+            )
+        from dimos.core.transport import ZenohTransport, pZenohTransport
+
+        zenoh_topic = f"dimos{topic}"
```
paul-nechifor: This prefix should be a constant. It's similar to `dimos/*` used in `zenohpubsub.py`.

dimos/protocol/service/zenohservice.py
paul-nechifor, line 33: Does a session not require cleanup at worker exit?

dimos/core/coordination/module_coordinator.py
bogwi: How about a small bool helper:

```python
def _blueprint_specifies_lcm_transport(blueprint: Blueprint) -> bool:
    """True when the merged blueprint wires at least one stream over LCM explicitly."""
    for transport in blueprint.transport_map.values():
        if isinstance(transport, (LCMTransport, pLCMTransport)):
            return True
    return False
```

then wire it:

```python
def _run_configurators(blueprint: Blueprint) -> None:
        ...
        
    need_lcm_system_config = global_config.transport == "lcm" or _blueprint_specifies_lcm_transport(
            blueprint
        )
    lcm_checks = lcm_configurators() if need_lcm_system_config else []
    configurators = [*lcm_checks, *blueprint.configurator_checks]
    
        ...
        
```

dimos/protocol/pubsub/impl/zenohpubsub.py
bogwi: Yes, under the lock:

```python
...

def unsubscribe() -> None:
        nonlocal undeclared
        with self._subscriber_lock:
            if undeclared:
                return
            undeclared = True
...
```

docs/development/testing.md
bogwi: removed

dimos/core/test_zenoh_transport.py
bogwi: Removed all `@pytest.mark.skipif(not ZENOH_AVAILABLE, reason="zenoh not installed")`.

Added a short note at the top too: "Requires the ``zenoh`` extra (``eclipse-zenoh``): ``ZenohTransport`` is only defined when that dependency is installed, so this module does not load without it."

docs/usage/cli.md
bogwi: Well, no - dimos topic echo does not follow Zenoh for the main stack. It is implemented only on top of LCM: LCMTransport / pLCMTransport and in inferred mode LCMPubSubBase and raw LCM subscribe. There is no branch on global_config.transport or any Zenoh pub/sub in dimos/robot/cli/topic.py.

Running DimOS with --transport=zenoh, modules talk over Zenoh and so the topic echo still listens on LCM and will not see that traffic.

I think implementing Zenoh in the topic CLI would be a separate feature definitely not something implied by --transport today


I will add a subsection to cli.md explaining this: ### `dimos topic echo` and `dimos topic send`

dimos/core/coordination/module_coordinator.py
bogwi: Will add this to constants.py and update call sites: `ZENOH_DIMOS_KEY_PREFIX = "dimos"`. If you object for constants.py then we can put it into transport.py or create a dedicated zenoh_keys.py I think.

dimos/protocol/service/zenohservice.py
bogwi: Strictly for correctness often no cause when a worker process exits the OS tears down its address space, file descriptors and threads. A leaked zenoh.Session in that process does not keep a zombie worker around or corrupt another process. 

We can add `close_all_zenoh_sessions()` foo into zenohservice.py:

```python
with _sessions_lock:
        to_close = list(_sessions.values())
        _sessions.clear()
    for session in to_close:
        try:
            session.close()
        except Exception:
            logger.error("Error closing Zenoh session", exc_info=True)
```

And then wire it into the python_worker._worker_entrypoin, and then update tests for consistency.

Give a thumb up if you agree on this

paul-nechifor: I think the easiest option would be to add an `atexit.register`.

dimos/core/coordination/module_coordinator.py
paul-nechifor: I don't think that's sufficient, because we sometimes use LCM outside of module transports. For example see `pLCMTransport("/human_input")` in `utils/cli/human/humancli.py`.

My assumption is that we need to always configure LCM until we remove it (if we do that).

bogwi: Then let's make lcm_checks = lcm_configurators(). (No bool helper proposed is needed)

```python
def _run_configurators(blueprint: Blueprint) -> None:
    from dimos.protocol.service.system_configurator.base import configure_system
    from dimos.protocol.service.system_configurator.lcm_config import lcm_configurators

    lcm_checks = lcm_configurators()
    configurators = [*lcm_checks, *blueprint.configurator_checks]

    try:
        configure_system(configurators)
    except SystemExit:
        labels = [type(c).__name__ for c in configurators]
        print(
            f"Required system configuration was declined: {', '.join(labels)}",
            file=sys.stderr,
        )
        sys.exit(1)
````

?

**Comment by spomichter** [comment]
THnaks @bogwi will need @leshy review here. ALso resolve conflicts when u get chance

File: dimos/protocol/pubsub/impl/test_zenohpubsub.py [hunk 13]
```diff
@@ -0,0 +1,206 @@
...
+
+        def callback_b(msg: bytes, t: Topic) -> None:
+            received_b.append(msg)
+            countdown.wait()
+
+        pubsub.subscribe(topic, callback_a)
+        pubsub.subscribe(topic, callback_b)
+        retry_until(both_received, lambda: pubsub.publish(topic, b"broadcast"))
+        assert received_a[-1:] == [b"broadcast"]
+        assert received_b[-1:] == [b"broadcast"]
```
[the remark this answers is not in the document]
bogwi: Not a P1 

Here’s the probe script I ran from the repo root:

```python
uv run python - <<'PY'

import threading
import time

from dimos.protocol.pubsub.impl.lcmpubsub import Topic
from dimos.protocol.pubsub.impl.zenohpubsub import ZenohPubSubBase
from dimos.protocol.service.zenohservice import close_all_zenoh_sessions

close_all_zenoh_sessions()
ps = ZenohPubSubBase()
ps.start()

try:
    topic = Topic("dimos/test/thread_probe")
    barrier = threading.Barrier(2)
    done = threading.Event()
    calls = []
    lock = threading.Lock()

    def make_cb(name: str):
        def cb(msg: bytes, t: Topic) -> None:
            with lock:
                calls.append(
                    (
                        name,
                        threading.get_ident(),
                        threading.current_thread().name,
                        time.monotonic(),
                    )
                )

            barrier.wait(timeout=2)

            if len(calls) >= 2:
                done.set()

        return cb

    ps.subscribe(topic, make_cb("a"))
    ps.subscribe(topic, make_cb("b"))

    for _ in range(50):
        ps.publish(topic, b"x")
        if done.wait(0.05):
            break

    print("done:", done.is_set())
    for call in calls[:4]:
        print(call)

finally:
    ps.stop()
    close_all_zenoh_sessions()
PY
```

```
done: True
('a', 6178910208, 'Thread-1 (pyo3-closure)', 2180490.299834208)
('b', 6195736576, 'Thread-2 (pyo3-closure)', 2180490.299891166)
```

callback_a ran on Thread-1 (pyo3-closure)
callback_b ran on Thread-2 (pyo3-closure)
The thread IDs differ.
done: True, meaning both callbacks reached the barrier and it released.

bogwi:
Superseded by #2106.

PR #1907, opened 2026-04-23
Jeff/fix/mac configurator

jeff-hykin:
## Problem

#1789 had an bug for MacOS. 

## Solution

Main fix is `>=` instead of `>` the rest is a test readability/consistency cleanup. 

## Breaking Changes

None.

## How to Test

MacOS only 

```bash
# Clear saved state to reproduce
echo '{}' > ~/.local/state/dimos/sysctl.json

# reboot
# Run tests (this is where the problem is worst)
bin/pytest-fast
```


dimos/protocol/service/test_system_configurator.py
```diff
@@ -38,6 +38,19 @@
 )
 from dimos.utils import prompt
 
+# Named constants for test readability (buffer sizes, powers of 2)
```
jeff-hykin, line 41: this comment is noise (will delete)

PR #1909, opened 2026-04-24
feat(api): peek_stream

paul-nechifor:
## Problem

We need an easy way to get a sample value from a stream. Useful for agents.

Closes DIM-822

## Solution

* Add `app.peek_stream`.

## Breaking Changes

None.

## How to Test

```python
# Grab the image.
img = app.peek_stream("color_image", 1.0)

# Display it in a window.
import cv2, numpy as np
cv2.imshow("color_image", np.array(img.data))
cv2.waitKey(0)
```

### DIFF
--- a/dimos/core/coordination/python_worker.py
+++ b/dimos/core/coordination/python_worker.py
@@ -409,6 +409,7 @@ def _handle_request(request: Any, state: _WorkerState) -> WorkerResponse:
                 protocol_config={
                     "allow_all_attrs": True,
                     "allow_public_attrs": True,
+                    "allow_pickle": True,
                 },
             )
             state.rpyc_thread = threading.Thread(target=state.rpyc_server.start, daemon=True)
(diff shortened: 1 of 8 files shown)


leshy:
I think this is attractive but directionally problematic and should be careful with `peek_stream` type functionality. 

(extension of my "I think we should be careful about py repl") - agents don't know about modules, process boundaries etc, having `peek_stream` makes it very attractive for agents to then write a while loop, do live data processing via this function, suddenly they'll write a module that's not a module at all.

A need for peek_stream shows me our module API should be better since we want to cheat.

We can make writing a temporary test module simpler, for example agent should be able to write a script that's a module that interacts with a blueprint, once script is killed, module is dead also. This already works (or worked) nothing used to stop you from running your module from a separate script

Not blocking this, just some thoughts, we will see experimentally, maybe it's ok

leshy (dismissed):

mustafab0 (approved):

PR #1920, opened 2026-04-26
feat(modules): async modules

paul-nechifor:
## Problem

Having so many threads in modules makes code hard to write.

Closes DIM-812

## Solution

* Multiple aspects of modules can be async now: handlers, RPCs, and observable callbacks all run on the per-module `self._loop` daemon-thread asyncio loop, removing the need for locks on per-instance state.
* Auto-binds `async def handle_x` for every `x: In[T]`; each handler gets a dedicated dispatcher task with LATEST-only semantics (drops intermediate messages, guarantees the most recent value is eventually processed).
* New `@arpc` decorator: async RPC body interchangeable with `@rpc` (same discovery + RPC machinery).
* Sync and async specs interoperate.
* `self.spawn(coro)` replaces bare `run_coroutine_threadsafe` and routes unhandled exceptions to the module logger instead of swallowing them in an unread Future.
* `self.process_observable(obs, async_handler)` bridges rxpy observables onto `self._loop`.
* `async def main(self)` async-generator-with-one-`yield` collapses paired `start()`/`stop()` resource setup/teardown into a single visually-adjacent block. 

## Breaking Changes

None.

### DIFF
--- a/dimos/core/core.py
+++ b/dimos/core/core.py
@@ -15,22 +15,56 @@
 
 from __future__ import annotations
 
+import asyncio
+import functools
+import inspect
 from typing import (
     TYPE_CHECKING,
+    Any,
+    ParamSpec,
     TypeVar,
+    cast,
 )
 
 if TYPE_CHECKING:
     from collections.abc import Callable
 
 T = TypeVar("T")
-
-from typing import ParamSpec, TypeVar
-
 P = ParamSpec("P")
 R = TypeVar("R")
 
 
 def rpc(fn: Callable[P, R]) -> Callable[P, R]:
-    fn.__rpc__ = True  # type: ignore[attr-defined]
-    return fn
+    """Mark a method as an RPC body callable across modules.
+
+    Sync methods are tagged in place. Async methods get a sync dispatcher that
+    runs the coroutine on `self._loop`:
+
+      * Caller is on self._loop (another async @rpc, a handle_*, or a
+        process_observable callback): returns the coroutine so the caller can
+        `await` it normally.
+      * Caller is on any other thread (RPC dispatcher, sync test, sync @rpc on
+        the same module): schedules the coroutine onto self._loop and blocks
+        until done.
+    """
+    if not inspect.iscoroutinefunction(fn):
+        fn.__rpc__ = True  # type: ignore[attr-defined]
+        return fn
+
+    @functools.wraps(fn)
+    def wrapper(self: Any, *args: Any, **kwargs: Any) -> Any:
+        loop = self._loop
+        if loop is None:
+            raise RuntimeError("async @rpc method called outside a running module loop")
+        try:
+            running = asyncio.get_running_loop()
+        except RuntimeError:
+            running = None
+        if running is loop:
+            return fn(self, *args, **kwargs)  # type: ignore[call-arg]
+        future = asyncio.run_coroutine_threadsafe(fn(self, *args, **kwargs), loop)  # type: ignore[call-arg, arg-type]
+        return future.result()
+
+    wrapper.__rpc__ = True  # type: ignore[attr-defined]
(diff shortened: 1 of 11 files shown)


dimos/core/core.py
```diff
@@ -34,3 +38,35 @@
 def rpc(fn: Callable[P, R]) -> Callable[P, R]:
     fn.__rpc__ = True  # type: ignore[attr-defined]
     return fn
+
+
+def arpc(fn: Callable[..., Coroutine[Any, Any, Any]]) -> Callable[..., Any]:
```
leshy: you can autodetect if function is async, so we can have a same @rpc decorator for both

dimos/core/module.py
```diff
@@ -396,6 +409,227 @@ def get_skills(self) -> list[SkillInfo]:
...
+        return future
+
+    def process_observable(
+        self,
+        observable: "Observable[Any]",
+        async_cb: Callable[[Any], Any],
+    ) -> "DisposableBase":
+        """Subscribe `async_cb` (an async function) to `observable`, dispatching
+        each emitted value onto self._loop. Invocations are serialized through a
+        per-subscription dispatcher task with LATEST coalescing. The subscription
```
leshy, line 439: this means we basically drop msgs if consumer is slow? that's good just making sure

docs/usage/modules.md
```diff
@@ -188,6 +187,145 @@ coordinator.load_module(CameraModule)
 coordinator.restart_module(CameraModule)
 ```
 
+## Async modules (lock-free state)
+
+Modules contain a per-instance asyncio loop on a daemon thread (`self._loop`). It is possible to write modules using only `async def` methods so that everything runs on the same thread and you don't need to use locks. The module's auto-bound input handlers, `@arpc` methods, and `process_observable` callbacks all run on `self._loop`, and each handler subscription is serialized through a dedicated dispatcher task.
+
+### Auto-bound input handlers
+
+For every declared `x: In[T]`, if the module defines `async def handle_x(self, msg: T)`, the handler is automatically subscribed at `start()` and dispatched onto `self._loop`. Subscriptions are cleaned up at `stop()`.
```
leshy, line 196: that's great!

docs/usage/modules.md
```diff
@@ -188,6 +187,145 @@ coordinator.load_module(CameraModule)
...
+class MovementManager(Module):
+    cmd_vel: Out[Twist]
+    goal: Out[PointStamped]
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        # No lock needed. `_teleop_active` is only mutated on `self._loop`.
+        self._teleop_active = False
+
+    async def handle_clicked_point(self, msg: PointStamped) -> None:
```
leshy, line 218: this is confusing to me, should't this module have

`clicked_point: In[PointStamped]` defined?

leshy (changes requested):
super awesome, small questions

dimos/core/module.py
paul-nechifor: Yes.

docs/usage/modules.md
paul-nechifor: It should. Initially I started with the real module which had all of these:

```
    clicked_point: In[PointStamped]
    nav_cmd_vel: In[Twist]
    tele_cmd_vel: In[Twist]

    goal: Out[PointStamped]
    way_point: Out[PointStamped]
    cmd_vel: Out[Twist]
    stop_movement: Out[Bool]
```

and I removed the ones which didn't seem relevant. But `*` in vim doesn't match partial sequences so `clicked_point` didn't match `handle_clicked_point` and I thought it had no uses.

I've added it back.

dimos/core/core.py
paul-nechifor: Done.

leshy (approved):

PR #1921, opened 2026-04-27
fix(types): resolve mypy 3.10 errors

jeff-hykin:
## Description

CI was failing with 14 mypy errors under `--python-version 3.10`. This PR resolves all of them without changing runtime behavior.

- `memory2/vis/plot/{elements,plot}.py` — replaced `enum.StrEnum` (3.11+) with the existing-in-codebase `(str, Enum)` mixin pattern (matches `Voice`, `Go2Mode`).
- `perception/detection/type/imageDetections.py` — replaced `typing.Self` (3.11+) with the version-guarded import pattern already used in `core/resource.py`.
- `simulation/engines/mujoco_shm.py` — bound `.copy()` results to typed locals so the `NDArray[np.float64]` return type isn't an `Any` leak.
- `protocol/rpc/pubsubrpc.py` — extended the existing `type: ignore[assignment]` to `[misc, assignment]` to cover the `hasattr`-guarded `super().__getstate__()` call.
- `msgs/sensor_msgs/Image.py` — added `# type: ignore[arg-type]` on `memoryview(np.ascontiguousarray(...))` (numpy implements the buffer protocol but isn't typed as `Buffer`).
- `perception/detection/module2D.py` — bound `imageDetections.filter(...)` to a typed local to drop the `Self` → `Any` leak.

## How to Test

Run mypy on both supported Python versions:

```
uv sync --frozen --all-extras --no-extra dds --no-extra cuda
uv run mypy --python-version 3.10 dimos
uv run mypy --python-version 3.12 dimos
```

Expect: clean on both.

Closes DIM-XXX

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
           git clean -ffdx -e .venv
 
       - name: Install Python dependencies
-        run: uv sync --all-extras --no-extra dds --frozen
+        run: uv sync --frozen --all-extras --no-extra dds --no-extra cuda
 
       - name: Remove pydrake stubs
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
           git clean -ffdx -e .venv
 
       - name: Install Python dependencies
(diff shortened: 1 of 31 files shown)


mustafab0 (dismissed):
👍

paul-nechifor (approved):

PR #1922, opened 2026-04-27
fix: rerun viewer not auto-opening

jeff-hykin:
misc rerun fix


dimos/visualization/rerun/bridge.py
```diff
@@ -291,7 +298,7 @@ def start(self) -> None:
             entity: 1.0 / hz for entity, hz in self.config.max_hz.items() if hz > 0
         }
 
-        rerun_init("dimos")
+        rr.init("dimos")
```
jeff-hykin: Can't use rerun_init because we need to do the grpc work before `register_colormap_annotation`

paul-nechifor (dismissed):

paul-nechifor (dismissed):

dimos/visualization/rerun/bridge.py
leshy: I need to init rerun without a bridge for memory stuff, which "grpc work" do you need to do? what's the issue?

leshy (changes requested):
need info here

PR #1958, opened 2026-05-01
feat(sim): MuJoCo sim teleop for xarm + piper

ruthwikdasyam:
## Problem

Quest teleop only had a MuJoCo sim variant for xarm7
Camera allignment correction req in sim

Closes DIM-XXX

## Solution

- Added support for xarm 6 and piper
- added environemnts with objects on table to pick
- added eye-in-hand camera for arms in sim
- corrected camera posiitoning with xarm

## Breaking Changes
None
## How to Test

Each command pops a MuJoCo viewer; with a Quest connected at
`https://<host-ip>:8443/teleop`:

```bash
dimos run teleop-quest-xarm6-sim
dimos run teleop-quest-piper-sim
dimos run teleop-quest-xarm7-sim
```


dimos/control/blueprints/teleop.py
```diff
@@ -167,6 +182,111 @@
     }
 )
 
+# XArm7 TeleopIK in MuJoCo sim
+coordinator_teleop_sim_xarm7 = autoconnect(
+    ControlCoordinator.blueprint(
+        tick_rate=100.0,
+        publish_joint_state=True,
```
paul-nechifor, line 189: Many of these are default values. There's no point in having defaults if we always specify them for everything.

paul-nechifor (approved):

dimos/control/blueprints/teleop.py
ruthwikdasyam: True. Lot of Blueprints in Maipulation has this pattern to specify every argument. Its because to underst every conig that is set, in-case we dont know what the default's are. But, will clean this.

PR #1990, opened 2026-05-06
go2 cli

leshy:
allows you discover robots 

both on your LAN network (find IP) 
and via BT (find robots that are not on wifi) and set the wifi settings via BT (no need for app, logins and unbinding, and claude is able to setup a robot for you)

new dep for bluetooth (under go2 extras)
`uv sync --all-extras --no-extra dds`

```sh
dimos go2tool discover
dimos go2tool connect-wifi
```


dimos/robot/unitree/go2/cli/landiscovery.py
```diff
@@ -0,0 +1,214 @@
...
+                break
+
+
+def discover(timeout: float = 2.0, iface_ip: str | None = None) -> list[Go2Device]:
+    """Probe the LAN for Go2 robots.
+
+    Args:
+        timeout: seconds to wait for replies after sending the probe.
+        iface_ip: pin multicast to this local IPv4 address. If None, probe every
+            non-tunnel interface in parallel.
```
mustafab0: It says probe interfaces in parallel. but we only do a for loop scan.

mustafab0 (dismissed):

dimos/robot/unitree/go2/cli/landiscovery.py
leshy: I know, didn't want to work on this too much, removed "in parallel" claim

mustafab0 (dismissed):
Tested the discover command, got the robot IP works 👍 

Did not setup the wifi. A bit more documentation on that would be nicer to know exactly what's going to happen.

leshy:
> Tested the discover command, got the robot IP works 👍
> 
> Did not setup the wifi. A bit more documentation on that would be nicer to know exactly what's going to happen.

didn't want to merge without main docs, added now

mustafab0 (approved):

PR #1992, opened 2026-05-06
go2 lidar timestamps bugfix

leshy:
GO2 webrtc stream has a bug with bizzare timestamps occasionally attached to lidar frames

this very simple hack corrects timestamps that are non monotonic because we know the expected lidar frequency

### DIFF
--- a/dimos/robot/unitree/type/lidar.py
+++ b/dimos/robot/unitree/type/lidar.py
@@ -14,13 +14,19 @@
 
 """Unitree WebRTC lidar message parsing utilities."""
 
-import time
-from typing import TypedDict
+from collections.abc import Callable
+import logging
+from typing import TypedDict, TypeVar
 
 import numpy as np
 import open3d as o3d  # type: ignore[import-untyped]
+from reactivex import operators as ops
+from reactivex.observable import Observable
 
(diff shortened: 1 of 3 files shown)


dimos/robot/unitree/type/lidar.py
```diff
@@ -68,7 +76,32 @@ def pointcloud2_from_webrtc_lidar(raw_message: RawLidarMsg, ts: float | None = N
...
         frame_id="world",
     )
+
+
+T = TypeVar("T", bound=Timestamped)
+
+
+def repair_stale_ts(default_period: float = 0.130) -> Callable[[Observable[T]], Observable[T]]:
+    """Repair Unitree's stale-stamp bug by forward-extrapolating non-monotonic stamps.
+
```
paul-nechifor, line 89: Instead of hardcoding 0.130 maybe you can compute the default period. Is this a unitree derived constant? Can we rely on it? Does it change over time?

paul-nechifor (approved):

dimos/robot/unitree/type/lidar.py
leshy: this is unitree lidar rate derived constant, it might change if they change lidar/dog firmware.. 
generally I agree with you, I just didn't want to spend much more time on this, we can create as low priority issue

PR #1996, opened 2026-05-06
Ivan/fix/go2recording

leshy:
memory2 recorder wasn't appending poses to Observations stored.
now Recorder module uses the transform system to find global coordinates and timestamp of a message (according to its frame ID)


dimos/memory2/codecs/test_codecs.py
```diff
@@ -130,15 +129,23 @@ def _jpeg_case() -> Case | None:
 
         TurboJPEG()  # fail fast if native lib is missing
 
-        replay = TimedSensorReplay("go2_bigoffice/color_image")
-        frames = [replay.find_closest_seek(float(i)) for i in range(1, 4)]
-        codec = JpegCodec(quality=95)
+        from dimos.memory2.store.sqlite import SqliteStore
+        from dimos.utils.data import get_data
```
paul-nechifor: Please move imports to the top. We always have get_data.

dimos/memory2/codecs/test_codecs.py
```diff
@@ -130,15 +129,23 @@ def _jpeg_case() -> Case | None:
...
+
+        db_path = get_data("go2_short.db")
```
paul-nechifor: Why would `get_data` throw `ImportError`? This try-except is just because turbojpeg might not be installed.

leshy: sorry should be below try/except

dimos/memory2/codecs/test_codecs.py
leshy: I'd like to leave SqliteStore which is not small, these tests do do very selective imports per case, for that reason I left get_data also since used only here. turbojpeg is in some kind of limbo right now, not provided by images etc, we need to resolve this, for now doing this stupid detection

PR #2000, opened 2026-05-07
feat(agents): agent-encode PoseStamped + position nav skills

### LINKED ISSUE #1693: agent encode for PoseStamped
if my skill returns some posestamped (and other stuff potentially)

PoseStamped needs to be rendered to an agent (maybe as text description relative to the robot body) so agent knows "this pose is 5 meters to my 3 o clock" and more then that - agent should be able to ask to navigate to it, rotate towards it etc (I guess can just support global coordinate nav)

This way I can ask the agent "what is the coolest place you saw last 5 minutes" this is a memory2 query - probably semantic query, returns images and poses, agent should be able to look at the image, be able to describe it, and navigate to pose

douglasswng:
## Problem

LLM agents can't act on raw `PoseStamped` data. When a skill or memory query returns a pose (e.g. memory2 returning the location of an interesting place), the agent has no way to render it for reasoning or hand it back to a navigation skill. This is the missing layer between perception/memory output and agent reasoning/action.

Closes #1693

## Solution

Two pieces, designed so the encode output and skill inputs share the same keys (agent can pipe the dict straight through):

1. `PoseStamped.agent_encode()` — returns a flat dict (`frame_id`, `x`, `y`, `z`, `yaw_deg`). Position rounded to 3 dp, yaw to 1 dp degrees, so the agent sees human-scale numbers instead of float noise.

2. Three new skills on `NavigationSkillContainer`:
   - `navigate_to_position(x, y, yaw_deg, frame_id)` — go to an absolute pose.
   - `rotate_toward_position(x, y, frame_id)` — face a target without translating; yaw computed from current odom.
   - `current_pose()` — return the robot's pose so the agent can reason about relative geometry (e.g. distance, bearing) without us baking that into the encode.

Also extracted a `_make_position_goal` helper now used by the three new skills and the existing `_get_goal_pose_from_result`.

**Why absolute, not body-relative.** MCP auto-renders skill returns via `result.agent_encode()` with no args ([mcp_server.py:130-132](dimos/agents/mcp/mcp_server.py#L130-L132)). Making the output body-relative requires an `observer` somewhere — and since the call site is MCP, that means coupling a transport layer to live odometry (per-tool subscriptions, hidden state between skill return and agent input). Absolute coords + a `current_pose` skill avoids that: the agent composes the relative view itself, the rendering layer stays context-free.

## Breaking Changes

None.

## How to Test

```bash
uv run pytest dimos/msgs/geometry_msgs/test_PoseStamped.py \
              dimos/agents/skills/test_navigation.py -v
```

Covers `agent_encode` (fields + rounding), each new skill (happy path), and the no-odom branches of `rotate_toward_position` / `current_pose`.

**Not tested end-to-end:** I don't have an OpenAI API key or local GPU, so I couldn't run an agentic blueprint and watch the LLM actually invoke these skills. Unit tests cover the skill internals; the agent-facing contract (docstring → schema, return-string interpretability) is untested. Happy to coordinate on a smoke test — e.g., spin up `unitree-go2-agentic` (or any blueprint with `McpClient`) and prompt the agent to fetch its current pose and navigate to (1, 0).


dimos/agents/skills/navigation.py
```diff
@@ -243,6 +252,78 @@ def _navigate_using_semantic_map(self, query: str) -> str:
...
+
+        if self._latest_odom is None:
+            return "No odometry data received yet, cannot rotate."
+
+        cur_x = self._latest_odom.position.x
+        cur_y = self._latest_odom.position.y
+        yaw_rad = math.atan2(y - cur_y, x - cur_x)
+
+        goal = _make_position_goal(cur_x, cur_y, yaw_rad, frame_id)
+        return self._navigate_to(goal, f"Rotating to face ({x:.2f}, {y:.2f}) in {frame_id}")
```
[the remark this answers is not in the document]
douglasswng: In dimos, map↔world is published as an identity TF (dimos/navigation/rosnav.py:187-194), and existing skills follow the same implicit-frame convention — tag_location reads odom (frame "world") and stamps the result "map" (navigation.py:162-166), and _get_goal_pose_from_result hardcodes "map" (navigation.py:357). A frame_id string-equality check would actively break the default Go2 flow (odom publishes frame_id="world", default skill arg is "map") without catching any real geometric mismatch. Proper handling needs a TF-tree lookup applied uniformly across all nav skills — separate PR.

dimos/msgs/geometry_msgs/PoseStamped.py
```diff
@@ -82,6 +82,23 @@ def __str__(self) -> str:
             f"euler=[{math.degrees(self.roll):.1f}, {math.degrees(self.pitch):.1f}, {math.degrees(self.yaw):.1f}])"
         )
 
+    def agent_encode(self) -> dict[str, Any]:
```
paul-nechifor: PoseStamped is not just for simple land robots. It's used for drones too, so roll and pitch should not be excluded.

dimos/agents/skills/navigation.py
```diff
@@ -243,6 +252,78 @@ def _navigate_using_semantic_map(self, query: str) -> str:
         message = f"Found a location in the semantic map matching '{query}'."
         return self._navigate_to(goal_pose, message)
 
+    @skill
+    def navigate_to_position(
+        self,
+        x: float,
+        y: float,
+        yaw_deg: float = 0.0,
+        frame_id: str = "map",
```
paul-nechifor: The coordinates are always absolute so I don't think it makes sense to give the option of selecting a `frame_id`.

dimos/msgs/geometry_msgs/PoseStamped.py
douglasswng: Good call — agent_encode lives on the generic msg type, so baking in yaw-only was a ground-robot assumption in the wrong place. Added roll_deg and pitch_deg in 2b10b9893.

dimos/agents/skills/navigation.py
douglasswng: Agreed — dropped `frame_id `from both new skills and documented the map-frame assumption in the docstrings (04bb0914d). Can bring it back when there's a concrete use case for non-map frames.

douglasswng:
Addressed both comments — see threads. Also retargeted base from `legacy-dev-dont-merge` to `main` (the original base was renamed after this PR opened; no code change, branch was already on top of current main). Ready for another look.

paul-nechifor:
@douglasswng I think the issue ( https://github.com/dimensionalOS/dimos/issues/1693 ) is suggesting that @skill functions should be able to return PoseStamped and it would be automatically transformed to a format usable by agents, not that agent_encode is called manually.

douglasswng:
> @douglasswng I think the issue ( #1693 ) is suggesting that @Skill functions should be able to return PoseStamped and it would be automatically transformed to a format usable by agents, not that agent_encode is called manually.

You're right, fixed.

**What changed:**
- `current_pose -> PoseStamped`. The MCP server's auto-render path (`hasattr(result, "agent_encode")` in `mcp_server.handle_tools_call`) renders it.
- No-odom raises `RuntimeError`; `mcp_server` surfaces it as `"Error running tool 'current_pose': ..."`.
- `PoseStamped.agent_encode()` was returning a flat `dict[str, Any]`, which would have crashed `McpClient` (it iterates `result["content"]` expecting a list of MCP content blocks). Fixed it to `[{"type": "text", "text": json.dumps({...})}]`, matching `Image.agent_encode()` and `SkillResult.agent_encode()`.

**Out of scope — separate PR:**
- `Vector3.agent_encode()` and `Object3D.agent_encode()` have the same flat-dict shape and would crash the auto-render path if returned from a skill. Currently latent — no `@skill` returns either type.

paul-nechifor:

Issue #1998, opened 2026-05-07
consider moving LFS download to S3


leshy:
LFS might be a good way to upload stuff for devs (version management etc)

but:
- seems slow to download
- we keep running into budget issues https://github.com/dimensionalOS/dimos/issues/1997
- it requires dev installation of lfs extension

we can consider moving LFS download (not upload) to a mirror like s3 (needs to be pluggable, is there a china mirror that supports s3 api?) devs should be able to easily point their system to an arbitrary mirror

keep in mind mirror needs to be version controled like lfs, it's possible for multiple branches to have different versions of the same file in LFS, so probably on the mirror level we store files named by their hashes? how do we deal with deletions? (if branch is deleted)

### quick hack?

is it possible to just own the LFS host? deploy our own LFS host in both China and west?

spomichter:
PR: https://github.com/dimensionalOS/dimos/pull/2001

PR #2001, opened 2026-05-07
feat(data): replace git LFS with S3 for test data downloads

spomichter:
## Summary

Replace git LFS with S3 for fetching test data archives, addressing recurring LFS budget exhaustion and slow downloads. Refs #1998.

- `dimos/utils/data.py`: lazy S3 download via `boto3` (with unsigned-access fallback for public read), replacing `git lfs pull` smudging
- `bin/s3_push`: new compress + upload script (replaces `bin/lfs_push`)
- `bin/hooks/lfs_check`: hook now directs devs to `bin/s3_push`
- `.gitattributes` / `.gitignore`: drop LFS filters; stop tracking `data/.lfs/` in git
- Bucket / prefix / region overridable via `DIMOS_DATA_S3_BUCKET`, `DIMOS_DATA_S3_PREFIX`, `DIMOS_DATA_S3_REGION`

The `LfsPath` API and `get_data()` function are unchanged — all callers continue working without modification.

## Notes

- Default bucket: `s3://dimos-github-lfs/.lfs/` (us-east-2)
- 71/75 archives mirrored from LFS to S3
- 4 files (`go2_sf_office`, `models_graspgen`, `piper_description`, `xarm_description`) remained as LFS pointer stubs locally and need to be sourced from a teammate's cache or a temp LFS budget bump before tests that depend on them can pass
- Branch-level data versioning (e.g. dev vs main pointing at different SHAs of the same archive) isn't handled by the flat-key S3 layout in this PR — see issue for discussion of moving to content-addressed keys later

## Test plan

- [x] \`pytest dimos/utils/test_data.py\` — 12/12 pass (10 mirrored from LFS + 2 new edge cases)
- [x] Full \`pytest\` suite: 1615 passed / 26 skipped / 1 failure + 3 errors that all pass in isolation (pre-existing flakiness, not S3-related)
- [x] End-to-end round-trip via \`bin/s3_push\` → S3 → \`get_data()\` → decompress → contents verified
- [x] Local SHA256 of 70/75 archives verified against current \`dev\` LFS pointers
- [ ] Verify CI passes
- [ ] Recover the 4 remaining pointer stubs

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,17 +4,13 @@
 # Ensure Python files always use LF for line endings.
 *.py text eol=lf
 # Treat designated file types as binary and do not alter their contents or line endings.
-*.png filter=lfs diff=lfs merge=lfs -text binary
-*.jpg filter=lfs diff=lfs merge=lfs -text binary
-*.jpeg filter=lfs diff=lfs merge=lfs -text binary
+*.png binary
+*.jpg binary
+*.jpeg binary
 *.ico binary
 *.pdf binary
-# Explicit LFS tracking for test files
-/data/.lfs/*.tar.gz filter=lfs diff=lfs merge=lfs -text
-*.onnx filter=lfs diff=lfs merge=lfs -text binary
-*.mp4 filter=lfs diff=lfs merge=lfs -text binary
-*.mov filter=lfs diff=lfs merge=lfs -text binary
-*.gif filter=lfs diff=lfs merge=lfs -text binary
-*.foxe filter=lfs diff=lfs merge=lfs -text binary
-docs/capabilities/memory/assets/** filter=lfs diff=lfs merge=lfs -text
-docs/capabilities/memory/assets/.gitattributes -filter -diff -merge text
+*.onnx binary
+*.mp4 binary
+*.mov binary
+*.gif binary
+*.foxe binary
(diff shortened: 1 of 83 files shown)


leshy:
## try S3 backed LFS self hosting first

to not re-invent LFS (branching, third party devs etc) benchmark self hosted LFS vs WGET to s3 to compare if there are speed diffs (LFS seems slow sometimes, idk if due to github) to decide if we need raw S3 mirroring or just our own LFS servers 

## branching, third party devs

this needs solving for above to be viable, how do I test a third party branch that added a new big file? how do I test a branch that overrides existing file? when do these files get deleted?

### we can solve all this:


keep using LFS as a branching/identity mechanism, defer to S3 as cache. store files as hashes in S3 and not filenames, 

```sh
cat data/.lfs/xarm6.tar.gz 
version https://git-lfs.github.com/spec/v1
oid sha256:71c9990ab779d20b878ec4c6b0ee21b29bbb7963b59375acf9dd635b7241009b
size 1863215
```
(actual .lfs stub files store a hash - use that hash for your s3 filename) 
keep using lfs but our dimos `get_data` function can try our S3 caching layer first, and auto-defer to LFS if it 404s) 

`GET our_bucket/71c9990ab779d20b878ec4c6b0ee21b29bbb7963b59375acf9dd635b7241009b`

this allows us to whenever we want - upload whatever we want. If we don't upload, everything still works.

Uploading just current dev or main to S3, 99% of traffic goes from S3

PR #2004, opened 2026-05-07
refactor(core): memory2 for dtop

aclauer:
## Problem

We want to plot metrics from dtop. Previously I saved the metrics to jsonl files and made a stand alone cli tool for the plotting, but a much better way is to just use memory2.

Closes DIM-XXX

## Solution

Add streams for the metrics tracked by dtop and add them to a memory2 Sqlite store. By default, running dtop will not save the metrics. Pass `dtop --log` to save. It will create a default db at `dtop_<date and time>.ignore.db`. Then use that as you would normally with memory2.

Minor change: move legend for memory2 plots off of axis so data isn't covered.

## Breaking Changes

dtop-plot no longer exists.

## How to Test

terminal 1
`dimos --replay --dtop run unitree-go2`

terminal 2
`dtop --log`

```python
from dimos.memory2.vis.plot.plot import Plot

store = SqliteStore(path="dtop_20260507_122933.ignore.db")
Plot().add(store.stream("GO2Connection_pss"), label="GO2Connection").to_svg("memory.svg")
```
dtop_20260507_150323 ignore_cpu_percent
Produces some nice plots, like this cpu usage on the Orin Nano

memory
As you can see, the memory usage of Go2 connection in replay has an issue.


leshy (approved):
this is great, question is how do we analize this data, we have this hacky jupyter notebook thing

```python
md-babel-py run yourfile.md
```

executes code blocks, produces

https://github.com/dimensionalOS/dimos/blob/dev/docs/capabilities/memory/plot.md

so you can build an analysis, agents are pretty good at this. idk if we should have a template file that people can run, or docs, or realtime rendererr for rerun. memory2 has structures for realtime rerun rendering in place, just didn't write it yet

on where we store,

need to merge https://github.com/dimensionalOS/dimos/pull/1996
`dimos run unitree-go2-memory` runs go2 that auto-records all sensors

so it would make sense for dtop also to record into the same db? can see later how

PR #2021, opened 2026-05-08
fix: adaptive timestamp correction

aclauer:
## Problem

Lidar timestamps on older go2 firmware doesn't get updated

Closes DIM-XXX

## Solution

Do a "calibration" for the first 10 lidar frames. If the timestamps are never updated, set all future timestamps to use system time. Otherwise for invalid timestamps increment by the default period.

## Breaking Changes

None?

## How to Test

Logs show the results of the calibration period. You can plot with memory2 to see the timestamps as well.


dimos/robot/unitree/type/lidar.py
```diff
@@ -84,24 +85,56 @@ def pointcloud2_from_webrtc_lidar(raw_message: RawLidarMsg, ts: float | None = N
...
+
+        if use_system_time:
+            item.ts = now()
+            return item
+
+        if not calibrated:
+            if prev_raw is not None and item.ts != prev_raw:
+                calibrated = True
+                logger.info(
+                    "repair_stale_ts: sensor stamps advancing — using sensor time after %d-frame calibration",
```
leshy: this is kinda default behaviour so let's not emit msgs

leshy (approved):

PR #2025, opened 2026-05-08
fixes replay mem leak

leshy:


aclauer:
Does in fact fix the issue

With this fix (note the scale is different):
post_fix

Before fix:
pre_fix

paul-nechifor (approved):

PR #2029, opened 2026-05-09
chore: add problems back

paul-nechifor:
## Description

* Add Problem/Solution back instead of Description


leshy (approved):
I don't think Problem/Solution fits many PRs but if you care about this not blocking from my side

mustafab0 (approved):

PR #2033, opened 2026-05-09
Pim/feat/g1 groot wbc

Nabla7:
## Problem

dimos doesn't have a whole-body locomotion controller for the Unitree G1 humanoid. The `ControlCoordinator` / `WholeBodyAdapter` machinery from #1954 is set up to host one — there just isn't a task that drives the legs and waist.

Closes DIM-XXX

## Solution

A `GrootWBCTask` that runs the GR00T balance + walk ONNX policies inside the coordinator tick loop, and two blueprints (`unitree-g1-groot-wbc` real-hw + `unitree-g1-groot-wbc-sim`) that compose it. Sim and real run identical task code; only the `WholeBodyAdapter` underneath differs — sim uses an in-process MuJoCo adapter, real hardware uses Mustafa's `transport_lcm` bridge to `G1WholeBodyConnection` (no new G1 low-level adapter).

The coordinator gains `activate` / `dry_run` / `twist_command` ports duck-typed to any task exposing `arm()` / `set_dry_run()` / `set_velocity_command()`. `TaskConfig` gains `hardware_id`, `default_positions`, `auto_arm`, `auto_dry_run`, `default_ramp_seconds`, `decimation`. `WholeBodyConfig` carries per-joint kp/kd; `ConnectedWholeBody` resolves them when building `MotorCommand`s. `WebsocketVisModule` exposes Arm + Dry-Run buttons that publish on the new ports — the operator activates real-hw runs from the dashboard.

Real-hardware safety profile: comes up unarmed + dry-run + 10-s ramp from current pose to the bent-knee default. Sim auto-arms with no ramp.

`G1WholeBodyConnection` patched for macOS: passive `Init(None, 0)` + `Read()`-per-tick (the callback-based variant doesn't fire reliably under cyclonedds on Darwin), and `mode_machine` hardcoded to the G1's static value with a one-shot warning if a future firmware reports something else.

## Breaking Changes

None. Additive — existing blueprints unchanged. `WholeBodyConfig` is new (`kp`/`kd` were previously direct fields on `HardwareComponent`; the migration is a constructor rename).

## How to Test

**Unit:**

```bash
uv run pytest dimos/control/tasks/test_groot_wbc_task.py \
              dimos/control/test_control.py \
              dimos/robot/unitree/g1/blueprints/basic/test_unitree_g1_groot_wbc.py \
              dimos/robot/test_all_blueprints_generation.py
```

**Sim** (no hardware needed):

```bash
DIMOS_MUJOCO_VIEW=1 dimos run unitree-g1-groot-wbc-sim
```

MuJoCo viewer pops, robot stands. Open http://localhost:7779/, WASD walks the robot.

**Real hardware** (requires a G1 + ethernet + the `[unitree-dds]` extra installed; `CYCLONEDDS_HOME` exported in the shell):

```bash
ROBOT_INTERFACE=<nic> dimos run unitree-g1-groot-wbc
```

Robot connects in dev-mode damping; dashboard at http://localhost:7779/. Click **Arm** (10-s ramp to default pose), inspect the dry-run command stream, then toggle **Dry Run** OFF to hand control to the policy. WASD = cmd_vel.


dimos/control/tasks/groot_wbc_task.py
```diff
@@ -0,0 +1,626 @@
...
+    def compute(self, state: CoordinatorState) -> JointCommandOutput | None:
+        if not self._active:
+            return None
+
+        # Read our 15 claimed joints' current positions — needed for the
+        # hold-pose / ramp-start / unarmed echo paths below.
+        current_15 = np.zeros(_NUM_ACTIONS, dtype=np.float32)
+        for i, jname in enumerate(self._joint_names_list):
+            pos = state.joints.get_position(jname)
+            current_15[i] = pos if pos is not None else 0.0
```
mustafab0: This will return `current_15` as `[0*15]` if there is a packet drop or a corrupted message.
This can cause the robot to snap because it thinks it is zero position when it is actually somewhere else.

Better to use a cached pose or handle this gracefully.

dimos/control/tasks/groot_wbc_task.py
```diff
@@ -0,0 +1,626 @@
...
+            )
+
+        # Read all 29 joints from CoordinatorState in DDS order.
+        q_29 = np.zeros(_NUM_MOTORS, dtype=np.float32)
+        dq_29 = np.zeros(_NUM_MOTORS, dtype=np.float32)
+        for i, jname in enumerate(self._all_joint_names):
+            pos = state.joints.get_position(jname)
+            vel = state.joints.get_velocity(jname)
+            q_29[i] = pos if pos is not None else 0.0
+            dq_29[i] = vel if vel is not None else 0.0
```
mustafab0: should use cached or safe values in case of None. Similar to above comment

dimos/control/tasks/g1_groot_wbc_task.py
```diff
@@ -0,0 +1,626 @@
...
+        for i, jname in enumerate(self._all_joint_names):
+            pos = state.joints.get_position(jname)
+            vel = state.joints.get_velocity(jname)
+            q_29[i] = pos if pos is not None else 0.0
+            dq_29[i] = vel if vel is not None else 0.0
+
+        # IMU comes from the adapter, not CoordinatorState.
+        imu = self._adapter.read_imu()
+        gyro = np.asarray(imu.gyroscope, dtype=np.float32)
+        gravity = self._projected_gravity(imu.quaternion)
```
mustafab0: Coordinator state should also hold IMU data.

Please create issue for this.

dimos/control/coordinator.py
```diff
@@ -75,29 +76,62 @@ class TaskConfig:
 
     Attributes:
         name: Task name (e.g., "traj_arm")
-        type: Task type ("trajectory", "servo", "velocity", "cartesian_ik", "teleop_ik")
+        type: Task type ("trajectory", "servo", "velocity", "cartesian_ik", "teleop_ik", "groot_wbc")
```
mustafab0: Need better task management system. Something similar to registry for adapters

dimos/control/coordinator.py
```diff
@@ -75,29 +76,62 @@ class TaskConfig:
...
+    # current pose toward a default on arming.
+    default_ramp_seconds: float = 10.0
+    # GR00T WBC only: run policy inference every N coordinator ticks.
+    # Effective policy rate = ``tick_rate / decimation``.  The model was
+    # trained at 50 Hz, so the original convention was tick_rate=500 +
+    # decimation=10.  At tick_rate=50 set decimation=1 (matches upstream
+    # `run_g1_control_loop.py` which spins the policy directly at 50 Hz).
+    # Mismatched rates make the policy hold actions for too long and
+    # the robot tips over.  ``None`` keeps the task's own default (10).
+    decimation: int | None = None
```
mustafab0, line 94: We can skip so much verbose. one line comments are good enough in coordinator.

dimos/control/coordinator.py
```diff
@@ -262,24 +322,30 @@ def _create_twist_base_adapter(self, component: HardwareComponent) -> TwistBaseA
...
+        sim adapters use it as the MJCF path (str).  We pass it under
+        both ``network_interface`` and ``address`` so each registered
+        adapter can pick whichever is meaningful — extras flow through
+        ``**component.adapter_kwargs``.
         """
         from dimos.hardware.whole_body.registry import whole_body_adapter_registry
 
-        addr: int | str | None = component.address
+        addr = component.address
+        net_iface: int | str = 0
```
mustafab0: `net_iface` and `addr` should be just one thing.

`net_iface` should be generalized just how we do for all other robot address connection

dimos/control/coordinator.py
```diff
@@ -166,11 +209,22 @@ class ControlCoordinator(Module):
...
+    # → task.arm(); False → task.disarm().  Routed to every task that
+    # duck-types an ``arm`` method (and ``disarm`` for False).
+    activate: In[Bool]
+
+    # Input: Toggle dry-run on velocity-policy tasks.  In dry-run the
+    # policy keeps computing but the coordinator forwards no command to
+    # the adapter — operators use this to sanity-check commands on real
+    # hardware before committing motor torques.
+    dry_run: In[Bool]
+
```
mustafab0: These shouldn't be stream inputs. activate and deactivate are event driven instances.

RPC calls should suffice here

dimos/control/coordinator.py
```diff
@@ -594,12 +714,52 @@ def _on_twist_command(self, msg: Twist) -> None:
...
+        enabled = bool(msg.data)
+        with self._task_lock:
+            for task in self._tasks.values():
+                handler = getattr(task, "set_dry_run", None)
+                if callable(handler):
+                    try:
+                        handler(enabled)
+                    except Exception:
+                        logger.exception(f"set_dry_run() raised on task {task.name!r}")
+
```
mustafab0: Remove replace with rpc calls

dimos/control/coordinator.py
```diff
@@ -729,6 +900,32 @@ def start(self) -> None:
...
+        if has_dry_run:
+            try:
+                self._dry_run_unsub = self.dry_run.subscribe(self._on_dry_run)
+                logger.info("Subscribed to dry_run for dry-run routing")
+            except Exception:
+                logger.warning(
+                    "Dry-run-capable task configured but could not subscribe to dry_run. "
+                    "Use task_invoke RPC or set transport via blueprint."
+                )
+
```
mustafab0, line 677: Same as mentioned above. These should be simple RPCs

dimos/simulation/adapters/whole_body/g1.py
mustafab0, line 1: Shouldn't this file live under dimos/simulation/?

dimos/hardware/whole_body/spec.py
```diff
@@ -64,6 +86,10 @@ def is_connected(self) -> bool: ...
     def read_motor_states(self) -> list[MotorState]: ...
     def has_motor_states(self) -> bool: ...
     def read_imu(self) -> IMUState: ...
+    def read_odom(self) -> PoseStamped | None:
+        # Default: no base pose available. Sim/estimator adapters override.
+        return None
```
mustafab0: Whole body adapters will likely never publish odom so we should modify the spec.

Should be addressed in separate PR

dimos/robot/unitree/g1/blueprints/basic/_groot_wbc_common.py
mustafab0, line 1: I feel this should just be part of the `groot-wbc_task.py`

Don't see a reason why we must have this here

dimos/robot/unitree/g1/blueprints/basic/unitree_g1_groot_wbc_sim.py
mustafab0, line 1: One blueprint for real and sim, with just a simulation flag change would be what we would like

dimos/robot/unitree/g1/wholebody_connection.py
```diff
@@ -201,6 +230,28 @@ def _publish_loop(self) -> None:
...
+                            actual = int(getattr(fresh, "mode_machine", -1))
+                            if actual != self._mode_machine:
+                                logger.warning(
+                                    f"mode_machine mismatch: hardcoded "
+                                    f"{self._mode_machine}, robot reports "
+                                    f"{actual}.  Commands may be silently "
+                                    f"rejected by firmware — set "
+                                    f"_MODE_MACHINE_G1 to {actual} for this "
+                                    f"variant."
+                                )
```
mustafab0: definitely needs a refactor. 2 tasks in 1 code block

`sub` is not none is not needed either.

dimos/simulation/engines/mujoco_view_subprocess.py
mustafab0, line 1: Don't like this at all. we already have a mujoco sim module. We must not use another implementation

mustafab0 (changes requested):

dimos/web/websocket_vis/websocket_vis_module.py
```diff
@@ -330,6 +338,38 @@ async def clear_gps_goals(sid: str) -> None:
                 await self.sio.emit("gps_travel_goal_points", self.gps_goal_points)
             logger.info("GPS goal points cleared and updated clients")
 
+        @self.sio.event  # type: ignore[untyped-decorator]
+        async def arm(sid: str, data: dict[str, Any] | None = None) -> None:
```
paul-nechifor: You need to send these commands from the command center, right? Or are you sending from rerun?

dimos/web/websocket_vis/websocket_vis_module.py
```diff
@@ -107,6 +108,13 @@ class WebsocketVisModule(Module):
     stop_explore_cmd: Out[Bool]
     cmd_vel: Out[Twist]
     movecmd_stamped: Out[TwistStamped]
+    # Arming / dry-run for locomotion-policy tasks (e.g. GrootWBCTask).
+    # Uses dimos.msgs.std_msgs.Bool to match the coordinator's
+    # ``activate`` / ``dry_run`` In[Bool] ports, rather than
+    # dimos_lcm.std_msgs.Bool used by ``explore_cmd`` — the LCM wire
+    # format is identical; what matters for autoconnect is type parity.
```
paul-nechifor:
```suggestion
```

dimos/web/websocket_vis/websocket_vis_module.py
```diff
@@ -107,6 +108,13 @@ class WebsocketVisModule(Module):
...
+    activate: Out[DimosBool]
+    dry_run: Out[DimosBool]
```
paul-nechifor: Nit: you might want more specific names. This module is for all robots and t's not clear what `activate` means for any one particular robot. Maybe `groot_activate`?

dimos/simulation/engines/mujoco_view_subprocess.py
```diff
@@ -0,0 +1,152 @@
...
+            for name, q in joints.items():
+                adr = name_to_qposadr.get(name)
+                if adr is not None:
+                    data.qpos[adr] = q
+            if has_freejoint and base_pos is not None and base_wxyz is not None:
+                data.qpos[free_qposadr : free_qposadr + 3] = base_pos
+                data.qpos[free_qposadr + 3 : free_qposadr + 7] = base_wxyz
+            mujoco.mj_kinematics(model, data)
+            v.sync()
+            time.sleep(1.0 / 60.0)
```
paul-nechifor: To achieve 60 hz you need to subtract the time it took to process the iteration of the loop.

dimos/simulation/engines/mujoco_view_subprocess.py
```diff
@@ -0,0 +1,152 @@
...
+
+    js_t.stop()
+    od_t.stop()
```
paul-nechifor: This is not good practice. You have to call stop even if the code fails in the middle. You need try-finally.

dimos/simulation/engines/mujoco_view_subprocess.py
```diff
@@ -0,0 +1,152 @@
...
+            latest["base_pos"] = np.array(
+                [msg.position.x, msg.position.y, msg.position.z], dtype=np.float64
+            )
+            latest["base_wxyz"] = np.array(
+                [msg.orientation.w, msg.orientation.x, msg.orientation.y, msg.orientation.z],
+                dtype=np.float64,
+            )
+        except Exception:
+            pass
+
```
paul-nechifor: These calls are coming on different threads. You need a lock to modify latest. You also need the lock in the loop below where you access latest.

dimos/simulation/engines/mujoco_view_subprocess.py
```diff
@@ -0,0 +1,152 @@
...
+    def dimos_joint_to_mjcf(name: str) -> str:
+        parts = name.split("/", 1)
+        suffix = parts[1] if len(parts) > 1 else parts[0]
+        return f"{suffix}_joint"
+
+    # G1 GR00T MJCF references meshes by bare filename (menagerie convention);
+    # from_xml_path can't find them on disk.  Inject the dimos-bundled mesh
+    # bytes by name — same trick MujocoEngine uses.
+    try:
+        from dimos.simulation.mujoco.model import get_assets
```
paul-nechifor: Please move all import everywhere in this file to the top.

dimos/simulation/engines/mujoco_view_subprocess.py
paul-nechifor: Technically we have 2 others.

* `simulation/mujoco/mujoco_process.py` is the first one. It used to be a class, but Jeff noticed that Mujoco can only run on the main thread in a Mac. So I converted it to a standalone process with shared memory communication.
* `simulation/engines/mujoco_engine.py` was added afterwards. It runs in side a Module, which means it's not running on the main thread.
* This PR adds `simulation/engines/mujoco_view_subprocess.py`. From the comment below, I assume it was added to fix the same issue of Mujoco not working on Macs. @Nabla7 is this so?

@mustafab0 I agree that we shouldn't be adding another one. The ideal solution would be to convert `simulation/engines/mujoco_engine.py` to work as the main thread in a new process if possible.

dimos/simulation/engines/mujoco_sim_module.py
```diff
@@ -202,6 +265,35 @@ def start(self) -> None:
...
+            self._engine.model,
+            "imu-pelvis-angular-velocity",
+            "imu-torso-angular-velocity",
+            "gyro_pelvis",
+            "imu_gyro",
+            dim=3,
+        )
+        self._imu_accel_slice = _find_sensor_slice(
+            self._engine.model,
+            "imu-pelvis-linear-acceleration",
```
paul-nechifor: These are G1 specific sensor names, right? But this module isn't just for G1. I think this should be abstracted some how.

dimos/robot/unitree/g1/blueprints/basic/unitree_g1_groot_wbc_sim.py
```diff
@@ -0,0 +1,185 @@
...
+    if sys.platform == "darwin":
+        _viewer_python = shutil.which("mjpython") or shutil.which("python")
+    else:
+        _viewer_python = sys.executable
+    if _viewer_python is None:
+        logger.warning(
+            "DIMOS_MUJOCO_VIEW=1: couldn't locate mjpython/python on PATH; viewer not launched"
+        )
+    else:
+        _viewer_proc = subprocess.Popen(
```
paul-nechifor: Blueprint files are for defining blueprints. You can't just start random process because a file is imported. In any case, you'd need to manage the process (shut it down when its not needed).

dimos/robot/unitree/g1/blueprints/basic/unitree_g1_groot_wbc_sim.py
```diff
@@ -0,0 +1,185 @@
...
+            joint_names=g1_arms,
+            priority=10,
+            default_positions=ARM_DEFAULT_POSE,
+            auto_start=True,
+        ),
+    ],
+).transports(
+    {
+        ("joint_state", JointState): LCMTransport("/coordinator/joint_state", JointState),
+        ("odom", PoseStamped): LCMTransport("/odom", PoseStamped),
```
paul-nechifor: This is the default. No need to define it.

dimos/robot/unitree/g1/blueprints/basic/unitree_g1_groot_wbc_sim.py
```diff
@@ -0,0 +1,185 @@
...
+        ("activate", DimosBool): LCMTransport("/g1/activate", DimosBool),
+        ("dry_run", DimosBool): LCMTransport("/g1/dry_run", DimosBool),
+    }
+)
+
+_g1_ws_vis = WebsocketVisModule.blueprint().transports(
+    {
+        ("cmd_vel", Twist): LCMTransport("/g1/cmd_vel", Twist),
+        ("activate", DimosBool): LCMTransport("/g1/activate", DimosBool),
+        ("dry_run", DimosBool): LCMTransport("/g1/dry_run", DimosBool),
```
paul-nechifor: `.transports` applies for all the matching modules in a blueprint so it doesn't make sense to define it twice (in `_g1_coordinator` too).

If you really want to change the default topic, it's better to only define it once:

```python
unitree_g1_groot_wbc_sim = autoconnect(_g1_engine, _g1_coordinator, _g1_ws_vis).transports(
    ("dry_run", DimosBool): LCMTransport("/g1/dry_run", DimosBool)
)
```

But I think you should remove both and embrace the defaults. This applies to all. Why not just use the defaults?

dimos/robot/unitree/g1/blueprints/basic/unitree_g1_groot_wbc_sim.py
```diff
@@ -0,0 +1,185 @@
...
+from dimos.utils.logging_config import setup_logger
+from dimos.web.websocket_vis.websocket_vis_module import WebsocketVisModule
+
+logger = setup_logger()
+
+# Resolved to an absolute path so MujocoSimModule (parent) and the
+# DIMOS_MUJOCO_VIEW=1 subprocess can both open the file regardless of
+# the shell's CWD.  get_data also auto-extracts the LFS tarball on
+# first run.
+_MJCF_PATH = str(get_data("mujoco_sim/g1_gear_wbc.xml"))
```
paul-nechifor: You shouldn't call `get_data` at import time. This blocks python's imports until the data is downloaded. Import-time is just for defining (functions/variables/etc), not doing work.

dimos/robot/unitree/g1/blueprints/basic/unitree_g1_groot_wbc.py
```diff
@@ -0,0 +1,148 @@
...
+    g1_joints,
+    g1_legs_waist,
+)
+from dimos.robot.unitree.g1.wholebody_connection import G1WholeBodyConnection
+from dimos.utils.data import get_data
+from dimos.web.websocket_vis.websocket_vis_module import WebsocketVisModule
+
+_g1_connection = G1WholeBodyConnection.blueprint(
+    release_sport_mode=True,
+    network_interface=os.getenv("ROBOT_INTERFACE", "enp86s0"),
```
paul-nechifor: There's no need for environment variables.

You can see all the configs for a blueprint with:

```bash
$ uv run dimos run unitree-g1-groot-wbc --help
Blueprint arguments:
    g1wholebodyconnection:
      * g1wholebodyconnection.default_rpc_timeout: float (default: 120.0)
      * g1wholebodyconnection.frame_id_prefix: str | None (default: None)
      * g1wholebodyconnection.frame_id: str (default: g1_pelvis)
      * g1wholebodyconnection.network_interface: str (default: enp86s0)
      * g1wholebodyconnection.release_sport_mode: bool (default: True)
      * g1wholebodyconnection.publish_rate_hz: float (default: 500.0)
    controlcoordinator:
      * controlcoordinator.default_rpc_timeout: float (default: 120.0)
      * controlcoordinator.frame_id_prefix: str | None (default: None)
      * controlcoordinator.frame_id: str | None (default: None)
      * controlcoordinator.tick_rate: float (default: 500.0)
      * controlcoordinator.publish_joint_state: bool (default: True)
      * controlcoordinator.joint_state_frame_id: str (default: coordinator)
      * controlcoordinator.publish_odom: bool (default: True)
      * controlcoordinator.log_ticks: bool (default: False)
    websocketvismodule:
      * websocketvismodule.default_rpc_timeout: float (default: 120.0)
      * websocketvismodule.frame_id_prefix: str | None (default: None)
      * websocketvismodule.frame_id: str | None (default: None)
      * websocketvismodule.port: int (default: 7779)
```

So if you want to pass `network_interface` you just send:

```bash
uv run dimos run unitree-g1-groot-wbc -o g1wholebodyconnection.network_interface=enp86s0
```

See `docs/usage/cli.md`

dimos/robot/unitree/g1/blueprints/basic/unitree_g1_groot_wbc.py
```diff
@@ -0,0 +1,148 @@
...
+            wb_config=WholeBodyConfig(kp=tuple(G1_GROOT_KP), kd=tuple(G1_GROOT_KD)),
+        ),
+    ],
+    tasks=[
+        TaskConfig(
+            name="groot_wbc",
+            type="groot_wbc",
+            joint_names=g1_legs_waist,
+            priority=50,
+            model_path=os.getenv("GROOT_MODEL_DIR", str(get_data("groot"))),
```
paul-nechifor: You'll want to use LfsPath instead of get_data so the file isn't downloaded at import time.

dimos/robot/unitree/g1/blueprints/basic/test_unitree_g1_groot_wbc.py
```diff
@@ -0,0 +1,117 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Composition smoke tests for ``unitree_g1_groot_wbc`` (real-hw blueprint).
+
+These tests do not exercise DDS or actually run anything — they just
```
paul-nechifor: > These tests do not [...] run anything

A very damning statement. :P

Please delete the file. Blueprints are just declarations, not logic. There's nothing to test.

dimos/control/tasks/g1_groot_wbc_task.py
```diff
@@ -0,0 +1,626 @@
...
+    ``groot_wbc_backend.GrootWBCBackend._compute_obs`` verbatim):
+
+        [0:3]    cmd_vel * cmd_scale                # scaled velocity command
+        [3]      height_cmd                         # fixed slot (0.74)
+        [4:7]    (0, 0, 0)                          # rpy_cmd, zeros
+        [7:10]   gyro * obs_ang_vel_scale           # body-frame ang vel
+        [10:13]  projected_gravity(quat)            # gravity in body frame
+        [13:42]  (q_29 - default_29) * dof_pos_scale
+        [42:71]  dq_29 * dof_vel_scale
+        [71:86]  last_action (15 dims)
```
paul-nechifor, line 251: Are these G1 specific values? If so, either abstract it to allow for other robots, or rename the class to make it clear it's only for G1.

dimos/robot/unitree/g1/wholebody_connection.py
leshy: don't PR code like this pls, this shifts the burden of your agent output review to others

Issue #2036, opened 2026-05-09
AprilTag marker 3D detector


leshy:
For Navigation loop closure evals, and multi-embodiment robot to robot recognition we need marker detector

write a module that takes color_image and emits a TF each time a marker pose is detected  (in 3d space) (use self.tf service available on all modules)

it will need to use camera_info to estimate tag pose through the camera distortion into 3D space.

then use current base_link -> marker transform to derive world -> markers -> marker_id transform

validate on real go2 robot

paul-nechifor:
Implemented in https://github.com/dimensionalOS/dimos/pull/2107

PR #2037, opened 2026-05-09
feat Aruco/Apriltag generator

leshy:
## Problem

I wanted a nice PDF generator for april tags, quickly vibed.
Size ruler for validation, intelligent packing

getting ready for https://github.com/dimensionalOS/dimos/issues/2036

## How to Test

dimos apriltag --help

example outputs

[tags_50.pdf](https://github.com/user-attachments/files/27552657/tags_50.pdf)

[tags_75.pdf](https://github.com/user-attachments/files/27552667/tags_75.pdf)


dimos/utils/cli/apriltag.py
```diff
@@ -0,0 +1,293 @@
...
+    "a0": A0,
+    "a1": A1,
+    "a2": A2,
+    "a3": A3,
+    "a4": A4,
+    "a5": A5,
+    "a6": A6,
+    "a7": A7,
+    "a8": A8,
+}
```
mustafab0, line 70: Should also add `letter` size to support 'murica!

mustafab0 (commented):
Do we need this to be a cli utility? 

Can be just a regular utility function/script. Wouldn't hurt to have it but also use cases seem not as frequent to warrant a cli tool.

mustafab0:
Also, we should add support for ArucoTags (a separate family) and also very widely used.

leshy:
> Do we need this to be a cli utility?
> 
> Can be just a regular utility function/script. Wouldn't hurt to have it but also use cases seem not as frequent to warrant a cli tool.

good point, I kinda like that dimos `something` can quickly access all tooling we have, but it's a bit stupid to modify a single cli entrypoint file for this, I think maybe `dimos something` should just delegate to some `tool/something.py` dir of individual scripts

let us think about it, and let me merge this for now

dimos/utils/cli/apriltag.py
leshy: added

leshy:
> Also, we should add support for ArucoTags (a separate family) and also very widely used.

added

mustafab0:
`--pack` argument not working for me. I can only get 1 aruco/April tag on one page. 

I run the following `dimos apriltag --ids 4 --pack 5`

Attached is the output
[apriltags.pdf](https://github.com/user-attachments/files/27619513/apriltags.pdf)

leshy:
> `dimos apriltag --ids 4 --pack 5`

you selected only id 4 with this command

`dimos apriltag --ids 4-10 --pack`

also pack is default so you don't need to include it


`dimos apriltag --help`

mustafab0 (approved):

PR #2055, opened 2026-05-12
feat(agents): add prompt-compaction middleware for McpClient

### LINKED ISSUE #1899: Agent Compaction
* Compact the agent history when it gets too large for the current model to handle. (This depends on which model is used.)
* This needs to be tested to see how well it works in practice with real commands.

---
*Synced from* [DIM-807](https://linear.app/dimensional/issue/DIM-807) *by paul.nechifor*

Mgczacki:
## Summary

Closes #1899

Caps the prompt the dimos agent sends to its LLM so the conversation history
never grows unbounded. Implemented as a langchain `AgentMiddleware` plugged into
`create_agent(middleware=...)`. Because the hook (`before_model`) fires before
every model invocation, the input-size bound becomes an invariant of the agent
loop — including intra-turn re-invocations (model → tool → tool result → model).

On long sessions the middleware quietly summarizes older turns once it detects
an oversized prompt. Behavior is unchanged for short sessions.

### `dimos_turn`

A new integer tag attached to each message's `additional_kwargs` dict.
Incremented once per `McpClient._process_message` call — that is, once per
user-facing turn (a human input from `agent-send`, or a tool-stream
notification that wakes the agent). Every message that flows through during
that turn — the input `HumanMessage`, intermediate `AIMessage`s with
`tool_calls`, the resulting `ToolMessage`s, the final `AIMessage` — all get
stamped with the same turn number.

This is what lets compaction:

1. **Group messages by turn** so tool_call/tool_response pairs always travel
   together (compaction selects entire turns, never partial ones — no orphan
   `tool_call_id` references).
2. **Identify the current turn** (the latest tag value plus any trailing
   untagged in-flight messages from the agent loop) and preserve it untouched
   regardless of threshold.
3. **Score / inspect** the history per-turn for future heuristics (e.g.,
   keep-N-most-recent strategies).

`dimos_turn` is metadata only — it lives in `additional_kwargs`, which
providers ignore but langchain serialization preserves. The compaction
summary itself is tagged with the max turn it covers (plus
`dimos_compacted: True`), so re-compaction folds the prior summary into the
next one cleanly.

### Current turn is sacred

`_current_turn_start` walks from the end of the message list to find the
boundary of the latest turn. Everything from that boundary forward is never
compacted — no image strip, no summary touch. This protects:

- The user's current query
- In-progress tool calls and their pending `ToolMessage` responses
- Fresh images from perception that the user might be asking about right now

## How it works

Two-stage compaction inside `before_model`:

1. **Strip images** in messages older than the current turn. Image content
   blocks are replaced with a small text placeholder. If this alone gets us
   below `target_tokens`, we stop here.

   > **Caveat:** this is an incomplete solution. Dropping the image with only
   > a `"[image removed]"` placeholder is destructive as the model can no
   > longer refer back to that perception. A more principled design would
   > follow progressive disclosure: keep the image addressable in a content
   > store and replace the inline block with a reference (e.g.,
   > `[image: ref://…]`) plus a tool the agent can call to re-fetch it on
   > demand. I am deferring this decision as it  needs a broader agent-harness
   > conversation about content addressability.

As to why I decided to strip images: LLM's visual reasoning capabilities are
currently noticeably worse than with text. Additionally, the way in which the
agent loop is set up right now makes it so the model gets to see the image at the
beginning of a new turn, and it tends to give a description of what's in the image.
This description is detailed enough for reasoning about the content of the image,
but it also causes a secondary effect: The model, when considering the image, will
default to anchor its perception (even if the image is available in chat history) to the
comment it gave at the moment. Keeping images that were already observed therefore
seems like a waste of tokens that we can save since we are already going to cause a
cache burst with our compaction process.

2. **Summarize older messages** into a single `SystemMessage` while keeping
   the most recent turns verbatim. The summarizer LLM is configurable;
   defaults to reusing the agent's own model. Output is hard-capped via
   `summarizer.bind(max_tokens=summary_size_tokens)`.

## See it in action

A public Langfuse trace captured with deliberately small defaults so
compaction fires after a handful of turns:

https://us.cloud.langfuse.com/project/cmp23t80n09ooad08jnw1lksy/traces/887630cfbf49bb97f1c5b4d2cc980ad1?observation=b73fcf77cb4f2dc5&timestamp=2026-05-12T07:54:34.311Z

Use the trace timeline to see the prompt that hits the LLM at each
`agent-turn-N` span — older turns get folded into a single summary
`SystemMessage` and the agent continues with a shrunk prompt.

## Configuration

All on by default via `McpClientConfig`, env-driven:

| Env var                          | Field                            | Default |
|----------------------------------|----------------------------------|--------:|
| `AGENT_COMPACTION_THRESHOLD`     | `agent_compaction_threshold`     | `40000` |
| `AGENT_COMPACTION_TARGET`        | `agent_compaction_target`        |  `3000` |
| `AGENT_COMPACTION_SUMMARY_SIZE`  | `agent_compaction_summary_size`  |  `1000` |
| `AGENT_COMPACTION_MODEL`         | `agent_compaction_model`         |    `None` (reuses agent's model) |

## Why a middleware

Two reasons, both documented in `compaction_middleware.py`'s module docstring:

1. **Middleware vs preprocessing.** External preprocessing on `_history`
   would only fire once per user turn, leaving every intra-turn re-invocation
   unprotected. Middleware fires before *each* model call.
2. **`before_model` vs `after_model` / `wrap_model_call`.** `before_model` is
   the minimal-intervention hook. `after_model` is too late (the model
   already errored on overflow); `wrap_model_call` conflates compaction with
   the model-call concerns (retries, error shaping, tool dispatch).

### New files

- **`dimos/agents/compaction_middleware.py`** — `DimosCompactionMiddleware`
  class (subclass of `langchain.agents.middleware.AgentMiddleware`),
  placeholder token counter (3 chars/token, 1000 tokens/image; memoized in
  `additional_kwargs[\"dimos_tokens\"]` for O(new-only) recompute), static
  token cache for `system_prompt` + tool schemas, and the algorithm helpers
  (`_strip_images`, `_split`, `_current_turn_start`, `_summarize`).
- **`dimos/agents/test_compaction_middleware.py`** — 15 pytest cases,
  hermetic (no API key needed). Coverage includes:
  - Token counter unit tests (text, image, memoization, static cache)
  - `before_model` no-op below threshold
  - Stage 1 alone suffices (image strip only)
  - Stage 2 summarization with FakeListChatModel summarizer
  - Protected SystemMessage prefix preserved
  - Mid-list untagged messages get summarized (not protected)
  - Prior summary re-folded into the next summary (no stacking)
  - Most-recent turns kept verbatim
  - Tool-call/tool-response pairs never split across summarize/keep boundary
  - Summarizer failure propagates after retries
  - Two **integration tests** that drive a real `create_agent` loop with a
    `RecordingFakeAgent` and assert: (a) the agent node receives a compacted
    prompt (proves langgraph's `add_messages` reducer interprets the
    `RemoveMessage(REMOVE_ALL_MESSAGES)` sentinel correctly), and
    (b) compaction can fire mid-turn between a tool result and the next
    model call.

### Modified: `dimos/agents/mcp/mcp_client.py`

- **Config:** four new fields on `McpClientConfig` reading the env vars in
  the table above. `_env_int` / `_env_str` helpers loaded via pydantic
  `Field(default_factory=...)`.
- **Turn tagging:** new `_turn: int` counter on `McpClient` (incremented at
  the top of `_process_message`), and a new module-level `_tag_turn(message,
  turn)` helper that stamps `additional_kwargs[\"dimos_turn\"]`. Every
  message flowing through a turn gets stamped — the incoming `HumanMessage`
  first, then every message emitted by the state graph.
- **History sync:** new `_apply_messages_update` method that mirrors
  langgraph's `add_messages` reducer semantics locally — honors
  `RemoveMessage(id=REMOVE_ALL_MESSAGES)` as \"wipe history, use what
  follows\" and specific-id `RemoveMessage` as targeted removal. This keeps
  `McpClient._history` in sync with the graph's internal state even when the
  middleware replaces the entire message list.
- **Middleware wiring:** in `on_system_modules`, construct the summarizer
  (either via `init_chat_model(agent_compaction_model)`, or
  `init_chat_model(model)` if the agent's model is a string, or reuse the
  agent's model object), build the middleware with the system prompt and
  tool JSON schemas (`t.args_schema.model_json_schema()`), and pass it as
  `create_agent(..., middleware=middleware)`.
- **Robustness in the stream loop:** the worker thread now guards against
  middleware no-op updates that yield `{node: None}` instead of `{node:
  {\"messages\": [...]}}`, which would previously crash with `'NoneType'
  object has no attribute 'get'`.

### Modified: `.gitignore`

Adds `MUJOCO_LOG.TXT` (MuJoCo runtime artifact written to the repo root on
every sim run; should never be committed).

## Test plan

- [x] `uv run pytest dimos/agents/test_compaction_middleware.py -v` — 15/15
      pass.
- [x] `uv run mypy dimos/agents/compaction_middleware.py
      dimos/agents/test_compaction_middleware.py
      dimos/agents/mcp/mcp_client.py` — clean.
- [x] Live verification: `dimos --simulation run unitree-go2-agentic` with
      `AGENT_COMPACTION_THRESHOLD=2000`, drive the agent until the threshold
      is crossed, confirm a `Compaction fired (summarize)` log line appears
      and the next prompt sent to the LLM contains the summary
      `SystemMessage` instead of the older turns.

## Known limitations

Documented in the module docstring as \"Known limitations\":

1. **Image stripping is destructive** — see caveat under stage 1 above.
   Progressive disclosure with a content store is the right long-term answer.
2. **Summarizer transcript size is unbounded** — a first-ever compaction on a
   very long session could exceed the summarizer model's own context window.
   Mitigation deferred to a follow-up (chunked summarization).
3. **`@retry(on_exception=Exception)` is intentionally broad** because the
   summarizer is duck-typed; permanent errors cost up to 3 attempts + 1s of
   sleeps before propagating.


dimos/agents/compaction_middleware.py
```diff
@@ -0,0 +1,533 @@
...
+        the caller will no-op).
+        """
+        max_turn: int | None = None
+        for m in messages:
+            t = (m.additional_kwargs or {}).get("dimos_turn")
+            if isinstance(t, int) and (max_turn is None or t > max_turn):
+                max_turn = t
+
+        if max_turn is None:
+            return len(messages)
```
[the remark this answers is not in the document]
Mgczacki: This turns the middleware off and keeps the problem intact. I have added an alternative split for any agents that don't have turn-tagged messages where the compressible section starts just before the latest human message. It's a good compromise for that edge case.

dimos/agents/mcp/mcp_client.py
```diff
@@ -41,12 +46,35 @@
 logger = setup_logger()
 
 
+def _env_int(name: str) -> int | None:
+    v = os.environ.get(name)
+    return int(v) if v else None
```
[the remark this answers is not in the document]
spomichter: seems useless

dimos/agents/mcp/mcp_client.py
```diff
@@ -41,12 +46,40 @@
...
+
 class McpClientConfig(ModuleConfig):
     system_prompt: str | None = SYSTEM_PROMPT
     model: str = "gpt-4o"
     model_fixture: str | None = None
     mcp_server_url: str = "http://localhost:9990/mcp"
 
+    # Compaction: env-driven, agent-scoped. On by default.
+    agent_compaction_threshold: int = Field(
+        default_factory=lambda: _env_int("AGENT_COMPACTION_THRESHOLD") or 40_000
```
paul-nechifor: I don't think these should be env vars. The specific values depend on the model that's in use. Also, there's no need to use env vars for defaults since we have a system to override config values from the CLI (see dimos/docs/cli.md). I think it should be `-o mcpclient.agent_compaction_threshold=1234`.

dimos/agents/compaction_middleware.py
```diff
@@ -0,0 +1,564 @@
...
+    return total
+
+
+def _has_image(content: Any) -> bool:
+    if not isinstance(content, list):
+        return False
+    return any(isinstance(b, dict) and b.get("type") in ("image_url", "image") for b in content)
+
+
+class DimosCompactionMiddleware(AgentMiddleware):  # type: ignore[misc]
```
paul-nechifor, line 195: Is there a reason you're not using https://reference.langchain.com/python/langchain/agents/middleware/summarization/SummarizationMiddleware ? It seems simpler to rely on langchain features.

dimos/agents/mcp/mcp_client.py
Mgczacki: Got it, will remove env vars. I saw we are using OPENAI_API_KEY as an env var so I put them as one.

As for the per-model config, it's a valid point. We should probably adapt it as a fraction of the model's max context length. EDIT: I just saw that langchain now ships this number, it's super doable. Got it.

dimos/agents/compaction_middleware.py
Mgczacki: For future process clarity, mainly. During development the compaction logic will be iterated on beyond what langchain's default offers as more first-class structures are added to the agent harness. Examples could include: Image, Plot, Task, Goal - As these are structures in the agent state, which are observed by the agent through the context window, but don't exist only in the context window. When you deal with such structures, most agent harnesses implement custom compaction logic as it makes for more controllable behavior.

This current version's behavior is slightly less naive than the default, but barely. It gives us a place to easily integrate logic changes as the agent development progresses. If you think this is a premature optimization, we could fall back to the default for this PR and keep this implementation as future reference for when custom compaction logic is needed.

dimos/agents/mcp/mcp_client.py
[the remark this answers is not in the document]
Mgczacki: Alright, environment vars are going, so this snippet is going too.

paul-nechifor:

PR #2065, opened 2026-05-12
Update docs to --all-groups

Dreamsorcerer:
Having moved development dependencies from the extras into groups, developers will now need to install the groups.


paul-nechifor (approved):

leshy:
some issue with CI with this @Dreamsorcerer

PR #2071, opened 2026-05-13
Swastika/mint docs/new pages

swstica:
**Merge order: 1/3**- This PR can be merged independently.

## Problem

Docs folder has no Mintlify configuration and is missing logo, introduction and quickstart pages.

## Solution

Added docs.json with Mintlify config (theme, colors, logo, sidebar) and full navigation structure accounting for upcoming docs restructuring.
Added Dimensional logo in assets folder for Mintlify light/dark theme. 
Added introduction.mdx - overview of DimOS with capabilities at a glance.
Added quickstart.mdx - install instructions, replay/simulation/real robot setup, CLI and MCP usage, featured runfiles.

## How to Test
```
cd docs
npm i -g mintlify
mintlify dev
```


docs/docs.json
```diff
@@ -0,0 +1,180 @@
+{
+  "theme": "mint",
+  "name": "Dimensional",
+  "description": "Official documentation for Dimensional (DimOS) \u2014 the modern operating system for generalist robotics: Python-first, agent-native, and hardware-agnostic.",
```
paul-nechifor, line 4: On the readme we call it "The Agentive Operating System for Physical Space" (although I think agentive should be agentic)

docs/docs.json
```diff
@@ -0,0 +1,180 @@
+{
```
paul-nechifor, line 1: This PR is nice, but I think there are a few things we need in the first PR.

* We need to use a local `mintlify`. We can't rely on it being installed globally because we don't have an idea about which version to use then. You need a `package.json` and mintlify installed in it.
* We need CI checks. There needs to be a GitHub CI job which runs `mintlify validate` whenever files in `docs/**` are changed. This ensures we don't break the docs even when we're not careful and build locally.
* Currently there are several parsing errors and warnings because `<!-- Result -->` is not supported. This is because the `.md` files are included in this `docs.json` file even though they're not valid at the moment. I think we need to only use `introduction.mdx` and `quickstart.mdx` for now. Then `md-babel-py` needs to be updated to support `{/* Result */}` ( https://discord.com/channels/1341146487186391173/1491987064294609038/1504274765698044106 ). After than, then we can update all the docs to use `{/* Results */}`

leshy:
I don't think we should write docs that are not readable on github or locally, do we have to use mdx? can we either stay on standard github markdown, or exclude this into mintlify specific dir?

docs/docs.json
swstica: Updated to address changes needed:

- Added a local Mintlify setup under `docs/` with `package.json` and `package-lock.json`, so validation uses a pinned local version instead of relying on a global install.
- Added a GitHub Actions workflow that runs `npm ci` and `npm run validate` for changes under `docs/**`.
- Scoped `docs.json` to only `introduction` and `quickstart` for this first PR.
- Added `.mintignore` to exclude the legacy `.md` pages until we update the docs generator / markdown output to emit MDX-safe comments.
- Updated links in `introduction.mdx` and `quickstart.mdx` so the visible pages do not point to unavailable Mintlify routes.

Also verified locally with: `cd docs && npm ci && npm run validate` and validation passes.

.github/workflows/docs-validate.yml
Dreamsorcerer, line 1: We should probably add this as a job inside ci.yml so it can be added to the ci-complete status check.

paul-nechifor (approved):

.github/workflows/docs-validate.yml
paul-nechifor: @swstica I agree, but it should probably be done in the next PR.

PR #2080, opened 2026-05-13
feat: rust performance and api improvements

aclauer:
## Problem

Rust module needs some performance improvements and the api could be better.

Closes DIM-XXX

## Solution

New api (refer to native/rust/README.md for more docs):
- Added proc macros in `dimos-module-macros` crate for message handling and boiler plate (no more select! or match, just write the handlers)
- inputs/outputs are declared as attributes of the module struct
- setup and teardown attributes allow for initialization and cleanup 

Performance changes:
- Publish and receive are moved to separate tokio tasks so they can't block each other
- Routing is done with hashmap instead of linear search through topics. Now routing takes constant time.

Tests:
- Added tests for the publish / receive concurrency blocking

Random:
- Renamed crates to `dimos-module` and `dimos-module-macros`
- Added `cargo clippy` and `cargo fmt` to precommit. In future should add CI testing.

## How to Test

`cargo test`
`python3 rust_ping_pong.py`


native/rust/dimos-module/src/module.rs
```diff
@@ -0,0 +1,522 @@
...
+        let recv_returned = transport.recv_returned.clone();
+
+        let (publish_tx, publish_rx) = mpsc::channel(PUBLISH_CHANNEL_CAPACITY);
+        let mut builder = Builder::new(topics(&[("slow", "/slow"), ("out", "/out")]), publish_tx);
+
+        // simulate slow processing function in a receive
+        let _input = builder.input("slow", |b| {
+            std::thread::sleep(Duration::from_millis(200));
+            Ok(b.to_vec())
+        });
```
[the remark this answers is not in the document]
aclauer: the whole point of this test is what happens when the runtime is blocked...

native/rust/dimos-module/src/module.rs
```diff
@@ -0,0 +1,522 @@
...
+
+    let pub_transport = Arc::clone(&transport);
+    tokio::spawn(async move {
+        while let Some((topic, data)) = publish_rx.recv().await {
+            if let Err(e) = pub_transport.publish(&topic, &data).await {
+                eprintln!("dimos_module: publish error on {topic}: {e}");
+            }
+        }
+    });
+}
```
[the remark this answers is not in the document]
aclauer: fixed

native/rust/README.md
```diff
@@ -0,0 +1,100 @@
+# dimos rust module SDK
```
paul-nechifor:
```suggestion
# DimOS Rust module SDK
```

.pre-commit-config.yaml
```diff
@@ -108,3 +108,17 @@ repos:
         entry: python bin/hooks/filter_commit_message.py --check
         language: python
         stages: [pre-commit]
+
+      - id: cargo-fmt
+        name: cargo fmt
+        entry: bash -c "cargo fmt --manifest-path native/rust/Cargo.toml --all --check && cargo fmt --manifest-path examples/native-modules/rust/Cargo.toml --check"
+        language: system
+        types: [rust]
+        pass_filenames: false
```
paul-nechifor, line 125: Nice!

native/rust/dimos-module/src/module.rs
```diff
@@ -0,0 +1,561 @@
...
+    }
+}
+
+pub async fn run<M, T>(transport: T) -> io::Result<()>
+where
+    M: Module,
+    T: Transport,
+{
+    let mut line = String::new();
+    io::stdin().lock().read_line(&mut line)?;
```
paul-nechifor: Since run is async, you probably want to use BufReader from tokio to read the line.

native/rust/dimos-module/src/module.rs
```diff
@@ -0,0 +1,561 @@
...
+                Ok((channel, data)) => {
+                    if let Some(rs) = routes.get(&channel) {
+                        for route in rs {
+                            route.try_dispatch(&data);
+                        }
+                    }
+                }
+                Err(e) => {
+                    eprintln!("dimos_module: recv error: {e}");
+                    tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
```
paul-nechifor: Why sleep?

.pre-commit-config.yaml
```diff
@@ -108,3 +108,17 @@ repos:
         entry: python bin/hooks/filter_commit_message.py --check
         language: python
         stages: [pre-commit]
+
+      - id: cargo-fmt
+        name: cargo fmt
+        entry: bash -c "cargo fmt --manifest-path native/rust/Cargo.toml --all --check && cargo fmt --manifest-path examples/native-modules/rust/Cargo.toml --check"
```
paul-nechifor: It's probably not sustainable to mention every path. You might want to run something like:

```
    entry: bash -c '
      set -euo pipefail
      git ls-files "*Cargo.toml" | while read -r m; do
        cargo locate-project --manifest-path "$m" --workspace --message-format plain
      done | sort -u | while read -r root; do
        cargo fmt --manifest-path "$root" --all --check
      done'

```

native/rust/dimos-module/src/module.rs
aclauer: If the transport breaks or recv() otherwise starts returning immediate errors, this prevents errors from spamming the error logs.

This could be an issue for other types of transports (I don't think LCM will necessarily have this issue bc it's udp), but if we have a websocket based transport for example and the connection is killed.

native/rust/dimos-module/src/module.rs
aclauer: True

.pre-commit-config.yaml
aclauer: Haha yeah good idea, I'll change that

native/rust/dimos-module/src/module.rs
paul-nechifor: If you don't want to spam logs, I think the solution is to store when the error was last printed and only log it again after a certain deadline has passed.

Adding sleep just adds an unnecessary delay.

aclauer: Ah yeah that's definitely better. But I actually think we should just remove it, probably complicating things too much for now. If it ever becomes a real issue we will find out very quickly.

native/rust/dimos-module/src/module.rs
```diff
@@ -0,0 +1,561 @@
...
+        loop {
+            match recv_transport.recv().await {
+                Ok((channel, data)) => {
+                    if let Some(rs) = routes.get(&channel) {
+                        for route in rs {
+                            route.try_dispatch(&data);
+                        }
+                    }
+                }
+                Err(e) => eprintln!("dimos_module: recv error: {e}"),
```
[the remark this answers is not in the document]
aclauer: We'll fix this if it is a problem

.pre-commit-config.yaml
leshy: @Dreamsorcerer example of why I care about single path for CI and local dev envs for linting/code modification, think we should do what we discussed with modifying vs validating pre-commit hooks?

native/rust/dimos-module/src/module.rs
```diff
@@ -0,0 +1,561 @@
...
+
+fn propagate_task_failure(name: &str, res: Result<(), tokio::task::JoinError>) {
+    match res {
+        Ok(()) => eprintln!("dimos_module: {name} task exited unexpectedly"),
+        Err(e) => {
+            eprintln!("dimos_module: {name} task panicked, propagating");
+            std::panic::resume_unwind(e.into_panic());
+        }
+    }
+}
```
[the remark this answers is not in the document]
aclauer: there's no code path that calls abort, and i don't think we ever will. if we do, then we can add this back

native/rust/dimos-module/src/module.rs
```diff
@@ -0,0 +1,561 @@
...
+    module.setup().await;
+
+    tokio::select! {
+        _ = module.handle() => {}
+        _ = tokio::signal::ctrl_c() => {}
+        res = &mut recv_handle => propagate_task_failure("recv", res),
+        res = &mut pub_handle => propagate_task_failure("publish", res),
+    }
+
+    module.teardown().await;
```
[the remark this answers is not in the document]
aclauer: fixed

.pre-commit-config.yaml
Dreamsorcerer: Yeah, sounds good.

leshy (approved):

PR #2091, opened 2026-05-15
Skill Error codes for pick and place manipulation

mustafab0:
## Problem

Manipulation skills return prose like `"Error: Robot not found"`. The agent has to parse strings to know what failed — blocks structured retries. Phase 1 of the pick-and-place reliability plan.

Closes DIM-786

## Solution

Skills now return a SkillResult dataclass with success, message, error_code, duration_ms. Error codes are domain enums (CommonSkillError, ManipulationError). MCP auto-detects the new agent_encode() method, so the LLM sees structured JSON. Every return logs SKILL <name> result=<code> duration_ms=<n> server-side. All 18 manipulation skills migrated. Drive-by: enabled gripper on xarm7-planner-coordinator.

## How to Test
```
uv run pytest dimos/agents/test_skill_result.py dimos/manipulation/
dimos run xarm7-planner-coordinator, then python -i -m dimos.manipulation.planning.examples.manipulation_client`
_client.set_gripper(0.85) returns a SkillResult dataclass; logs show SKILL set_gripper result=OK ...
```

- [ x] I have read and approved the [CLA](https://github.com/dimensionalOS/dimos/blob/main/CLA.md).


dimos/agents/skill_result.py
```diff
@@ -0,0 +1,126 @@
...
+            return stamp(SkillResult.ok("done"))
+    """
+    t0 = time.monotonic()
+
+    def stamp(result: SkillResult) -> SkillResult:
+        result.duration_ms = (time.monotonic() - t0) * 1000.0
+        if name is not None:
+            code = result.error_code.name if result.error_code is not None else "OK"
+            logger.info(f"SKILL {name} result={code} duration_ms={result.duration_ms:.1f}")
+        return result
```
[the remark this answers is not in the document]
mustafab0: fixed

dimos/manipulation/pick_and_place_module.py
```diff
@@ -419,87 +429,95 @@ def get_scene_info(self, robot_name: str | None = None) -> str:
...
+            detections = self._detection_snapshot
+            if not detections:
+                return stamp(
+                    SkillResult.fail(
+                        ManipulationError.NO_OBJECTS_VISIBLE,
+                        "No objects visible from current position",
+                    )
+                )
+
+            lines = [f"Currently see {len(detections)} object(s):"]
```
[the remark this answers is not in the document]
mustafab0: fixed

dimos/manipulation/test_manipulation_unit.py
```diff
@@ -119,18 +119,21 @@ def test_cancel_only_during_execution(self):
 
     def test_reset_not_during_execution(self):
         """Reset works in any state except EXECUTING."""
+        from dimos.agents.skill_result import CommonSkillError
```
paul-nechifor: Move to top. Same for the rest.

dimos/agents/skill_result.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+@dataclass
+class SkillResult:
+    """Structured outcome of a ``@skill`` call.
+
+    ``error_code`` accepts any ``Enum`` so each domain can supply its own
+    code namespace (manipulation, locomotion, perception, ...). The wire
+    representation uses ``Enum.name``, which is supported by all enums.
+    """
```
paul-nechifor: This sounds like SkillResult should be a generic class, where the enum type is one of the inputs. I also think you should use Literals instead of Enum, because we have to convert to strings anyway. Something like this:

```python
  from typing import Generic, Literal, TypeVar

  CommonSkillError = Literal[
      "ROBOT_NOT_FOUND",
      "INVALID_INPUT",
      "INVALID_STATE",
      "NOT_CONFIGURED",
      "EXECUTION_FAILED",
      "EXECUTION_TIMEOUT",
  ]

  E = TypeVar("E", bound=str, default=str)

  @dataclass
  class SkillResult(Generic[E]):
      ...
      error_code: E | None = None
      ...
      @classmethod
      def ok(cls, message: str = "", **metadata: Any) -> "SkillResult[E]":
          return cls(success=True, message=message, metadata=dict(metadata))
```

dimos/agents/skill_result.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def __str__(self) -> str:
+        if self.success:
+            return f"OK: {self.message}" if self.message else "OK"
+        code = self.error_code.name if self.error_code is not None else "ERROR"
+        return f"{code}: {self.message}" if self.message else code
+
+
+@contextmanager
+def skill_timing(name: str | None = None) -> Iterator[Callable[[SkillResult], SkillResult]]:
```
paul-nechifor: Since all SkillResults have duration_ms, then maybe `@skill` should measure the time and set it instead of manually using `with skill_timing(...)`. Otherwise, all places which don't use `skill_timing` will have a misleading duration of 0.0.

dimos/agents/test_skill_result.py
```diff
@@ -0,0 +1,195 @@
...
+    """
+
+    def __init__(self) -> None:
+        self.messages: list[str] = []
+
+    def __call__(self, msg: str, *args, **kwargs) -> None:
+        self.messages.append(msg if not args else msg % args)
+
+
+def _patch_logger(monkeypatch) -> _ListHandler:
```
paul-nechifor: This should be a fixture. It takes monkeypatch which is a fixture, so it should also be a fixture.

(Also, move the import to the top)

dimos/manipulation/test_manipulation_unit.py
mustafab0: Done. MB

dimos/agents/skill_result.py
mustafab0: modified

dimos/agents/skill_result.py
mustafab0: Applied suggestions

dimos/agents/test_skill_result.py
mustafab0: Done

paul-nechifor (dismissed):

dimos/agents/annotation.py
```diff
@@ -38,6 +43,24 @@ def current_skill_context() -> dict[str, Any] | None:
...
+    """
+    # Lazy import to avoid a hard dependency cycle on package init.
+    from dimos.agents.skill_result import SkillResult
+
+    if isinstance(result, SkillResult):
+        result = replace(result, duration_ms=elapsed_ms)
+        code = result.error_code if result.error_code is not None else "OK"
+    else:
+        code = "OK"
+    logger.info(f"SKILL {func_name} result={code} duration_ms={elapsed_ms:.1f}")
```
Dreamsorcerer: There's a few places in here, but logger calls should not use pre-formatted strings:
```suggestion
    logger.info("SKILL %s result=%s duration_ms=%.1f", func_name, code, elapsed_ms)
```

This allows logger formatters greater flexibility, and avoids overhead of string formatting when the line doesn't actually get logged.

dimos/agents/annotation.py
```diff
@@ -59,10 +88,16 @@ def sync_context_wrapper(*args: Any, **kwargs: Any) -> Any:
...
             previous = getattr(_SKILL_CONTEXT, "context", None)
             _SKILL_CONTEXT.context = context
+            t0 = time.monotonic()
             try:
-                return func(*args, **kwargs)
+                result = func(*args, **kwargs)
+            except BaseException:
+                elapsed_ms = (time.monotonic() - t0) * 1000.0
+                logger.info(f"SKILL {func.__name__} result=EXCEPTION duration_ms={elapsed_ms:.1f}")
+                raise
```
Dreamsorcerer, line 102: Do you think this will become common enough to warrant a helper? e.g.

```
@contextlib.context_manager
def log_exception_timing():
    start_s = time.monotonic()
    try:
        yield
    except BaseException:
        elapsed_ms = (time.monotonic() - start_s) * 1000.
        logger.info("SKILL %s result=EXCEPTION duration_ms=%.1f", func.__name__, elapsed_ms)
        raise

...

with log_exception_timing():
    result = func(*args, **kwargs)
```

dimos/agents/test_skill_result.py
```diff
@@ -0,0 +1,182 @@
...
+    The dimos logger sets ``propagate=False``, so pytest's ``caplog`` fixture
+    can't see its records via the root logger. Instead we monkeypatch the
+    annotation module's logger info method with this collector.
+    """
+
+    def __init__(self) -> None:
+        self.messages: list[str] = []
+
+    def __call__(self, msg: str, *args, **kwargs) -> None:
+        self.messages.append(msg if not args else msg % args)
```
Dreamsorcerer: Don't think the conditional is needed.
```suggestion
        self.messages.append(msg % args)
```

dimos/agents/test_skill_result.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+@pytest.fixture
+def captured_logs(monkeypatch) -> _ListHandler:
+    """Capture ``logger.info`` calls from ``dimos.agents.annotation``."""
+    handler = _ListHandler()
+    monkeypatch.setattr(annotation.logger, "info", handler)
+    return handler
```
Dreamsorcerer: The existing caplog fixture should be used for this:
https://docs.pytest.org/en/stable/how-to/logging.html#caplog-fixture

Dreamsorcerer (dismissed):

dimos/agents/annotation.py
mustafab0: fixed

dimos/agents/annotation.py
mustafab0: Not yet I think. Timing is also needed in success path logging. and only being used in 2 places for now.

dimos/agents/test_skill_result.py
mustafab0: fixed

dimos/agents/test_skill_result.py
mustafab0: switched to caplog. deleted this code snippet

Dreamsorcerer (approved):

paul-nechifor (approved):

dimos/agents/annotation.py
paul-nechifor: Technically, we're using structlog, so this would be better:

 ```python
logger.info("SKILL", func=func_name, result=code, duration_ms=elapsed_ms)`.
```

But no point in changing it now.

PR #2094, opened 2026-05-15
feat: remove rpyc

paul-nechifor:
## Problem

* We don't want RPyC anymore.

## Solution

* Excise it.


dimos/core/coordination/blueprints.py
```diff
@@ -23,6 +24,23 @@
 
 from pydantic import create_model
 
+
+# `Blueprint` carries `mappingproxy` fields (e.g. `global_config_overrides`,
+# `remapping_map`) that are not picklable by default — `MappingProxyType` lives
+# at `builtins.mappingproxy` which isn't a real builtins attribute, so pickle
```
Dreamsorcerer: Shouldn't it just be able to get it from `types.MappingProxyType` the same as we do?

dimos/core/coordination/test_module_coordinator.py
```diff
@@ -775,26 +775,22 @@ def test_restart_preserves_remapped_streams(dynamic_coordinator) -> None:
...
+    from dimos.protocol.rpc.pubsubrpc import LCMRPC
+
+    dynamic_coordinator.start_rpc_service()
+    client = LCMRPC()
+    client.start()
+    try:
+        result, _unsub = client.call_sync(f"{COORDINATOR_RPC_NAME}/ping", ([], {}), rpc_timeout=2.0)
+        assert result == "pong"
+    finally:
+        client.stop()
```
Dreamsorcerer: This is probably a good oppurtunity to improve the class to work as a context manager?
```suggestion
    with LCMRPC() as client:
        result, _unsub = client.call_sync(f"{COORDINATOR_RPC_NAME}/ping", ([], {}), rpc_timeout=2.0)
        assert result == "pong"
```

dimos/core/module.py
```diff
@@ -767,6 +779,21 @@ def set_transport(self, stream_name: str, transport: Transport) -> bool:  # type
...
+        sentinel if no such stream exists, or `None` on timeout/error.
+
+        Used by `Dimos.peek_stream` to scan running modules.
+        """
+        stream = self.outputs.get(stream_name) or self.inputs.get(stream_name)
+        if stream is None:
+            return PeekNotFound()
+        try:
+            return stream.get_next(timeout)
+        except Exception:
```
Dreamsorcerer, line 799: There should be a lint rule for this, CC keeps doing this far too frequently.
We should be clear what exception we expect here. This masks mistakes in spelling the attribute etc.

dimos/core/coordination/blueprints.py
paul-nechifor: `types.MappingProxyType` is the real location. But it incorrectly reports its location as `builtins.mappingproxy` (if you use `__module__` and `__qualname__`). Looks like this is incorrect by choice as they don't want you to serialize/deserialize mapping proxies.

But it's true that this is ugly. I'll change Blueprints to be serializable instead.

dimos/core/coordination/test_module_coordinator.py
paul-nechifor: I've tried to use context managers before, but wasn't really liked.

The class itself isn't really used as a context manager, and I think it's not a good idea to add something if it's only meant for tests.

dimos/core/module.py
paul-nechifor: I agree, but that's what `get_next` actually raises:

```python
    def get_next(self, timeout: float = 10.0) -> T:
        try:
            return (  # type: ignore[no-any-return]
                self.observable()
                .pipe(ops.first(), *([ops.timeout(timeout)] if timeout is not None else []))
                .run()
            )
        except Exception as e:
            raise Exception(f"No value received after {timeout} seconds") from e
```

paul-nechifor: Reactivex actually throws `Exception("Timeout")`. So there's nothing that can be done here. Initially I thought the above code was wrong, but you can't capture something more finegrained than that.

Dreamsorcerer: That's... kinda insane..

leshy (approved):
very clean

Arch thoughts on this though.

I understand this is a very abstract ask but we've been careful to not commit to transport protocols at all levels (TF, stream, rpc etc). in theory (at most with a very small intervention) we can talk to different modules via different RPC protocols easily.

Would be nice for this "remote dimos interaction proto" was done with spec.py, and impl/lcm.py (only lcm) for now, so that we can switch these out or run parallel protocols in the close future, even HTTP for web dashboards, Zenoh for remote debugging etc.

This is just a note, merging this since it's a strict improvement

PR #2096, opened 2026-05-15
reworked code block result format in the docs

leshy:
updated md-babel-py to not use

> 
> something

but 

> \`\`\`results
> something
> \`\`\`

which is compatible with mintlify


docs/capabilities/memory/index.md
```diff
@@ -160,10 +159,9 @@ drawing.add(matches)
 drawing.to_svg("assets/embedding_focused.svg")
 ```
 
-<!--Result:-->
-```
+```results
 Stream("color_image_embedded") | vector_search(k=30)
-08:19:54.129 [inf][dimos/mapping/voxels.py       ] VoxelGrid using device: CUDA:0
+13:15:15.190 [inf][dimos/mapping/voxels.py       ] VoxelGrid using device: CUDA:0
```
paul-nechifor, line 164: Won't this change on every check?

docs/capabilities/navigation/nav_stack.md
```diff
@@ -78,6 +78,10 @@ create_nav_stack(
 )
 ```
 
+```results
```
paul-nechifor, line 81: Are you sure we want this? It doesn't seem useful.

docs/usage/data_streams/reactivex.md
```diff
@@ -58,12 +56,11 @@ rx.of(1, 2, 3).pipe(
...
 ```
 
-<!--Result:-->
-```
+```results
 item_1
 item_2
 item_3
-<reactivex.disposable.disposable.Disposable object at 0x7fcedec40b90>
+<reactivex.disposable.disposable.Disposable object at 0x7f5fa44d0d10>
```
paul-nechifor, line 63: ~~We probably need a way to deal with this as memory locations will always be different.~~

paul-nechifor (approved):

docs/capabilities/navigation/nav_stack.md
leshy: it's not useful but this is the output of that code block @jeff-hykin

docs/capabilities/memory/index.md
leshy: it does, I think slight confusion, we don't execute and commit docs blocks in every PR. we just execute and see if they fail.

If we do want to auto-commit back (and I think we do) we'll need some rudimentary filters like memory addresses or time above

leshy:
@swstica @paul-nechifor is good with this but lmk if any issues in docs, I actually just executed and PRed, needs a more careful reveiw

docs/capabilities/memory/index.md
paul-nechifor: Oh, I see. My impression was that the output was compared to see if it's the same.

PR #2106, opened 2026-05-16
Default to Zenoh transport on macOS and document replay workflow

bogwi:
Supersedes #1906 so self-hosted CI can run. Prior review discussion is preserved there.
---
Supersedes #1787.

This PR carries forward the Zenoh transport integration from #1787 and wraps it into a merge-ready branch that fixes the remaining macOS Big Office replay gap.

## Validation

Typical **replay on macOS** when Zenoh is installed (default is already Zenoh, so no transport flag is required):

```bash
dimos --dtop --replay run unitree-go2
```

The same workload on **Linux** (default remains `lcm` until you opt in):

```bash
dimos --transport=zenoh --dtop run unitree-go2
```

## Notes

- this PR is intended as the wrapped successor to #1787, not a separate redesign
- Linux behavior remains unchanged by default: explicit Zenoh still works, and the default transport remains LCM

### DIFF
--- a/dimos/agents/conftest.py
+++ b/dimos/agents/conftest.py
@@ -84,7 +84,7 @@ def agent_setup(request, mcp_url: str, lcm_url: str):
             AgentTestRunner.blueprint(messages=messages),
         )
 
-        global_config.update(viewer="none")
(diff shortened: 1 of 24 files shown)


paul-nechifor:

PR #2117, opened 2026-05-16
feat(unitree): forward aes_128_key to WebRTC driver for G1 firmware >=1.5.1

mihai-chiorean:
## Motivation

Unitree G1 firmware 1.5.1 introduced a `data2=3` WebRTC handshake that requires a per-device AES-128 key. Without it the connection fails inside `unitree_webrtc_connect` with `RSA key format is not supported` (pycryptodome trying to parse ciphertext bytes as an RSA key).

The per-device key is fetched from Unitree's cloud once via `unitree-fetch-aes-key --email YOU --sn <serial>` and then cached locally. The upstream driver `legion1581/unitree_webrtc_connect@v2.1.1` accepts it through an `aes_128_key=` kwarg on `UnitreeWebRTCConnection`. The dimos wrapper at `dimos/robot/unitree/connection.py` never forwarded it, so anyone on G1 firmware >=1.5.1 cannot use the dimos Unitree integration today.

## Change

`UnitreeWebRTCConnection.__init__` now accepts an optional `aes_128_key: str | None = None` kwarg and falls back to the `UNITREE_AES_128_KEY` environment variable. The value is forwarded to the underlying `LegionConnection` only when non-empty, so the call is byte-identical to the previous behaviour when unset.

```python
def __init__(self, ip: str, mode: str = "ai", aes_128_key: str | None = None) -> None:
    ...
    if aes_128_key is None:
        aes_128_key = os.environ.get("UNITREE_AES_128_KEY")
    extra: dict[str, Any] = {"aes_128_key": aes_128_key} if aes_128_key else {}
    self.conn = LegionConnection(WebRTCConnectionMethod.LocalSTA, ip=self.ip, **extra)
```

## Verification

Tested on a Unitree G1 EDU+ (model string `G1_Edu+`) at firmware 1.5.1.1. With `UNITREE_AES_128_KEY` exported in the shell, `dimos run unitree-g1-basic` proceeds past the WebRTC handshake on `192.168.123.161:9991` and the robot accepts motion-switcher / wireless-controller commands as before. Without the env var, the same command reproduces `RSA key format is not supported` and the handshake never completes — same behaviour as `main`.

## No breaking change

The kwarg is optional and defaults to `None`. The env-var lookup is opt-in (only consulted when the kwarg is `None`). Callers on G1 firmware <1.5.1 and all Go2 robots see no behavioural change — the `**extra` is empty and the `LegionConnection` call is byte-identical to the old one.

## Optional follow-up (not in this PR)

`pyproject.toml` currently pins `unitree-webrtc-connect-leshy>=2.0.7`. The `aes_128_key` kwarg + the `_decrypt_data1_v3` path that consumes it landed in `legion1581/unitree_webrtc_connect@v2.1.1`. If `unitree-webrtc-connect-leshy` is a downstream rebuild of that repo and is already on >=2.1.1, no change is needed and this PR is enough on its own. If it's still tracking 2.0.x, the dep will also need to be bumped (e.g. via `[tool.uv.sources]` pointing at `git+https://github.com/legion1581/unitree_webrtc_connect.git@v2.1.1`). Happy to follow up in a separate PR if maintainers confirm the situation.


leshy:
reason for legion client fork is that unitree-webrtc-connect-leshy had a more efficinet lidar data parser and we had some issues with aiortc package pinning and had to fork it as well

https://github.com/legion1581/unitree_webrtc_connect/compare/master...leshy:unitree_webrtc_connect:master
I assume lidar stuff is merged in the legion lib and all is resolved now

for FDEs (or anyone else here in the thread:) - we need to validate this on current g1s/go2s (ensure my lidar changes are now merged to legion client, transition to legion lib, test on our robots, make sure installation works, merge this)

dimos/robot/unitree/connection.py
```diff
@@ -85,12 +86,17 @@ def to_ndarray(self, format=None):  # type: ignore[no-untyped-def]
...
     _SPORT_API_ID_RAGEMODE: int = 2059
 
-    def __init__(self, ip: str, mode: str = "ai") -> None:
+    def __init__(self, ip: str, mode: str = "ai", aes_128_key: str | None = None) -> None:
         self.ip = ip
         self.mode = mode
         self.stop_timer: threading.Timer | None = None
         self.cmd_vel_timeout = 0.2
-        self.conn = LegionConnection(WebRTCConnectionMethod.LocalSTA, ip=self.ip)
+        # Per-device AES-128 key required by G1 firmware >= 1.5.1 (data2=3 WebRTC handshake).
```
natcl, line 102: We could also include GO2 firmware 1.1.15 as it has the same issue and this should fix it for the GO2 too !

natcl:
> had a more efficinet lidar data parser

I did notice a degradation in navigation performance on our go2 edu when I switched to the upstream library, might be worth checking if the changes are there.

mihai-chiorean:
I Was at a hackathon, but I'll get this sorted out shortly.

spomichter:
Thanks! @ruthwikdasyam or @krishna can assist with any testing here

mihai-chiorean:
Force-pushed `6bff60d44` addressing the open review items:

- **@natcl** — Go2 firmware ≥1.1.15 coverage added. `aes_128_key` now flows through `go2/connection.py` (`ConnectionConfig` field → `make_connection(..., aes_128_key=...)` → `UnitreeWebRTCConnection`) mirroring the G1 pattern, plus through `go2/fleet_connection.py` so fleet followers get the same forwarding (with an inline comment noting per-device-key heterogeneous fleets are a future per-IP-mapping follow-up).
- **@greptile-apps** — The previous `if aes_128_key is None` → `if not aes_128_key` truthiness fix landed in `fceccd7c0`. Updated the now-misleading docstring on `test_empty_string_kwarg_*` and added a sibling `test_empty_string_kwarg_uses_env_when_set` covering the YAML/JSON `aes_128_key: ""` regression you flagged.
- Added a targeted unit test `go2/test_connection.py::test_make_connection_webrtc_forwards_aes_128_key` to pin the new make_connection ↔ UnitreeWebRTCConnection contract.

Ran the changes through `codex review --commit HEAD` after each iteration. Codex caught the missed `Go2FleetConnection.start()` forwarding (now fixed) and signed off on the amended commit.

@leshy — re. the upstream legion lib transition + lidar perf, leaving that for a separate PR (also flagged on this PR thread by @natcl who noticed degraded nav on Go2 EDU after the upstream switch). Happy to scope that follow-up once you confirm the lidar parser changes have landed in legion upstream.

leshy:
I think ideally should bump to latest unitree_webrtc_connect for this to be usable.

my lidar changes have landed upstream looks like (https://github.com/legion1581/unitree_webrtc_connect/pull/46) but let's still compare perforamnce between https://github.com/leshy/unitree_webrtc_connect (what we use now) and upgraded latest official unitree_webrtc_connect

because of
> I did notice a degradation in navigation performance on our go2 edu when I switched to the upstream library, might be worth checking if the changes are there.

@ruthwikdasyam has experience with lidar data profiling so can look into this. if performance is the same, we can commit a bump in pyproject to official lib and merge all together

mihai-chiorean:
> I think ideally should bump to latest unitree_webrtc_connect for this to be usable.
> 
> my lidar changes have landed upstream looks like ([legion1581/unitree_webrtc_connect#46](https://github.com/legion1581/unitree_webrtc_connect/pull/46)) but let's still compare perforamnce between https://github.com/leshy/unitree_webrtc_connect (what we use now) and upgraded latest official unitree_webrtc_connect
> 
> because of
> 
> > I did notice a degradation in navigation performance on our go2 edu when I switched to the upstream library, might be worth checking if the changes are there.
> 
> @ruthwikdasyam has experience with lidar data profiling so can look into this. if performance is the same, we can commit a bump in pyproject to official lib and merge all together

want me to do that here or is that for a separate PR...? happy to update it, a little worried about the blast radius of updating a package like that since I'm new to the repo.

mihai-chiorean:
> > I think ideally should bump to latest unitree_webrtc_connect for this to be usable.
> > my lidar changes have landed upstream looks like ([legion1581/unitree_webrtc_connect#46](https://github.com/legion1581/unitree_webrtc_connect/pull/46)) but let's still compare perforamnce between https://github.com/leshy/unitree_webrtc_connect (what we use now) and upgraded latest official unitree_webrtc_connect
> > because of
> > > I did notice a degradation in navigation performance on our go2 edu when I switched to the upstream library, might be worth checking if the changes are there.
> > 
> > 
> > @ruthwikdasyam has experience with lidar data profiling so can look into this. if performance is the same, we can commit a bump in pyproject to official lib and merge all together
> 
> want me to do that here or is that for a separate PR...? happy to update it, a little worried about the blast radius of updating a package like that since I'm new to the repo.

@leshy?

leshy:
> > want me to do that here or is that for a separate PR...? happy to update it, a little worried about the blast radius of updating a package like that since I'm new to the repo.
> 
> @leshy?

thx for the ping, I think bump is good, then we will do IRL reviews of the whole feature and can merge right after

mihai-chiorean:
I don't want to step into some larger thing that's beyond the scope of what I was trying to fix. I'm going to close this and let someone else take care of it.

natcl:
Please leave open, this is critical otherwise newer firmwares won't work at all with dimOS..

natcl:
Hello !
If this can help speed up things, I was the one who reported a slow down in navigation when updating to the upstream untree_webrtc_connect library, I did some further testing and I don't believe the slowdowns were coming from the library change but rather from my machine. I feel this is safe to merge !

spomichter:
Hey @natcl we can prioritize this. Can you share data on any profiling you did on navigation speed / latency from your machine. And then @KrishnaH96 can test and merge

natcl:
Hello, thanks for getting back to me ! I sadly don't have any data to share right now, since my robot is on the new firmware I can't easily test with the current library, let me know however if there's something I can test with the new firmware !

PR #2107, opened 2026-05-16
AprilTag marker 3D detector 

bogwi:
Supersedes #2098 so self-hosted CI can run. Prior review discussion is preserved there.

---

This ship AprilTag marker 3D detector as per #2036 

Goes in two parts:

I ) included a camera calibration tool - dedicated utility to calibrate camera and obtain camera_info.yaml
dimos/dimos/utils/cli/cameracalibrate

`uv run pytest dimos/utils/cli/cameracalibrate`

How to calibrate is explained in the dimos/docs/usage/camera_calibration.md

---

II ) The detector module here dimos/dimos/perception/fiducial
The module has been tested in real-life detecting 
- individual tags 
- a group of twelve tags
all tried on on distances up to 4m, and various yaw, pitch, roll changes, slant changes.

verified streaming into rerun.
APRILTAG_36h11_12_A4_rerun_screenshot_2026-05-16 at 7 06 52

or partial detection, notice correctly detected tags on the bottom

APRILTAG_36h11_12_A4_rerun_screenshot_2026-05-16 at 7 17 08

---

If you have obtained for your camera a camera_info.yaml after the calibration step  then you can sub it here dimos/dimos/perception/fiducial/blueprints/fixtures/camera_info.yaml and reproduce testing steps as

Manual sequence (two terminals from `dimos` repo, `uv run dimos`):

1. term1: `uv run dimos stop` then `uv run dimos run desk-marker-tf --daemon` — note the printed `Log:` path.
2. term2: `uv run dimos rerun-bridge` — default opens a native viewer (`--rerun-open web` for browser, `none` if headless). Waits until Ctrl+C.
3. In the Rerun timeline / 3D view, expand entities under `world/tf/` and confirm `base_link`, `camera_optical`, `marker_tf/markers`, `marker_tf/marker_<id>` when the printed tag is in view (markers appear in bursts matching detection).
4. End: Ctrl+C on the bridge, `uv run dimos stop`.

Python tests are based on:
- fixture PNGs load; (obtained from rerun)
- OpenCV detects expected AprilTag IDs;
- detected corners match the generated 12-tag PDF layout via homography;
- swapped IDs/layout mismatches fail;
- MarkerTfModule publishes expected marker frames with finite transforms;
- AprilTag PDF generator layout remains stable;
- existing MarkerTf unit behavior still passes.

`uv run pytest dimos/perception/fiducial`

PR #2123, opened 2026-05-17
Ivan/feat/cameracalibrator go2

leshy:
builds upon april tag camera calibrator (https://github.com/dimensionalOS/dimos/pull/2107) makes it usable with go2

  - **Pubsub topic string registry** - very quickly vibed, we need string -> Pubsub subscribable topic function for specifying topics via cli, config etc, will iterate on this, let us ignore for now
  - **Live topic source.** `--source topic --topic <proto>:<channel>` subscribes to a running pubsub stream 
  - **Fisheye distortion model.** `--distortion-model {plumb_bob,fisheye}` (default `plumb_bob` for back-compat)
  - **New Go2 front camera calibration.** `dimos/robot/unitree/go2/front_camera_720.yaml`, fit with the new fisheye path and consumed by `GO2Connection` via `CameraInfo.from_yaml`. Moves out of `params/` and into the `go2` package directory so it ships as a packaged resource
  - **Removes references to base_link from marker detector module** - it just needs to know about world -> optical_frame
  - **Passes emit_every arg to global mapper** reduces speed to 1-2hz for global map just to chill a bit
  - **Switches go2 sensor data to system timestamps** having annoying issues aligning frames (image from webrtc with sensor data from go2, go2 onboard timestamps are also very flakey depending on a firmware)
  - **Fixes dimos topic echo** - was broken for a while 

```
uv run dimos cameracalibrate --source topic --topic "jpeg_lcm:/color_image" --cols 9 --rows 6 --square-size-m 0.02 --out ./camera_info.yaml
```


dimos/robot/unitree/go2/blueprints/smart/unitree_go2.py
```diff
@@ -53,9 +54,14 @@ class Go2Memory(Recorder):
     config: Go2MemoryConfig
 
 
-unitree_go2_memory = autoconnect(
+unitree_go2_markers = autoconnect(
     unitree_go2,
+    MarkerTfModule.blueprint(marker_length_m=0.1),
+).global_config(n_workers=11, robot_model="unitree_go2")
```
paul-nechifor, line 60: Are you sure we need that many workers? Seems rather excessive.

dimos/protocol/tf/tf.py
```diff
@@ -156,12 +166,20 @@ def get_transform(
...
+
         if simple is not None:
             return simple
 
         complex = self.get_transform_search(*args, **kwargs)
 
         if complex is None:
+            parent_frame = args[0] if len(args) > 0 else kwargs.get("parent_frame", "?")
+            child_frame = args[1] if len(args) > 1 else kwargs.get("child_frame", "?")
+            time_point = args[2] if len(args) > 2 else kwargs.get("time_point") or time.time()
```
paul-nechifor: This is odd. Just define all the expected arguments in the method, removing `*args , **wargs`:

```python
        parent_frame: str,
        child_frame: str,
        time_point: float | None = None,
        time_tolerance: float | None = None,
```

dimos/robot/unitree/connection.py
```diff
@@ -255,7 +263,8 @@ def lidar_stream(self) -> Observable[PointCloud2]:
         return backpressure(
             self.raw_lidar_stream().pipe(
                 ops.map(pointcloud2_from_webrtc_lidar),
-                repair_stale_ts(),
+                ops.map(time_is_now),
+                # repair_stale_ts(),
```
paul-nechifor, line 267: What's the issue with repair_stale_ts?

paul-nechifor (dismissed):

dimos/robot/unitree/go2/blueprints/smart/unitree_go2.py
leshy: yeah it is.. issue is that i can't control which modules get their own workers, for example memory module, rerun bridge module, connection module cannot overlap at all, it's super bad for the system.

and if you have 10 modules, and deploy 11th, it always ends up on the same worker with rerun bridge :D 

so we need some way to adjust which modules get their own process, some simple config setting perhaps

dimos/protocol/tf/tf.py
leshy: it is pretty odd will fix

dimos/robot/unitree/connection.py
leshy: different firmwares have differet weird issues that I'm investigating now, differnet firmwares behave differently, it interacted with memory in weird ways, I would like to return to repair_stale_ts asap, but for now to continue on nav evals I supressed it.

paul-nechifor (approved):

PR #2137, opened 2026-05-17
Autoresearch on relocalization

leshy:
First pass but pretty good, sub 1s alignment on a CPU from 5s of lidar data

## Next steps

better data - needs lidar frames collected from multiple independant loop closed runs. We can use this same relocalzation to align them, or use fiduciary markers for ground truth. With multiple recordings we can validate relocalization with unique never seen lidar frames into a global map.

Global mapper will try to align what it can see to a loaded map. When do we know we are ready to import loaded map into current ground truth? we need match confidence measure

better confidence measure autoresearch - we need a match confidence measure that corresponds to ground truth match quality - this way we can keep ever expanding local map, until we find a way to insert it into a lodaded global map

## How does amount of lidar points affect our alignment quality / exec time?

graph runtime or numbers of points collected vs alignment quality

autoresearch on a better dataset that includes different pc sizes

## Continuous alignment

We could still incorrectly align local map into a loaded global map - can we keep those permanently separate? always search for a better match? Can mapper change it's mind about where the loaded map should sit? How do we deal with env updates in that case?

### DIFF
--- a/dimos/mapping/relocalization/results.tsv
+++ b/dimos/mapping/relocalization/results.tsv
@@ -65,3 +65,4 @@ e7cb4ca	0.8545	0.032	1.159	60.8	discard	FPFH max_nn 100->50; lost 8 frames - fea
 f5da9da	0.9818	0.029	0.194	54.7	discard	RANSAC confidence 0.999->0.99; lost frame 1292
 f94e8ce	0.9455	0.029	0.253	54.1	discard	per-scale iters 0.2m@300k; lost 3 frames
 ba73c3f	1.0000	0.028	0.087	55.8	keep	remove final tight ICP refine; median 0.029->0.028 (slight improvement)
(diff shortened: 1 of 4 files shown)


paul-nechifor:

leshy:
I'd actually merge this, it's an artefact of work that's a good example for others, just shared with candidate.. should we keep this stuff in-repo? I think for now next to features it's ok

dimos/mapping/relocalization/data/global_map.npy
paul-nechifor, line 1: You need to put these in LFS.

dimos/mapping/relocalization/results.tsv
```diff
@@ -0,0 +1,92 @@
+commit	success_rate	median_distance	average_distance	total_seconds	status	description
```
paul-nechifor, line 1: Should this be committed?

paul-nechifor:

PR #2143, opened 2026-05-18
relocalization spec

leshy:
relocalization as a memory2 transform

Relocalization transform consumes current global map, when ready it outputs a `map -> world` transform

loaded map is already published at startup

### DIFF
--- a/dimos/mapping/relocalization.py
+++ b/dimos/mapping/relocalization.py
@@ -1,0 +1,93 @@
+# Copyright 2026 Dimensional Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


dimos/mapping/relocalization.py
```diff
@@ -0,0 +1,93 @@
...
+
+    # gives you relative pose of object in base_link frame, or None if not found
+    def lookup(self, query: str) -> Transform | None:
+        if not self._map:
+            # no relocalization until we have a map
+            return None
+
+        return Transform.from_pose(self.object_locations[query], frame_id="base_link")
+
+        return Transform.from_pose(self.object_locations[query], frame_id="base_link")
```
paul-nechifor, line 86: A few too many returns.

paul-nechifor:
I assume this isn't needed as the code is invalid. Closed.

PR #2171, opened 2026-05-19
Create backport workflow for release branches

Dreamsorcerer:
This is a workflow to create backports from main to a stable release branch when preparing for a release.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,0 +1,27 @@
+name: backport
+
+on:
+  pull_request_target:
+    types: [closed, labeled]
+
+permissions: {}
+
+jobs:
+  backport:
+    name: Open backport PR
+    if: >-
+      github.event.pull_request.merged == true &&
+      (
+        github.event.action == 'closed' ||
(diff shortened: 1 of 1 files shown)


Dreamsorcerer:
> The third-party action (`korthout/backport-action@v3`) is referenced by a mutable major-version tag rather than a pinned commit SHA, which is a supply-chain concern given the job's `contents: write` / `pull-requests: write` permissions under `pull_request_target`.

We'll resolve this across all workflows in a few months time, when Github release dependency pinning for workflows.

.github/workflows/backport.yml
```diff
@@ -0,0 +1,27 @@
+name: backport
```
paul-nechifor, line 1: Looks good. Can you document the workflow in docs/? That is, how we're supposed to add the labels in order to make the backport.

Dreamsorcerer: Coming in a separate PR, along with the actual release workflow in another PR.

paul-nechifor (approved):

PR #2081, opened 2026-05-14
feat: include DimSim

paul-nechifor:
## Problem

Closes DIM-XXX

## Contributor License Agreement

- [x] I have read and approved the [CLA](https://github.com/dimensionalOS/dimos/blob/main/CLA.md).

Issue #1691, opened 2026-03-27
Simulation Editing


leshy:
original issue provided
[https://github.com/dimensionalOS/dimos/issues/1120](<https://github.com/dimensionalOS/dimos/issues/1120>)

# Model importing

We'd like access to standard threejs model loading primitives

[https://threejs.org/manual/#en/loading-3d-models](<https://threejs.org/manual/#en/loading-3d-models>)

```js
import { GLTFLoader } from 'three/addons/loaders/GLTFLoader.js';
const loader = new GLTFLoader();

loader.load( 'path/to/model.glb', function ( gltf ) {
  scene.add( gltf.scene );
})
```

# Standard threejs API

as well as all the standard threejs API

Primitives
Scenegraph
Materials
Textures
Lights
Cameras
Shadows
Fog

etc, I'm assuming you won't need to write extra code for this, threejs api is already available.

# Optional physics for models

If loading a global map, we'll want to enable colisions so that robot doesn't fall through the map.
for random visual elements (skybox) we won't need physics, but I can imagine adding a ball, or a car model (obstacle) which we do want colisions on

# Movable characters

We are testing stuff like person follow etc, I assume giving us threejs API access we can easily animate a character in the world, just making sure to mention here

# Robot definition API

we should be able to code simple control interface and a simple robot (drone, holonomic, plane, car) etc

# System Validation

To validate the delivery, we'd like to see these examples to validate how easy the API is to use. all of below should be reasonably performant as well

### ThreeJS API exposed

Deploy a robot sim in a random threejs example environment:

[https://threejs.org/examples/#webgl_lights_physical](<https://threejs.org/examples/#webgl_lights_physical>)
[https://threejs.org/examples/#webgl_loader_ifc](<https://threejs.org/examples/#webgl_loader_ifc>)
[https://threejs.org/examples/#webgpu_performance](<https://threejs.org/examples/#webgpu_performance>)

### Third party map load

load something like 
[https://sketchfab.com/3d-models/tdm-map-3-fps-game-environment-low-poly-4c645286d8554eb0b8cf2bee6642d710](<https://sketchfab.com/3d-models/tdm-map-3-fps-game-environment-low-poly-4c645286d8554eb0b8cf2bee6642d710>)
[...]

PR #2187, opened 2026-05-20
DimSim scene editing/authoring in JS + moving inside dimos

Viswa4599:
Brings DimSim into `misc/DimSim/`, wires it up as a first-class simulation backend, and replaces the legacy JSON authoring + sticky-state plumbing with a standard Three.js JS dev cycle for scenes and a JS-native eval system.

## What this does

- **DimSim lives inside dimos.** Continues Paul's #2081 vendoring with a cleaned-up source tree (`misc/DimSim/{src,cli,evals,scenes,public,docs}`) — no separate repo to clone, dimos drives it directly via `--simulation dimsim`.
- **Scene authoring is now JS.** Each scene is one file at `scenes/<name>/index.js` that default-exports `async build(api)` and uses the engine's THREE / Rapier / physics helpers — no JSON, no editor sidebar plumbing.
- **Apartment ported to JS.** 97 MB `apt.json` decomposed into `scenes/apartment/{data/*.js, textures/*}` (~6 MB total) and fed through `loadLevel`; full interactivity (pickables, door states, TV toggle) preserved.
- **Eval system collapsed to one file per workflow.** `scenes/<env>/evals/<name>.js` is a runnable program that `import { runEval } from '@dimsim/eval'` and calls it; works both in the browser (via Vite-pinned harness chunk) and via `deno run` (via Deno import map) — same file, two runtimes.
- **`dimsim` CLI shipped.** `dimsim dev`, `dimsim eval list`, `dimsim eval <workflow>` (auto-connect), `dimsim eval --headless ...`.

## Docs

- [`misc/DimSim/docs/getting-started.md`](https://github.com/dimensionalOS/dimos/blob/antim/sim-authoring-js/misc/DimSim/docs/getting-started.md) — 5-minute tour
- [`misc/DimSim/docs/scenes.md`](https://github.com/dimensionalOS/dimos/blob/antim/sim-authoring-js/misc/DimSim/docs/scenes.md) — create + edit scenes
- [`misc/DimSim/docs/evals.md`](https://github.com/dimensionalOS/dimos/blob/antim/sim-authoring-js/misc/DimSim/docs/evals.md) — author eval workflows

## Validation checklist (against [dimos#1691](https://github.com/dimensionalOS/dimos/issues/1691))

- [x] Write a scene from scratch in JS (`scenes/warehouse/index.js`)
- [x] Edit the existing apartment scene in JS (`scenes/apartment/`)
- [x] Robot embodiment API — `setEmbodiment({...})` scene-side API + bridge `ServerPhysics`/`ServerLidar` live reconfigure
- [x] Load a third-party map (Sketchfab GLB) into a scene
- [x] Load an arbitrary GLB via `loadGLTF`
- [x] Load assets from LFS

## Follow-ups (separate)

- [x] Decompose the apartment into pure Three.js + a registerable-object library so its interactive contents (pickables, doors, TV) can be authored in JS like everything else, not as apt-shape `loadLevel` data.
- [x] Move objects and textures to LFS


misc/DimSim/cli/vendor/lcm/transport.ts
```diff
@@ -0,0 +1,352 @@
+// LCM UDP Multicast Transport (vendored from @dimos/lcm@0.2.0)
+// FIX: Added joinMulticastV4() call in UdpMulticastSocket.listen()
```
Viswa4599, line 2: @leshy @paul-nechifor — flagging the vendored fix against @dimos/lcm@0.2.0.
  
So the deno dimos/lcm binds the UDP port but never actually subscribes to the multicast group, so no packets are received. On macOS this means Deno↔Python LCM doesn't work at all. 

Additions : 

 1. joinMulticastV4(host, iface) - upstream omits this entirely, so the socket binds the port but never subscribes to the multicast group. No inbound packets are ever delivered.

2. setLoopback(true) - sets IP_MULTICAST_LOOP on the socket. Without it, Deno's outgoing multicast packets go out to the network but the kernel skips local delivery, so Python dimos on the same host never sees the sensor channels Deno publishes (camera, lidar, odom).

dimos/simulation/dimsim/dimsim_process.py
```diff
@@ -126,28 +130,30 @@ def _kill_port_holder(port: int) -> None:
...
-            "1",
-            "--branch",
-            _DIMSIM_REPO_BRANCH,
-            _DIMSIM_REPO_URL,
-            str(repo_dir),
-        ],
-        check=True,
-    )
-    return repo_dir
+def _resolve_dimsim_dir() -> Path:
```
paul-nechifor: This shouldn't be needed. It should always use _DIMSIM_DIR

paul-nechifor:
Please run `pre-commit run --all-files` to fix the linting issues.

misc/DimSim/scenes/empty/index.js
```diff
@@ -0,0 +1,19 @@
+// scenes/empty/index.js — empty starter scene.
```
paul-nechifor, line 1: It looks like this does not contain a ground. I specifically included my fork of DimSim because I added a ground plane to the empty scene. E.g.: https://github.com/paul-nechifor/DimSim/commit/050aa1ec96e71524cdc9d84575ba5bc46d82993e

paul-nechifor: This also means some of the tests do not work, because the robot falls through the ground. E.g.: `uv run pytest -m dimsim dimos/e2e_tests/test_dimsim_path_replaning.py`

misc/DimSim/cli/agent.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env python3
```
paul-nechifor: Is this used for anything? It should probably be deleted. We already have agent blueprints.

misc/DimSim/docs/getting-started.md
```diff
@@ -0,0 +1,88 @@
+# Getting started
+
+5-minute tour. For deeper guides see [scenes.md](scenes.md) and [evals.md](evals.md).
+
+## The two ways DimSim runs
+
+```bash
+# 1. dimos drives DimSim (production / agentic / CI)
+.venv/bin/dimos --simulation dimsim --dimsim-scene=apartment run unitree-go2-agentic
```
paul-nechifor:
```suggestion
uv run dimos --simulation dimsim --dimsim-scene=apartment run unitree-go2-agentic
```

dimos/simulation/dimsim/scene_client.py
Viswa4599, line 1: @paul-nechifor I think we should probably remove the scene client? I see you're testing adding walls etc but that's doable directly in js now.

paul-nechifor: So how would you re-write `test_dimsim_path_replaning.py`?

Viswa4599: Hmm on second thought might make tests a headache, will keep it and slim down some unused methods.

paul-nechifor:
@Viswa4599 Did you run the tests? `uv run pytest -m dimsim dimos`

At least one of the dimsim tests isn't working because the robot exists outside of the walls now.

Viswa4599:
@paul-nechifor Fixed the empty scene spawn regression, dimsim tests run now. I also bumped the `spawn_wall_on_pose` threshold 1.5 → 2.5m in `test_dimsim_path_replaning.py` - at 1.5 the wall spawned right on
 the robot, I think the add_wall full-snapshot rebuild lags behind the moving robot. Possibly worse on my machine, can revert if that's the case.

.github/workflows/dimsim-check.yml
Dreamsorcerer, line 1: If I'm reading right, this only does a type check. If so, that should be in ci.yml, possibly just added as a step to the existing lint job.

Viswa4599: Deferring to @paul-nechifor.  The sim json scene template validation it used to carry is gone now (scenes moved to JS), so what's left is just build + type-check.

misc/DimSim/docs/validation.md
```diff
@@ -0,0 +1,242 @@
...
+}
+```
+
+- `physics.staticCollider(mesh, "trimesh" | "box" | "sphere")` — fixed obstacle the
+  robot collides with (and LiDAR sees). Created at boot → captured by the Rapier
+  snapshot the bridge restores.
+- `physics.dynamicCollider(mesh, { shape, mass, restitution })` — body that responds
+  to gravity (ball, knock-around obstacle).
+- `spawnPoint` is in **Three.js** coords (Y-up). It maps to ROS/odom `(x, y) = (z, x)`
+  via the bridge's `(x,y,z)->(z,x,y)` swap.
```
leshy: what is the workflow for placing a robot in a specific place after loading a new map? struggling with this atm

misc/DimSim/docs/scenes.md
```diff
@@ -0,0 +1,186 @@
...
+  const wall = new THREE.Mesh(
+    new THREE.BoxGeometry(20, 3, 0.2),
+    new THREE.MeshPhysicalMaterial({ color: 0xc4c1b8, roughness: 0.8 }),
+  );
+  wall.position.set(0, 1.5, -10);
+  scene.add(wall);
+  physics.staticCollider(wall, 'box');
+
+  // lighting
+  scene.add(new THREE.HemisphereLight(0xffffff, 0x404040, 0.6));
```
leshy: your example environment in validation.md suggests
```js

  scene.add(new THREE.AmbientLight(0xffffff, 0.2));
  const sun = new THREE.DirectionalLight(0xffffff, 1.5);

  sun.position.set(20, 20, 20); sun.castShadow = true; scene.add(sun);

```

any lighting added is added on top of this lighting here, so we get ultra bright scenes


Image

leshy: shadows don't seem to work in general, example suggests
```js
  map.traverse((c) => { if (c.isMesh) { c.castShadow = c.receiveShadow = true; } });
```
but this doesn't do anything

leshy:
what do these buttons do?

leshy:
unable to run

```sh
dimsim dev --scene apartment
```

misc/DimSim/docs/validation.md
leshy: guessing numbers easily makes robot float in the air, stuck in a wall etc

Viswa4599:
@leshy added helpers for lighting , shadows and robot placing. 
Also dimsim dev commands should work. Interact and spawn buttons were deprecated , somehow leaked back in. Removed now

paul-nechifor (dismissed):

paul-nechifor (approved):

.github/workflows/ci.yml
```diff
@@ -613,6 +613,17 @@ jobs:
...
+        # The checkout above doesn't fetch LFS, so the dimsim scene/embodiment
+        # glbs are pointer stubs and the browser sim fails to load them. Pull
+        # ONLY misc/DimSim assets — a full `lfs: true` would also drag in main's
+        # multi-GB data/.lfs/* objects (some missing on the server → 404).
+        # persist-credentials is false above, so supply the token for this pull.
+        run: |
+          git config --local http.https://github.com/.extraheader \
+            "AUTHORIZATION: basic $(printf 'x-access-token:%s' '${{ github.token }}' | base64 -w0)"
+          git lfs pull --include="misc/DimSim/**"
+          git config --local --unset-all http.https://github.com/.extraheader
```
Dreamsorcerer, line 626: @paul-nechifor Don't the LFS object get pulled automatically? Thought that was how the other tests were working...

paul-nechifor: No, the python `get_data` stuff only pulls in LFS objects from `/data`

PR #2189, opened 2026-05-20
fix(mcp): return tool images via langgraph Command (no more UUID side-channel)

Mgczacki:
## Context

I hit this while developing #2188 (the memory2 agent + Space raster backend) and figured the fix would be worth porting to the MCP client, which had the same problem and was working around it in a less elegant way.

## The underlying constraint

Many OpenAI chat-completion models **don't accept image content inside a tool result** — you can't just stuff an image into a `ToolMessage`. The canonical workaround is to deliver the image as a `HumanMessage` after the tool returns. That part isn't avoidable.

What we *can* avoid is delivering that HumanMessage in a **separate, later** agent turn — which is what the MCP client was doing.

## What was happening before

When an MCP tool returned non-text content (an image), `mcp_client.py`:

1. Minted a UUID for the image.
2. Returned a placeholder `ToolMessage` to the agent: *"Tool call started with UUID: X. You will be updated with the result soon"*.
3. Queued the actual image as a fresh `HumanMessage` via `add_message`, which got picked up on the **next** agent turn.

The agent therefore saw a fake "wait for it" tool result, kept reasoning with no real data, and only got the image one turn later. Two extra round-trips per image, and the tool result the agent reasons over isn't the actual content.

## What we do now

We're already on langchain's `create_agent`, which is a thin wrapper over LangGraph. LangGraph lets a tool return a `Command(update={...})` instead of a single message — i.e. the tool call can append **multiple messages to state** as the result of one call. The state's `messages` reducer (the tool-aware one in `agent.py`) then re-orders so each image-carrying `HumanMessage` lands next to its matching `ToolMessage`, paired by `additional_kwargs["tool_call_id"]`.

Concretely, the MCP tool now returns:

```python
Command(update={
    "messages": [
        ToolMessage(content="<text part or stub>", tool_call_id=...),
        HumanMessage(content=[image_block], additional_kwargs={"tool_call_id": ...}),
    ]
})
```

Both messages land in the same turn. The `HumanMessage`-with-image workaround for OpenAI's API constraint is still there — but it's now *part of the tool's response*, not a fake-out + late delivery. The reducer guarantees ordering even when multiple parallel tool calls return images simultaneously.

## Bonus fix: `InjectedToolCallId` for JSON-Schema tools

LangChain auto-injects `InjectedToolCallId` into tool arguments when `args_schema` is a Pydantic model. MCP gives us its schema as a plain JSON-Schema dict (the authoritative LLM contract on the MCP side), which langchain *doesn't* auto-inject for. Added `_McpStructuredTool`, a tiny `StructuredTool` subclass that does the injection for the JSON-Schema path. Without this, the tool wouldn't know its own `tool_call_id` and the reducer pairing breaks.

## Future steps

The same `Command(update=...)` pattern unlocks more than just "tools that return images". Two observations from driving the memory2 agent against this fix:

- **The agent skips steps.** Skills that prescribe verification sub-procedures (e.g. "call `frames_facing` to refine each candidate position before merging") get partially-followed even when the prose is strict. The current harness has no way to *enforce* that a step ran before the agent can submit an answer — the LLM is the only authority. Closing this needs explicit task management in the harness: a way to gate the final-answer tool on whether the prerequisite tool calls actually fired.

- **Long-horizon tasks benefit from subagent delegation.** Some questions (e.g. room segmentation across a 5-minute recording, exhaustive frontier exploration) are big enough that a single agent context fills up. The cleanest way to scale is to let the top-level agent dispatch self-contained sub-tasks to subagents — each with its own context window — and integrate their results. That keeps the parent agent's context focused on synthesis instead of bookkeeping.

These two together — harness-level task management + subagent delegation — are most naturally implemented as **another module**: a long-running, compute-heavier agent specialised for complex tasks, that the existing `McpClient`-style short-horizon agents can defer to in a **true multi-agent setting**. This PR is the prerequisite (image-returning tools work cleanly across agents); the multi-agent module is the follow-up.

## Test plan

- [x] `pytest dimos/agents/mcp/test_mcp_client_unit.py` — covers the new Command-return path and the multi-image / parallel-tool-call ordering.
- [x] Manual: drive an MCP tool that returns an image; verify the agent sees the image in the same turn as the tool call (no "started with UUID" placeholder).


dimos/agents/mcp/mcp_client.py
```diff
@@ -334,17 +352,26 @@ def _process_message(
...
+    to apply the same injection from the function signature — letting tools
+    return a `langgraph` `Command` that references their own `tool_call_id`
+    without having to invent a side-channel for it.
+    """
+
+    def _to_args_and_kwargs(
+        self, tool_input: str | dict[str, Any], tool_call_id: str | None
+    ) -> tuple[tuple[Any, ...], dict[str, Any]]:
+        args, kwargs = super()._to_args_and_kwargs(tool_input, tool_call_id)
+        if "tool_call_id" in self._injected_args_keys:
```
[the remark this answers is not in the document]
Mgczacki: We can update that if updating the langchain version changes the way the library does it. That's why pinned libraries exist.

dimos/agents/mcp/mcp_client.py
```diff
@@ -168,24 +169,41 @@ def _mcp_tool_to_langchain(self, mcp_tool: dict[str, Any]) -> StructuredTool:
...
+            text = "\n".join(c.get("text", "") for c in content if c.get("type") == "text")
+            image_blocks = [c for c in content if c.get("type") != "text"]
+
+            if not image_blocks:
+                return text
+
+            # Vision content can't be embedded inside a ToolMessage for OpenAI
+            # (and others), so we use Command to append a follow-up HumanMessage
+            # carrying the image blocks within the same agent turn. Mirrors the
+            # pattern used by examples/memory2_agent/tools.py.
```
paul-nechifor: What is `memory2_agent`? I don't see this in the codebase.

Mgczacki: Apologies for the lack of context, it's the agent in this PR: https://github.com/dimensionalOS/dimos/pull/2188 - will clean up this comment and others that reference it.

dimos/agents/mcp/mcp_client.py
```diff
@@ -168,24 +169,41 @@ def _mcp_tool_to_langchain(self, mcp_tool: dict[str, Any]) -> StructuredTool:
...
+            #
+            # The HumanMessage is tagged with `additional_kwargs["tool_call_id"]`
```
paul-nechifor, line 242: Where is `additional_kwargs["tool_call_id"]` used?

In `langchain_core` it seems like it's meant to be used for additional fields used by some providers.

Mgczacki: It seems like I stashed my reducer that read it just before uploading the PR 😢. Will update shortly when I reapply it and double-check everything. Answering your question, it is read as a condition for the reducer to be able to re-order the messages once it gets them. Human Message with tool_call_id in additional_kwargs will be the indicator to reorder it just after the tool response that matches said tool_call_id.

dimos/agents/mcp/mcp_client.py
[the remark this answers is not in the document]
paul-nechifor: It's better to write a future-proof version now than wait for the private API to break and implement the fix then.

`_to_args_and_kwargs` and `_injected_args_keys` should not be used.

If the goal is to capture `tool_call_id` you can do so by defining something like this:

```python
class _McpStructuredTool(StructuredTool):
    def run(
        self, tool_input: str | dict[str, Any], *args: Any, **kwargs: Any
    ) -> Any:
        tool_input = _inject_tool_call_id(tool_input, kwargs.get("tool_call_id"))
        return super().run(tool_input, *args, **kwargs)

    async def arun(
        self, tool_input: str | dict[str, Any], *args: Any, **kwargs: Any
    ) -> Any:
        tool_input = _inject_tool_call_id(tool_input, kwargs.get("tool_call_id"))
        return await super().arun(tool_input, *args, **kwargs)


def _inject_tool_call_id(
    tool_input: str | dict[str, Any], tool_call_id: str | None
) -> str | dict[str, Any]:
    if not isinstance(tool_input, dict):
        return tool_input
    if tool_call_id is None:
        raise ValueError(
            "MCP tool requires a tool_call_id; invoke via a ToolCall, not a bare dict."
        )
    return {**tool_input, "tool_call_id": tool_call_id}

```

Mgczacki:
Closing and reopening as a branch within the repo.

PR #2192, opened 2026-05-20
Disable GIL in astar module

Dreamsorcerer:
In free-threaded builds, this should allow the GIL to remain disabled and improve threading performance.


jeff-hykin:
I'm no expert on pybind, but it looks to me like `grid` can be modified on the python side and cause a race condition on the C++ side (ex: C++ read height, python changes `grid`, C++ reads width).

With the GIL enabled that shouldn't happen. 

Is this for performance? or is it related to CI issues?

jeff-hykin:
(also feel free to ping me on discord, I often forget to check these)

PR #2194, opened 2026-05-20
Release workflow

Dreamsorcerer:
This is the workflow that will actually create and publish a new release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -169,7 +169,7 @@ jobs:
         with:
           path: dist
           merge-multiple: true
-      - name: Attestation
+      - name: Publish to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           attestations: true
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -53,11 +53,14 @@ jobs:
           echo "Releasing version: $version"
       - name: Verify version not already on PyPI
         run: |
-          if pip install "dimos==${{ steps.read-version.outputs.version }}" \
-              --dry-run --no-deps --ignore-installed --quiet 2>/dev/null; then
-            echo "::error::dimos ${{ steps.read-version.outputs.version }} already exists on PyPI"
-            exit 1
-          fi
+          version="${{ steps.read-version.outputs.version }}"
+          http_code=$(curl -sS -o /dev/null -w '%{http_code}' \
+            "https://pypi.org/pypi/dimos/${version}/json" 2>/dev/null || true)
+          case "$http_code" in
+            200) echo "::error::dimos ${version} already exists on PyPI"; exit 1 ;;
+            404) echo "ok: ${version} not on PyPI" ;;
+            *)   echo "::error::PyPI version check failed (HTTP code: '${http_code}')"; exit 2 ;;
+          esac
       - name: Verify tag does not exist
         run: |
           if git ls-remote --tags origin "refs/tags/v${{ steps.read-version.outputs.version }}" | grep -q .; then
@@ -152,6 +155,7 @@ jobs:
         with:
           repository-url: https://test.pypi.org/legacy/
           attestations: true
+          skip-existing: true
 
   publish-pypi:
     if: ${{ ! inputs.dry-run }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,250 @@
...
+    permissions:
+      id-token: write      # For pypi-publish's OIDC trusted-publisher exchange (PyPI)
+      attestations: write  # For PEP 740 attestations attached to each uploaded file
+    steps:
+      - name: Download wheels/sdist
+        uses: actions/download-artifact@v4
+        with:
+          path: dist
+          merge-multiple: true
+      - name: Attestation
```
Dreamsorcerer:
```suggestion
      - name: Publish to PyPI
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,254 @@
...
+        uses: actions/attest-build-provenance@v2
+        with:
+          subject-path: 'dist/*.tar.gz'
+      - name: Upload sdist
+        uses: actions/upload-artifact@v4
+        with:
+          name: sdist
+          path: dist/*.tar.gz
+
+  build-wheels:
```
paul-nechifor, line 97: We probably want to make a test release for this to make sure it works. Locally, it doesn't allow me to publish anything other than the source dist because PyPI complains that my binary is not compatible with enough architectures (or something like that, I don't remember the exact error)

.github/workflows/release.yml
```diff
@@ -0,0 +1,254 @@
...
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read  # For checkout
+    outputs:
+      version: ${{ steps.read-version.outputs.version }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
```
paul-nechifor, line 32: I think it's best practice to write the exact hash for these `uses:` just in case later versions get compromised. Maybe for linting is not as important, but it probably is for publishing.

.github/workflows/release.yml
```diff
@@ -0,0 +1,254 @@
...
+          git tag -s -a "$tag" -m "Release $tag"
+          git push origin "$tag"
+      - name: Create GitHub Release
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh release create "v${{ needs.validate.outputs.version }}" \
+            --verify-tag \
+            --title "v${{ needs.validate.outputs.version }}" \
+            --notes "See CHANGELOG / commit history for changes." \
```
paul-nechifor, line 216: We don't maintain a CHANGELOG file, though we probably should.

.github/workflows/release.yml
Dreamsorcerer: Yes, that's what the dry-run is for. Once we cut a new branch, I'll test it via the dry-run and make any necessary fixes.

.github/workflows/release.yml
Dreamsorcerer: My plan is to pin them in a few months time when Github releases the dependency pinning feature. That would work more like a lock file then, so we still have versions here, but pins at the bottom of the file. At that point we can also enable the Github option that requires that all actions are pinned (a major issue here without that feature, is that anything any of these actions uses may not be pinned, so we're not gaining much).

Also, pinning anything from actions/ isn't much of a concern. If Github is compromised, then all is lost anyway..

.github/workflows/release.yml
Dreamsorcerer: Yeah, this was kind of a placeholder. How do you create the changelog currently? Any suggestions on how you'd like to do it in future?

paul-nechifor: The current "changelog" is the text on the release pages.

The notes are composed by Claude by going through every PR description and analyzing what's important. I then go through it to make sure everything is alright (if you ask it to be succinct there are few mistakes).

I guess the idea would be for us to have a proper CHANGELOG.md file.

paul-nechifor (approved):

.github/workflows/release.yml
Dreamsorcerer: Well, we could tell the releaser to use Claude. Or, as an example, aiohttp requires change fragments to be added under CHANGES/ with each PR (a label is used to bypass the check for irrelevant things), then `towncrier` is used generate the final changelog during the release process.

PR #2198, opened 2026-05-20
feat(teleop): stream robot camera into Quest VR headset

ruthwikdasyam:
## Problem

Adding streaming to the VR, for teleop - sim camera, realsense camera addition, or go2 camera view will be displayed in the headset

Closes DIM-850

## Solution

- QuestTeleopModule has an optional color_image: In[Image]. When the blueprint wires it, frames are JPEG-encoded with OpenCV and cached; otherwise the subscribe is skipped and /video stays empty.
- Wired into teleop_quest_xarm7 so dimos --simulation run teleop-quest-xarm7 shows MuJoCo's color_image out of the box; for real hardware, compose at launch: dimos run teleop-quest-xarm7 real-sense-camera.

## How to Test
``` python
# login through quest, you can see the camera view
dimos --simulation run teleop-quest-xarm7

# for go2
dimos run teleop-quest-go2
```

https://github.com/user-attachments/assets/a7db5c88-0a5c-4b6f-a959-85a740f4346f

https://github.com/user-attachments/assets/692c77c8-d5ef-4619-94ae-e700eb6c7dae


leshy:
yoooo can you add Yaw control for robot body?

dimos/teleop/quest/quest_teleop_module.py
```diff
@@ -143,6 +148,8 @@ async def teleop_index() -> HTMLResponse:
         @self._web_server.app.websocket("/ws")
         async def websocket_endpoint(ws: WebSocket) -> None:
             await ws.accept()
+            self._ws_loop = asyncio.get_running_loop()
```
paul-nechifor, line 154: `_ws_loop` is set here, but it's never used. Instead it's used in `_push_jpeg`. But `_push_jpeg` shouldn't use it because it's private. But even if you make it public, setting it here to use it in a random place is a bit convoluted.

`_push_jpeg` just needs a loop to run in, right? You don't have to use the websocket loop, just use the loop of the module: `module._loop`.

BUT... now that we have support for async modules, you don't have to reference the loop at all. See `### Auto-bound input handlers` in `docs/usage/modules.md`.

If you declare `async def handle_color_image` it automatically listens to `color_image` without needing to be registered/unregistered.

---

In general, this is something we should start doing more. We shouldn't call `asyncio.run_coroutine_threadsafe`. We should use async functions from the root and use `await` for corutines, not `asyncio.run_coroutine_threadsafe`.

dimos/teleop/quest/quest_extensions.py
```diff
@@ -148,3 +192,108 @@ def _publish_button_state(
...
+        logger.info("Quest teleop: Go2 camera subscribed → /ws push")
+
+    @rpc
+    def stop(self) -> None:
+        # Send one zero Twist so the base halts if teleop dies mid-motion.
+        try:
+            zero = Twist()
+            zero.linear = Vector3(0.0, 0.0, 0.0)
+            zero.angular = Vector3(0.0, 0.0, 0.0)
+            self.cmd_vel.publish(zero)
```
paul-nechifor:
```suggestion
            self.cmd_vel.publish(Twist.zero())
```

dimos/teleop/quest/web/static/index.html
```diff
@@ -126,6 +127,23 @@ <h1>DimOS Quest-3 Teleop</h1>
         let lastSendTime = 0;
         const sendInterval = 1000 / 80; // ~80Hz target
 
+        // Video panel state
```
paul-nechifor: This file is getting quite large. We should probably break up the JS and CSS into separate files.

dimos/teleop/quest/quest_extensions.py
```diff
@@ -17,18 +17,62 @@
...
+            return
+        jpeg = buf.tobytes()
+    except Exception:
+        logger.exception("Failed to encode camera frame")
+        return
+
+    loop = module._ws_loop
+    if loop is None or not module._connected_clients:
+        return
+    for ws in list(module._connected_clients):
```
paul-nechifor: When I see `for x in list(xs):` my spidey sense is tingling. I assume you're using `list` because `_connected_clients` is being modified from another thread. You need to use a lock.

If you switch to using a single loop, then you don't need locks.

dimos/teleop/quest/quest_extensions.py
```diff
@@ -17,18 +17,62 @@
...
+    """JPEG-encode an Image and push it to all of module's connected /ws clients.
+
+    Runs on the RX thread; sends are scheduled on the asyncio loop captured by
+    QuestTeleopModule when the first client connected.
+    """
+    try:
+        bgr = msg.to_opencv()
+        ok, buf = cv2.imencode(".jpg", bgr, [int(cv2.IMWRITE_JPEG_QUALITY), quality])
+        if not ok:
+            return
```
paul-nechifor: I think TurboJPEG is the fastest encoder. We use that in `Image.py`. Unfortunatelly, it doesn't look like we have `msg.to_jpeg_bytes()`. You might want to add that. (That is what we're doing in `lcm_jpeg_encode`, but we don't have access to the raw bytes.)

paul-nechifor: Also, before encoding, check if there are clients. If there are no clients, exit early without encoding.

dimos/teleop/quest/quest_teleop_module.py
ruthwikdasyam: - async works great. changed it, and removed manual registrations.

- will do a followup PR on removing _ws_loop, 
created issue on this - https://github.com/dimensionalOS/dimos/issues/2321

dimos/teleop/quest/quest_extensions.py
ruthwikdasyam: THanks, added lock

paul-nechifor (approved):

PR #2200, opened 2026-05-20
Rust warn on dropped LCM packets

jeff-hykin:
Backpressure from heavy faster-than-real-time lidar benchmarks is kind of evitable. Right now its a silent failure. 

This includes a warning message, and some higher default rust buffer sizes to handle more throughput to match. 

Tests added to make sure rust native modules aren't bottlenecking LCM.

Run the tests:

```sh
cd native/rust/dimos-module
cargo test
```

### DIFF
--- a/native/rust/dimos-module/src/module.rs
+++ b/native/rust/dimos-module/src/module.rs
@@ -1,6 +1,7 @@
 use std::collections::HashMap;
 use std::fmt::Debug;
 use std::io;
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
 use tokio::io::{AsyncBufReadExt, BufReader};
 use tokio::sync::mpsc;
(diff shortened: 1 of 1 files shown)


native/rust/dimos-module/src/module.rs
```diff
@@ -22,15 +23,26 @@ struct TypedRoute<T: Send + 'static> {
...
-            // If the input channel is full, the newest message is dropped.
-            Ok(msg) => {
-                let _ = self.sender.try_send(msg);
-            }
+            Ok(msg) => match self.sender.try_send(msg) {
+                Ok(()) => {}
+                Err(mpsc::error::TrySendError::Full(_)) => {
+                    let n = self.drops.fetch_add(1, Ordering::Relaxed) + 1;
+                    // Rate-limited: log at drop counts 1, 2, 4, 8, ...
+                    if n.is_power_of_two() {
```
aclauer: What if we just did n % 10 or something like that, probably would be simpler than what we were discussing. No reason to over complicate the logging unless it becomes an issue in my mind

aclauer: Also if we switch to the tracing crate, you can control the log level (error, warn, info, debug), so things like this could be debug and you can turn them off/on nicely.

native/rust/dimos-module/src/module.rs
```diff
@@ -564,4 +588,24 @@ mod tests {
...
+            last_log: Mutex::new(None),
+        };
+        // Fill the queue, then force a drop.
+        route.try_dispatch(&[1u8]);
+        route.try_dispatch(&[1u8]);
+        assert_eq!(route.drop_count.load(Ordering::Relaxed), 1);
+        assert!(
+            route.last_log.lock().unwrap().is_some(),
+            "drop must trigger a log",
+        );
```
paul-nechifor, line 609: The test doesn't actually check if the log is written. In Python, you'd actually check that `logger` object actually had something written to it. I'm guessing that's not possible with `eprintln!`, right?

paul-nechifor (approved):

native/rust/dimos-module/src/module.rs
aclauer: Yeah I don't think eprintln! can do that, other than checking stderr. I created a task to switch the logging to use the tracing crate, and then we can do what you mentioned

leshy (approved):

aclauer (approved):

PR #2205, opened 2026-05-21
chore: Rust format and testing in CI

### LINKED ISSUE #2203: Add Rust tests to CI
*Synced from* [DIM-938](https://linear.app/dimensional/issue/DIM-938) *by andrew*

aclauer:
## Problem

Need to run Rust formatting checks and tests in CI

Closes DIM-938

## Solution

Added auto discovery job to find all Rust workspaces in the repository based on `Cargo.lock`. Then run `cargo fmt`, `cargo clippy`, `cargo test` in all work spaces in parallel.

## How to Test

Run in CI, result of all three commands matches local results. All workspaces are automatically detected without any hardcoded paths.


.github/workflows/ci.yml
```diff
@@ -36,6 +36,65 @@ jobs:
...
+          echo "Found Rust workspace roots:"
+          echo "$crates" | jq
+
+  rust:
+    needs: rust-discover
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        crate: ${{ fromJSON(needs.rust-discover.outputs.crates) }}
```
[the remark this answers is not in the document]
aclauer: this would only happen if we remove all the rust code, and in that case we should just remove the tests, so I don't think this is an issue

leshy:
@Dreamsorcerer can you review?

.github/workflows/ci.yml
```diff
@@ -36,6 +36,65 @@ jobs:
       - name: Run pre-commit
         uses: pre-commit/action@v3.0.1
 
+  rust-discover:
+    # Auto discover the Rust workspaces by searching for Cargo.lock files
+    timeout-minutes: 1
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
```
Dreamsorcerer: Let's keep reasons for each permission documented. I assume this is just for checkout.
```suggestion
      contents: read  # For checkout
```

.github/workflows/ci.yml
```diff
@@ -36,6 +36,65 @@ jobs:
...
+    outputs:
+      crates: ${{ steps.find.outputs.crates }}
+    steps:
+      - uses: actions/checkout@v6
```
Dreamsorcerer: Trying to keep a convention of naming every step currently.
```suggestion
      - name: Checkout
        uses: actions/checkout@v6
```

.github/workflows/ci.yml
```diff
@@ -36,6 +36,65 @@ jobs:
...
+  rust:
+    needs: rust-discover
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        crate: ${{ fromJSON(needs.rust-discover.outputs.crates) }}
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
```
```suggestion
      contents: read  # For checkout
```

leshy (dismissed):

.github/workflows/ci.yml
Dreamsorcerer: One blocking detail here I think. I don't think we should be using a matrix here, atleast from what I see of the current run.

Currently, this results in 3 separate jobs, requiring 3 workers. Each job only takes ~20s with about 50% of that time doing the setup/teardown of the job.

So currently this is basically 3 x 20s, instead of 1 x 40s. The difference today is neglible, but if we start adding in more crates over time, then this will lead to worker starvation and delay the other (slower) test jobs.

leshy: oops I approved, but disabled auto-merge, letting you guys decide, just ping me when you need approval

aclauer: That's a good point, I'll change that

aclauer: should be fixed now, also removed the separate discover job

Dreamsorcerer (approved):
Can't comment on the rust specifics, but the workflow looks solid now, assuming that the checks can't easily be incorporated into pre-commit.

.github/workflows/ci.yml
Dreamsorcerer: Appears that the setup/teardown was substantially more than I estimated, as the new job does everything in 15s.

leshy (approved):

PR #2207, opened 2026-05-21
Native modules should build during build step

jeff-hykin:
## Problem

Replay modules (all non-native modules) don't wait for native modules to build before they start replaying.

## Solution

Have native modules build at build stage instead of start

## How to Test

On a fresh G1

```
uv run dimos run --rerun-host 0.0.0.0 unitree-g1-nav-onboard
```

### DIFF
--- a/dimos/core/native_module.py
+++ b/dimos/core/native_module.py
@@ -180,6 +180,11 @@ class NativeModule(Module):
             self.config.executable = str(Path(self.config.cwd) / self.config.executable)
 
     @rpc
+    def build(self) -> None:
+        super().build()
+        self._maybe_build()
+
+    @rpc
     def start(self) -> None:
         super().start()
         if self._process is not None and self._process.poll() is None:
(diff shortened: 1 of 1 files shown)


leshy:
generally the idea is that `start()` should block and once start passes, your module is considered started, whatever you need to do in start you should do in order to be ready, releasing that function means the module is good to go, that's why start was also building - in case of native modules.

Otherwise this is leaking abstractions, there are many things modules might want to do, maybe heavy ML model warmup. we shouldn't add `.warmup()` calls to module coordinator, another category of modules needs to boot up a robot arm, so we wouldn't add `.bootup()` calls to a module coordinator.. this is why `start()` is supposed to deal with all those concerns. externally all other systems care about is "are you started"

issue here is the opposite - that there is "dimos is running" signal at dimos that module could start replying on, that robot can start moving on, etc.

we can allow this for now but in reality we need a start signal @paul-nechifor and then can remove build() from anywhere else but the native moudules themselves

leshy (approved):

Issue #1266, opened 2026-02-16
Unify Docker Modules and Native Modules


leshy:
Native modules run any binary and provide CLI args to them 

read more here https://github.com/dimensionalOS/dimos/blob/507651c798eaa8134cb9de338ed0c70e6b042a6a/docs/usage/native_modules.md

This seems mega related to docker modules that communicate via ENV vars

Probably Native module system should host Docker modules? just need to switch it to use ENV vars? 

Unified serious config system would help here https://github.com/dimensionalOS/dimos/issues/1082 (since it can translate py config objects (pydantic, dataclass) into ENV vars naturally) but it's ok to do the Module unification first

PR #2224, opened 2026-05-22
fix: delete docker_module parallel hierarchy (#2038)

### LINKED ISSUE #2038: Delete docker_module — re-add as NativeModule subclass if/when needed
`dimos/core/docker_module.py` (693 LOC) is a parallel `ModuleProxyProtocol` impl — `DockerModuleProxy`, `DockerModuleConfig`, `DockerModuleInner` — none of it extends `NativeModule`. Same in coordination: `worker_manager_docker.py` sits next to `worker_manager_python.py` doing largely the same thing, both wired up in `module_coordinator.py:56`.

bad arch to keep around. just delete it.

if/when we want docker support back, it goes in as a thin subclass of `NativeModule` that only overrides how the child process is launched (`docker run` instead of plain exec). image build / cache logic moves to a separate helper that NativeModule doesn't need to know about.

scope:
- delete `dimos/core/docker_module.py`
- delete `dimos/core/coordination/worker_manager_docker.py`
- drop `WorkerManagerDocker` from `module_coordinator.py:56`
- delete `examples/docker_hello_world/`
- two real consumers to migrate or hold: `dimos/manipulation/grasping/graspgen_module.py` (`GraspGenConfig(DockerModuleConfig)`) and `dimos/manipulation/pick_and_place_module.py` (uses `DockerModuleProxy`)

related: #1266 — we still want unification; but delete this first to make sure we don't build upon it. (re-add as a NativeModule subclass instead of refactoring the parallel hierarchy in place). also #1082 (config system).

KrishnaH96:
## Problem

`docker_module.py` is a 693-line parallel `ModuleProxyProtocol` implementation that duplicates what `WorkerManagerPython` + `RPCClient` already provide. `WorkerManagerDocker` and `DockerModuleProxy` re-implement the same host-side proxy pattern just to swap `docker run` for a plain subprocess. The right design (proven by `NativeModule`) is a thin subclass that overrides process launch — not a separate hierarchy.

Closes #2038. Related: #1266 (re-add as NativeModule subclass if/when needed).

## Solution

Delete the parallel Docker hierarchy and hold the two real consumers:

- Delete `docker_module.py`, `worker_manager_docker.py`, their tests, and `examples/docker_hello_world/`
- Drop `WorkerManagerDocker` from `ModuleCoordinator`
- `graspgen_module.py`: swap `DockerModuleConfig` → `ModuleConfig`, `deployment = "python"`, remove docker-specific config fields
- `pick_and_place_module.py`: remove `DockerModuleProxy` usage, stub `generate_grasps` with `NotImplementedError` pointing to #1266

## How to Test

```bash
python3 -c "from dimos.manipulation.grasping.graspgen_module import GraspGenModule; from dimos.manipulation.pick_and_place_module import PickAndPlaceModule; print('imports ok')"
```

## Notes

`demo_grasping.py` still passes docker-specific kwargs to `graspgen()` — these are silently ignored at blueprint creation and only fail at deploy time. Left as-is since the demo cannot run regardless (graspgen libs were Docker-only). Will be cleaned up when Docker support is re-added in #1266.


spomichter:
legend

dimos/manipulation/grasping/graspgen_module.py
```diff
@@ -63,7 +57,7 @@ class GraspGenConfig(DockerModuleConfig):
 
 class GraspGenModule(Module):
```
mustafab0: This needs docker module.

As per @paul-nechifor's suggestion, we can delete this module for now.

mustafab0 (changes requested):

mustafab0 (approved):

paul-nechifor (approved):

PR #2225, opened 2026-05-22
Pim/feat/g1 groot wbc part 3

Nabla7:
## Problem

Closes DIM-XXX


dimos/robot/cli/dimos.py
```diff
@@ -245,8 +245,7 @@ def run(
 
     cli_config_overrides: dict[str, Any] = ctx.obj
 
-    # this is a workaround until we have a proper way to have delayed-module-choice in blueprints
-    # ex: vis_module(viewer=global_config.viewer) is wrong (viewer will always be default value) without this patch
+    # Apply CLI config before importing blueprints that branch on global_config.
```
Nabla7, line 248: why is this here, unnecessary comment. Undo.

dimos/control/task.py
Nabla7, line 1: @mustafab0 is it ok if imu lives here?

dimos/control/tick_loop.py
```diff
@@ -174,7 +175,8 @@ def _tick(self) -> None:
         self._tick_count += 1
 
         joint_states = self._read_all_hardware()
-        state = CoordinatorState(joints=joint_states, t_now=t_now, dt=dt)
+        imu_states = self._read_all_imu()
```
Nabla7, line 178: @mustafab0 , same here getting imu state is now part of the control tick loop because I need to get it from somewhere and controlcoordinator is the canonical interface for hardware, it makes sense that it would provide imu bit it does extend the responsibility of controlcoordinator beyond that of pure joint control, what are your thoughts here?

dimos/simulation/engines/mujoco_sim_module.py
```diff
@@ -71,10 +87,101 @@ def _default_identity_transform() -> Transform:
     )
 
 
+class _WholeBodySimHooks:
```
Nabla7, line 90: this does bloat mujoco sim module, probably needs to be moved elsewhere along with other robot specific logic.

PR #2229, opened 2026-05-23
Enable TF service to wait for future transforms

### LINKED ISSUE #2212: TF service should take the match tolerance into account not just into the past but in the future
if transform match isn't found, and we are searching around a current ts, block and wait a bit, it might arrive

use default tolerance, unless a special forward_tolerance kwarg is set (we often want to search further in the past then block in the future)

this is an issue that caused https://github.com/dimensionalOS/dimos/issues/2215 and also april tag marker having issues in replays but not in real runs

```python
git checkout feat/ivan/village_recordings
uv run dimos --dtop --replay --replay-db hk_village4 run unitree-go2-markers
```

arkluc:
## Problem

Closes https://github.com/dimensionalOS/dimos/issues/2212

## Solution

Add a forward_tolerance param to TF service get functions. Default value is 0, keeps old behaviour. 

When the lookup cannot find a transform and forward_tolerance > 0, it re-checks under a threading.Condition lock and waits until the deadline.

## How to Test

uv run pytest dimos/protocol/tf/test_tf.py

### DIFF
--- a/dimos/protocol/tf/test_tf.py
+++ b/dimos/protocol/tf/test_tf.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 import math
+import threading
 import time
 
 import pytest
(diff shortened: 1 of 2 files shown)


dimos/protocol/tf/test_tf.py
```diff
@@ -371,6 +372,103 @@ def test_time_tolerance(self) -> None:
...
+
+    def test_forward_tolerance_times_out(self) -> None:
+        ttbuffer = MultiTBuffer()
+        t0 = time.monotonic()
+        result = ttbuffer.get(
+            "world", "robot", time_point=time.time(), forward_tolerance=0.1
+        )
+        elapsed = time.monotonic() - t0
+        assert result is None
+        assert 0.05 < elapsed < 1.0
```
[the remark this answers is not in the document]
arkluc: mhm 50ms is a lot
we will see if ci is all green

dimos/protocol/tf/tf.py
```diff
@@ -179,13 +185,40 @@ def get(
...
+            deadline = time.monotonic() + forward_tolerance
+            with self._cv:
+                while True:
+                    result = self._get(parent_frame, child_frame, time_point, time_tolerance)
+                    if result is not None:
+                        return result
+                    remaining = deadline - time.monotonic()
+                    if remaining <= 0:
+                        break
+                    self._cv.wait(timeout=remaining)
```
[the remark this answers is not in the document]
arkluc: updated

dimos/protocol/tf/tf.py
```diff
@@ -112,25 +115,28 @@ class MultiTBuffer:
...
+                key = (transform.frame_id, transform.child_frame_id)
+                if key not in self.buffers:
+                    self.buffers[key] = TBuffer(self.buffer_size)
+                self.buffers[key].add(transform)
+            self._cv.notify_all()
 
     def get_frames(self) -> set[str]:
         frames = set()
-        for parent, child in self.buffers:
+        for parent, child in list(self.buffers):
```
paul-nechifor: You're using `list` here presumably because we're mutating `self.bluffers` from different threads. `list` holds the GIL so it does prevent concurrent read/writes of `self.buffers`. But Python is moving away from having a GIL so using `list` is not a good idea long term.

I see you're already using `with self._cv` in some functions. Presumably all changes should be done with that.

dimos/protocol/tf/tf.py
```diff
@@ -179,13 +185,40 @@ def get(
...
+    def get(
+        self,
+        parent_frame: str,
+        child_frame: str,
+        time_point: float | None = None,
+        time_tolerance: float | None = None,
+        *,
+        forward_tolerance: float = 0.0,
+    ) -> Transform | None:
+        result = self._get(parent_frame, child_frame, time_point, time_tolerance)
```
paul-nechifor, line 226: Shouldn't `_get` be under `with self._cv` too? Because it too accesses `self.buffers` without a Lock/Condition.

I think you need to check whereever `self.buffers` is used and make sure you use `self._cv`.

dimos/protocol/tf/tf.py
```diff
@@ -179,13 +185,40 @@ def get(
...
+                    result = self._get(parent_frame, child_frame, time_point, time_tolerance)
+                    if result is not None:
+                        return result
+                    remaining = deadline - time.monotonic()
+                    if remaining <= 0:
+                        break
+                    self._cv.wait(timeout=remaining)
+
+        logger.warning(
+            f"No direct transform found between '{parent_frame}' and '{child_frame}' at '{to_human_readable(time_point or time.time())}', {self}"
```
paul-nechifor: Might be better to drop `self` from the log. You need a lock for `__str__` as well because it too accesses `self.buffers` and logging this is not worthwhile.

dimos/protocol/tf/tf.py
```diff
@@ -179,13 +185,40 @@ def get(
...
 
         if complex is None:
-            logger.warning(
-                f"No direct transform found between '{parent_frame}' and '{child_frame}' at '{to_human_readable(time_point or time.time())}', {self}"
-            )
             return None
 
         return reduce(lambda t1, t2: t1 + t2, complex)
 
+    def get(
```
paul-nechifor, line 217: Adding `forward_tolerance` to `get` just makes it more complicated. It's 2 functions masquerading as 1 function.

I think it's better to leave `get` as it is and introduce a new function: `wait_for_transform(..., *, timeout: float)`

That makes the semantics clear: one just gets, the other can block for a while.

leshy:
`dimos --dtop --replay --replay-db hk_village4 run unitree-go2-markers`

verified apriltag transforms visible in replay now

dimos/protocol/tf/tf.py
```diff
@@ -112,25 +115,28 @@ class MultiTBuffer:
...
-            if key not in self.buffers:
-                self.buffers[key] = TBuffer(self.buffer_size)
-            self.buffers[key].add(transform)
+        with self._cv:
+            for transform in args:
+                key = (transform.frame_id, transform.child_frame_id)
+                if key not in self.buffers:
+                    self.buffers[key] = TBuffer(self.buffer_size)
+                self.buffers[key].add(transform)
+            self._cv.notify_all()
```
leshy, line 127: fancy

leshy (dismissed):

dimos/protocol/tf/tf.py
leshy: in most cases when using get, you do want forward tolerance get.
we can have a double function internally, but externally as a user I want a single enrypoint for this

paul-nechifor: "Forward tolerance get" is not "get". "Get" implies something is there and you're retrieving it. When you're blocking and waiting for something, it's better to communicate the semantics by using a different verb in the name, liike "wait".

Also, "forward tolerance" is a complicated way of saying "wait".

leshy: Yup we can rename the function - but important for me to stay the same function

tf2 in ROS has the same API: 

lookupTransform(target, source, time, timeout)
lookupTransform(target, source, time)

`self.tf.lookup` seems pretty good to me

dimos/protocol/tf/tf.py
arkluc: good catch we should

dimos/protocol/tf/tf.py
arkluc: good advice, dropped. there is enough information from other pieces in this logging call

dimos/protocol/tf/tf.py
arkluc: fixed. now we also use condition for reads

dimos/protocol/tf/tf.py
arkluc: mhm i think we can seperating into 2 functions `_get` and `_wait_get`. These function has same behaviour as their name suggest

Then we use `get` to call them

leshy (approved):

PR #2234, opened 2026-05-23
feat(mcp): parallel tool dispatch with per-lane locks

johnny-kantaros:
## Problem

Tool calls dispatched by the LLM always execute sequentially even when they are fully independent. On a robot, this means "take a photo" must wait for "move forward" to complete before starting — even though the two have nothing to do with each other. At the same time, some tools genuinely must not overlap: two concurrent motion commands can produce conflicting actuator inputs.

Closes DIM-2204

## Solution

Adds a lane parameter to the @skill decorator:

```
  @skill(lane="motion")
  def move_forward(self): ...
  
  @skill(lane="motion")
  def turn_left(self): ...   # locked with move_forward

  @skill    
  def take_photo(self): ...  # runs freely in parallel with anything
```

The lane travels through the full stack:
  
  @skill(lane=...) → __skill_lane__ on the function
  → `SkillInfo.lane`
  → `McpServer` tools/list response ({"lane": "motion"})
  → `McpClient._tool_registry`
  → `call_tool()` acquires `threading.Lock` per lane at execution time

  On the client side, two changes enable parallel dispatch:

1. `call_tool` now uses `asyncio.to_thread` to run the blocking HTTP call off the event loop, so multiple tool calls can be in-flight simultaneously. Tools in the same lane acquire a shared `threading.Lock` before proceeding; tools in different lanes or with no lane run concurrently without waiting.
2. `_process_message` switches from `state_graph.stream()` to `asyncio.run(state_graph.astream())`. This is what actually enables LangGraph to dispatch all pending tool calls as concurrent async tasks in a single agent turn.

The MCP server handles concurrent requests natively via FastAPI's async request handler and `run_in_executor`, so no server-side changes were needed.

  ---
### Backwards Compatibility
  
Existing @skill methods with no lane argument are unaffected — they continue to work exactly as before. The only behavioral change is that the agent now dispatches independent tools concurrently within a single turn instead of sequentially.
  
For modules where multiple skills have implicit ordering dependencies (e.g. a write_state skill that must complete before read_state), concurrent dispatch could produce unexpected results if the LLM happens to call both in the same turn.

The fix is to assign them to a shared lane:
```
  @skill(lane="state")
  def write_state(self, value: str) -> str: ...

  @skill(lane="state")
  def read_state(self) -> str: ...
```
  In practice, LLMs generally reason through dependent tool chains step-by-step and won't call both in a single turn — but for safety-critical sequences, adding a lane is recommended and needs to be well documented. See Next Steps for a `default_skill_lane config` that avoids annotating every method individually.

---

## How to Test

 Run the unit tests — the timing in the two new tests directly proves the behavior:
  
  `uv run pytest dimos/agents/mcp/test_mcp_client_unit.py dimos/agents/mcp/test_mcp_server.py -v`

  - `test_same_lane_tools_execute_serially` completes in ~0.11s (two 0.05s sleeps run back-to-back)
  - `test_different_lane_tools_run_in_parallel` completes in ~0.05s (both sleeps run simultaneously)
  - `test_lane_stored_in_tool_registry` verifies lane propagates end-to-end from tools/list into the client registry
  - `test_tools_list_includes_lane_when_set` verifies the server includes/omits lane correctly

The existing integration tests (`test_can_call_tool`, `test_multiple_tool_calls_with_multiple_messages`) cover backwards compatibility — they pass unchanged through the stream → astream migration.

---
### Next Steps
  
`default_skill_lane` module config: For modules where every skill belongs to the same lane (e.g. a `MotionModule` where every method is a motion command), it is verbose to annotate each method with `@skill(lane="motion")`. A `default_skill_lane: str | None = None` field on `ModuleConfig` would let a module declare a lane that applies to all its unannotated skills, with an explicit `@skill(lane=...)` overriding it per method. 

Implementation is two changes: add the field to `ModuleConfig`, and fall back to it in `get_skills()` when `__skill_lane__` is None.

### DIFF
--- a/dimos/agents/annotation.py
+++ b/dimos/agents/annotation.py
@@ -66,7 +66,17 @@ def _stamp_and_log(func_name: str, result: Any, elapsed_ms: float) -> Any:
     return result
 
 
-def skill(func: F) -> F:
+def skill(func: F | None = None, *, lane: str | None = None) -> F | Callable[[F], F]:
+    """Mark a method as an agent-callable skill, optionally assigned to a lane."""
+    if func is None:
+        def decorator(f: F) -> F:
+            return _make_skill(f, lane=lane)
+        return decorator  # type: ignore[return-value]
+
+    return _make_skill(func, lane=lane)
+
+
+def _make_skill(func: F, *, lane: str | None) -> F:
     if inspect.iscoroutinefunction(func):
 
         @functools.wraps(func)
@@ -108,4 +118,5 @@ def skill(func: F) -> F:
 
     wrapped = rpc(context_wrapper)
     wrapped.__skill__ = True  # type: ignore[attr-defined]
+    wrapped.__skill_lane__ = lane  # type: ignore[attr-defined]
     return cast("F", wrapped)
--- a/dimos/agents/mcp/mcp_client.py
+++ b/dimos/agents/mcp/mcp_client.py
@@ -143,6 +143,9 @@ class McpClient(Module):
 
         raw_tools = result.get("tools", [])
         self._tool_registry = {t["name"]: t for t in raw_tools}
+        for t in raw_tools:
+            if lane := t.get("lane"):
+                self._lane_locks.setdefault(lane, threading.Lock())
         tools = [self._mcp_tool_to_langchain(t) for t in raw_tools]
 
         if not tools:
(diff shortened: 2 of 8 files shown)


paul-nechifor:

PR #2239, opened 2026-05-24
Refactor coordinator for whole-body control

Nabla7:
## Problem

ControlCoordinator needs to support whole-body control before the G1 walking policy can land. Today the coordinator is mostly shaped around arms and simple joint tasks. Whole-body control needs per-joint gains, IMU state, and a cleaner way to create tasks without adding more special cases to the coordinator.

## Solution

This PR adds the generic coordinator changes needed for whole-body control.

Task-specific settings now live in task params instead of directly on `TaskConfig`. Task creation goes through a small lazy registry, so the coordinator no longer imports and constructs every task type itself. Whole-body hardware can now carry per-joint `kp` and `kd` gains, and the coordinator includes IMU readings in the state passed to tasks each tick.

This does not add the G1 GR00T policy itself.

## How to Test

```bash
python -m pytest dimos/control -q

### CODE REVIEW

File: dimos/hardware/whole_body/registry.py [hunk 1]
```diff
@@ -14,14 +14,21 @@
...
+  (``g1.py``, etc.). Sim engines live under ``dimos/simulation/`` so
+  their adapter glue lives there too instead of under ``hardware/``.
 
 Usage:
     from dimos.hardware.whole_body.registry import whole_body_adapter_registry
 
-    adapter = whole_body_adapter_registry.create("unitree_go2")
-    print(whole_body_adapter_registry.available())  # ["unitree_go2"]
+    adapter = whole_body_adapter_registry.create("sim_mujoco_g1")
+    print(whole_body_adapter_registry.available())  # ["sim_mujoco_g1", ...]
```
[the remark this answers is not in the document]
Nabla7: We'll have a sim adapter in the groot wbc PR.

dimos/control/tasks/registry.py
```diff
@@ -0,0 +1,97 @@
...
+    """Registry for control-task factories with lazy imports."""
+
+    def __init__(self) -> None:
+        self._factory_paths: dict[str, str] = {
+            "trajectory": "dimos.control.tasks.trajectory_task:create_task",
+            "servo": "dimos.control.tasks.servo_task:create_task",
+            "velocity": "dimos.control.tasks.velocity_task:create_task",
+            "cartesian_ik": "dimos.control.tasks.cartesian_ik_task:create_task",
+            "teleop_ik": "dimos.control.tasks.teleop_task:create_task",
+            "g1_groot_wbc": "dimos.control.tasks.g1_groot_wbc_task:create_task",
```
mustafab0: We have to maintain the `_factory_path` dict manually for each task that is added.

Consider moving to hardware adapter pattern for auto discovery. But you will need to maintain the lazy imports for the tasks, because tasks use heavy library dependencies.

We can make a manifest `__register__` file that is automatically discovered. So a user/developer can plugin a new task with a simple sub-directory rather than having to make changes to `registry.py`.

```
control
-tasks
 -servo_task
  -__registry__.py
  -servo_task.py
```

mustafab0 (changes requested):

dimos/control/tasks/registry.py
Nabla7: That is indeed cleaner, I updated the code to follow the pattern you suggested.

mustafab0 (approved):

PR #2241, opened 2026-05-24
Dimos map tool

leshy:
Purpose of this tool is to visualize maps directly from mem2 databases, do off-line loop closure and store them as pointclouds one can relocalize on

TODO for the future, add KVstore to actual mem2 database, relocalization module can consume mem2 database directly, and recompute loop closure/global map only if needed. if not, just load from kvstore. elimites the need for tooling like this. Tooling like this will stay for analysis and development, but not as a required part of relocalization flow


docs/capabilities/mapping/relocalization.md
arkluc: Since the output filename is {db_stem}.pc2.lcm, also change the name of saved premap from `{DB_NAME}_twopass_map` to `{DB_NAME}` for consistency.

dimos/msgs/sensor_msgs/PointCloud2.py
```diff
@@ -729,7 +729,10 @@ def to_rerun(
             point_colors = colors
         else:
             z = points[:, 2]
-            class_ids = ((z - z.min()) / (z.max() - z.min() + 1e-8) * 255).astype(np.uint8)
+
+            # Clip to robust percentiles so a few outliers don't squash the colormap range.
+            z_lo, z_hi = np.percentile(z, [1, 99])
```
arkluc: it's a good one

dimos/msgs/sensor_msgs/PointCloud2.py
paul-nechifor, line 688: Unrelated to this PR, but I'd like to see what you think about this.

I think it's not sustainable to have so many methods on messages. Having `to_rerun` on messages (or other such complicated functions) makes the messages very heavy and mixes concerns quite a bit.

I think it would be a lot easier to have messages just be "structs" with a few precise utility functions. (By this I'm referring to things which don't use external libs and are un-opinionated, for example there's only one way to convert a radian to a degree, but there are many choices to make when converting a occupancy map to an RGB representation of it. This particular function is both opinionated and uses external libs).

An alternative would be to have `dimos/msgs/conversions/rerun/to_rerun.py` . It would be a single function which takes any message, uses `match` and spits out a converted rerun object.

This also has the effect of containing all rerun imports to `dimos/msgs/conversions/rerun` and not needing to `import rerun` inside functions like we do now to avoid unnecessary imports when just using a message class for its data with no intention to use rerun.

I'm not saying we should do this now or anytime soon, just checking the temperature, to see if this is something we should pursue latter, or maybe when new methods are added to messages.

dimos/utils/cli/map.py
```diff
@@ -0,0 +1,157 @@
...
+            pass1_on_frame=progress(total, "pgo pass 1 (optimizing)"),
+            pass2_on_frame=progress(total, "pgo pass 2 (rebuilding)"),
+        )
+
+    global_map = (
+        lidar.tap(collect_path)
+        .transform(VoxelMapTransformer(voxel_size=voxel, device=device))
+        .tap(progress(lidar.count(), "reconstructing global map"))
+        .last()
+        .data
```
paul-nechifor, line 121: `VoxelMapTransformer` emits every transform. `last` is implemented as `.order_by("ts", desc=True).first()`.

Would it not be simpler to just emit one (with `emit_every=0`)and use `.first()` so no need for sorting? Something like:

```suggestion
    global_map = (
        lidar.tap(collect_path)
        .tap(progress(lidar.count(), "reconstructing global map"))
        .transform(VoxelMapTransformer(voxel_size=voxel, device=device, emit_every=0))
        .first()
        .data
```

dimos/mapping/relocalization/pgo.py
```diff
@@ -420,24 +431,35 @@ def pgo_then_voxels(
...
+                if pass2_on_frame is not None:
+                    pass2_on_frame(obs)
                 grid.add_frame(obs.data)
             return grid.get_global_pointcloud2()
 
-        kf_ts = np.array([kp.timestamp for kp in pgo._key_poses])
+        # Sort + dedupe by timestamp: Slerp requires strictly increasing ts.
+        kps = sorted(pgo._key_poses, key=lambda kp: kp.timestamp)
+        kps = [kp for i, kp in enumerate(kps) if i == 0 or kp.timestamp > kps[i - 1].timestamp]
+        kf_ts = np.array([kp.timestamp for kp in kps])
```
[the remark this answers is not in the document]
paul-nechifor: I think this is correct, right? @leshy

dimos/mapping/relocalization/pgo.py
```diff
@@ -420,24 +431,35 @@ def pgo_then_voxels(
     n_kf = pgo.num_key_poses
     print(f"  Pass 1: {n_frames} frames, {n_kf} keyframes")
 
+    if corrected_path_out is not None:
+        corrected_path_out.extend(
+            (float(kp.t_global[0]), float(kp.t_global[1]), float(kp.t_global[2]))
+            for kp in pgo._key_poses
```
paul-nechifor: Nit: please rename `_key_poses` to `key_poses` since it's used all over.

dimos/msgs/sensor_msgs/PointCloud2.py
```diff
@@ -729,7 +725,10 @@ def to_rerun(
...
```
paul-nechifor: Since this is purely cosmetic, you might want to add a small optimization for very large point clouds: only process a small subset. For example: only process the first 20000 points, or only process a random sample of the points `z = z if len(z) <= 20_000 else np.random.choice(z, 1000, replace=False)`

dimos/mapping/relocalization/pgo.py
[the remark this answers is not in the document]
jeff-hykin: this also seems correct to me

dimos/msgs/sensor_msgs/PointCloud2.py
leshy: yeah I don't mind this personally :)

dimos/msgs/sensor_msgs/PointCloud2.py
leshy: True, actually will exclude my PC2 changes from this PR, shouldn't change efficiency

dimos/msgs/sensor_msgs/PointCloud2.py
paul-nechifor: Sorry, can you be more specific. :sweat_smile:

leshy: I don't have a very strong opinion but I think you are probably right about everything above and won't mind if we start moving towards this

dimos/mapping/relocalization/pgo.py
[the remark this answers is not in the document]
leshy: pgo.py is super tempoary / vibed code already totally rewritten, let's ignore it for now, was used just for relocalization tests

dimos/utils/cli/map.py
leshy: oof I hate the sort here, didn't realize it creates a permanent buffer.. will look into it.

Your suggestion won't work as it will literally take the first message out of global map transformer, global map transformer has an internal state and we need it to eat all the lidar frames, that's why I ask for last() message.

This whole `emit_every=10` and inefficiency of this pipeline I'm aware of (emitting global map every 10 frames) but right now not optimizing but rushing through to deploy.. sort by ts I honestly wasn't aware of is happening here though

dimos/mapping/relocalization/pgo.py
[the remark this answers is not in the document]
leshy: https://github.com/dimensionalOS/dimos/pull/2242/ rewriten totally here and still iterating

dimos/mapping/relocalization/pgo.py
leshy: renamed, but as I said this is a temp file to be trashed

paul-nechifor (dismissed):

paul-nechifor (dismissed):

paul-nechifor (approved):

PR #2242, opened 2026-05-24
Loop closure / map Reconstruction first pass

leshy:
- pgo rewrite, map tool vis aditions
- full global map reconstruction takes a few seconds, we use memory2 to fetch 20% of total lidar frames based on time and space (latest only, not all)

- marker detector as a mem2 transform `dimos map hk_village5 --pgo --markers`

- marker detection debugger
```sh
cd dimos/utils/cli
python markers_rrd.py hk_village3 --out test3.rrd && rerun test3.rrd

# or for full map

dimos map hk_village3 --pgo --markers

```


dimos/memory2/type/observation.py
```diff
@@ -46,28 +50,123 @@ def __repr__(self) -> str:
...
+    """Sentinel for "argument not supplied", distinct from None."""
+
+    def __repr__(self) -> str:
+        return "<missing>"
+
+
+_MISSING = _MissingType()
+
+
+def _to_tuple(p: Any) -> PoseTuple | None:
```
paul-nechifor, line 63: Not a fan of this type of "guessing what the input is"... Why can't you use the actual message classes? Why avoid the import? If the objects already exist then it's already been imported.

dimos/memory2/stream.py
```diff
@@ -177,7 +178,11 @@ def at_relative(self, t: float, tolerance: float = 1.0) -> Stream[T, O]:
         return self.at(t0 + t, tolerance=tolerance)
 
     def near(self, pose: Any, radius: float) -> Stream[T, O]:
-        return self._with_filter(NearFilter(pose, radius))
+        # Accept Pose/PoseStamped (any object with `.position`), Vector3,
+        # numpy arrays, or (x, y, z) tuples — Vector3() handles the rest.
+        if hasattr(pose, "position"):
```
paul-nechifor, line 183: Instead of using hasattr and having this comment you can use `@overload` and use the actual types in multiple definitions.

paul-nechifor (approved):

dimos/memory2/stream.py
leshy: will adress in a followup

dimos/memory2/type/observation.py
leshy: me niether, will adress in a followup

PR #2244, opened 2026-05-25
Fastlio Refactor

LuigiVan01:
## Problem

  `main.cpp` of fastlio2  did not allow to drop in other source of LiDAR and IMU in a clean and easy way

Closes DIM-XXX

## Solution
Split `main.cpp` into headers and introduced a polymorphic base so alternate LiDAR sources can be dropped in

## How to Test

  Run the existing fastlio2 blueprint on an X2 and confirm SLAM output is unchanged:
  dimos run x2_slam

### DIFF
--- a/dimos/hardware/sensors/lidar/fastlio2/cpp/cloud_filter.hpp
+++ b/dimos/hardware/sensors/lidar/fastlio2/cpp/cloud_filter.hpp
@@ -12,10 +12,20 @@
 #include <pcl/point_cloud.h>
 #include <pcl/point_types.h>
 
+#include "dimos_native_module.hpp"
+
 struct CloudFilterConfig {
     float voxel_size = 0.1f;
     int sor_mean_k = 50;
     float sor_stddev = 1.0f;
+
+    static CloudFilterConfig from_args(const dimos::NativeModule& mod) {
+        CloudFilterConfig c;
+        c.voxel_size = mod.arg_float("voxel_size", c.voxel_size);
+        c.sor_mean_k = mod.arg_int  ("sor_mean_k", c.sor_mean_k);
+        c.sor_stddev = mod.arg_float("sor_stddev", c.sor_stddev);
+        return c;
+    }
 };
 
 /// Apply voxel grid downsample + statistical outlier removal in-place.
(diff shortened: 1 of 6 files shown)


dimos/hardware/sensors/lidar/fastlio2/cpp/pose_transform.hpp
```diff
@@ -0,0 +1,62 @@
+// Copyright 2026 Dimensional Inc.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Init-pose value type and quaternion helpers used by FastlioPublisher to
+// transform SLAM-origin outputs into the user's chosen world frame.
+
+#ifndef POSE_TRANSFORM_HPP_
```
leshy, line 7: hi tnx,

let's completely get rid of these transforms, transforms are for a downstream TF system, fastlio shouldn't care about this, it should only name it's frame_id, from that point on, TF system deals with positions.. 
I know you didn't implement this, just making sure we delete and don't continue building on this

dimos/hardware/sensors/lidar/fastlio2/cpp/cloud_filter.hpp
```diff
@@ -12,10 +12,20 @@
...
 #include <pcl/point_types.h>
 
+#include "dimos_native_module.hpp"
+
 struct CloudFilterConfig {
     float voxel_size = 0.1f;
     int sor_mean_k = 50;
     float sor_stddev = 1.0f;
+
+    static CloudFilterConfig from_args(const dimos::NativeModule& mod) {
```
leshy, line 22: For context, this fastlio module was created as my weekend experiment with a global map voxelizer, and a quick vibed connection to mid360 sdk as I didn't want to deal with raw lidar msgs.. have low respect for structures here and feel free to completely clean up this up, we should have JUST a nice fastlio integration. IDK what CloudFilter does at all.

IDK if we want to keep gluing drivers directly to fastlio or not (without pipeline msgs) I don't have a strong opinion at this time.

this module also has a bunch of global voxel mapper experiments that can simply be deleted, let's try and make this just a standard lean clean fastlio integration. 
(Potentially can leave current frame voxelizer in if you choose to)

as long as `dimos run mid360-fastlio` works, and andrew voxelizer (pls ask Andrew what's his blueprint) the rest here can go away

PR #2261, opened 2026-05-27
fix(g1): enable MuJoCo navigation with GR00T policy

raulbastidas1203:
## Summary

- Fix `unitree-g1-basic-sim` navigation wiring by adding `MovementManager`, so keyboard teleop and map-selected goals are muxed into the `cmd_vel` stream consumed by `G1SimConnection`.
- Add an opt-in GR00T ONNX locomotion policy for Unitree G1 MuJoCo simulation via `--mujoco-g1-policy groot`.
- Filter the simulated G1's own body points from MuJoCo lidar before publishing the point cloud, so click-to-nav does not immediately cancel due to self-obstacles in the costmap.
- Keep existing behavior as the default (`mujoco_g1_policy=default`) and add tests for blueprint wiring, policy selection, and G1 self-point filtering.

## Why

In G1 MuJoCo simulation, keyboard commands and clicked map goals could publish upstream, but the basic sim stack did not include the movement manager that bridges `tele_cmd_vel` / `nav_cmd_vel` into `cmd_vel`. As a result, the robot did not follow keyboard or map navigation commands as expected.

After wiring the movement manager, keyboard teleop worked, but map-selected navigation could still stop early because the MuJoCo depth/lidar output could include parts of the humanoid itself. Those self-points polluted the global map/costmap, causing the local planner to report an obstacle ahead and cancel/replan.

The GR00T policy is included as an optional simulation locomotion backend because it tracks velocity commands better for the humanoid G1 in MuJoCo while preserving the same DimOS navigation interface.

## Scope

Simulation only. This PR does not change the real G1 hardware path, and hardware validation is still pending.

## How to Test

```bash
source .venv/bin/activate
ruff check dimos/core/global_config.py \
  dimos/robot/unitree/g1/blueprints/basic/unitree_g1_basic_sim.py \
  dimos/robot/unitree/g1/blueprints/basic/test_unitree_g1_basic_sim.py \
  dimos/simulation/mujoco/model.py \
  dimos/simulation/mujoco/mujoco_process.py \
  dimos/simulation/mujoco/policy.py \
  dimos/simulation/mujoco/test_model_policy_selection.py \
  dimos/simulation/mujoco/test_mujoco_process.py

pytest dimos/simulation/mujoco/test_mujoco_process.py \
  dimos/robot/unitree/g1/blueprints/basic/test_unitree_g1_basic_sim.py \
  dimos/simulation/mujoco/test_model_policy_selection.py \
  dimos/navigation/movement_manager/test_movement_manager.py \
  dimos/robot/test_all_blueprints_generation.py -q
```

Manual sim check:

```bash
dimos --simulation --mujoco-g1-policy groot run unitree-g1-sim
```

Verified locally in simulation that:
- keyboard teleop moves the G1;
- publishing a `/clicked_point` goal produces `Found path`, `path_following`, and `Arrived at goal`;
- lidar/map visualization remains active.


raulbastidas1203:
Real G1 hardware validation is still pending, and I’ll try to complete it next week and update the PR with the results.

Nabla7:
Thanks for the PR. One important clarification: this is adding GR00T to the older unitree_g1_basic_sim / G1SimConnection MuJoCo path, whereas GR00T has already been implemented in the newer ControlCoordinator + MujocoSimModule stack via unitree-g1-groot-wbc which also supports running on the real robot already.

The main direction is the newer sim architecture, and we’re already updating that path to support nav/scene-loading as well. Because of that, I don’t think it makes sense to add a parallel GR00T implementation to the legacy sim stack, especially since that older path is likely to be removed soon.

For now I'm closing this PR. Take a look at the newer GR00T sim path instead, since that’s the version we expect to carry forward.

PR #2268, opened 2026-05-27
Add arm to test matrix

### LINKED ISSUE #2263: ARM CI
we'd like to test dimos on arm

Dreamsorcerer:
Fixes #2263.


leshy (approved):

leshy (changes requested):
actually @Dreamsorcerer "skip if no sqlite vec" is a bit unclear on where this happens,

can we have standardized skip tag per OS? like this would have skip_arm tag? so that we know which OS can cover which functionality

Dreamsorcerer:
> can we have standardized skip tag per OS? like this would have skip_arm tag? so that we know which OS can cover which functionality

We can, but this needs to be skipped in macos, so I was merging this skip. Can change to adding 2 skip markers if you prefer?

leshy (approved):
don't want to block too much on details, can move and reassess, but yeah generally a preferece for per-os ignores on tests, since then we can easily see our support  per OS, could we fix in the follow ups?

PR #2280, opened 2026-05-28
skip two nav tests

bogwi:
skip two nav tests


leshy:
for context, `gtsam-extended` was added as a dep for mapping tests, which implicitly enabled run of these tests in CI, which didn't work due to lfs, which blocked merges.

we tried tagging as tool, but it was an import level issue, so renamed for now @jeff-hykin

leshy (approved):

PR #2300, opened 2026-05-28
Pim/feat/g1 groot wbc part 2

Nabla7:
## Problem

Part 1 (#2239) added the coordinator pieces needed for whole-body control but did not include the policy. This PR lands the G1 GR00T walking policy as a first-class `ControlCoordinator` task and ships a blueprint to run it.

## Solution

Adds a `g1_groot_wbc` `ControlTask` that runs the balance and walk ONNX policies, with the arming / dry-run / ramp lifecycle wired through the coordinator.

Adds the `unitree-g1-groot-wbc` blueprint, which selects the backend with `--simulation`:

* Real hardware uses `G1WholeBodyConnection` + `transport_lcm` at 500 Hz
* Simulation uses `MujocoSimModule` + `sim_mujoco_g1` at 50 Hz

Twists on `/g1/cmd_vel` flow through the coordinator into the task's velocity command.

## How to Test

Real hardware:

```bash
dimos run unitree-g1-groot-wbc -o g1wholebodyconnection.network_interface=<iface>
```

In a Python shell:

```python
from dimos import Dimos

app = Dimos.connect()
app.ControlCoordinator.set_dry_run(False)
app.ControlCoordinator.set_activated(True)
```

Wait 10 seconds for the ramp. Then publish a `Twist` to `/g1/cmd_vel` to drive it.

Simulation:

```bash
dimos --simulation mujoco run unitree-g1-groot-wbc
```

* [ ] I have read and approved the CLA: https://github.com/dimensionalOS/dimos/blob/main/CLA.md


dimos/control/tasks/g1_groot_wbc_task/g1_groot_wbc_task.py
```diff
@@ -0,0 +1,788 @@
...
+if TYPE_CHECKING:
+    from dimos.hardware.whole_body.spec import WholeBodyAdapter
+    from dimos.msgs.geometry_msgs.Twist import Twist
+
+logger = setup_logger()
+
+
+# The 29 DDS motor names + their kp/kd for the GR00T-trained policies.
+# Lifted verbatim from g1-control-api/configs/g1_groot_wbc.yaml, which
+# itself copies GR00T-WBC's g1_29dof_gear_wbc.yaml. Diverging from these
```
mustafab0: What are these files?

Nabla7: The original groot wbc control policy from the original repo ships with specific kp kd values against which the policy was trained, this is a leftover comment. Will remove.

dimos/robot/unitree/g1/wholebody_connection.py
```diff
@@ -51,7 +51,12 @@
 
 _NUM_MOTORS = 29
 _NUM_MOTOR_SLOTS = 35  # G1 hg LowCmd has 35 slots; only 29 are used
-_MODE_MACHINE_WAIT_S = 10.0
+# mode_machine is a static value identifying the G1 firmware/hardware
+# variant.  Older code read it back from the first LowState frame and
+# echoed it into LowCmd; that callback path is unreliable on macOS
+# cyclonedds, and the value never changes for a given robot, so we
+# hardcode it.  29-DOF G1 (gear) reports 5.
+_MODE_MACHINE_G1: int = 5
```
mustafab0, line 58: Does this mean the `_MODE_MACHINE_G1` will change for different users/robot? 

If that is the case hardcoding it is not ideal. Should be user configurable

dimos/simulation/adapters/whole_body/g1.py
```diff
@@ -0,0 +1,205 @@
...
+        self._shm = None
+        self._connected = False
+
+    def is_connected(self) -> bool:
+        return self._connected and self._shm is not None
+
+    # IO (WholeBodyAdapter protocol)
+
+    def read_motor_states(self) -> list[MotorState]:
+        if not self._connected or self._shm is None:
```
mustafab0: you already have a `has_motor_states` helper method that does this check 

line 154

mustafab0: Should just use the helper everywhere, or delete the helper

dimos/control/tasks/g1_groot_wbc_task/g1_groot_wbc_task.py
```diff
@@ -0,0 +1,788 @@
...
+]
+G1_GROOT_KD: list[float] = [
+    2.0,
+    2.0,
+    2.0,
+    4.0,
+    2.0,
+    2.0,  # left leg
+    2.0,
+    2.0,
```
mustafab0, line 104: I am seeing Kp, Kd and tau everywhere. And they seem to be set/changed together everywhere.

we should make a custom message type that encodes Kp, Kd and Tau into it just how  we have `Twist` or `Pose` that encode `x y z wx wy wz` 

list[float] is unsightly for all joints. 

To not block this PR from being merged. Please create an issue to track this refactor.

mustafab0 (dismissed):

dimos/control/tasks/g1_groot_wbc_task/__registry__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2025-2026 Dimensional Inc.
```
paul-nechifor, line 1: I don't think `__registry__` is an established dunder name

dimos/robot/unitree/g1/blueprints/basic/unitree_g1_groot_wbc.py
```diff
@@ -0,0 +1,238 @@
...
+
+# Lazy data handles. LfsPath only triggers the LFS pull on first
+# str()/open(); using ``get_data(...)`` at import time would block the
+# whole CLI on a multi-GB download every time the module is imported.
+_GROOT_MODEL_DIR = LfsPath("groot")
+_MJCF_PATH = LfsPath("mujoco_sim/g1_gear_wbc.xml")
+
+_adapter_address: str | Path
+
+if global_config.simulation:
```
paul-nechifor: simulation is no longer a bool. You have to use `global_config.simulation == "mujoco"`

dimos/robot/unitree/g1/wholebody_connection.py
```diff
@@ -167,6 +173,29 @@ def stop(self) -> None:
...
+                        self._low_cmd.motor_cmd[i].mode = 0x00  # disable
+                        self._low_cmd.motor_cmd[i].q = POS_STOP
+                        self._low_cmd.motor_cmd[i].dq = VEL_STOP
+                        self._low_cmd.motor_cmd[i].kp = 0
+                        self._low_cmd.motor_cmd[i].kd = 0
+                        self._low_cmd.motor_cmd[i].tau = 0
+                    self._low_cmd.crc = self._crc.Crc(self._low_cmd)
+                    self._publisher.Write(self._low_cmd)
+                logger.info("Sent safe-stop lowcmd (motors disabled)")
+            except (OSError, RuntimeError, AttributeError) as e:
```
paul-nechifor: This exception is just for `self._publisher.Write(self._low_cmd)`, right?

Wrap just that line in try-except.

dimos/robot/unitree/g1/wholebody_connection.py
```diff
@@ -190,56 +219,114 @@ def stop(self) -> None:
...
+        self,
+    ) -> (
+        tuple[
+            list[float],
+            list[float],
+            list[float],
+            tuple[float, float, float, float],
+            tuple[float, float, float],
+            tuple[float, float, float],
+        ]
```
paul-nechifor: This inline type is a bit long. It should be a sparate type where the complex structure of floats is explained.

Presumably it's this, right?

```
        positions: list[float],
        velocities: list[float],
        efforts: list[float],
        quat: tuple[float, float, float, float],
        gyro: tuple[float, float, float],
        accel: tuple[float, float, float],
```

Maybe make that a dataclass.

data/.lfs/groot.tar.gz
```diff
@@ -0,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
```
paul-nechifor, line 1: (Line not related.)

I see no tests in this PR. It would be useful to add some meaningful tests to make sure the functionality doesn't get easily broken in the future.

dimos/control/tasks/g1_groot_wbc_task/__registry__.py
Nabla7: This is following the repo-local task manifest convention introduced in #2239 from @mustafab0 's review, where ControlTaskRegistry discovers <task>/__registry__.py. Should we use a different naming convention for this?

dimos/control/tasks/g1_groot_wbc_task/g1_groot_wbc_task.py
Nabla7: Tracked separately in #2436 so this does not block the GR00T WBC PR.

dimos/robot/unitree/g1/wholebody_connection.py
Nabla7: Addressed: `_MODE_MACHINE_G1` is now only the default value for the 29-DOF G1 gear variant. `G1WholeBodyConnectionConfig` exposes `mode_machine`, and `start()` writes `self.config.mode_machine` into `LowCmd`, so users can override it with:

```bash
dimos run unitree-g1-groot-wbc -o g1wholebodyconnection.mode_machine=<value>
```

dimos/control/tasks/g1_groot_wbc_task/g1_groot_wbc_task.py
Nabla7: Removed the stale references to external source files that are not in this repo. The comments now describe the local policy contract directly instead of pointing at nonexistent paths.

dimos/control/tasks/g1_groot_wbc_task/__registry__.py
mustafab0: @paul-nechifor will change this name as clean up work.

Issue added here: https://github.com/dimensionalOS/dimos/issues/2473

mustafab0 (commented):
Please address pending comments and mark them as resolved. 

wbc task looks okay to merge.

data/.lfs/groot.tar.gz
Nabla7: Added tests in three areas:

`test_g1_groot_wbc_task.py`

Behavioral coverage for the G1 WBC task with ONNX stubbed out, so it runs without weights or hardware and finishes in milliseconds.

- Pins the full 86-float observation vector against the policy contract: command scaling, default-position offsets, gyro/dq scaling, and projected gravity. This is the main regression guard, since a bad observation layout usually won’t crash, it just makes the policy behave wrong.
- Covers walk/balance policy selection, including fallback to balance when the velocity command goes stale.
- Checks safety modes: unarmed holds the measured pose and skips inference; arming ramps from the current pose to policy defaults; dry-run keeps inference warm but emits no commands.
- Verifies projected gravity against analytic references in `w,x,y,z` order, catching the usual `wxyz`/`xyzw` quaternion mixup.
- Checks that partial joint-state updates keep both the 29-joint cache and the 15-claimed-joint cache consistent.

`test_robot_sim_binding.py`

Runs real MuJoCo against an intentionally awkward MJCF: the scene freejoint is declared before the robot, and the actuators are in swapped order. The binding has to find the robot floating base and use policy joint order rather than document order, which is exactly what tends to break in composed scenes.

Also checks that missing IMU config or bad sensor dimensions fail loudly instead of producing garbage.

`test_mujoco_sim_module.py`

Covers the SHM ready signal. It is only raised after both joint state and IMU data have been written, so a consumer waking on `ready` can’t read a half-written frame.

dimos/robot/unitree/g1/wholebody_connection.py
Nabla7: Done, turned it into a frozen dataclass.

mustafab0 (approved):

PR #2303, opened 2026-05-28
Remove * from CODEOWNERS

Dreamsorcerer:
I've not heard confirmation from Ivan, but I think this was added by mistake. This line blocks all PRs from being merged without approval from one of those 4 people.

We already have a pr-approvers team that must approve _every_ PR, regardless of files changed. So, this is basically just a duplicate.


Dreamsorcerer:
Maybe it makes sense to just remove the pr-approvers team instead...

paul-nechifor:
> I've not heard confirmation from Ivan

I talked to Ivan before merging.

Dreamsorcerer:
I've removed pr-approvers then. We can use CODEOWNERS alone.

Issue #2204, opened 2026-05-21
Agent skills: distinguish parallel-safe vs sequential-only execution


spomichter:
Currently there is no distinction between skills that can be dispatched in parallel (e.g. perception queries, lookups) and skills that **must** run sequentially because they are physical actions (e.g. navigate, pick, place). The agent has no way to know which skills are safe to chain concurrently and which require ordered execution.

We need a mechanism (decorator flag, capability annotation, or similar) so the agent can parallelize independent queries while still enforcing sequential ordering for action skills that mutate world state.

---
*Synced from* [DIM-939](https://linear.app/dimensional/issue/DIM-939) *by summer*

leshy:
proposal for API here is to specify different lanes for skills, two skills in the same lane, wait for each other
```python
@skill(lane="motion")
@skill(lane="audio_out")
```

PR #2308, opened 2026-05-29
Add skill lanes for MCP tool serialization

RitwijParmar:
## Problem

Addresses #2204. DimOS skills currently do not expose whether a tool can run concurrently with another tool. That is risky for robot/world-mutating skills such as motion or manipulation: two MCP calls can be dispatched at the same time even when they should share an execution lane.

Closes DIM-XXX

## Solution

- Add a parameterized `@skill(lane="motion")` form while preserving the existing bare `@skill` decorator.
- Carry the lane through `SkillInfo` discovery so module skill metadata knows which lane a skill belongs to.
- Include lane metadata in MCP `tools/list` responses under tool `_meta`.
- Serialize concurrent MCP `tools/call` requests that share the same non-empty lane, while leaving unlaned or differently-laned skills concurrent.

## How to Test

- `/tmp/dimos/.venv/bin/python -m ruff check dimos/agents/annotation.py dimos/core/module.py dimos/agents/mcp/mcp_server.py dimos/agents/test_skill_result.py dimos/agents/mcp/test_mcp_server.py`
- `/tmp/dimos/.venv/bin/python -m pytest dimos/agents/test_skill_result.py dimos/agents/mcp/test_mcp_server.py`

Note: I used Python 3.12 through `uv`. A full `uv run` sync initially hit local `pyaudio`/PortAudio headers on this Mac, so I provisioned a minimal env for the focused files above.

### DIFF
--- a/dimos/agents/annotation.py
+++ b/dimos/agents/annotation.py
@@ -18,7 +18,7 @@ import functools
 import inspect
 import threading
 import time
-from typing import Any, TypeVar, cast
+from typing import Any, TypeVar, cast, overload
 
 from dimos.core.core import rpc
 from dimos.utils.logging_config import setup_logger
@@ -66,7 +66,10 @@ def _stamp_and_log(func_name: str, result: Any, elapsed_ms: float) -> Any:
     return result
 
 
-def skill(func: F) -> F:
+def _decorate_skill(func: F, *, lane: str | None = None) -> F:
+    if lane == "":
+        raise ValueError("skill lane must be a non-empty string or None")
+
     if inspect.iscoroutinefunction(func):
 
         @functools.wraps(func)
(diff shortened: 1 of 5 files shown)


RitwijParmar:
Addressed the test-isolation concern from the Greptile review.

What changed:
- `handle_request` now accepts an optional `lane_locks` store, while the FastAPI endpoint still defaults to `app.state.lane_locks` for production behavior.
- The same-lane serialization test now passes a local lock store.
- Added a regression test that exercises the same lane name across two separate `asyncio.run()` event loops with fresh lock stores, so loop-bound `asyncio.Lock` objects do not leak between tests.

Validation run locally:
- `/Users/ritwij/.cache/codex-runtimes/codex-primary-runtime/dependencies/python/bin/python3 -m compileall dimos/agents/mcp/mcp_server.py dimos/agents/mcp/test_mcp_server.py`

I could not run the full pytest target locally because the repo's `uv`/`pixi` test environment is not installed on this machine.

paul-nechifor:

PR #2316, opened 2026-06-01
worktree command, docs/agents moved to docs/coding-agents

leshy:
adds a `./bin/worktree` cmd (totally vibed) that quickly creates a dimos worktree and installs dimos in it (pytest/mypy passes)

renames docs/agents to docs/coding-agents (there was some confusion here, these are docs for coding agents working on dimos, not docs ABOUT agents in dimos)


spomichter (approved):

bin/worktree
```diff
@@ -0,0 +1,165 @@
+#!/usr/bin/env bash
+# Manage ready-to-work git worktrees for parallel/agent development.
```
Dreamsorcerer, line 2: Don't you just do `claude --worktree`?

AGENTS.md
Dreamsorcerer, line 1: We still need to move and replace this file. This entire file will end up in the context for most coding agents.

Issue #2119, opened 2026-05-17
Validate readme/install examples


leshy:
Someone recently on discord had quite an unhappy time trying to onboard on dimos
[https://discord.com/channels/1341146487186391173/1344622457327321150/1505059108800299078](<https://discord.com/channels/1341146487186391173/1344622457327321150/1505059108800299078>)

As well as me on osx with an in-person candidate

Please vaidate our instructions per OS, both installed via uv and used as a repo checkout. Go through our install flow per robot, make sure it's reasonable. Go2 needs to be checked for sure, G1? Probably all platforms we claim are Alpha or better supported.

Pay attention not just to stuff breaking but to details that are wrong. For example should go2 install flow pull all deps or just unitree, etc? are we mentioning system libraries people need to install (turbojpeg, some audio stuff)

This needs to be checked by hand for now but automation work welcome [https://github.com/dimensionalOS/dimos/issues/2118](<https://github.com/dimensionalOS/dimos/issues/2118>)

---
*Synced from* [DIM-892](https://linear.app/dimensional/issue/DIM-892) *by summer*

arkluc:
Also noticed some inconsistency between the main and internal READMEs. Please refer to 5c3005 @ main.

https://github.com/dimensionalOS/dimos/blob/5c30059c4e8846f34d3d2ae15a778dd513074ad6/README.md

```bash
# Replay a recorded quadruped session (no hardware needed)
# NOTE: First run will show a black rerun window while ~75 MB downloads from LFS
dimos --replay run unitree-go2
```

https://github.com/dimensionalOS/dimos/blob/5c30059c4e8846f34d3d2ae15a778dd513074ad6/docs/platforms/quadruped/go2/index.md

```bash
# Replay a recorded Go2 navigation session
# First run downloads ~2.4 GB of LiDAR/video data from LFS
dimos --replay run unitree-go2
```

PR #2320, opened 2026-06-01
docs(install): fix required system libs, replay size, and library-ins…

KrishnaH96:
Part of #2119 (validate README / install examples). Small, hand-validated install-doc fixes (tested on Ubuntu 24.04).

## Changes
- **`ubuntu.md`** — add `libgl1 libegl1` (OpenGL runtime for the always-installed `open3d` + `rerun-sdk` core deps; matches `scripts/install.sh`). Without them the visualizer fails at runtime with `libGL.so.1: cannot open shared object file` on minimal/headless/Docker installs. Also drop `pre-commit` (a dev `[dependency-groups]` tool, not needed by library users).
- **`osx.md`** — drop `pre-commit`.
- **`ubuntu.md` / `osx.md`** — note the inline `export PATH` is session-only (uv persists via the shell profile; open a new terminal).
- **`go2/index.md`** — correct the replay first-run download to **~85 MB LFS / ~220 MB SQLite DB** (was "~2.4 GB of LiDAR/video"), resolving the README-vs-`go2` size inconsistency raised on #2119.

## Notes
Draft — description to be expanded. Broader #2119 findings (agentic-runfile import crash on the PyPI wheel, `base[unitree]` pulling torch, dev-install command divergence) will be summarized separately.


docs/installation/osx.md
```diff
@@ -4,10 +4,12 @@
 # install homebrew
 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
 # install dependencies
-brew install gnu-sed gcc portaudio git-lfs libjpeg-turbo python pre-commit
+brew install gnu-sed gcc portaudio git-lfs libjpeg-turbo python
```
leshy: why drop pre-commit?

leshy: in both os install instructions pre-commit is used if using as a git-checkout

docs/installation/ubuntu.md
```diff
@@ -2,10 +2,20 @@
 
 ```sh skip
 sudo apt-get update
-sudo apt-get install -y curl g++ portaudio19-dev git-lfs libturbojpeg python3-dev pre-commit
+# Required system libraries. libturbojpeg + portaudio19-dev cover image/audio; libgl1 +
+# libegl1 are the OpenGL runtime for open3d and rerun-sdk (both always-installed core deps).
+# Without libgl1/libegl1 the visualizer fails at runtime with
```
leshy: why all these comments?

dimos/robot/unitree/go2/blueprints/agentic/_common_agentic.py
```diff
@@ -27,6 +27,11 @@
     UnitreeSkillContainer.blueprint(),
     WebInput.blueprint(),
     SpeakSkill.blueprint(),
+).requirements(
+    # SpeakSkill (OpenAI TTS) needs OPENAI_API_KEY — fail fast at blueprint build with a
+    # clear message instead of an ExceptionGroup deep in start_all_modules. Mirrors the
+    # ollama_installed gate used by the -ollama variant.
+    openai_api_key_set,
```
leshy: @Dreamsorcerer if we put this key as required config value for a module, will we be able to auto-crash? not to have a special case here

docs/installation/ubuntu.md
```diff
@@ -2,10 +2,20 @@
...
+sudo apt-get install -y curl g++ portaudio19-dev git-lfs libturbojpeg libgl1 libegl1 python3-dev
+
+# optional: graphviz enables blueprint-graph visualization. Without it dimos logs
+# "graphviz not found, skipping blueprint graph" at startup (everything else still works).
+# sudo apt-get install -y graphviz
 
 # install uv
 curl -LsSf https://astral.sh/uv/install.sh | sh && export PATH="$HOME/.local/bin:$PATH"
+# NOTE: the `export` above only affects the current shell. Open a new terminal (or run
+# `source ~/.bashrc`) so `uv` is on PATH in later sessions — the installer also adds it there.
```
paul-nechifor: I don't think this comment is necessary. Are there cases where uv is not accessible in other terminals after installation?

dimos/robot/unitree/go2/blueprints/agentic/_common_agentic.py
Dreamsorcerer: Yes, so it can either be a key for the module specifically, or we could add it to the global config and link to it.
Roughly:
`openai_key: str = Field(default_factory=lambda d: d["g"].openai_key)`

The global config openai_key can still be `str | None`.

You might also need to add `validate_default=True` to the pydantic config (maybe add it to BaseConfig or ModuleConfig to apply everywhere?)

PR #2351, opened 2026-06-03
feat: FFI for Rust modules

### LINKED ISSUE #2332: Rust FFI
Create Rust bindings for Python so we can run core module logic in mem2 pipelines. Voxel mapper is a good one to start with.

---
*Synced from* [DIM-974](https://linear.app/dimensional/issue/DIM-974) *by andrew*

aclauer:
## Problem

For testing and dev with mem2, it's easier to call core module functionality as functions instead of deploying a blueprint. For non-Python modules, we need an FFI to expose the module functions.

Closes DIM-974

## Solution

Refactor the ray trace mapper and 3d planner to include Python Rust bindings. Now we can call the core functionality of each in Python.

## How to Test
Build .so files for both
```
uv run maturin develop --uv --release -m dimos/mapping/ray_tracing/rust/Cargo.toml
uv run maturin develop --uv --release -m dimos/navigation/nav_3d/mls_planner/rust/Cargo.toml
```
Run tests on imported FFI
```
uv run pytest dimos/mapping/ray_tracing/test_voxel_map.py dimos/navigation/nav_3d/mls_planner/test_mls_planner.py
```


pyproject.toml
```diff
@@ -40,6 +45,7 @@ exclude = [
     "**/*.css",
     "**/*.js",
     "**/*.tcss",
+    "**/*.pyi",
```
paul-nechifor, line 48: Why exclude pyi files?

aclauer: This includes them, no?

paul-nechifor: Oops. I read Greptile's comment above and thought these was `exclude`. 🤦‍♂️

dimos/mapping/ray_tracing/test_voxel_map.py
```diff
@@ -0,0 +1,87 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numpy as np
+import pytest
+
+pytest.importorskip("dimos_voxel_ray_tracing")
```
paul-nechifor, line 18: Personally, I don't like this. If the module fails to import, these tests are never ran. And if they actually fail, you'd never know because they're skipped.

dimos/navigation/nav_3d/mls_planner/mls_planner.py
```diff
@@ -0,0 +1,27 @@
...
+"""Python interface to the Rust multi-level surface path planner."""
+
+from __future__ import annotations
+
+try:
+    from dimos_mls_planner import MLSPlanner
+except ImportError as e:
+    raise ImportError(
+        "dimos_mls_planner is not built. Run: "
+        "uv run maturin develop --uv -m dimos/navigation/nav_3d/mls_planner/rust/Cargo.toml"
```
paul-nechifor, line 24: Do we have to run this manually? Does CI compile the extensions?

aclauer: Yeah, so I didn't actually add these to CI yet, not sure where the best place to put them is. I guess I should run them in the existing Rust job now that I think of it.

The reason I included that import error is because when you're developing you have to rebuild the .so manually for your changes to be picked up. Not a great work flow, but I'm not sure if there's a way around it?

dimos/mapping/ray_tracing/test_voxel_map.py
aclauer: True, let me add the tests to CI and then I'll drop this

dimos/navigation/nav_3d/mls_planner/mls_planner.py
aclauer: Added to CI

dimos/mapping/ray_tracing/test_voxel_map.py
aclauer: I added to the Rust job, but I need to leave these because otherwise all the other jobs will fail unless they build the FFI, which isn't necessary.

The Rust test will fail if it doesn't build or can't be imported, so no worries about the tests getting silently skipped.

paul-nechifor (approved):

PR #2355, opened 2026-06-04
feat: auto fix common review issues

paul-nechifor:
## Problem

Reviewing takes a long time. Some issues come up multiple times, so we can automate them.

Closes DIM-978

## Solution

* Add a script to automatically fix common review issues.
* Instead of just saying what the issues are, make fixes for them and allow the author to drop commits he doesn't agree with.


leshy:
discussed in private, should move to agents/auto-fix or something like this

leshy (approved):

PR #2362, opened 2026-06-05
Integrate Zenoh

paul-nechifor:
## Problem

We need to support Zenoh as well.

Closes DIM-955

## How to Test

Run a blueprint with Zenoh communication:
```bash
uv run dimos --transport=zenoh --simulation run unitree-go2-agentic
```

Start `humancli`, also with Zenoh:

```bash
uv run humancli --transport=zenoh
```


dimos/core/transport_factory.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+    if g.transport == "zenoh":
+        return "dimos/" + name.lstrip("/")
+    return name if name.startswith("/") else "/" + name
+
+
+def make_transport(
+    name: str, msg_type: type | None = None, *, g: GlobalConfig = global_config
+) -> PubSubTransport[Any]:
+    """Construct the active-backend pub/sub transport for a logical channel.
```
leshy, line 77: Few things, Transport isn't neccessarily a PubSubTransport - we can use TCP and IP address as a setting here etc (in theory, haven't implemented)

in case of PubSubTransport, a string doesn't define a full topic, there is a reason why Topic for LCM and Zenoh is a different object. Zenoh offers QoS settings etc per channel. maybe specific router config etc.

So I'm thinking when global switch zenoh or LCM is used, for lcm that can literally be just LCM(topic_string) but zenoh probably wants reliable delivery for RPC specifically or specific per topic configuration (Image can be unreliable, but not agent messages)

I'm not sure right now what to suggest here - if we can normalize transport requirements across topics "this is reliable", "this is unreliable" or if we need per transport global blueprint config overlay that this global config switch just applies? global overlay seems better to me

dimos/protocol/rpc/pubsubrpc.py
```diff
@@ -321,6 +322,27 @@ def topicgen(self, name: str, req_or_res: bool) -> Topic:
         return Topic(topic=topic)
 
 
+class ZenohRPC(PubSubRPCMixin[Topic, Any], PickleZenoh):
```
leshy, line 329: We can do this initially, but Zenoh actually supports RPC on their protocol level, so we dont need to piggyback to pubsub here.

dimos/protocol/pubsub/impl/zenohpubsub.py
```diff
@@ -0,0 +1,280 @@
...
+from collections.abc import Callable, Iterable
+from functools import lru_cache
+import threading
+from typing import Any
+
+import zenoh
+
+from dimos.msgs.helpers import resolve_msg_type
+from dimos.protocol.pubsub.encoders import LCMEncoderMixin, PickleEncoderMixin
+from dimos.protocol.pubsub.impl.lcmpubsub import Topic as Topic
```
leshy: you should implement your own `Topic` which configures Zenoh-specific topic settings like QoS, instead of adding another global config layer for qos

`moule.image.transport = Zenoh(Topic("/bla", qos=...))`

I think simplifies this a lot

dimos/robot/unitree/go2/connection.py
```diff
@@ -382,9 +383,9 @@ def deploy(dimos: ModuleCoordinator, ip: str, prefix: str = "") -> "ModuleProxy"
         f"{prefix}/image", default_capacity=DEFAULT_CAPACITY_COLOR_IMAGE
     )
 
-    connection.cmd_vel.transport = LCMTransport(f"{prefix}/cmd_vel", Twist)
+    connection.cmd_vel.transport = make_transport(f"{prefix}/cmd_vel", Twist)
 
-    connection.camera_info.transport = LCMTransport(f"{prefix}/camera_info", CameraInfo)
+    connection.camera_info.transport = make_transport(f"{prefix}/camera_info", CameraInfo)
```
leshy: can remove these deploy() functions, not to mention transports within modules..

leshy (dismissed):

bogwi:
@paul-nechifor there is no camera in rerun on mac.

On Darwin unitree_go2_basic.py forces color_image onto pSHMTransport.

Under --transport=zenoh the rerun bridge subscribes to Zenoh + LCM only but not shared memory. So camera frames go to pSHM and the bridge never sees them. Linux is ok because it skips the Mac override. You keep the line that does this unconditionally. 

@jeff-hykin code, `unitree_go2_basic.py` 32-38
```python
_mac_transports: dict[tuple[str, type], pSHMTransport[Image]] = {
    ("color_image", Image): pSHMTransport(
        "color_image", default_capacity=DEFAULT_CAPACITY_COLOR_IMAGE
    ),
}

_transports_base = (
    autoconnect() if platform.system() == "Linux" else autoconnect().transports(_mac_transports)
```

We can compute _transport_base:
```python
def _platform_transports() -> Any:
    if platform.system() == "Linux":
        return autoconnect()
    if global_config.transport == "lcm":
        #Darwin: pSHM for color_image only under LCM
        return autoconnect().transports(_mac_transports)
    # under Zenoh, do not override color_image to pSHM.
    return autoconnect()


_transports_base = _platform_transports()
```

leshy:
plz benchmark, seems slow
`python -m pytest -svm tool dimos/protocol/pubsub/benchmark/test_benchmark.py`

dimos/robot/unitree/go2/connection.py
paul-nechifor: removed.

dimos/protocol/rpc/pubsubrpc.py
paul-nechifor: Will do this in a separate issue.

leshy (dismissed):

leshy (approved):

PR #2306, opened 2026-05-29
dimos map tooling, stream alignment

leshy:
# `dimos map` utilities for dataset collection team

you can ignore those, they are temporary,

changing pose source for observation stream, rendering global map, loop closures etc. lots of tooling (temporary and will change a lot)

dataset validation instructions document

# stream alignment

a very simple stream.align tool - this needs much more work. multi stream input and output, interpolation of poses/transforms etc

PR #2364, opened 2026-06-05
Fan-io

bogwi:
# Fan-io

## Problem
`StreamModule` only supports 1 input and 1 output.

## Solution
* We want to support multiple inputs (which can be aligned on time).
* And multiple outputs emitted at once.
* The first declared `In` regulates all emission times. The pipeline must construct a Bundle which is used to publish messages to multiple outputs. 

## Proof the platform works

- `memory2/test_module.py` fusion tests (literal N:M)
- `MarkerModule` migration (clearest before/after)
- `Detection3DModule` migration (Rx -> pipeline, compute-once multi-out)

## Out of scope

module-as-transform in arbitrary outer chains (will be new mem2-native-modules PR).

## What this branch is (the core deliverable)

| Piece                      | File                       | What it adds                                                             |
| -------------------------- | -------------------------- | ------------------------------------------------------------------------ |
| Generalized `StreamModule` | `memory2/module.py`        | N in / M out; `ingest()` seam; `self.streams.<port>`; no 1:1 gate        |
| Fan-out primitives         | `memory2/fanio.py`         | `Bundle`, `scatter_to_ports`, `normalize_to_bundle`                      |
| Cross-stream pairing       | `memory2/stream.py`        | `Stream.align`, optional `interpolator=`                                 |
| Stock interpolators        | `memory2/interpolators.py` | `lerp_pose`, `interp_odom`                                               |
| Contract tests             | `memory2/test_module.py`   | `TwoInputFusion`, `ChainedFusion`, multi-out compute-once, interpolation |

## Why the design

Before fan-I/O:

- MarkerModule copied 24 lines of start() for one ingestion change
- Detection3DModule was not a mem2 pipeline at all - Rx alignment in start()
- Multi-out meant manual subscribe/publish or duplicate compute
- Every new fusion module picked its own private wiring idiom

After fan-I/O:

- Port count is data: declare N in, M out
- One authoring model: align -> transform -> Bundle ingest() for ingress enrichment without copying start()
- One subscribe, one compute, scatter to all outs
- Migrated modules share the same mental model as offline/replay streams

## Summary

Fan-I/O is the wrapper layer: mem2 fusion inside deployed StreamModules. That works today - and an important step toward **modules that are just mem2 transforms**: the same pipeline() you deploy on a robot should run in any mem2 chain, on a recorded SqliteStore, faster than realtime.

What survives from fan-I/O into that follow-up will be the core - every input as a mem2 stream, .align() / interpolator=, Bundle scatter, ingest() - but not the legacy port/transport wrapper around it.


dimos/memory2/fanio.py
```diff
@@ -0,0 +1,154 @@
...
+    A pipeline that already yields a :class:`Bundle` is passed through untouched -
+    including a single-``Out`` module whose tail is a ``Bundle`` (marker-style
+    1->1): its keys still route by port name, so scatter publishes
+    ``bundle[port_name]`` rather than the whole bundle. With multiple ``Out``
+    ports there is no single key to wrap a raw payload under, so the stream is
+    returned unchanged and the pipeline must yield a ``Bundle`` by contract
+    (scatter raises otherwise).
+    """
+    if len(ports) != 1:
+        return stream
```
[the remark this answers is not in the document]
bogwi: That behavior is intentional and already spelled out:

With multiple ``Out`` ports there is no single key to wrap a raw payload under, so the stream is returned unchanged and the pipeline must yield a ``Bundle`` by contract (scatter raises otherwise).

dimos/memory2/module.py
```diff
@@ -90,42 +92,117 @@ def pipeline(self, stream: Stream[PointCloud2]) -> Stream[PointCloud2]:
...
+    def ingest(self, name: str, stream: Stream[Any], msg: Any) -> None:
+        """Append an incoming *msg* from In port *name* to its backing *stream*.
+
+        Override to enrich (e.g. anchor a robot pose), tag, or drop messages
+        before they enter the pipeline. This is the seam that replaces copying
+        ``start()``. The default stamps each observation with the message's own
+        ``.ts`` when present (so cross-port ``.align()`` is meaningful), falling
+        back to arrival time for unstamped payloads such as bare ints.
+        """
+        stream.append(msg, ts=getattr(msg, "ts", None) or time.time())
```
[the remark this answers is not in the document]
bogwi: suggested fix (is not None) would actually preserve 0.0 and treat it as a real timestamp

Not worth changing at all unless you explicitly want epoch-zero timestamps to be first-class - you do not  and the rest of the codebase does not either.

dimos/perception/detection/moduleDB.py
```diff
@@ -139,14 +141,14 @@ def to_pose(self) -> PoseStamped:
 class ObjectDBModule(Detection3DModule, TableStr):
```
leshy: oh sorry actually this ObjectDBModule module is really dumb (stores in memory, infinitely) and we can delete it, memory is replaced entirely now with mem2.. idk if easy to delete (what dependancies there are) - if not easy we will create a separate issue

bogwi: will check the boom it brings, if it's small - I'll remove it

dimos/perception/detection/module3D.py
```diff
@@ -12,53 +12,160 @@
...
+    Carries both the original 2D detections and the projected 3D detections so a
+    single fused tick can feed every ``Out`` (2D array, 3D array, per-object
+    pointclouds and cropped images) without recomputing anything.
+    """
+
+    detections_2d: ImageDetections2D
+    detections_3d: ImageDetections3DPC
+
+
+class Detection3DModule(StreamModule[Image, Bundle]):
```
leshy, line 98: Still thinking about this, first thoughts - should FusedDetections be a Bundle? two of these types seem very similar, so we don't have `_to_bundle` converter but can pass FusedDetections directly to output

bogwi: they look similar but are different layers and you cannot pass FusedDetections to outputs today, if I get your intent right

StreamModule.start() always ends in scatter_to_ports, which requires obs.data to be a Bundle. This is very intentional.

if you want less ceremony - then FusedDetections.to_bundle(config) -> Bundle instead of a module-level _to_bundle.  Same step, though hat does not remove the conversion step.

*upd: Bundle also features in in Paul's reviews

dimos/robot/unitree/go2/blueprints/smart/unitree_go2_detection.py
```diff
@@ -37,9 +38,12 @@
...
         {
             # Detection 3D module outputs
-            ("detections", Detection3DModule): LCMTransport(
+            ("detections_2d", Detection3DModule): LCMTransport(
                 "/detector3d/detections", Detection2DArray
             ),
+            ("detections_3d", Detection3DModule): LCMTransport(
+                "/detector3d/detections_3d", Detection3DArray
+            ),
             ("detected_pointcloud_0", Detection3DModule): LCMTransport(
```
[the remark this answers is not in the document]
bogwi: resolved [77af32f](https://github.com/dimensionalOS/dimos/pull/2364/commits/77af32fd0eca734324c780d4211476a43b082180)

dimos/robot/unitree/go2/blueprints/smart/test_unitree_go2_markers.py
paul-nechifor: This is a useless test. Blueprints are just configuration. There's no functionality to test.

dimos/robot/unitree/go2/blueprints/smart/test_unitree_go2_detection.py
```diff
@@ -0,0 +1,46 @@
...
+    ("detected_pointcloud_0", PointCloud2): "/detector3d/pointcloud/0",
+    ("detected_pointcloud_1", PointCloud2): "/detector3d/pointcloud/1",
+    ("detected_pointcloud_2", PointCloud2): "/detector3d/pointcloud/2",
+    ("detected_image_0", Image): "/detector3d/image/0",
+    ("detected_image_1", Image): "/detector3d/image/1",
+    ("detected_image_2", Image): "/detector3d/image/2",
+}
+
+
+def test_detector_transports_resolve_at_runtime() -> None:
```
paul-nechifor, line 36: Please delete this test. It just tests config.

dimos/robot/unitree/g1/blueprints/perceptive/unitree_g1_detection.py
```diff
@@ -43,62 +36,41 @@ def _person_only(det: Any) -> bool:
...
     .remappings(
         [
-            # Connect detection modules to camera and lidar
-            (Detection3DModule, "image", "color_image"),
-            (Detection3DModule, "pointcloud", "pointcloud"),
-            (ObjectDBModule, "image", "color_image"),
-            (ObjectDBModule, "pointcloud", "pointcloud"),
-            (PersonTracker, "image", "color_image"),
+            # Currently only the person tracker's detections feed needs routing
+            # to the renamed 2D output.
```
paul-nechifor:
```suggestion
```

dimos/robot/unitree/g1/blueprints/perceptive/test_unitree_g1_detection.py
```diff
@@ -0,0 +1,48 @@
...
+    ("detected_pointcloud_1", PointCloud2): "/detector3d/pointcloud/1",
+    ("detected_pointcloud_2", PointCloud2): "/detector3d/pointcloud/2",
+    ("detected_image_0", Image): "/detector3d/image/0",
+    ("detected_image_1", Image): "/detector3d/image/1",
+    ("detected_image_2", Image): "/detector3d/image/2",
+    ("target", PoseStamped): "/person_tracker/target",
+}
+
+
+def test_detection_transports_resolve_at_runtime() -> None:
```
paul-nechifor: Please delete. Just tests config.

dimos/perception/detection/module3D.py
```diff
@@ -12,53 +12,160 @@
...
-
-    # just for visualization, emits latest top 3 detections in a frame
     detected_image_0: Out[Image]
     detected_image_1: Out[Image]
     detected_image_2: Out[Image]
 
-    detection_3d_stream: Observable[ImageDetections3DPC] | None = None
+    def __init__(self, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self.detector = self.config.detector()  # type: ignore[call-arg, misc]
```
paul-nechifor, line 127: What is `# type: ignore[call-arg, misc]` for?

dimos/perception/detection/module3D.py
```diff
@@ -12,53 +12,160 @@
...
+            image.transform(
+                QualityWindow(lambda img: img.sharpness, window=1.0 / self.config.max_freq)
+            )
+            .transform(Detect2D(self.detector, self.config.filter))
+            .align(self.streams.pointcloud, tolerance=0.25)
+            .map_data(self._fuse)
+        )
+
+    def _fuse(self, obs: Observation[Any]) -> FusedDetections:
+        # obs.data is the AlignedPair; its fields are named after the backing
```
paul-nechifor: If obs.data is AlignedPair, why use Any in Observation[Any]

dimos/memory2/fanio.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def __getitem__(self, key: str) -> Any:
+        return self.values[key]
+
+    def get(self, key: str, default: Any = None) -> Any:
+        return self.values.get(key, default)
+
+    def with_(self, **updates: Any) -> Bundle:
+        """Return a new ``Bundle`` with *updates* layered over the current values."""
+        return Bundle({**self.values, **updates})
```
paul-nechifor: Use the standard `from dataclasses import replace` instead of this custom method.

paul-nechifor: This is never actually used.

dimos/memory2/fanio.py
```diff
@@ -0,0 +1,154 @@
...
+
+    from dimos.core.stream import Out
+    from dimos.memory2.stream import Stream
+    from dimos.memory2.type.observation import Observation
+
+logger = setup_logger()
+
+
+@dataclass(frozen=True)
+class Bundle:
```
paul-nechifor: Isn't this basically just `frozendict`? They're adding it to Python builtins soon, but you have to use the package until then.

bogwi: No, this is not equivalent to frozendict. 

They have different scope,
dataclass(frozen=True) only prevents reassignment of direct attributes but It does not make mutable fields inside the dataclass immutable.

Also, dataclass(frozen=True) is bundled yet from Python 3.7.

The frozendict is a new feature and will come whenever the Python community decide.

dimos/perception/detection/test_module3D.py
```diff
@@ -0,0 +1,253 @@
...
+    the second tick would publish a 3D box fused against the wrong scene."""
+    detector = _OneBoxDetector()
+    tf = _StaticTf()
+    module = _make_module(detector, tf)
+    try:
+        # Fan-I/O contract of the migrated module: 2 In, 2+ Out, primary first,
+        # and the wiring comes from the base class - not a copied start().
+        assert list(module.inputs) == ["color_image", "pointcloud"]
+        assert {"detections_2d", "detections_3d"} <= set(module.outputs)
+        assert Detection3DModule.start is StreamModule.start
```
paul-nechifor: I don't see why this is important to assert.

dimos/perception/detection/module3D.py
bogwi: Will be resolved from https://github.com/dimensionalOS/dimos/pull/2371, [92e3301](https://github.com/dimensionalOS/dimos/pull/2371/commits/92e3301490f669796bdb321c1e355eaee76b2f71)

dimos/memory2/fanio.py
paul-nechifor: > dataclass(frozen=True) only prevents reassignment of direct attributes but It does not make mutable fields inside the dataclass immutable.

Isn't that what `frozendict` does?

```
>>> from frozendict import frozendict
>>> fd = frozendict({"items": []})
>>> fd["items"].append(1) 
>>> fd["items"]
[1]
```

> The frozendict is a new feature and will come whenever the Python community decide.

Yes, they've decided on 3.15: https://docs.python.org/3.15/library/stdtypes.html#frozendict

We have to support 3.10, so we can't use the builtin version yet, but we can use the package.

dimos/perception/detection/module3D.py
paul-nechifor: Ah. Can you merge that PR here please? (Unless there are things you disagree with in it.)

dimos/memory2/fanio.py
bogwi: Paul: Isn't that what frozendict does?
No, it is not. A frozen dataclass is closer to a fixed-shape record, it gives an immutable object but not a frozen dictionary.

Also dataclass(frozen=True) expression is used already across Dimos repo at least 65 times at the moment, and is bundled yet from Python 3.7, so we can use this at our will.

paul-nechifor: What I'm saying is that: I don't understand the purpose of this class.

* It only holds one field: `values`, which is a `MappingProxyType` meaning its data cannot be changed, i.e a frozen dictionary.

* It implements `__getitem__`/`get` which are just helper functions for what a dict does.

* It implements `with_` which is never used. `with_` never mutates the data, but instead creates a new Bundle... which is how you change `frozendict`.

* It's used like this:
   ```python
   Bundle({"low": obs.data, "high": obs.data * 10})
   Bundle({"present": "x", "absent": None})
   Bundle({"doubled": obs.data * 2, "unrouted": obs.data})
   ```
   all of these could just as easily be:
   ```python
   frozendict({"low": obs.data, "high": obs.data * 10})
   frozendict({"present": "x", "absent": None})
   frozendict({"doubled": obs.data * 2, "unrouted": obs.data})
   ```

This class does nothing important. It can be replaced with `frozendict`. The less code the better.

bogwi: Now, that is clear.
---
Agree on keeping the name Bundle as a type alias (Bundle = frozendict) ?

and a small comment:

```python
Bundle = frozendict  # immutable port-name -> payload map per tick
```

passing bare frozendict looks very off to me


or, even better this:

```python
class Bundle(frozendict):
    """Immutable port-name -> payload map per tick."""
 ```
 
 You can hover and know what that is.

File: dimos/perception/detection/type/detection2d/bbox.py [hunk 14]
```diff
@@ -321,16 +326,21 @@ def from_ros_detection2d(cls, ros_det: ROSDetection2D, **kwargs) -> Self:  # typ
...
+            ObjectHypothesisWithPose(
+                ObjectHypothesis(
+                    class_id=str(self.class_id),
+                    score=self.confidence,
+                )
+            )
+        ]
+        # results_length governs how many hypotheses LCM encodes; leaving it 0
+        # silently drops the hypothesis on the wire (and from results-length-
+        # bounded readers like Detection*Array._label_for_detection).
```
paul-nechifor: I assume this comment is added by the LLM after it detected and fixed the problem.

There's no need to leave the comment behind.

However you should fix the bug in other places. For example, `dimos/perception/detection/type/detection2d/point.py` also does the same, it populates `results`, but not `results_length`. I assume there are others.

dimos/perception/fiducial/marker_transformer.py
```diff
@@ -302,107 +303,120 @@ def __call__(
...
+    ) -> Iterator[Observation[Bundle]]:
+        for obs, detections in _group_markers_per_frame(upstream):
+            image = _frame_image(obs, detections)
+            d3d = ImageDetections3D(image, detections).to_ros_detection3d_array(
+                frame_id=self.frame_id
             )
-        return obs.pose
+            d2d = ImageDetections2D(image, detections).to_ros_detection2d_array()
+            yield obs.derive(
+                data=Bundle({"detections_3d": d3d, "detections_2d": d2d}),
```
paul-nechifor, line 421: If `d3d` or `d2d` are empty, the the stream publishes empty detections. Wouldn't it be better to only add to the dict if at least one detection is present?

Also maybe don't yield at all if both lists are empty.

dimos/perception/detection/module3D.py
bogwi: This PR inherits AlignedPair from #2306, 

dimos/dimos/memory2/stream.py, 73-80
```python
@cache
def _pair_class(field_a: str, field_b: str) -> type:
    """Cached ``AlignedPair`` namedtuple with fields named after the two streams.

    ``rename=True`` so any non-identifier or duplicate name degrades to ``_0`` /
    ``_1`` rather than raising — index access (``pair[0]``) always works.
    """
    return namedtuple("AlignedPair", [field_a, field_b], rename=True)  # type: ignore[misc]
```


If clarity is wanted then option A:
```python
def _fuse(
    self, obs: Observation[_pair_class("color_image", "pointcloud")]
) -> FusedDetections:
```

or option B: update/remove the comment to have no AlignedPair mentioned
You call, @paul-nechifor

paul-nechifor: Yeah, if option A works, use that.

dimos/perception/detection/module3D.py
```diff
@@ -158,57 +265,30 @@ def nav_vlm(self, question: str) -> str:
         print("No 3d detections, projecting 2d")
 
         center = detections[0].get_bbox_center()
+        from dimos.msgs.geometry_msgs.PoseStamped import PoseStamped
```
paul-nechifor: Why was this moved here?

dimos/perception/detection/module3D.py
```diff
@@ -158,57 +265,30 @@ def nav_vlm(self, question: str) -> str:
...
-    def _publish_detections(self, detections: ImageDetections3DPC) -> None:
-        if not detections:
-            return
-
-        for index, detection in enumerate(detections[:3]):
-            pointcloud_topic = getattr(self, "detected_pointcloud_" + str(index))
-            pointcloud_topic.publish(detection.pointcloud)
-
 
 def deploy(  # type: ignore[no-untyped-def]
```
paul-nechifor, line 278: @leshy Are these functions still used?

dimos/perception/detection/module3D.py
bogwi: This is caught by your auto-bot already. I have picked the commit, [c1d28df](https://github.com/dimensionalOS/dimos/pull/2371/commits/c1d28df7d253a632e1a6fc2c7f91610c9d21d266)

dimos/perception/fiducial/marker_module.py
```diff
@@ -99,24 +121,23 @@ def pipeline(self, stream: Stream[Image]) -> Stream[Detection3DArray]:
...
+                "MarkerModule: distortion_model=%r may be unsupported; using D as-is.",
                 camera_info.distortion_model,
             )
             self._warned_distortion_model = True
 
     def _append_image_with_pose(self, stream: Stream[Image], image: Image) -> None:
         info = self.config.camera_info
         if info is None:
-            logger.debug("MarkerDetectionStreamModule: no CameraInfo yet; skipping frame")
+            logger.debug("MarkerModule: no CameraInfo yet; skipping frame")
```
paul-nechifor: I don't think anyone ever sees these debug messages. Might was well remove them.

dimos/perception/fiducial/marker_transformer.py
bogwi: A) Wouldn't it be better to only add to the dict if at least one detection is present?
B) Also maybe don't yield at all if both lists are empty.

Here A equals B, because in MarkersToBundle.__call__, both arrays are built from the same detections list and they both resolve to detections == []

Empty publishes are OK here. MarkersToBundle is documented and tested to emit empty but present arrays. Tests in test_marker_module.py assert that and the 2D overlay path relies on empty arrays to clear stale boxes.

Basically the real question here is whether empty frames should publish at all and the current answer in this module is yes.

dimos/perception/fiducial/marker_transformer.py
```diff
@@ -302,107 +303,120 @@ def __call__(
...
+    image = obs.tags.get("marker_frame_image")
+    if isinstance(image, Image):
+        return image
+    raise ValueError("marker frame grouping requires marker_frame_image for empty frames")
+
+
+def _frame_pose(
+    obs: Observation[Detection3DMarker | None],
+    detections: list[Detection3DMarker],
+) -> Any | None:
```
paul-nechifor: Please use the actual type.

dimos/memory2/test_module.py
```diff
@@ -129,3 +143,589 @@ def test_e2e_runtime_wiring(module_cls: type[StreamModule]) -> None:
...
+
+    def pipeline(self, stream: Stream[int]) -> Stream[Bundle]:
+        def _to_bundle(upstream: Iterator[Observation[int]]) -> Iterator[Observation[Bundle]]:
+            for obs in upstream:
+                yield obs.derive(data=Bundle({"doubled": obs.data * 2, "unrouted": obs.data}))
+
+        return stream.transform(_to_bundle)
+
+
+@pytest.mark.tool
```
paul-nechifor: Why is this a tool test? It seems like a max 5 second test with no human input.

dimos/memory2/test_module.py
```diff
@@ -129,3 +143,589 @@ def test_e2e_runtime_wiring(module_cls: type[StreamModule]) -> None:
...
+
+def _wire_inputs(
+    module: StreamModule, store: MemoryStore, dtype: type = int, **points: list
+) -> dict[str, Stream]:
+    """Populate ``module._in_streams`` from ``name=[(ts, value), ...]`` point lists.
+
+    Mirrors what ``StreamModule.start()`` assembles, so ``pipeline()`` can reach
+    siblings through ``self.streams.<port>`` without standing up the live
+    transport machinery. Each secondary's last ts must reach the primary's last
+    ts so the two-pointer merge never blocks waiting on a (never-arriving) later
```
paul-nechifor: > without standing up the live transport machinery

Isn't it better to do that? You're essentially testing test code here. Why not test the actual code. If that code breaks we won't know because we're not testing the actual code.

dimos/memory2/test_module.py
```diff
@@ -129,3 +143,589 @@ def test_e2e_runtime_wiring(module_cls: type[StreamModule]) -> None:
...
+            return True
+        time.sleep(0.01)
+    return predicate()
+
+
+class TestFanOutScatter:
+    """A multi-output pipeline yields a :class:`Bundle` per tick and
+    :func:`scatter_to_ports` fans it to the matching ports in one subscribe."""
+
+    @pytest.mark.tool
```
paul-nechifor: Why are all of these tool tests. Nobody runs them.

dimos/memory2/test_stream.py
```diff
@@ -1084,3 +1084,193 @@ def test_relative_bounds_on_empty_yield_empty(self, make_stream):
...
+        a.append(value, ts=ts)
+    b = session.stream(secondary, dtype)
+    for ts, value in secondary_pts:
+        b.append(value, ts=ts)
+    return a, b
+
+
+class TestAlignNearestNeighbor:
+    """`.align(other, tolerance=)` pairs each primary observation with the
+    nearest-in-time secondary within tolerance (PR #2306 nearest-neighbor)."""
```
paul-nechifor: Why mention the PR in the description?

dimos/memory2/test_module.py
bogwi: .tool labels will be removed.

dimos/memory2/test_stream.py
bogwi: Because the PR was integrated much *before* it was merged in main. So upon any final changes I would know where I had it. Will remove.

dimos/perception/fiducial/marker_transformer.py
bogwi: Will use it

dimos/memory2/test_module.py
bogwi: Agree with your framing, and yes - _wire_inputs is test-only, 

yes, it sets module._in_streams directly and the tests call module.pipeline(streams["lidar"]).to_list() instead of pressing module.start().

But that gap is smaller than it sounds actually. This is very sound test for fusion logic.
No need to spin a new e2e test. The docstring is misleading here, overclaims. I will mend it.

dimos/perception/detection/module3D.py
bogwi: I have already did by cherry-picking to my local, I picked those three only: https://github.com/dimensionalOS/dimos/commit/c1d28df7d253a632e1a6fc2c7f91610c9d21d266 https://github.com/dimensionalOS/dimos/commit/da70a663475ff583e11c7740ba13633e1b2bcfbb https://github.com/dimensionalOS/dimos/commit/92e3301490f669796bdb321c1e355eaee76b2f71

@paul-nechifor If you insist, I can merge here in full, then fetching to my local I will discard what I don't need.

dimos/perception/detection/module3D.py
bogwi: @paul-nechifor  A works yet it trips mypy entirely. 

Will do this, want to keep _pair_class; dynamic namedtuples are invisible to static typing.
```python 
obs: Observation[_pair_class("color_image", "pointcloud")],  # type: ignore[valid-type]
```

paul-nechifor (dismissed):

leshy:
Discussed, will close

PR #2366, opened 2026-06-05
sudo clean

Dreamsorcerer:
This should keep the large tests running correctly, though there appears to still be some flakiness in the dimsim tests themselves.


.github/workflows/ci.yml
```diff
@@ -355,7 +355,7 @@ jobs:
       - name: Fix permissions
         run: |
           git config --global --add safe.directory '*'
```
```suggestion
          sudo git config --global --add safe.directory '*'
```

paul-nechifor (approved):

PR #2368, opened 2026-06-05
feat: ray tracer and 3d planner mem2 transforms

aclauer:
## Problem

mem2 transforms allow for faster algo dev

Closes DIM-XXX

## Solution

Wire the ray tracer and 3d planner together using the Rust bindings in to a mem2 pipeline.

## How to Test

build FFI
```
uv run maturin develop --uv --release -m dimos/mapping/ray_tracing/rust/Cargo.toml
uv run maturin develop --uv --release -m dimos/navigation/nav_3d/mls_planner/rust/Cargo.toml
```

run
```
python -m dimos.navigation.nav_3d.mls_planner.utils.plan_rrd ~/Downloads/mem2-003.db --robot-height 0.3
```


dimos/navigation/nav_3d/mls_planner/utils/plan_rrd.py
```diff
@@ -0,0 +1,184 @@
...
+        rr.spawn(connect=False, memory_limit="16GB", server_memory_limit="16GB")
+        rr.set_sinks(rr.GrpcSink(), rr.FileSink(str(out)))
+    elif out is not None:
+        rr.save(str(out))
+    else:
+        rr.spawn()
+    register_colormap_annotation("turbo")
+
+    store = SqliteStore(path=str(db_path))
+    with store:
```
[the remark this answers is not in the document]
aclauer: fixed

.github/workflows/ci.yml
```diff
@@ -117,17 +117,22 @@ jobs:
...
           source .venv/bin/activate
           maturin develop -m dimos/mapping/ray_tracing/rust/Cargo.toml
           maturin develop -m dimos/navigation/nav_3d/mls_planner/rust/Cargo.toml
           python -c "import dimos_voxel_ray_tracing, dimos_mls_planner"
           python -m pytest -c /dev/null --rootdir . --noconftest --import-mode=importlib -p no:cacheprovider -v \
             dimos/mapping/ray_tracing/test_voxel_map.py \
-            dimos/navigation/nav_3d/mls_planner/test_mls_planner.py
+            dimos/mapping/ray_tracing/test_transformer.py \
+            dimos/navigation/nav_3d/mls_planner/test_mls_planner.py \
+            dimos/navigation/nav_3d/mls_planner/test_transformer.py
```
Dreamsorcerer, line 135: Are these tests to verify the rust binary when used as part of the Python library works correctly?

I'm starting to think that we should just build the binaries in one job and install them in all the test jobs, so they are used for all tests...

Dreamsorcerer: i.e. I assume we are currently running these tests twice, and most of the time in this job is duplicating what is done in the other jobs.

Most optimal might be:
- Build binaries in one job (<30 seconds), upload them.
- All test jobs depend on this job and download the artifacts.
- Another new test job runs the rust tests/lint (upto 1 min).

aclauer: Yeah that makes sense. But at least right now this is the only job that uses the Rust binaries, so not sure if we need to split to a different job yet?

Dreamsorcerer: Maybe I'm misunderstanding, are these tests getting skipped somehow in the regular test jobs? At a glance, I'm assuming this is running the same tests that are already being run in the other test jobs.

aclauer: Oh yeah, the test files have `importorskip()` at the top. The Rust job is the only one that builds the binaries, so the other jobs all just skip these tests.

If the build fails in the Rust job, it fails the test, so it won't silently skip anything.

Idk if there's any value in testing the Rust stuff in all the other jobs? (maybe would be good to add in a macos or arm job?)

aclauer: https://github.com/dimensionalOS/dimos/pull/2368/changes#diff-bc031c2b3b0bc4aade41012fda27f8947ea1f251ef65456a84c332c31a86e39aR21

This is where that happens ^

dimos/mapping/ray_tracing/test_transformer.py
```diff
@@ -0,0 +1,65 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+import numpy as np
+from numpy.typing import NDArray
+import pytest
+
+pytest.importorskip("dimos_voxel_ray_tracing")
```
paul-nechifor, line 21: Noooooo, don't skip the tests 😭

.github/workflows/ci.yml
aclauer: But do you think we should run these tests in the other jobs as well? The FFI should work the exact same on all python3 versions since it's abi3. But you'd know more though so I'll defer to you

Dreamsorcerer: > Idk if there's any value in testing the Rust stuff in all the other jobs? (maybe would be good to add in a macos or arm job?)

Yeah, I think it'd still be good to merge into the regular test runs and keep the tests together, it'll reduce maintenance with the job. e.g. this job is currently missing codecov uploads, and the matrix of different platforms/version to test against. It's also now needing to reinstall dependencies that are already setup in the test jobs.

Dreamsorcerer: Not going to block this PR, but we should look at refactoring this as a separate task.

dimos/mapping/ray_tracing/test_transformer.py
Dreamsorcerer: Think this'll be necessary if it's an optional dependency (also see the other thread about merging into the regular test job).

Dreamsorcerer (approved):
Approving the CI change per the discussion.

.github/workflows/ci.yml
aclauer: Ah I see, that's a good point. I can make an issue for that

dimos/mapping/ray_tracing/test_transformer.py
paul-nechifor: But if the test is never executed then why have it at all?

aclauer: The tests do get ran in the `rust` job. First it builds the binaries, then does the pytest, so the `importorskip` imports properly.

https://github.com/dimensionalOS/dimos/pull/2368/changes#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR127

aclauer: They just get skipped in all the other jobs that don't build the binaries

paul-nechifor: But they do get skipped locally by default, right? That is, if I run this locally, do they get executed:

```
git clone ...
CYCLONEDDS_HOME=/opt/cyclonedds uv sync --all-extras --all-groups
uv run pytest -m 'not (tool or mujoco)'
```

Dreamsorcerer: Because you need to build the dependency, yeah.

aclauer: Yeah they'll get skipped by default. But part of the dev process is building the binaries, so I think that's alright at least for now? Once we do what Sam mentioned, then I think we should be able to remove this

paul-nechifor (approved):

PR #2371, opened 2026-06-05
Auto-fixes for danvi/feat/fan-io

paul-nechifor:
These are automated fixes. Each fix is a separate commit. Use `git rebase -i` to drop any you disagree with.

### DIFF
--- a/dimos/memory2/fanio.py
+++ b/dimos/memory2/fanio.py
@@ -137,7 +137,7 @@ def scatter_to_ports(stream: Stream[Any], ports: dict[str, Out[Any]]) -> Disposa
     """
 
     def _on_error(e: Exception) -> None:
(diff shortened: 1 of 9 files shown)


dimos/memory2/fanio.py
```diff
@@ -137,7 +137,7 @@ def scatter_to_ports(stream: Stream[Any], ports: dict[str, Out[Any]]) -> Disposa
     """
 
     def _on_error(e: Exception) -> None:
-        logger.error("scatter_to_ports() pipeline error: %s", e, exc_info=True)
```
bogwi, line 140: Actually the real fix will be: logger.error("scatter_to_ports() pipeline error: %s", e, exc_info=e)

bogwi:
Picked those: c1d28df da70a66 92e3301

PR #2376, opened 2026-06-06
fix(go2): route phone teleop command and video streams

kezaer:
## Summary
- Route browser and Rerun `tele_cmd_vel` commands to the Go2 `cmd_vel` input in the direct `teleop-phone-go2` stack.
- Expose Go2 `color_image` over JPEG-LCM so the Rerun bridge can render camera frames.
- Add focused blueprint transport tests for command and video routing.

## Why
The direct Go2 phone teleop stack should be independently usable: the browser control path needs to reach the Go2 command stream, and the camera stream needs to be visible to Rerun without depending on a larger navigation stack.

## Safety
- No robot motion commands were sent while validating this change.
- This PR changes stream wiring only; it does not auto-enable any browser control UI or publish nonzero velocity commands.

## Validation
- `uv run pytest dimos/teleop/phone/test_blueprints.py -q`
- `uv run ruff check dimos/teleop/phone/blueprints.py dimos/teleop/phone/test_blueprints.py`

### DIFF
--- a/dimos/teleop/phone/blueprints.py
+++ b/dimos/teleop/phone/blueprints.py
@@ -14,6 +14,9 @@
 # limitations under the License.
 
 from dimos.core.coordination.blueprints import autoconnect
+from dimos.core.transport import JpegLcmTransport, LCMTransport
+from dimos.msgs.geometry_msgs.Twist import Twist
+from dimos.msgs.sensor_msgs.Image import Image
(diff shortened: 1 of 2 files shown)


ruthwikdasyam:
Thanks for this PR. I see the addition of movement-manager creates conflict, and your fix resolves it.
But, we wouldn't want rerun and def not use keyboard controls from rerun with this blueprint. The better fix would be to just replace unitreeGo2basic with Go2Connection

PR #2380, opened 2026-06-06
fix(ci): fix git-lfs-guard compat and add LFS pull retry

spomichter:
## Problem

`test_make_path_mask` and other LFS-dependent tests are failing intermittently on GitHub-hosted runners since PR #2287 removed `self_hosted` markers and PR #2352 switched LFS to `lfs.dimensionalos.com`.

**Root cause (two bugs):**
1. `git-lfs-guard` uses `--json` flag for `ls-files` which requires git-lfs 3.6+. On older versions this fails silently (guard warns and allows through, but the size check is bypassed)
2. `git lfs pull` against the self-hosted LFS server occasionally returns exit code 2 (transient server error), with no retry logic

## Solution

1. **Fix `git-lfs-guard`**: Replace `--json` + `jq` parsing with text `--size` output parsing that works on all git-lfs versions. Parses the standard `ls-files --size` format (`* = downloaded`, `- = pointer`, size in parentheses with unit suffix)
2. **Add retry to `_lfs_pull()`**: 3 attempts with linear backoff (1s, 2s) for transient LFS server errors

## How to Test

```bash
pytest dimos/mapping/occupancy/test_path_mask.py -v
```

CI should pass without the flaky LFS failures.

### DIFF
--- a/dimos/utils/data.py
+++ b/dimos/utils/data.py
@@ -22,6 +22,7 @@ import subprocess
 import sys
 import tarfile
 import tempfile
+import time
 
 from dimos.constants import DIMOS_PROJECT_ROOT
 from dimos.utils.logging_config import setup_logger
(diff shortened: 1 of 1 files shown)


bin/git-lfs-guard
Dreamsorcerer, line 1: Do we want this change (seems notably more complex)? Description says it's an issue with git-lfs <3.6, but I tested it in the runners and that's the only place we'll use it.

dimos/utils/data.py
```diff
@@ -191,23 +191,32 @@ def _is_lfs_pointer_file(file_path: Path) -> bool:
...
+                env=env,
+            )
+            return
+        except subprocess.CalledProcessError as e:
+            last_err = e
+            if attempt <= retries:
+                import time as _time
 
-    return None
+                _time.sleep(attempt)  # 1s, 2s backoff
```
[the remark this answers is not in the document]
Dreamsorcerer: The retry code looks perfectly reasonable. Only change I'd make is moving the import as recommended here.

bin/git-lfs-guard
spomichter: yeah like i mentioned this is an AI slop PR probably can remove and just have the retry logic

spomichter: Reverted this change

paul-nechifor (approved):

PR #2381, opened 2026-06-06
fix: now publishes z=0 to odom in DroneConnectionModule for 2D navigation stack

knikou01:
The A* planner checks arrival using 3D distance. DroneConnectionModule was publishing real altitude in odom, so the drone was never arriving at the destination, that always had z=0. It failed to find a path there because 3D pathfinding isn't implemented. With this change it publishes z=0 to odom and keeps real z only for visualization, so it can at least run a 2D pathfinding, and the developers can handle the drone's z axis movements independently.


knikou01:
The three test failures (test_mcp_client::test_image, test_OccupancyGrid::test_from_pointcloud, test_data::test_lfs_path_with_real_file) aren't related to this change. This PR only modifies _publish_tf in dimos/robot/drone/connection_module.py.

paul-nechifor:
I don't think we want to use the 2d nav stack with the drone...

knikou01:
I understand. We can talk about it on Discord.

PR #2382, opened 2026-06-06
feat: new openarm adapter

TomCC7:
## Solution

Adds an opt-in `openarm_rs` manipulator adapter for OpenArm hardware using the Rust-backed `can-motor-control` Python binding. (https://github.com/TomCC7/can-motor-control)

The existing `openarm` adapter remains the default production path. `openarm_rs` is selected explicitly through hardware config or the new OpenArm RS blueprints, and is intended for binding-backed bring-up, state monitoring, MIT command validation, gravity compensation, and trajectory-control validation.

This also updates manipulator adapter discovery to use lightweight `__registry__.py` manifests, so listing available adapters does not import unselected hardware SDKs or optional bindings.

## User-facing behavior

- `adapter_type="openarm"` continues to use the existing in-tree SocketCAN OpenArm adapter.
- `adapter_type="openarm_rs"` selects the Rust-backed OpenArm adapter.
- Missing `can_motor_control` now fails only when `openarm_rs` is selected/connected, with a clear install hint.
- Adapter listing remains healthy in partial installs.
- The OpenArm RS path defaults to CAN-FD and supports staged validation before active trajectory control.

## How to Test

```bash
uv run pytest \
  dimos/hardware/manipulators/test_registry.py \
  dimos/hardware/manipulators/openarm_rs/test_adapter.py \
  dimos/hardware/manipulators/damiao/test_base_adapter.py -q

uv run mypy \
  dimos/hardware/manipulators/damiao/base_adapter.py \
  dimos/hardware/manipulators/openarm_rs/test_adapter.py
For hardware bring-up:
dimos run coordinator-openarm-rs

### CODE REVIEW

File: openspec/changes/archive/2026-06-06-add-dm-motor-arm-adapter/tasks.md [hunk 1]
```diff
@@ -0,0 +1,50 @@
+## 1. Adapter Implementation
```
paul-nechifor: I assume this shouldn't be committed.

dimos/hardware/manipulators/openarm_rs/test_adapter.py
```diff
@@ -0,0 +1,469 @@
...
+        damiao_base_adapter,
+        "_can_motor_control_import_error",
+        ImportError("can_motor_control"),
+    )
+    adapter = OpenArmRSAdapter(use_mock_bus=True)
+    with pytest.raises(OpenArmRSBindingUnavailableError, match="openarm_rs.*can-motor-control"):
+        adapter.connect()
+
+
+def test_lifecycle_read_write_disable() -> None:
```
paul-nechifor: There are many low value tests in this file. Tests should check functionality.

A good test is structured like this:

1. setup the test
2. execute the functionality
3. check that the desired result was achieved

But a test like this one just constructs one object and asserts every minor aspect. It's not clear what behavior is even desired given that so much is asserted.

Tests that over-assert make the system hard to change and introduce uncertainty.

docs/development/openspec.md
```diff
@@ -0,0 +1,102 @@
+# OpenSpec Workflow
+
+DimOS uses OpenSpec as the checked-in planning layer for behavior changes. OpenSpec artifacts live under `openspec/` and should describe what the system is supposed to do, why it is changing, and how contributors or agents should validate the work.
```
paul-nechifor: If we want to use OpenSpec, that should be a separate PR, it shouldn't be coupled to this openarm stuff.

paul-nechifor:
@TomCC7  My auto fixer created this PR: https://github.com/dimensionalOS/dimos/pull/2426/changes

Do you agree with the changes? If so, please merge them into this PR.

docs/development/openspec.md
TomCC7: Sorry for the confusion added here. I based all my branches on an openspec initialization commit. I cleaned that out and created this pr: https://github.com/dimensionalOS/dimos/pull/2428. will remove openspec stuff in this branch

TomCC7:
close in lieu of #3388

Issue #2372, opened 2026-06-06
Hosted Teleop Overview


ruthwikdasyam:
# Hosted Teleop Overview

Cloud-relayed VR/keyboard teleop. Operator's browser → broker (CF SFU) → robot, all over WebRTC. No port forwarding, no LAN, no public IP — the robot is just an outbound HTTPS/UDP client.

> **Two ways to wire it today:**
>
> 1. **Transport blueprints** (preferred) — the broker DataChannels/video are DimOS *transports*; the robot driver and the hosted control plane live in **one module on one process → one CF session**. `teleop-hosted-go2-transport`.
> 2. **Legacy module wrapper** — `HostedTeleopModule` owns the whole WebRTC + control plane itself. Still shipped (`teleop-hosted-go2`, `teleop-hosted-xarm7`) but the transport swap is preferred for Go2.

## 1. Big picture

```
┌─────────────┐    HTTPS    ┌───────────────────┐   HTTPS+REST   ┌──────────────┐
│ Operator    │ ──────────► │ dimensional-teleop│ ─────────────► │ Cloudflare   │
│ (browser)   │   X-Bearer  │   broker (EC2)    │   CF app key   │ Realtime SFU │
└──────┬──────┘             │   FastAPI         │                └──────┬───────┘
       │                    └───────────────────┘                       │
       │  HTTPS (X-Robot-API-Key) ────┘                                  │
       │                                                                 │
       │           WebRTC (data + video) ◄──────────────────────────────►│
       ▼                                                                 ▼
   Both peers connect to CF; CF bridges them.            ┌──────────────────────────┐
   Broker only handles SETUP — once channels are         │ Go2HostedConnection      │
   bridged, broker is off the data path.                 │ (driver + control plane) │
                                                         └──────────────────────────┘
```

**Two roles for the broker:**

* Auth (operator JWT login + robot API key validation).
* Session setup (creates CF sessions, bridges DataChannels, pulls video).

**Everything operational** (PoseStamped, Joy, Twist, video, telemetry, robot commands) flows direct: operator's browser ↔ Cloudflare edge ↔ robot. Broker is **not** a WebRTC peer.

## 2. One CF session (why the consolidation matters)

The CF broker provider is a **per-process singleton** keyed by config, and the Go2 driver runs `dedicated_worker=True` (its own process). So **all** hosted broker transports (cmd, video, state, state_back) must live on **that same module** to share one CF session.

[...]

PR #2411, opened 2026-06-06
feat: hosted teleop over Cloudflare RT

ruthwikdasyam:
Detailed in #2372 

## Problem

Hosted Teleop setup for DimOS

Closes DIM-957
Closes DIM-926
Closes DIM-846

## Solution

Add hosted teleop over WebRTC via a Cloudflare Realtime SFU broker: a `HostedTeleopModule` with cmd/state datachannels + a camera video track that drops into existing blueprints, plus a shared `teleop/utils` recording &
transport-stats stack (recorder, latency/jitter/loss report, `VideoStats`).

## How to Test

``` python
dimos run teleop-hosted-go2 teleop-recorder
```


dimos/msgs/geometry_msgs/TwistStamped.py
```diff
@@ -39,18 +39,21 @@ class TwistStamped(Twist, Timestamped):
     msg_name = "geometry_msgs.TwistStamped"
     ts: float
     frame_id: str
+    seq: int
```
leshy: this isn't in dimos-lcm, this is some old entry from ros1, we use ros2 msgs
seq is a "monotonically increasing per-publisher counter"

seq is sequence counter for message dropping which is a transport layer concern, not application layer concern, this is why it was removed

so this won't work in dimos (try lcm roundtrip - will confirm seq is lost) but also - why did you introduce this?

dimos/memory2/module.py
```diff
@@ -320,21 +325,25 @@ def _port_to_stream(self, name: str, input_topic: In[Any], stream: Stream[Any])
...
-                    "[%s] No tf available for frame '%s' at time %s (msg ts: %s), storing without pose",
+                    "[%s] No tf for frame '%s' — recording without pose "
+                    "(further such messages on this stream are not logged)",
                     name,
                     frame_id,
-                    ts,
-                    getattr(msg, "ts", None),
                 )
-            stream.append(msg, ts=ts, pose=pose)
+            stream.append(msg, ts=ts, pose=pose, tags={"recv_ts": recv_ts})
```
leshy: tags are unindexed json fields, IDK if you want to search based on recv ts? 

we can allow you to add random tags like this, including this recv_ts, but you shouldn't modify ALL memory streams in dimos in order to achieve this, idk if we need some API change on mem2 recorder level

also recv ts vs msg.ts is a legitimate concern and a thing to store broadly - I didn't deal with this yet but can review and make sure we can support this by default

dimos/memory2/module.py
```diff
@@ -320,21 +325,25 @@ def _port_to_stream(self, name: str, input_topic: In[Any], stream: Stream[Any])
...
         def on_msg(msg: Any) -> None:
-            ts = getattr(msg, "ts", None) or time.time()
+            recv_ts = time.time()
+            ts = getattr(msg, "ts", None) or recv_ts
             frame_id = getattr(msg, "frame_id", None) or default_frame_id
             transform = self.tf.get("world", frame_id, time_point=ts, time_tolerance=tf_tolerance)
             pose = transform.to_pose() if transform is not None else None
 
-            if not pose:
+            if not pose and frame_id not in warned_frames:
```
leshy: this might be wrong design decision on my end, mem2 maybe shouldn't insist on a pose for EVERY stream, this is here for efficient spatial search, for example you can search for all moments in which end effector or base_link was in some global pose.. so actually I'd consider storing something here in your case.. either end effector or base_link

alternatively I'll add a way to supress this warning PER stream, not for all streams, I do want to get warnings (and not one warning) if some TF tree is broken during a run

dimos/teleop/quest_hosted/hosted_extensions.py
```diff
@@ -0,0 +1,122 @@
...
+        # VR thumbsticks → base velocity. Left Y = fwd/back, left X = strafe,
+        # right X = yaw. Stick conventions are opposite ROS, so negate.
+        super()._on_joy_bytes(data)
+        with self._lock:
+            right = self._controllers.get(Hand.RIGHT)
+            left = self._controllers.get(Hand.LEFT)
+        fwd = -left.thumbstick.y if left is not None else 0.0
+        strafe = -left.thumbstick.x if left is not None else 0.0
+        yaw = -right.thumbstick.x if right is not None else 0.0
+        self._publish_twist(fwd, strafe, yaw, time.time(), "vr", 0)
```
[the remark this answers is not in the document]
ruthwikdasyam: will add ts for buttons in next PR

dimos/msgs/geometry_msgs/TwistStamped.py
ruthwikdasyam: On the round-trip detail: on the currently installed dimos-lcm build, std_msgs/Header still has a seq field, so TwistStamped(seq=42) actually does survive encode→decode today. but, once dimos-lcm is the proper ROS2 Header, it breaks. Either way your core point stands.
I introduced it to compute command-link loss/reorder for the teleop transport-benchmarking stats.

Looks like stash's WebRTC PR helps with this. So dropping .seq entirely now.

dimos/memory2/module.py
ruthwikdasyam: yes, it was mutating the shared recorder to tag every memory stream in dimos. The only thing that consumed it was offline latency analysis (recv_ts − msg.ts)

True that recv_ts vs msg.ts is a legitimate thing to store broadly. Either you can add, or I can do a follow-up PR on this. Will revert the changes for this PR.

dimos/memory2/module.py
ruthwikdasyam: suppressing per stream should solve the issue. I can add a frame_id for these messages from my end.
I did undo my changes.

dimos/teleop/quest_hosted/hosted_teleop_module.py
```diff
@@ -0,0 +1,633 @@
...
+    def start(self) -> None:
+        super().start()
+        self._stop_event.clear()
+        unsub = self.color_image.subscribe(self._video_track.set_latest)
+        self.register_disposable(Disposable(unsub))
+        self._connect_blocking()
+        self._start_heartbeat()
+        self._start_telemetry()
+        self._start_control_loop()
+        logger.info("HostedTeleopModule started")
```
paul-nechifor:
```suggestion
```

The module coordinator already logs this stuff.

dimos/teleop/quest_hosted/hosted_teleop_module.py
```diff
@@ -0,0 +1,633 @@
...
+
+    @rpc
+    def stop(self) -> None:
+        self._stop_event.set()
+        if self._control_loop_thread is not None:
+            self._control_loop_thread.join(timeout=1.0)
```
paul-nechifor: Use DEFAULT_THREAD_JOIN_TIMEOUT for all.

dimos/teleop/quest_hosted/hosted_teleop_module.py
```diff
@@ -0,0 +1,633 @@
...
+            self._control_loop_thread = None
+        if self._heartbeat_thread is not None:
+            self._heartbeat_thread.join(timeout=2.0)
+            self._heartbeat_thread = None
+        if self._telemetry_thread is not None:
+            self._telemetry_thread.join(timeout=2.0)
+            self._telemetry_thread = None
+        if self._loop is not None and self._loop.is_running():
```
paul-nechifor: `spawn` already checks this

dimos/teleop/quest_hosted/hosted_teleop_module.py
```diff
@@ -0,0 +1,633 @@
...
+            try:
+                self.spawn(self._disconnect()).result(timeout=5.0)
+            except Exception:
+                logger.exception("Error during disconnect")
+        # The base Module owns the event loop; it's torn down on module close.
```
paul-nechifor:
```suggestion
```

dimos/teleop/quest_hosted/hosted_teleop_module.py
```diff
@@ -0,0 +1,633 @@
...
+                resp.text[:1000],
+            )
+        resp.raise_for_status()
+        data = resp.json()
+        self._session_id = data["session_id"]
+
+        answer_sdp = propagate_bundle_candidates(data["sdp_answer"])
+        await self._pc.setRemoteDescription(RTCSessionDescription(sdp=answer_sdp, type="answer"))
+
+        _ = sctp_init  # intentionally left open
```
paul-nechifor: What is the point of this statement? Delete it and move the comment to where the channel is created.

dimos/teleop/quest_hosted/hosted_teleop_module.py
```diff
@@ -0,0 +1,633 @@
...
+            # escape into the datachannel callback and stall ping/clock-sync.
+            try:
+                stats = VideoStats.from_dict(msg)
+            except (TypeError, ValueError):
+                logger.warning("state_reliable: malformed video_stats, dropping")
+                return
+            logger.info("video: %s", stats)
+            self.video_stats.publish(stats)
+        else:
+            logger.debug(f"state_reliable: unknown message type {kind!r}")
```
paul-nechifor:
```suggestion
            logger.warning(f"state_reliable: unknown message type {kind!r}")
```

dimos/teleop/quest_hosted/README.md
```diff
@@ -0,0 +1,120 @@
...
+  the candidate block into every m-section that lacks one. **Do not remove.**
+
+- **`makeXRCompatible()` on real hardware.** The operator side, not us, but
+  worth knowing: `xrCompatible: true` at context creation is not enough on
+  Quest — `await gl.makeXRCompatible()` is required before building the
+  `XRWebGLLayer`.
+
+## Sidecar files
+
+- **`/tmp/dimos_netem_profile`** — written by `data/notes/benchmarks/netem/apply.sh`
```
paul-nechifor: What is `data/notes/benchmarks/netem/apply.sh`? I see this mentioned in several places.

dimos/teleop/quest_hosted/README.md
```diff
@@ -0,0 +1,120 @@
...
+  for aiortc + httpx; coroutines are scheduled onto it via `self.spawn(...)`.
+  Datachannel `send()` calls **must** happen here (aiortc datachannels aren't
+  thread-safe). The pong path is already on the loop (it fires from the
+  channel's `message` callback); the telemetry thread uses
+  `loop.call_soon_threadsafe` for the same reason.
+- **Heartbeat thread** — HTTP polls the broker; reacts to channel-id changes.
+- **Telemetry thread** — pushes `robot_telemetry` JSON on `state_reliable_back`
+  at `telemetry_hz` (default 3 Hz).
+- **Control loop** — `control_loop_hz` (default 50 Hz). Calls the
+  subclass-overridable engage / publish hooks.
```
paul-nechifor: This sounds like code documentation. It should be in the code, not here.

dimos/teleop/quest_hosted/video_track.py
```diff
@@ -0,0 +1,98 @@
...
+            self._armed = True
+
+    def set_latest(self, img: Image) -> None:
+        with self._lock:
+            self._latest = img
+            self._frame_seq += 1
+
+    async def recv(self) -> av.VideoFrame:
+        while True:
+            with self._lock:
```
paul-nechifor: This is bad async code.

Async is cooperative concurrency. Using `threading.Lock` means that when it blocks, every other thing that's running on the async loop also blocks.

Eliminate the lock. Use `call_soon_threadsafe` to set the value and use an `asyncio.Event` to notify `recv`.

```
      def set_latest(self, img: Image) -> None:                                                                                                                                                                                                                                                                                                                                                                                           
          loop = self._loop                                                                                                                                                                                                                                                                                                                                                                                                               
          if loop is None or not loop.is_running():                                                                                                                                                                                                                                                                                                                                                                                       
              return                                                                                                                                                                                                                                                                                                                                                                                                                      
          def _set() -> None:                                                                 
              self._latest = img                                                                                                             
              self._frame_seq += 1                                                                                                                                                                                                                                                                                                                                                                                       
              self._new_frame.set()                                                                                                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                                                                                          
          loop.call_soon_threadsafe(_set) 

...

          while True:
              await self._new_frame.wait()
              self._new_frame.clear()
```

dimos/teleop/utils/recorder.py
```diff
@@ -0,0 +1,94 @@
...
+    right_controller_output: In[PoseStamped]
+    buttons: In[Buttons]
+    cmd_vel_stamped: In[TwistStamped]
+    video_stats: In[VideoStats]
+    config: TeleopRecorderConfig
+
+    @rpc
+    def start(self) -> None:
+        # Append a per-run timestamp to the stem so each run is its own file.
+        base = getattr(self, "_db_path_base", None)
```
paul-nechifor: Don't use `getattr`.

Define it as none in `__init__` or on the class.

dimos/teleop/utils/recorder.py
```diff
@@ -0,0 +1,94 @@
...
+        if base is None:
+            base = Path(self.config.db_path)
+            self._db_path_base = base
+        timestamp = datetime.now().strftime("%Y%m%d_%H%M%S")
+        self.config.db_path = base.with_name(f"{base.stem}_{timestamp}{base.suffix}")
```
paul-nechifor: greptile is correct. Don't mutate the config. Use `self._db_path` or something.

dimos/teleop/utils/report.py
```diff
@@ -0,0 +1,293 @@
...
+}
+
+
+def generate_report(db_path: Path, out_dir: Path | None = None) -> Path:
+    """Write ``report.md`` for the recording at *db_path*.
+
+    Output lands in *out_dir* if given, else next to the .db. Returns the
+    written report.md path. Raises if the .db is missing or unreadable.
+    """
+    db_path = Path(db_path)
```
paul-nechifor:
```suggestion
```

dimos/teleop/utils/report.py
```diff
@@ -0,0 +1,293 @@
...
+    if not db_path.exists():
+        raise FileNotFoundError(f"Recording not found: {db_path}")
+    if out_dir is None:
+        out_dir = db_path.parent
+    out_dir = Path(out_dir)
```
paul-nechifor:
```suggestion
```

dimos/teleop/quest_hosted/hosted_teleop_module.py
```diff
@@ -0,0 +1,633 @@
...
+        self._state_channel_id: int | None = None
+        self._state_back_channel: RTCDataChannel | None = None
+        self._state_back_channel_id: int | None = None
+
+        self._video_track = CameraVideoTrack()
+        self._cmd_stats = LiveStreamStats()
+
+        self._control_loop_thread: threading.Thread | None = None
+        self._heartbeat_thread: threading.Thread | None = None
+        self._telemetry_thread: threading.Thread | None = None
```
paul-nechifor, line 131: This module has a lot of threads. It could be simplified if you used async... but it would require rewriting multiple things. 😬

dimos/teleop/quest_hosted/hosted_teleop_module.py
```diff
@@ -0,0 +1,633 @@
...
+                        if not self._should_publish(hand):
+                            continue
+                        output_pose = self._get_output_pose(hand)
+                        if output_pose is not None:
+                            self._publish_msg(hand, output_pose)
+                    left = self._controllers.get(Hand.LEFT)
+                    right = self._controllers.get(Hand.RIGHT)
+                    self._publish_button_state(left, right)
+            except Exception:
+                logger.exception("Error in control loop")
```
paul-nechifor, line 579: An error here could potentially spam the logs with the same error over and over.

dimos/teleop/quest_hosted/README.md
ruthwikdasyam: using netem - Network simulator to simulate degraded network conditions and testing hosted teleop metrics change. But, I am not planning on including it in this PR. Thanks for pointing this, will remove references

dimos/teleop/quest_hosted/hosted_teleop_module.py
ruthwikdasyam: yeah, if this fails.. the teleop flow is broken, and need to exit.
Mabe can leave it as is, or Throttled logging might make sense. This pattern is repeated across all teleop module control loops, i'll handle them togther in another PR. added an issue

dimos/teleop/quest_hosted/hosted_teleop_module.py
ruthwikdasyam: Module might need a refactor once WebRTC transport support is in. 
I'll try simplifying this too.

dimos/teleop/quest_hosted/hosted_teleop_module.py
paul-nechifor: Yes, throttled makes sense.

paul-nechifor (approved):

spomichter (approved):

dimos/memory2/module.py
spomichter: resolving. reverted

dimos/memory2/module.py
spomichter: resolving

PR #2443, opened 2026-06-10
feat: proposed architecture document for memory2 hosted

### LINKED ISSUE #2442: Make memory2 hosted on cloud
### Description

Current memory2 system is designed to run locally. List changes needed to be made to host the memory and handle batch uploads. Do we need to change the memory backend? And then what about pulling from a hosted memory system? How do we pull memory if its not local and its hosted?

sasagarw:
## Problem

The current memory system is designed to run locally. That creates several issues for hosted memory:

- SQLite is embedded and local; clients cannot safely use it as a shared network database.
- Current append behavior is observation-oriented. Remote upload needs batch-oriented ingest to avoid excessive network round trips.
- Large payloads such as images and point clouds should not always be pulled eagerly over the network.
- Hosted memory needs auth, tenant/robot scoping, pagination, live tails, retries, and operational visibility.
- Multiple clients and robots need a shared API instead of direct access to local DB files.

Closes #2442 

## Solution

- Preserve the current `Store` and `Stream` ergonomics for users.
- Keep `SqliteStore` for local memory, replay, demos, and offline robot operation.
- Add `RemoteStore` for hosted shared memory.
- Support batch upload as a first-class remote operation.
- Support remote query, vector search, lazy blob reads, and live tailing.
- Start with an EC2 MVP that reuses existing `SqliteStore`.
- Move production storage to scalable managed services without changing client usage.

## How to Test

Phase 1 MVP will be used to test if the hosted memory works and how the API contracts work.


leshy:
> Add RemoteStore for hosted shared memory.
> Support batch upload as a first-class remote operation.
> Support remote query, vector search, lazy blob reads, and live tailing.
> Start with an EC2 MVP that reuses existing SqliteStore.

Robots query their memory during runs, you can't feed stuff directly into a RemoteStore and expect to query it off the cloud. best case - you pay for the roundtrip and you don't have indexing and can't query recent data (seconds) etc. worst case - your network is slower then your upload, you will fill some internal buffer and crash the machine.

Data needs to be stored locally, into a full mem2 database, and used locally through it. with async remote sync.

(I imagine "remote sync" there doesn't just mean our self hosted cluster but other robots, controller machine, compute machine etc)

Sqlite is an initial super fast sketch, I don't think we should continue with it even locally and especially in terms of remote. For a first step instead of generically named "RemoteStore" - add "PostgresStore". this can be initially tested locally for mature data injestion by running posgres as a dimos native module - nix can provide postgres

As mentioned during upload we will have issues with bandwidth, unreliable connections etc, so we need a protocol robust to this, this needs to be investigated by someone familiar with this caregory of problems.

initial questions:

- do we use psql protocol or zenoh for data sync? Postgres has a system for syncing databases in a psql cluster, is some of this usable for injestion from robots as a POC or do we have to own the protocol from the start?
- Zenoh gives us robot->robot sync and network robustness, likely mid-term this is what we use for mem2? what is the content of the protocol? just database change log like clusters use?
- Are we configuring different priorities based on memory2 streams? this includes QOS (robot->robot mem sync maybe needs to be fast but not exact)
- does binaryblob store in case of remote live in postgresql at all? ok for POC? relational db isn't a binary data store. 


probably we need flexibility around all above and is deployment specific.

sasagarw:
@leshy Thanks for taking a look at it.

That makes sense. I agree the architecture should be local-first, not cloud-first.

The active robot run path should be a full local memory2 database that the robot writes to and queries directly. Remote/cloud should be an async sync target, not the query path. I’ll reframe the proposal around `PostgresStore` as the next serious local backend instead of a generic `RemoteStore`.

The way I understand the direction is:

1. Build `PostgresStore` first and run Postgres locally, ideally as a Dimos-native module with Nix providing the binary.
2. Keep the robot querying local memory during the run.
3. Add an append-only memory2 outbox/change log for async sync.
4. Treat cloud, other robots, controller machines, and compute machines as sync peers.
5. Investigate sync transport separately:
   - Postgres logical replication may be useful for a controlled Postgres-to-Postgres POC.
   - Zenoh may be better mid-term for robot-to-robot/controller/flaky network sync.
   - We probably still need memory2-level sync semantics above either transport.
6. Add per-stream sync priority/QoS so high-value or low-latency streams do not compete equally with raw bulk blobs.
7. For blobs, use Postgres only as a POC if needed. Longer term, keep blob payloads outside Postgres and sync blob manifests/chunks separately.

I’ll update the design doc to remove the cloud-first `RemoteStore` framing and make the main proposal: local `PostgresStore` + async memory2 sync protocol investigation.

PR #2446, opened 2026-06-10
feat(learning): teleop + dataset collection + dataprep pipeline

ruthwikdasyam:
## Problem

We had no path from a teleop session to a trainable dataset — robot streams were recorded ad hoc and there was no standard, format-agnostic way to segment episodes and export them for policy learning.

Closes DIM-803
Closes DIM-589

## Solution

Adds an end-to-end data-collection → dataset-prep pipeline:

- **Collection** (`dimos/learning/collection/`): `EpisodeMonitorModule`   turns Quest buttons/keyboard into `start`/`save`/`discard` episode  events; `CollectionRecorder` captures the obs/action/status streams to
  a SQLite session DB. Two ready blueprints: `learning-collect-quest-xarm7` and `learning-collect-quest-piper`.
- **DataPrep** (`dimos/learning/dataprep/`): reads the session DB,   extracts episodes (`episode_status` or explicit `ranges`), time-syncs   obs/action streams onto a common timeline, and writes **LeRobot v2** or
  **HDF5** datasets with streaming per-feature stats + a `dimos_meta.json` sidecar. Pure core / impure build split for testability.
- **CLI**: `dimos dataprep build` and `dimos dataprep inspect`.

## How to Test

```bash
# 1. Collect (drive the arm; A=start, B=save, X=discard)
dimos run learning-collect-quest-xarm7

# 2. Build a LeRobot dataset from the session
dimos dataprep build -c dimos/learning/dataprep/example_config.json

# 3. Inspect it
dimos dataprep inspect data/datasets/session
```


dimos/learning/collection/episode_monitor.py
```diff
@@ -0,0 +1,185 @@
...
+        for event_name, alias_or_attr in self.config.button_map.items():
+            attr = BUTTON_ALIASES.get(alias_or_attr, alias_or_attr)
+            try:
+                pressed = bool(getattr(msg, attr))
+            except AttributeError:
+                continue
+            prev = self._prev_bits.get(attr, False)
+            self._prev_bits[attr] = pressed
+            if pressed and not prev:  # rising edge
+                self._transition(event_name, ts)
```
mustafab0: There is a concept of pre-roll and post-roll in data collection. Where we send a stream of 0s before activation (rising edge) in this case and also after deactivation. 

This helps mark the exact start and stop point of an episode just from the data.

0s are good for Twist commands, for joint position probably need to stream current joint positions

dimos/learning/collection/episode_monitor.py
```diff
@@ -0,0 +1,185 @@
...
+    "MENU_R": "right_menu",
+}
+
+
+class EpisodeStatus(BaseModel):
+    state: Literal["idle", "recording"]
+    episodes_saved: int
+    episodes_discarded: int
+    current_episode_start_ts: float | None
+    last_event: Literal["start", "save", "discard", "init"] = "init"
```
mustafab0: A little confused here. Why is the EpisodeStatus counting episodes_saved/discarded ?

A single session would have multiple episodes inside it

dimos/learning/collection/episode_monitor.py
ruthwikdasyam: on_buttons method is a callback function, to read button presses.

Irrespective of data we are collecting, this marks start and stop checkpoints in the stream of data - which helps us to trim entire session into episodes.

dimos/learning/collection/episode_monitor.py
ruthwikdasyam: Yes. once we start a blueprint, we collect multiple episodes.
when we keep collecting episodes..  This msg will be live indication of how many we collected, and how many discarded.

dimos/learning/collection/recorder.py
```diff
@@ -0,0 +1,51 @@
...
+    db_path: str | Path = "data/recordings/session.db"
+
+
+class CollectionRecorder(Recorder):
+    """Records the streams DataPrep consumes from a teleop session."""
+
+    config: CollectionRecorderConfig
+
+    color_image: In[Image]  # observation (camera)
+    joint_state: In[JointState]  # observation + action (measured/next state)
```
mustafab0:
```suggestion
    coordinator_joint_state: In[JointState]  # observation + action (measured/next state)
```

dimos/learning/collection/episode_monitor.py
```diff
@@ -0,0 +1,215 @@
...
+    }
+    keyboard_map: dict[Literal["start", "save", "discard", "toggle"], str] = {}
+    default_task_label: str | None = None
+
+
+class EpisodeMonitorModule(Module):
+    config: EpisodeMonitorModuleConfig
+
+    buttons: In[Buttons]
+    # TODO: no KeyPress producer exists yet — add a pygame keyboard module that
```
mustafab0, line 78: Please create a ticket for this todo so we can track this

dimos/learning/collection/episode_monitor.py
```diff
@@ -0,0 +1,215 @@
...
+        with self._lock:
+            self._state = "idle"
+            self._saved = 0
+            self._discarded = 0
+            self._prev_bits = {}
+            status = self._snapshot("init", time.time())
+        return self._emit(status)
+
+    @rpc
+    def get_status(self) -> EpisodeStatus:
```
mustafab0: We are already publishing EpisodeStatus topic. Interested listeners should just get the latest EpisodeStatus

dimos/learning/dataprep/core.py
```diff
@@ -0,0 +1,382 @@
...
+            )
+
+    shift = max(0, sync.action_shift)
+    if shift == 0 or not action_keys:
+        yield from _build_frames()
+        return
+
+    # frame i keeps its obs but takes frame i+shift's action; tail dropped.
+    frames = list(_build_frames())
+    for i in range(len(frames) - shift):
```
mustafab0, line 356: for small episodes if the `shift` = `len(frames)` then we produce 0 samples. But this still gets indexed into episodes.

dimos/learning/dataprep/formats/_stats.py
```diff
@@ -0,0 +1,123 @@
...
+        self, image_subsample: int = 10, quantile_reservoir: int = 10_000, seed: int = 0
+    ) -> None:
+        self.image_subsample = image_subsample
+        self.quantile_reservoir = quantile_reservoir
+        self._rng = random.Random(seed)
+        self.aggs: dict[str, FeatureAggregator] = {}
+
+    def update(self, name: str, value: np.ndarray) -> None:
+        a = np.asarray(value)
+        is_image = a.ndim >= 3
```
mustafab0: grayscale images will be mis categorized as not iamges

mustafab0 (dismissed):

dimos/learning/collection/episode_monitor.py
```diff
@@ -0,0 +1,203 @@
...
+
+class KeyPress(BaseModel):
+    """Single keypress event from a keyboard input source."""
+
+    key: str
+    ts: float
+
+
+class EpisodeMonitorModuleConfig(ModuleConfig):
+    button_map: dict[Literal["start", "save", "discard", "toggle"], str] = {
```
paul-nechifor: You are using `Literal["start", "save", "discard", "toggle"]` in so many places. Extract it as its own type.

dimos/learning/collection/test_episode_monitor.py
```diff
@@ -0,0 +1,167 @@
...
+    """The EpisodeStatus objects published on the monitor's `status` port."""
+    return [call.args[0] for call in monitor.status.publish.call_args_list]  # type: ignore[attr-defined]
+
+
+def _press(monitor: EpisodeMonitorModule, alias: str, ts: float) -> None:
+    """Rising edge: release-then-press the given Quest button alias."""
+    attr = BUTTON_ALIASES[alias]
+    released = Buttons()
+    pressed = Buttons()
+    setattr(pressed, attr, True)
```
paul-nechifor: I don't like this way of setting random fields on a class. It's better to use a method like:

```python
pressed.set_attribute(attr, True)
```

dimos/learning/collection/test_episode_monitor.py
```diff
@@ -0,0 +1,167 @@
...
+
+def _events(monitor: EpisodeMonitorModule) -> list[EpisodeStatus]:
+    """The EpisodeStatus objects published on the monitor's `status` port."""
+    return [call.args[0] for call in monitor.status.publish.call_args_list]  # type: ignore[attr-defined]
+
+
+def _press(monitor: EpisodeMonitorModule, alias: str, ts: float) -> None:
+    """Rising edge: release-then-press the given Quest button alias."""
+    attr = BUTTON_ALIASES[alias]
+    released = Buttons()
```
paul-nechifor, line 78: It's frowned upon to have a class be a plural since classes are supposed to be nouns and you can't have a plural of a plural (i.e. a list of Buttons classes would have to be `buttonses: list[Buttons] = [...]`

The easiest solution is to use `ButtonSet`, but I guess something more appropriate here is `ButtonBitfield`.

dimos/learning/collection/blueprint.py
```diff
@@ -0,0 +1,68 @@
...
+from dimos.learning.collection.recorder import CollectionRecorder
+from dimos.teleop.quest.blueprints import (
+    teleop_quest_piper,
+    teleop_quest_xarm7,
+)
+
+
+def _session_db(robot: str) -> str:
+    """Timestamped session DB path under the state dir, namespaced by robot."""
+    return str(STATE_DIR / "recordings" / f"session_{robot}_{datetime.now():%Y%m%d_%H%M%S}.db")
```
TomCC7, line 40: might want to create the directory if not exist

dimos/learning/collection/test_episode_monitor.py
ruthwikdasyam: makes sense, I'll do rename through a seperate PR, Thanks

dimos/learning/dataprep/formats/test_lerobot.py
```diff
@@ -0,0 +1,225 @@
...
+from collections.abc import Iterator
+import json
+from pathlib import Path
+
+import numpy as np
+import pytest
+
+pytest.importorskip("pyarrow")
+pytest.importorskip("pandas")
+cv2 = pytest.importorskip("cv2")
```
paul-nechifor:
```suggestion
```

Always run the tests.

dimos/learning/dataprep/formats/test_hdf5.py
```diff
@@ -0,0 +1,86 @@
...
+
+from __future__ import annotations
+
+from collections.abc import Iterator
+from pathlib import Path
+
+import numpy as np
+import pytest
+
+h5py = pytest.importorskip("h5py")
```
paul-nechifor:
```suggestion
```

dimos/learning/dataprep/formats/lerobot.py
```diff
@@ -0,0 +1,491 @@
...
+                continue
+            out[f"stats/{feat}/{k}"] = _nest_image_stat(v) if is_video else v
+        out[f"stats/{feat}/count"] = int(entry["count"])
+        for q in ("q01", "q99"):
+            if q in entry:
+                out[f"stats/{feat}/{q}"] = _nest_image_stat(entry[q]) if is_video else entry[q]
+    return out
+
+
+def write(samples: Iterator[Sample], output: OutputConfig) -> Path:
```
paul-nechifor: This function is enormous. It could use some simplification.

dimos/learning/dataprep/formats/lerobot.py
```diff
@@ -0,0 +1,491 @@
...
+    total_episodes = len(episode_rows)
+    total_frames = global_index
+    if data_path.exists() and data_path.stat().st_size > DATA_FILE_SIZE_MB * 1e6:
+        logger.warning(
+            "[dataprep] data file exceeds %d MB (single-file writer, no rolling): %s",
+            DATA_FILE_SIZE_MB,
+            data_path,
+        )
+
+    _write_meta(
```
paul-nechifor: When a function has this many params, that's an indication that you should be using some sort of structure to keep the values together (a dict, dataclass, a class with the associated methods, etc)

dimos/learning/dataprep/formats/lerobot.py
TomCC7: Also are we not using lerobot dataset lib here directly because we don't want to bundle that dependency?

ruthwikdasyam: yeah, dont wanna import lerobot, it pulls torch, torchcodec, datasets etc.
If we do in future, we can just simplify lerobot.py file

TomCC7: We will need their package ultimately if we do lerobot model inference later. So it's kind of inevitable.

ruthwikdasyam: True. Its a single file refactor, we can do it when we push inference code
Another reason is, if someone wants to do data collection and formatting - they need not to install lerobot dep

paul-nechifor (dismissed):

paul-nechifor (approved):

PR #2460, opened 2026-06-11
feat: vis protocol for manipulation

TomCC7:
# Summary
- Add VisualizationSpec and remove visualization methods from WorldSpec.
- Route visualization URL, publish, preview show/hide, animation, and close through WorldMonitor.
- Update ManipulationModule.preview_path() to use monitor-level visualization APIs.
- Improve preview playback: default duration now follows generated trajectory timing, with optional explicit duration override and target-FPS densification.
 
## How to Test
Preview should be much smoother.
```bash
uv run --extra manipulation dimos run xarm7-planner-coordinator
uv run python -i -m dimos.manipulation.planning.examples.manipulation_client
```


dimos/manipulation/planning/utils/path_utils.py
```diff
@@ -87,6 +87,56 @@ def interpolate_path(
...
+        q_end = np.array(path[i + 1].position, dtype=np.float64)
+        segment_deltas.append(float(np.max(np.abs(q_end - q_start))))
+
+    total_delta = sum(segment_deltas)
+    if total_delta <= 0:
+        return list(path)
+
+    interpolated: list[JointState] = [path[0]]
+    for i, segment_delta in enumerate(segment_deltas):
+        q_start = np.array(path[i].position, dtype=np.float64)
```
paul-nechifor: You're constructing the same arrays as in the first loop. Not efficient.

dimos/manipulation/planning/utils/path_utils.py
```diff
@@ -87,6 +87,56 @@ def interpolate_path(
...
+        q_end = np.array(path[i + 1].position, dtype=np.float64)
+        diff = q_end - q_start
+
+        num_steps = 1
+        if segment_delta > 0:
+            num_steps = max(1, int(np.ceil((segment_delta / total_delta) * min_intervals)))
+
+        for step in range(1, num_steps + 1):
```
paul-nechifor: Doing a for inside a for for numpy array seems like a waste. It might be worth measuring how long this takes and asking Claude to vectorize it to remove the slow Python loops.

dimos/manipulation/test_manipulation_unit.py
```diff
@@ -296,6 +299,28 @@ def _make_module_with_monitor(*configs: RobotModelConfig) -> ManipulationModule:
...
+def _make_joint_state(positions: list[float], name: list[str] | None = None) -> JointState:
+    return JointState(name=name or [f"j{i}" for i in range(len(positions))], position=positions)
+
+
+def _make_path(*points: list[float]) -> list[JointState]:
+    return [_make_joint_state(list(point)) for point in points]
+
+
+def _make_world_monitor_with_viz(viz: object | None) -> WorldMonitor:
+    monitor = WorldMonitor.__new__(WorldMonitor)
```
paul-nechifor: Using `__new__` is bad practice. Just construct the object normally, don't duplicate what `__init__` does in test code.

dimos/manipulation/planning/utils/path_utils.py
TomCC7: Can actually be replaced by np.interp, removed whole function

mustafab0 (approved):
Tested, works 👍

PR #2472, opened 2026-06-11
h264 transport with memory2 support

TomCC7:
## Summary

Adds an opt-in H.264 image path for DimOS typed `Image` streams.

This PR keeps the public module contract as `Out[Image]` / `In[Image]`, preserves JPEG as the default memory2 image codec, and adds H.264 as an explicit transport/storage codec path:

- `H264LcmTransport` compresses live `Image` streams over LCM, should be straightforward to extend to other transport as well
- `Image` can now carry raw pixels or encoded H.264 access-unit bytes through `encoding` and `codec_metadata`.
- `H264ImageCodec` stores encoded H.264 `Image` values through the existing memory2 `Backend` + `Codec` path.
- Demo blueprints exercise synthetic, webcam, replay, Rerun, and JPEG-vs-H.264 storage benchmark flows.

### Preserve `Image` as the public stream type

Modules still publish and subscribe to `Image`. H.264 is an opt-in codec detail, not a new default module payload type.

`Image` now has explicit codec fields:

```python
encoding: str = "raw"
codec_metadata: dict[str, Any] = {}
```

- `encoding="raw"`: `data` is a NumPy pixel array.
- `encoding="h264"`: `data` is one complete H.264 Annex B access unit for one source frame.

Pixel operations require raw images and fail clearly for encoded images.

### Keep memory2 generic

H.264 storage uses the normal memory2 codec path:

```python
store.stream("color_image", Image, codec="h264")
```

There is no H.264-specific memory2 backend, payload-strategy framework, or GOP index table in `Store` or `Backend`.

Default image storage remains unchanged:

```python
store.stream("color_image", Image)  # JPEG
```

### Let transport own live encode/decode state

`H264LcmTransport` has two receive modes:

- `decode_images=True`: normal subscribers receive decoded raw `Image` values.
- `decode_images=False`: subscribers receive encoded `Image(encoding="h264")` values, suitable for memory2 recording.

### Match Foxglove-style packet shape

Each encoded image contains all H.264 Annex B NAL units emitted for one encoder input frame. This removes RTP-fragment handling from memory2 and LCM consumers.

A complete encoded frame packet is not necessarily independently decodable. P-frames still require prior GOP state.

## Current limitations

The first version is best effort over LCM. H.264 delta frames depend on prior GOP state, so a subscriber that joins late, or a replay decoder that starts from an arbitrary timestamp, may first receive frames that cannot be decoded on their own. In that case, the decoder suppresses output until the next keyframe arrives. Video starts showing from that keyframe onward.

This is the same underlying problem for live late-join and memory2 random seek: DimOS does not yet have a uniform way to request, resend, or cache the keyframe needed to bootstrap decoding. A future DimOS QoS layer should solve this at the transport/session level by expressing reliability, keyframe request/resend, durable keyframe cache, or PLI-like recovery policy outside the H.264 codec itself.

ROS2 has a durability QoS setting that can resend old messages when new subscriber joins. Implementing something like that should be able to resolve the current issue.

Other current limits:
- H.264 currently supports uint8 RGB, BGR, and grayscale images only.
- memory2 replay of H.264 streams emits encoded images; consumers that need pixels must decode through an H.264 decoder session.
- The memory2 path has the strongest automated coverage today. We still need more end-to-end connection tests for the full live path across module, transport, H.264 encode/decode, subscriber/replay behavior, and visualization.

## Benchmark

Adds `demo-h264-storage-benchmark`, which records the same frames into two memory2 DBs:

- `benchmark_jpeg.db` uses the default JPEG image codec.
- `benchmark_h264.db` uses H.264 encoded images.

On the OpenCV `vtest.avi` sample, resized to 320 x 240 and recorded for 150 frames:

| Codec | DB size |
|---|---:|
| JPEG | 1,884,160 bytes |
| H.264 | 1,126,400 bytes |

H.264 used 59.8% of the JPEG storage size for this sample.

## Verification

Run
```bash
uv run dimos run demo-h264-webcam-rerun
```


dimos/experimental/security_demo/security_module.py
```diff
@@ -299,7 +299,7 @@ def _patrol_step(self) -> None:
         )
 
         annotated = draw_bounding_box(
-            image.data.copy(),
+            image.require_raw("SecurityModule._detection_step").copy(),
```
leshy, line 302: why change image type? doesn't this break now if using diff transport? you should transport the actual type modules require, feel free to use some shim that reconstructs the image on .data access

paul-nechifor: @TomCC7 I think it's better to leave Image as is. If you want to have an encoded image message, can't you create a new message type? I think this change complicates the current Image class too much. We had this in the past as well where Image supported CUDA storage and the Image was overly complex.

dimos/core/transport.py
```diff
@@ -162,6 +162,38 @@ def stop(self) -> None:
...
 
+class H264LcmTransport(LCMTransport):  # type: ignore[type-arg]
+    def __init__(
+        self,
+        topic: str,
+        type: type,
+        config: Any | None = None,
+        decode_images: bool = True,
+        **kwargs: Any,
+    ) -> None:  # type: ignore[no-untyped-def]
```
paul-nechifor, line 173: The current transports have `type: ignore` for legacy reasons. You shouldn't copy that. Use proper types.

dimos/msgs/sensor_msgs/Image.py
```diff
@@ -82,82 +84,189 @@ class AgentImageMessage(TypedDict):
...
+
+    @property
+    def is_raw(self) -> bool:
+        return self.encoding == RAW_IMAGE_ENCODING
+
+    @property
+    def is_encoded(self) -> bool:
+        return not self.is_raw
+
+    def require_raw(self, operation: str = "operation") -> np.ndarray[Any, np.dtype[Any]]:
```
paul-nechifor, line 198: This is a strange API. Why would `require_raw` need `operation` just to print a message in case it errors. The callee is already available in the traceback.

dimos/experimental/security_demo/security_module.py
TomCC7: For module-to-module transport cases, since all encode-decode are handled within transport layer, the image shouldn't require an overload. This addition was required to store encoded videoframe directly into recorder so that we enjoy smaller file size. And since we are doing encoding in transport layer, the message type can't be changed in the current design.

The other possible choice is to add a dedicated module to convert type, or directly modify recorder's storage process to encode directly. But then we'll need to duplicate the encode logic into separate places.

dimos/msgs/sensor_msgs/Image.py
TomCC7: yeah these are purely vibed to pass the mypy check. need to be addressed together with how we handle image overloading

dimos/experimental/security_demo/security_module.py
TomCC7: @paul-nechifor @leshy created a new pr that patches to this pr with a new design. Can check if that's the right direction: https://github.com/dimensionalOS/dimos/pull/2504. It basically mimic the current way of handling jpeg for transport/storage.

Generally I would say the core conflict here is that we are handling two usecases - a new transport that leverage video as a more efficient encoding during transportation and a new storage format for recorder - both need compression but they are in different layer (so we have to somewhat duplicate the encode/decode logic in two places).

One possible future cleanup I mentioned to @leshy before would be to reposition the recorder closer to the transport/import layer, so it can naturally record the already-encoded transport payload instead of receiving decoded module messages and encoding again for storage.

That would look more like:
```
             +--------------------+
             | recorder/storage   |
             | stores encoded pkt |
             +---------^----------+
                       |
Module Out[Image] -> encode once -> wire packet -> decode -> Module In[Image]
```

That may be a cleaner long-term architecture because transport and recording could share the same encoded packet path directly. But it is a larger architectural change. This patch keeps the existing module/recorder model and follows the repo’s current JPEG-style layering while avoiding encoded Image semantics.

paul-nechifor: Sorry! I forgot about this.

For me, the biggest issue is that I don't know what this is solving. When the [JPEG transport was added](https://github.com/dimensionalOS/dimos/pull/693) (BTW, notice that is was only 600 lines), that was because it improved performance with Foxglove (which we no longer use). Foxglove is a chromium app, so we wasted a lot of resources transporting un-encoded bytes. It was faster to encode to JPEG (TurboJPEG is super fast) and to decode in Foxglove than it is to transport un-encoded bytes.

Why do we need h264? I'm sure it encodes to smaller sizes, but how long does it take to encode? Is the tradeoff in speed for space worth it? I see you have a "Benchmark" section in the description but it doesn't mention time and doesn't use realistic images and sizes. Perhaps it would be best to use the frames from one of the office recordings.

I've looked a bit at #2504, but many of the changes are reverts from this PR, so it would be easiest to review if merged into this so I can compare with `main` instead.

Also, why specifically h264? This is from a previous benchmark:

```
Format           Library                    Time       Size  vs orig
--------------------------------------------------------------------
JPEG q50         TurboJPEG                0.003s   147.3 KB    20.3%
WebP q50         webp (pywebp)            0.078s   100.5 KB    13.9%
JPEG XL q50      pillow-jxl-plugin        0.021s    86.0 KB    11.9%
AVIF q50         pillow-avif-plugin       0.257s    87.6 KB    12.1%
```

You can see how TurboJPEG has almost no cost.

PR #2413, opened 2026-06-06
Viser Manipulation Module

TomCC7:
## Problem

Going to replace the drake meshcat and provide a moveit style console.

Closes #2412 

## Solution

1. Viser as a dimos module
2. most of communication through rpc
3. moveit style dragging and plan-review-execute
4. small issue: ik kind of slow (currently through rpc to manipulation module), not sure it's rpc or ik impl issue
image

## How to Test
```bash
uv run --extra manipulation-viser dimos run xarm7-viser-panel-mock
```

## Contributor License Agreement

- [x] I have read and approved the [CLA](https://github.com/dimensionalOS/dimos/blob/main/CLA.md).

PR #2475, opened 2026-06-12
feat: add viser as new manipulation vis backend

### LINKED ISSUE #2412: Viser Manipulation Control Panel
Going to replace drake meshcat and provide a moveit style console

---
*Synced from* [DIM-984](https://linear.app/dimensional/issue/DIM-984) *by che*

TomCC7:
Closes #2412 and supersedes #2413 

Adds Viser as a manipulation planning visualizer and rewires manipulation visualization config around a single validated `visualization` field. The new visualizer is more feature-rich than the original drake-bundled meshcat vis and provides a moveit style UI for plan/preview/execute.

image

Main changes:
- Add in-process Viser manipulation visualization: runtime, scene rendering, GUI panel, adapter, animation, and tests.
- Replace loose `visualization_backend` / `visualization_options` config with a Pydantic discriminated `visualization` config for `none`, `meshcat`, and `viser`.
- Clean up spec creation so manipulation startup creates:
  1. `WorldSpec`
  2. `PlanningSpecs`
  3. optional `VisualizationSpec`
- Keep Meshcat as an embedded Drake-world visualizer and create Viser as a separate visualization backend.
- Sync planning-scene metadata to external visualizers so Viser can initialize robot visuals without Viser-specific hooks in `WorldMonitor`.
- Include `viser[urdf]` in the `manipulation` extra and move Viser usage docs into the manipulation docs.

## Notes

- Viser panel execution is opt-in via `allow_plan_execute`.
- The earlier `set_planning_target` / `clear_planning_target` `VisualizationSpec` API was reverted; target controls remain internal to the Viser GUI/scene.

## How to Test
```bash
uv run dimos run xarm7-planner-coordinator \
  -o manipulationmodule.visualization.backend=viser \
  -o manipulationmodule.visualization.allow_plan_execute=true
```


dimos/manipulation/visualization/viser/adapter.py
```diff
@@ -0,0 +1,360 @@
...
+    if joint_state is None:
+        return None
+    return JointState(
+        {
+            "name": list(joint_state.name),
+            "position": list(joint_state.position),
+            "velocity": list(joint_state.velocity),
+            "effort": list(joint_state.effort),
+        }
+    )
```
mustafab0: You can just do 
```
def copy_joint_state(joint_state: JointState | None) -> JointState | None:
    return None if joint_state is None else JointState(joint_state)
```

since JointState already has a copy constructor

mustafab0 (dismissed):

dimos/manipulation/visualization/viser/adapter.py
TomCC7: yes, removed the helper

TomCC7:
@mustafab0 made some further cleanup and make pink the default ik module. can check again

mustafab0 (commented):
👍 

@paul-nechifor could you also please take a glance at this once?

Its clear from my end.

docs/usage/visualization.md
```diff
@@ -1,6 +1,7 @@
 # Viewer Backends
 
-Dimos supports Rerun as its visualization backend (`rerun` default, or `none` to disable).
+Dimos stream visualization uses `GlobalConfig.viewer` and `vis_module(...)` to render typed
```
paul-nechifor: Nitpick: This sentence has become quite hard to parse.

When writing documentation, you should view it from the perspective of someone unfamiliar with the system. It would be more accessible to say:

> Dimos uses Rerun for visualizations. It can be disabled by using `dimos --viewer none ...`.

And then explain what `vis_module` is for.

(There's no need to mention `--viewer rerun`, the default. That made sense when we also supported foxglove.)

docs/capabilities/manipulation/readme.md
```diff
@@ -76,6 +76,68 @@ preview()               # Preview in Meshcat
...
+            "allow_plan_execute": False,  # keep panel execution blocked by default
+        },
+    )
+)
+```
+
+Viser support is included in the `manipulation` extra:
+
+```bash
+uv sync --extra manipulation
```
paul-nechifor:
```suggestion
uv sync --extra manipulation --inexact
```

docs/capabilities/manipulation/adding_a_custom_arm.md
```diff
@@ -568,11 +568,14 @@ Add this to your `dimos/robot/yourarm/blueprints.py` alongside the coordinator b
...
+).transports(
+    {
+        ("joint_state", JointState): LCMTransport("/coordinator/joint_state", JointState),
+    }
 )
-# The planner's `coordinator_joint_state` input auto-connects to the
-# ControlCoordinator's output on the default `/coordinator_joint_state`
-# topic, so no `.transports(...)` override is needed.
+# The `.transports(...)` override subscribes the planner to the coordinator's
+# published joint-state topic.
```
paul-nechifor:
```suggestion
)
```

Note that I have merged https://github.com/dimensionalOS/dimos/pull/2515

It renames `joint_state` to `coordinator_joint_state` and you rely on defaults, so no need to use `.transports(...)` here.

dimos/manipulation/test_manipulation_unit.py
```diff
@@ -466,12 +492,11 @@ def _make_trajectory(*points: tuple[float, list[float]]) -> JointTrajectory:
...
-    world = viz if viz is not None else object()
-    with patch(
-        "dimos.manipulation.planning.monitor.world_monitor.create_world",
-        return_value=world,
-    ):
-        return WorldMonitor(enable_viz=viz is not None)
+    world = MagicMock()
+    return WorldMonitor(
+        world=world,
+        visualization=cast("VisualizationSpec | None", viz),
```
paul-nechifor: No need for casting. We don't typecheck in tests. But even so, the problem is the bad type in the function definition: `viz: object | None`.

dimos/manipulation/test_manipulation_unit.py
```diff
@@ -202,56 +204,80 @@ def test_multiple_robots_require_name(self, robot_config):
...
+        mock_world = MagicMock()
         mock_world_monitor = MagicMock(spec=WorldMonitor)
         mock_world_monitor.add_robot.return_value = "robot_id"
+        planning_specs = MagicMock(
+            world_monitor=mock_world_monitor,
+            planner=MagicMock(),
+            kinematics=MagicMock(),
+        )
 
         with (
```
paul-nechifor: This very long with statement is identical to the one above. Use a pytest fixture to remove duplication.

dimos/manipulation/manipulation_module.py
```diff
@@ -760,6 +804,97 @@ def get_init_joints(self, robot_name: RobotName | None = None) -> JointState | N
             return None
         return self._init_joints.get(robot[0])
 
+    def evaluate_joint_target(
+        self, joints: JointState | None, robot_name: RobotName
+    ) -> dict[str, object]:
```
paul-nechifor: When something is of various types, you're supposed to use `Any`, not `object`.

But it looks like this is a precise structure. All the dicts have the same keys. It might be better to replace it with a dataclass or TypedDict.

dimos/manipulation/manipulation_module.py
```diff
@@ -760,6 +804,97 @@ def get_init_joints(self, robot_name: RobotName | None = None) -> JointState | N
...
+        return {
+            "success": True,
+            "status": "FEASIBLE" if collision_free else "COLLISION",
+            "message": "Target is collision-free" if collision_free else "Target is in collision",
+            "collision_free": collision_free,
+            "ee_pose": self._world_monitor.get_ee_pose(robot_id, target),
+            "joint_state": target,
+        }
+
+    def evaluate_pose_target(self, pose: Pose, robot_name: RobotName) -> dict[str, object]:
```
paul-nechifor: You can use this for both places:

```python
  class TargetEvaluation(TypedDict, total=False):
      success: bool
      status: str
      message: str
      collision_free: bool
      joint_state: JointState | None
      ee_pose: PoseStamped | None
      position_error: float
      orientation_error: float
```

dimos/manipulation/blueprints.py
```diff
@@ -168,26 +172,34 @@
...
+            distance_threshold=0.08,
+            min_detections_for_permanent=3,
+            max_distance=1.0,
+            use_aabb=True,
+            max_obstacle_width=0.06,
+        ),
+    )
+    .transports(
+        {
+            ("joint_state", JointState): LCMTransport("/coordinator/joint_state", JointState),
```
paul-nechifor: Same as the other place. Delete this. Please verify that it works though.

dimos/manipulation/blueprints.py
```diff
@@ -84,7 +84,11 @@
         _right_arm_cfg.to_robot_model_config(),
     ],
     planning_timeout=10.0,
-    enable_viz=True,
+    visualization={"backend": "meshcat"},
+).transports(
+    {
+        ("joint_state", JointState): LCMTransport("/coordinator/joint_state", JointState),
```
paul-nechifor: Here too.

dimos/manipulation/visualization/test_factory.py
```diff
@@ -0,0 +1,199 @@
...
+        create_manipulation_visualization(
+            NoManipulationVisualizationConfig(), world=MagicMock(), world_monitor=MagicMock()
+        )
+        is None
+    )
+
+
+def test_create_visualization_meshcat_accepts_structural_world() -> None:
+    fake_world = FakeWorld()
+    assert isinstance(fake_world, VisualizationSpec)
```
paul-nechifor, line 209:
```suggestion
```

dimos/manipulation/visualization/test_factory.py
```diff
@@ -0,0 +1,199 @@
...
+    try:
+        world_monitor = MagicMock()
+        create_manipulation_visualization(
+            MeshcatVisualizationConfig(),
+            world=cast("WorldSpec", object()),
+            world_monitor=world_monitor,
+        )
+        raise AssertionError("expected ValueError")
+    except ValueError as exc:
+        assert "implements VisualizationSpec" in str(exc)
```
paul-nechifor:
```suggestion
def test_create_visualization_meshcat_rejects_non_visualization_world():
    with pytest.raises(ValueError, match="implements VisualizationSpec"):
        create_manipulation_visualization(
            MeshcatVisualizationConfig(),
            world=object(),
            world_monitor=MagicMock(),
        )
```

dimos/manipulation/visualization/test_factory.py
```diff
@@ -0,0 +1,199 @@
...
+        return None
+
+    def animate_path(self, robot_id, path, duration: float = 3.0) -> None:
+        return None
+
+    def close(self) -> None:
+        return None
+
+
+class FakeWorld(FakeVisualization):
```
paul-nechifor: You already define `class FakeWorld` in the first file in this review.

But I don't think you should at all, can't you do something like:

```
@runtime_checkable
class FakeWorld(WorldSpec, VisualizationSpec, Protocol):
    ...
```

dimos/manipulation/visualization/viser/visualizer.py
```diff
@@ -0,0 +1,138 @@
...
+        WorldRobotID,
+    )
+
+logger = setup_logger()
+
+
+class ViserManipulationVisualizer:
+    """In-process Viser implementation of the manipulation VisualizationSpec."""
+
+    def __init__(
```
paul-nechifor, line 58: This is a very heavy init. Contructors should be for constructing objects, not doing work.

I think most of this should be moved to a `start` method or  something like that.

dimos/manipulation/visualization/viser/theme.py
```diff
@@ -0,0 +1,122 @@
...
+class _ViserThemeModule(Protocol):
+    TitlebarButton: Callable[..., object]
+    TitlebarConfig: Callable[..., object]
+    TitlebarImage: Callable[..., object]
+
+
+def apply_dimos_theme(server: object) -> bool:
+    """Apply the default DimOS Viser theme without blocking visualization startup."""
+    try:
+        gui = cast("_GuiServer", server).gui
```
paul-nechifor: If you make `apply_dimos_theme` take `server: _GuiServer` instead of `server: object` then you wouldn't need to cast.

dimos/manipulation/visualization/viser/theme.py
```diff
@@ -0,0 +1,122 @@
...
+        configure_theme = gui.configure_theme
+    except AttributeError:
+        return False
+    if not callable(configure_theme):
+        return False
+
+    kwargs: dict[str, object] = {
```
paul-nechifor: You probably mean `dict[str, Any]`.

dimos/manipulation/visualization/viser/theme.py
```diff
@@ -0,0 +1,122 @@
...
+        return True
+    if "titlebar_content" in kwargs:
+        kwargs.pop("titlebar_content")
+        return _configure_theme(configure_theme, kwargs)
+    return False
+
+
+def _dimos_titlebar_content() -> object | None:
+    try:
+        viser_theme = cast("_ViserThemeModule", importlib.import_module("viser.theme"))
```
paul-nechifor: Why `importlib.import_module` instead of a proper import?

dimos/manipulation/visualization/viser/theme.py
```diff
@@ -0,0 +1,122 @@
...
+    except (ImportError, AttributeError, TypeError):
+        logger.warning("Skipping Viser titlebar content; theme API unavailable", exc_info=True)
+        return None
+
+
+def _dimos_logo_data_url() -> str | None:
+    try:
+        logo = DIMOS_LOGO_PATH.read_bytes()
+    except OSError:
+        logger.warning("Skipping DimOS Viser logo; asset unavailable", exc_info=True)
```
paul-nechifor: The file should always exist, right?

dimos/manipulation/visualization/viser/test_viser_visualization.py
```diff
@@ -0,0 +1,1352 @@
...
+                config=ViserVisualizationConfig(panel_enabled=False),
+            )
+        except RuntimeError as e:
+            assert str(e) == "missing viser_urdf"
+        else:
+            raise AssertionError("expected Viser construction failure")
+        assert closed == ["runtime"]
+    finally:
+        cast("Any", visualizer_module).ViserRuntime = original_runtime
+        cast("Any", visualizer_module).import_viser_urdf = original_import
```
paul-nechifor:
```suggestion
def test_visualizer_closes_runtime_when_construction_fails(monkeypatch: Any) -> None:
    closed = []

    class FakeRuntime:
        url = "http://localhost:8095"

        def __init__(self, config) -> None:
            self.config = config

        def start(self) -> FakeServer:
            return FakeServer()

        def close(self) -> None:
            closed.append("runtime")

    def fail_import_viser_urdf():
        raise RuntimeError("missing viser_urdf")

    monkeypatch.setattr(visualizer_module, "ViserRuntime", FakeRuntime)
    monkeypatch.setattr(visualizer_module, "import_viser_urdf", fail_import_viser_urdf)

    with pytest.raises(RuntimeError, match="missing viser_urdf"):
        ViserManipulationVisualizer(
            world_monitor=cast("Any", object()),
            manipulation_module=None,
            config=ViserVisualizationConfig(panel_enabled=False),
        )

    assert closed == ["runtime"]

```

dimos/manipulation/visualization/viser/test_viser_visualization.py
```diff
@@ -0,0 +1,1352 @@
...
+    assert server.theme_kwargs["brand_color"] == (22, 130, 163)
+    assert server.theme_kwargs["dark_mode"] is True
+    assert server.theme_kwargs["show_logo"] is False
+    assert server.theme_kwargs["show_share_button"] is False
+    assert server.theme_kwargs["control_layout"] == "collapsible"
+    assert server.theme_kwargs["control_width"] == "medium"
+
+
+def test_dimos_theme_configures_titlebar_when_supported(monkeypatch: Any) -> None:
+    class FakeThemeModule:
```
paul-nechifor: This is rather unusual. You're using a class as a fake module and methods as fake classes.

Use agents to use more idiomatic pytest mocking.

dimos/manipulation/visualization/viser/adapter.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+class InProcessViserAdapter:
+    """Small in-process boundary between Viser callbacks and manipulation internals."""
+
+    def __init__(
+        self,
+        *,
+        world_monitor: WorldMonitor,
+        manipulation_module: ManipulationModule | None,
```
paul-nechifor: Almost every method here checks `if self._module is None`. Are you sure you want to allow for `manipulation_module` to be None? It seems to be needed a lot.

dimos/manipulation/visualization/viser/gui.py
```diff
@@ -0,0 +1,718 @@
...
+            self.state.target_status = TargetStatus.EMPTY
+            self._build_joint_sliders()
+        self._sync_robot_dropdown(robots)
+        self._refresh_selected_robot_state()
+        self._ensure_scene_controls()
+        self._sync_preset_dropdown()
+        self._update_status_text()
+        self._update_control_state()
+
+    def _gui(self) -> Any | None:
```
paul-nechifor: I don't understand why you're using all this defensive code around checking if `gui` exists on server, checking if gui has add_folder or add_markdown... All of these are guaranteed to exist. There's no need to check.

Even this `def _gui` method shouldn't exist. Just use `self.server.gui.add_folder(...)`, etc.

dimos/manipulation/visualization/viser/gui.py
```diff
@@ -0,0 +1,718 @@
...
+    def _add_button(self, gui: Any, label: str, *, disabled: bool = False) -> Any:
+        try:
+            handle = gui.add_button(label, disabled=disabled)
+        except TypeError:
+            handle = gui.add_button(label)
+            try:
+                handle.disabled = disabled
+            except Exception:
+                logger.warning("Could not set button '%s' disabled state", label, exc_info=True)
+        return handle
```
paul-nechifor: I also don't understand why all of these try-excepts are needed. `add_button` is defined as:

```python
    def add_button(
        self,
        label: str,
        *,
        disabled: bool = False,
        visible: bool = True,
        hint: str | None = None,
        color: LiteralColor | tuple[int, int, int] | None = None,
        icon: IconName | None = None,
        order: float | None = None,
    ) -> GuiButtonHandle:
```

It's pretty clear that `disabled` does exist, so I don't think you need `except  TypeError`.

dimos/manipulation/visualization/viser/gui.py
```diff
@@ -0,0 +1,718 @@
...
+# Fallback joint-slider range (radians) when a robot config omits joint limits.
+DEFAULT_JOINT_LIMITS = (-3.14, 3.14)
+
+
+class ViserPanelGui:
+    """Optional operator panel with parity for the original cc/viser-vis panel."""
+
+    def __init__(
+        self,
+        server: Any,
```
paul-nechifor: The type here is `viser.ViserServer`.  Please instruct Caude to use the actual types.

dimos/manipulation/visualization/viser/theme.py
```diff
@@ -0,0 +1,122 @@
...
+DIMOS_LOGO_PATH = Path(__file__).with_name("assets") / "dimensional-logo.svg"
+
+logger = setup_logger()
+
+
+class _ThemeGui(Protocol):
+    configure_theme: Callable[..., object]
+
+
+class _GuiServer(Protocol):
```
paul-nechifor: The proper type here is `viser.ViserServer`, right? Are you defining this here because `viser` is optional?

I think that for type checking we always have to install optional packages, otherwise that defeats the purpose of having type annotations.

Using `if TYPE_CHECKING:` will make sure the code works at runtime even when package is not installed (if that's needed).

dimos/manipulation/visualization/viser/runtime.py
```diff
@@ -0,0 +1,89 @@
...
+
+VISER_INSTALL_HINT = (
+    "Viser manipulation visualization requires Viser. Install it with: uv sync --extra manipulation"
+)
+VISER_URDF_INSTALL_HINT = (
+    "Viser URDF support requires yourdfpy. Install it with: uv sync --extra manipulation"
+)
+
+
+class _ViserModule(Protocol):
```
paul-nechifor: (As in a previous comment), please use the proper types provided by `viser`.

dimos/manipulation/visualization/viser/runtime.py
```diff
@@ -0,0 +1,89 @@
...
+
+class _ViserExtrasModule(Protocol):
+    ViserUrdf: object
+
+
+class _Stoppable(Protocol):
+    def stop(self) -> None: ...
+
+
+def import_viser() -> ModuleType:
```
paul-nechifor: You should import viser regularly. Since the package is optional, you might need to use `except ImportError`. Please look at how we do it in other places to support optional packages.

dimos/manipulation/visualization/viser/gui.py
```diff
@@ -0,0 +1,718 @@
...
+            pass
+        try:
+            return (
+                self._to_float(value[0]),
+                self._to_float(value[1]),
+                self._to_float(value[2]),
+                self._to_float(value[3]),
+            )
+        except (AttributeError, IndexError, TypeError, ValueError):
+            return None
```
paul-nechifor: In this file, you need to be more precise with your types.

This function looks like it has no idea what `value` could be and just accesses some properties and hopes for the best. Please define what you expect `value` to be and then you can use it confidently.

Also, why is `_to_float` needed? It just calls `float` too.

dimos/manipulation/visualization/viser/gui.py
```diff
@@ -0,0 +1,718 @@
...
+
+    def _feasibility_status(
+        self, result: dict[str, object], success: bool, collision_free: bool
+    ) -> FeasibilityStatus:
+        status = str(result.get("status", "")).upper()
+        if success and collision_free:
+            return FeasibilityStatus.FEASIBLE
+        if "COLLISION" in status:
+            return FeasibilityStatus.COLLISION
+        if "IK" in status:
```
paul-nechifor: This looks very error prone. What if there's some status called `IK_SUCCEEDED`. With `if "IK" in status` you return `IK_FAILED`.

Isn't there a more precise way of checking what these status strings are expected to be so you can match them better to `FeasibilityStatus`?

docs/usage/visualization.md
TomCC7: Addressed in 4556497d7: clarified Rerun visualization wording and the `--viewer none` disable path.

dimos/manipulation/blueprints.py
TomCC7: Addressed in 4556497d7: removed this stale `.transports(...)` override as part of the blueprint cleanup.

docs/capabilities/manipulation/adding_a_custom_arm.md
TomCC7: Addressed in 4556497d7: removed the stale transport override and documented the `coordinator_joint_state` auto-connect path.

dimos/manipulation/blueprints.py
TomCC7: Addressed in 4556497d7: removed stale manipulation joint-state transport overrides and unused transport imports.

docs/capabilities/manipulation/readme.md
TomCC7: Addressed in 4556497d7: updated the install command to `uv sync --extra manipulation --inexact`.

dimos/manipulation/visualization/test_factory.py
TomCC7: Addressed in 4556497d7: split the fake world from the visualization protocol fake so the structural checks are explicit without casts.

dimos/manipulation/visualization/test_factory.py
TomCC7: Addressed in 4556497d7: changed the meshcat rejection test to use `pytest.raises`.

dimos/manipulation/visualization/test_factory.py
TomCC7: Addressed in 4556497d7: cleaned up the factory test fake setup and removed the unnecessary line.

dimos/manipulation/visualization/viser/visualizer.py
TomCC7: Addressed in 5a70a4492: made the Viser visualizer constructor lightweight and moved runtime/scene/GUI startup behind lazy `_ensure_started()` with exception-safe cleanup.

dimos/manipulation/visualization/viser/adapter.py
TomCC7: Addressed in 5a70a4492: Viser now requires a real manipulation module at the factory, visualizer, and adapter boundaries, so the adapter no longer carries optional-module fallbacks.

dimos/manipulation/visualization/viser/test_viser_visualization.py
TomCC7: Addressed in 5a70a4492: replaced the construction-failure cleanup coverage with focused lazy-start lifecycle tests, including GUI-start and URDF-import failure cleanup.

dimos/manipulation/visualization/viser/test_viser_visualization.py
TomCC7: Addressed in 5a70a4492: moved the new visualizer lifecycle coverage into focused pytest tests using monkeypatch instead of direct module/object reassignment patterns.

dimos/manipulation/visualization/viser/theme.py
TomCC7: Addressed in 8b89f9fdf: typed the theme server boundary and switched theme kwargs to `dict[str, Any]`.

dimos/manipulation/visualization/viser/theme.py
TomCC7: Addressed in 8b89f9fdf: replaced dynamic theme loading with direct `viser.theme` imports behind the lazy optional path.

dimos/manipulation/visualization/viser/theme.py
TomCC7: Addressed in 8b89f9fdf: logo loading now treats the packaged asset as expected to exist instead of silently omitting it.

dimos/manipulation/visualization/viser/theme.py
TomCC7: Addressed in 8b89f9fdf: tightened theme typing/imports and kept non-blocking fallback only around optional theme API changes.

dimos/manipulation/visualization/viser/gui.py
TomCC7: Addressed in 8b89f9fdf: typed the panel server as `viser.ViserServer` and use `server.gui` directly.

dimos/manipulation/visualization/viser/gui.py
TomCC7: Addressed in 8b89f9fdf: removed defensive `hasattr` branches for guaranteed Viser GUI APIs and call the GUI methods directly.

dimos/manipulation/visualization/viser/gui.py
TomCC7: Addressed in 8b89f9fdf: buttons now use `gui.add_button(..., disabled=...)` directly.

dimos/manipulation/visualization/viser/theme.py
TomCC7: Addressed in 8b89f9fdf: switched the theme path to proper Viser types/imports while preserving optional startup behavior.

dimos/manipulation/visualization/viser/runtime.py
TomCC7: Addressed in 8b89f9fdf: runtime imports now use lazy direct Viser imports with type-checking annotations instead of module-level optional imports.

dimos/manipulation/visualization/viser/runtime.py
TomCC7: Addressed in 8b89f9fdf: narrowed optional Viser import handling and added tests for missing `ViserUrdf` versus unrelated import failures.

dimos/manipulation/visualization/viser/gui.py
TomCC7: Addressed in 8b89f9fdf: made transform value conversion explicit and removed the redundant `_to_float` helper.

dimos/manipulation/manipulation_module.py
TomCC7: Addressed in d994343bf: replaced the loose target-evaluation dict/object typing with a shared `TargetEvaluation` TypedDict.

dimos/manipulation/manipulation_module.py
TomCC7: Addressed in d994343bf: added a shared `TargetEvaluation` TypedDict and applied it to both module evaluation paths.

dimos/manipulation/visualization/viser/gui.py
TomCC7: Addressed in d994343bf: replaced substring matching with exact status sets and added coverage for `IK_SUCCEEDED` staying invalid.

dimos/manipulation/visualization/viser/gui.py
```diff
@@ -0,0 +1,823 @@
...
+        self._operation_worker = OperationWorker(
+            self._set_error,
+            timeout_seconds=lambda: self.config.preview_request_timeout,
+        )
+
+    def start(self) -> None:
+        self._worker.start()
+        self._operation_worker.start()
+        try:
+            self.state.runtime = PanelRuntime.RUNNING
```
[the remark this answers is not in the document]
TomCC7: Addressed in f175d531f: preview now passes its own timeout, while plan/execute/cancel/clear run without inheriting `preview_request_timeout`.

dimos/manipulation/visualization/viser/gui.py
```diff
@@ -0,0 +1,823 @@
...
+                self.state.plan_state.planned_path = path
+            else:
+                self.state.plan_state.status = PlanStatus.FAILED
+            self._finish_operation(f"plan_to_joints={ok}")
+
+        self._operation_worker.submit(operation)
+
+    def _submit_preview(self) -> None:
+        robot_name = self.state.selected_robot
+        if robot_name is None:
```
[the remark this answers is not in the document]
TomCC7: Addressed in f175d531f: operation completions are now guarded by operation IDs, and plan target failures finish back to IDLE while preserving the error.

dimos/manipulation/visualization/viser/gui.py
```diff
@@ -0,0 +1,718 @@
...
+        except Exception:
+            self.close()
+            self.state.runtime = PanelRuntime.FAILED
+            raise
+
+    def close(self) -> None:
+        self._closed = True
+        self.state.runtime = PanelRuntime.STOPPING
+        self._worker.stop()
+        self._operation_worker.stop(timeout=None)
```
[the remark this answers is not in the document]
TomCC7: Addressed in f175d531f: GUI close now uses a bounded operation-worker stop instead of `timeout=None`.

dimos/manipulation/visualization/viser/state.py
```diff
@@ -0,0 +1,352 @@
...
+            while True:
+                try:
+                    request = self._requests.get_nowait()
+                except queue.Empty:
+                    break
+            result = self._handler(request)
+            self._apply_result(request, result)
+
+
+class OperationWorker:
```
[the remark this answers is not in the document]
TomCC7: Addressed in f175d531f: kept request coalescing under a submit lock and extended the same request-object/coalescing pattern to operation submissions.

dimos/manipulation/test_manipulation_unit.py
TomCC7: Addressed in a4e1cde22: removed the cast by typing the helper as `VisualizationSpec | None` and using a concrete fake visualization in the test.

dimos/manipulation/test_manipulation_unit.py
TomCC7: Addressed in a4e1cde22: replaced the duplicated long patch blocks with a pytest fixture using `mocker.patch`.

dimos/manipulation/visualization/viser/gui.py
```diff
@@ -0,0 +1,686 @@
...
+                float(value.w),
+                float(value.x),
+                float(value.y),
+                float(value.z),
+            )
+        except (AttributeError, TypeError, ValueError):
+            pass
+        try:
+            return (
+                float(value[0]),
```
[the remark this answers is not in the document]
TomCC7: Addressed in a4e1cde22: guard failures now record a recoverable error and stay IDLE without submitting an operation, so the controls are not left disabled.

dimos/manipulation/visualization/viser/adapter.py
TomCC7: Follow-up in d77459b71: tightened this further so Viser factory/visualizer/adapter signatures require `ManipulationModule` directly instead of accepting `None` and raising at runtime.

dimos/manipulation/visualization/viser/runtime.py
TomCC7: Follow-up in d77459b71: Viser runtime/scene/visualizer now use top-level Viser imports with explicit DimOS install hints for missing `viser` / `yourdfpy`, and the helper import layer is gone.

dimos/manipulation/visualization/viser/gui.py
```diff
@@ -0,0 +1,763 @@
...
+    def _wxyz_from_value(self, value: Sequence[float]) -> tuple[float, float, float, float] | None:
+        try:
+            return (
+                float(value[0]),
+                float(value[1]),
+                float(value[2]),
+                float(value[3]),
+            )
+        except (AttributeError, IndexError, TypeError, ValueError):
+            return None
```
TomCC7: whole logic over complication, reduce

dimos/manipulation/visualization/viser/runtime.py
```diff
@@ -0,0 +1,62 @@
...
+import webbrowser
+
+from dimos.manipulation.visualization.viser.config import ViserVisualizationConfig
+
+VISER_INSTALL_HINT = (
+    "Viser manipulation visualization requires Viser. Install it with: uv sync --extra manipulation"
+)
+VISER_URDF_INSTALL_HINT = (
+    "Viser URDF support requires yourdfpy. Install it with: uv sync --extra manipulation"
+)
```
TomCC7: Can actually merge prompt

dimos/manipulation/visualization/viser/scene.py
```diff
@@ -0,0 +1,402 @@
...
+            pass
+        try:
+            handle.wxyz = (
+                float(pose.orientation.w),
+                float(pose.orientation.x),
+                float(pose.orientation.y),
+                float(pose.orientation.z),
+            )
+        except (AttributeError, TypeError):
+            pass
```
TomCC7: avoid try catch here as it should always suceed

dimos/manipulation/visualization/viser/scene.py
```diff
@@ -0,0 +1,402 @@
...
+        self._set_urdf_mesh_material(target, mesh_color, mesh_opacity)
+        for handle in handles:
+            if handle is None:
+                continue
+            for attr in ("color", "material_color"):
+                if hasattr(handle, attr):
+                    try:
+                        setattr(handle, attr, color)
+                    except Exception:
+                        logger.warning("Could not set Viser handle %s", attr, exc_info=True)
```
TomCC7: since handles are properly typed now, should use type check instead of hasattr stuff

dimos/manipulation/visualization/viser/scene.py
```diff
@@ -0,0 +1,402 @@
...
+        wrapped_urdf = getattr(urdf, "_urdf", None)
+        if wrapped_urdf is None:
+            return ()
+        actuated_joint_names = getattr(wrapped_urdf, "actuated_joint_names", None)
+        if actuated_joint_names is not None:
+            return tuple(str(name) for name in actuated_joint_names)
+        joint_map = getattr(wrapped_urdf, "joint_map", None)
+        if joint_map is not None:
+            return tuple(str(name) for name in joint_map)
+        return ()
```
TomCC7: again too defensive, they do have proper type

dimos/manipulation/visualization/test_factory.py
mustafab0: The assert  line still exists here

paul-nechifor (approved):

PR #2484, opened 2026-06-13
feat: robot network discovery daemon

johnny-kantaros:
## Problem

An agentic harness has no current mechanism for detecting robots running on the same network. This PR addresses that limitation by installing a persistent background service that broadcasts each robot's identity over mDNS, making them more easily discoverable.

## Solution

Adds dimwizard, a daemon that makes robots discoverable on the local network via mDNS. Includes a new hook in `dimos run` which prompts the user to install the background service (launchd on macOS, systemd on Linux).

The service will continuously broadcasts the robot's name and LCM URL under `_dimensional._tcp.local`. A follow-up harness PR will use this to automatically detect + connect to robots on the same network without manual config.
  
## How to Test

```
  uv sync
  dimos run <blueprint>  # wizard should prompt on first invocation
  dimwizard status          # confirm beacon is running
  dimwizard kill               # tear down
```

### DIFF
--- a/dimwizard/pyproject.toml
+++ b/dimwizard/pyproject.toml
@@ -1,0 +1,21 @@
+[build-system]
+requires = ["setuptools>=70", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "dimwizard"
+version = "0.1.0"
+description = "Makes a Dimensional robot permanently discoverable on the local network"
+requires-python = ">=3.10"
+dependencies = [
+    "zeroconf>=0.131.0",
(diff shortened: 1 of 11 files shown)


dimwizard/pyproject.toml
```diff
@@ -0,0 +1,21 @@
+[build-system]
```
paul-nechifor, line 1: Why a separate `pyproject.toml`? Does this get included in the `dimos` package?

paul-nechifor:
Closed to due to inactivity. Can be reopened later.

PR #2504, opened 2026-06-16
feat: add opt-in h264 image storage

TomCC7:
## Summary

This patch updates the existing H.264 image PR so `Image` is not overloaded with encoded-video semantics.

`Image` now remains a decoded, pixel-addressable raster type: `Image.data` is always a NumPy array. H.264 bytes are used only as a transport/storage representation, matching the way the repo already treats JPEG: modules keep using `In[Image]` / `Out[Image]`, while codecs encode/decode behind the transport or storage boundary.

## What changed

- Restored raw-only `Image` semantics.
  - Removed public encoded-Image fields and helpers such as `encoding`, `codec_metadata`, `require_raw()`, and `Image.encoded(...)`.
  - Updated call sites back to direct `.data` access.

- Kept H.264 transport as an end-to-end wire codec.
  - Module-facing API stays `Image -> Image`.
  - `H264LcmTransport` encodes `Image` to H.264 packet bytes on publish and decodes back to `Image` before subscribers receive it.
  - Encoded packets are not exposed as `Image`.

- Added opt-in H.264 storage for logical `Image` streams.
  - `store.stream(name, Image, codec="h264")`
  - `RecorderConfig(stream_codecs={"color_image": "h264"})`
  - Default `Image` storage remains JPEG.

- Reused one shared H.264 implementation.
  - Transport and storage both call the same core `H264Encoder`, `H264Packet`, and `H264Decoder` logic.
  - Storage adds stream-specific behavior around that core: append locking, keyframe metadata, keyframe seeking, lazy decode, and id-ordered sequential decode.

- Updated demos/docs/tests for the split semantics.
  - Webcam recorder now uses raw transport into the recorder and lets the storage backend do the H.264 encode.
  - Playback reads H.264-backed memory2 observations as decoded `Image` values.

## Why

The earlier version made `Image.data` sometimes mean decoded pixels and sometimes mean encoded H.264 bytes. That made the type harder to reason about and forced unrelated image consumers to handle encoded-video cases.

This patch follows the repo’s existing JPEG pattern instead:

```text
public API:       Image
physical layer:   encoded bytes
public API:       Image
The encode/decode logic lives behind transport/storage boundaries, not in the Image message type.

### DIFF
--- a/dimos/protocol/video/demo_h264_video_e2e.py
+++ b/dimos/protocol/video/demo_h264_video_e2e.py
@@ -26,11 +26,12 @@ import time
 import cv2
 import numpy as np
 
+from dimos.constants import DEFAULT_CAPACITY_COLOR_IMAGE
 from dimos.core.coordination.blueprints import autoconnect
 from dimos.core.core import rpc
 from dimos.core.module import Module, ModuleConfig
 from dimos.core.stream import In, Out
-from dimos.core.transport import H264LcmTransport
+from dimos.core.transport import H264LcmTransport, pSHMTransport
 from dimos.hardware.sensors.camera.module import CameraModule
 from dimos.hardware.sensors.camera.webcam import Webcam
 from dimos.memory2.module import OnExisting, Recorder
@@ -552,10 +553,9 @@ demo_h264_webcam_record = autoconnect(
     ),
 ).transports(
     {
-        ("color_image", Image): H264LcmTransport(
+        ("color_image", Image): pSHMTransport(
             "/demo_h264_webcam_record/color_image",
-            Image,
-            config=_webcam_h264_config,
+            default_capacity=DEFAULT_CAPACITY_COLOR_IMAGE,
         )
     }
 )
--- a/dimos/protocol/video/demo_h264_video_e2e.py
+++ b/dimos/protocol/video/demo_h264_video_e2e.py
@@ -22,7 +22,6 @@ import sqlite3
 import tempfile
 import threading
 import time
-from typing import ClassVar, cast
 
 import cv2
 import numpy as np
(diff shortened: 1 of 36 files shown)

### CODE REVIEW

File: dimos/protocol/video/demo_h264_video_e2e.py [hunk 1]
```diff
@@ -0,0 +1,627 @@
...
+        for seq in range(self.config.frame_count):
+            if self._stop_event.is_set():
+                break
+            frame = self._make_frame(seq)
+            self.color_image.publish(frame)
+            next_publish += period
+            time.sleep(max(0.0, next_publish - time.monotonic()))
+        logger.info("Synthetic H.264 source finished publishing frames")
+
+    def _make_frame(self, seq: int) -> Image:
```
paul-nechifor, line 102: We have `from dimos.utils.fast_image_generator import random_image` to produce random frames for testing encodings.

There's also `demo_image_encoding.py`, although it looks like it's broken currently and doesn't use blueprints (it was added before that).

dimos/protocol/video/demo_h264_video_e2e.py
```diff
@@ -0,0 +1,627 @@
...
+            )
+        frame_rgb = cv2.cvtColor(frame_bgr, cv2.COLOR_BGR2RGB)
+        return Image(
+            data=frame_rgb,
+            format=ImageFormat.RGB,
+            frame_id=self.config.output_frame_id,
+            ts=time.time(),
+        )
+
+    def _make_frame(self, seq: int) -> Image:
```
paul-nechifor, line 259: This is the same as the other `_make_frame`. Please don't duplicate code. 🥺

TomCC7: Hi Paul, pls don't look at this pr yet 😂 It's only a directional thing for now and contain a bunch of reverting change. I guess if you agree with the direction I'll first merge this back to the main pr and polish

paul-nechifor: I think it's best to just merge it there and I'll look at it again, but it would be nice to change the description too because I don't know if H.264 is needed for storage reasons, for web transport reasons, or just as an alternative.

PR #2505, opened 2026-06-16
feat: in-house robot asset manager

TomCC7:
## Summary

- Replace the checked-in A1Z, A750, Piper, and xArm LFS descriptions with pinned, lazily resolved Git sources. Cached checkouts preserve local changes and local-only commits; `dimos cache clean --force` explicitly discards them.
- Add an immutable `RobotModel` loading boundary. Xacro expansion, package URI resolution, fixed-frame additions, and joint-limit overrides happen in memory, with no generated URDF cache.
- Update Drake, RoboPlan, Pink/Pinocchio, Viser, planning-group discovery, control tasks, and blueprints to consume the loaded model instead of retaining backend-specific URDF processing.
- Preserve the A1Z centered gripper frame, make the A750 neutral gripper state valid, and wire A750 trajectory execution and Viser visualization.
- Remove the obsolete model parser, legacy description fields, and bundled robot-description archives.

## Test

```bash
./bin/pytest-fast
uv run --extra manipulation dimos run keyboard-teleop-a750
uv run --extra manipulation dimos run dual-xarm6-planner-coordinator
```


dimos/robot/assets/declarations.py
```diff
@@ -0,0 +1,79 @@
...
+from __future__ import annotations
+
+from dimos.robot.assets.manager import RobotAssetDeclaration
+
+XARM_ROS2_REPO = "https://github.com/xArm-Developer/xarm_ros2"
+PIPER_DESCRIPTION_REPO = "https://github.com/agilexrobotics/agx_arm_urdf"
+A750_DESCRIPTION_REPO = "https://github.com/adob/a750_description"
+
+
+ROBOT_ASSETS: dict[str, RobotAssetDeclaration] = {
```
TomCC7: One issue with this design is that it's still centralized. Ideally we want to let this new system support robot sideloading - user can use our repo as a python package and just define their own robot adapter with customized robot urdf/whatever

dimos/robot/assets/git_cache.py
```diff
@@ -0,0 +1,171 @@
...
+from pathlib import Path
+import shutil
+import tempfile
+from urllib.parse import urlparse
+import warnings
+
+from filelock import FileLock
+from git import GitCommandError, InvalidGitRepositoryError, NoSuchPathError, Repo
+
+DEFAULT_GIT_ASSET_CACHE_ROOT = Path.home() / ".cache" / "dimos" / "robot_assets"
```
TomCC7: should use some lib to make it canonical - we need to support windows/linux/mac and they might have different cache folder

dimos/robot/assets/git_cache.py
```diff
@@ -0,0 +1,171 @@
...
+DEFAULT_ROBOT_ASSET_CACHE_ROOT = DEFAULT_GIT_ASSET_CACHE_ROOT
```
TomCC7: should just unify, since they are the same

dimos/robot/assets/manager.py
```diff
@@ -0,0 +1,294 @@
...
+        instance: RobotAssetPath = super().__new__(cls, ".")
+        object.__setattr__(instance, "_robot_asset_model", model)
+        object.__setattr__(instance, "_robot_asset_role", _role_key(role))
+        object.__setattr__(
+            instance, "_robot_asset_relative_parts", tuple(str(p) for p in relative_parts)
+        )
+        object.__setattr__(
+            instance, "_robot_asset_manager", manager or default_robot_asset_manager()
+        )
+        object.__setattr__(instance, "_robot_asset_resolved_cache", None)
```
TomCC7: these seems creepy, why we need these?

dimos/robot/assets/git_cache.py
```diff
@@ -0,0 +1,171 @@
...
+@dataclass(frozen=True)
+class GitAssetCheckout:
+    """Resolved local checkout information."""
+
+    path: Path
+    repo_url: str
+    ref: str
+    updated: bool = False
+    used_cached_fallback: bool = False
+    skipped_dirty_update: bool = False
```
mustafab0: Should be default to true

mustafab0 (dismissed):

pyproject.toml
```diff
@@ -116,11 +116,12 @@ dependencies = [
     "pydantic",
     "python-dotenv",
     "packaging>=24.0",
-    "filelock>=3.16,<4",
+    "filelock>=3.20,<4",
     "annotation-protocol>=1.4.0",
     "lazy_loader",
     "typing_extensions>=4.7; python_version < '3.13'",
     "plum-dispatch==2.5.7",
+    "GitPython>=3.1.45",
```
TomCC7, line 124: should we put this in core dependency?

dimos/utils/cache.py
```diff
@@ -54,30 +56,106 @@ class CacheCleanResult:
...
-        result.failed.append(CacheIssue(target, str(error)))
-    else:
+    protected = {} if force else _protected_robot_checkouts()
+    result.skipped.extend(CacheIssue(path, reason) for path, reason in protected.items())
+    if _remove_except(target, set(protected), result):
         result.cleaned.append(target)
     return result
 
 
+def _protected_robot_checkouts() -> dict[Path, str]:
```
TomCC7: dont like how this file is organized, can we organize logic by kind of file we are cleaning?

dimos/robot/manipulators/piper/config.py
```diff
@@ -36,11 +37,16 @@
...
+_PIPER_REPO = RobotDescriptionSource(url=PIPER_DESCRIPTION_REPO, ref=PIPER_DESCRIPTION_REF)
+PIPER_MODEL_PATH = _PIPER_REPO / "piper" / "urdf" / "piper_with_gripper_description.xacro"
 PIPER_PACKAGE_PATHS: dict[str, Path] = {
-    "piper_description": LfsPath("piper_description"),
-    "piper_gazebo": LfsPath("piper_description"),
+    # Upstream URIs are package://agx_arm_description/agx_arm_urdf/..., so the
+    # package root is the parent of the preserved agx_arm_urdf checkout folder.
+    "agx_arm_description": _PIPER_REPO.parent,
 }
+PIPER_FK_MODEL = _PIPER_REPO / "piper" / "urdf" / "piper_description.urdf"
```
TomCC7, line 50: this seems a new variable is that ever been used

dimos/robot/assets/processing.py
```diff
@@ -0,0 +1,336 @@
...
+        return getattr(object.__getattribute__(self, "_resolve")(), name)
+
+    def __str__(self) -> str:
+        return str(self._resolve())
+
+    def __fspath__(self) -> str:
+        return str(self._resolve())
+
+
+def rendered_robot_description(
```
TomCC7: do we really need to process and cache the generated urdf from xacro? it seems very fast we can just redo everytime we need them. this also aligns with ros standard

dimos/robot/assets/processing.py
```diff
@@ -0,0 +1,336 @@
...
+
+    if removed_joint_names:
+        urdf_content = remove_joint_subtrees(urdf_content, removed_joint_names)
+
+    rendered_urdf.write_text(urdf_content)
+    logger.info(f"Rendered URDF cached at: {rendered_urdf}")
+    return rendered_urdf
+
+
+class RenderedRobotDescriptionPath(type(Path())):  # type: ignore[misc]
```
TomCC7: again don't like this

dimos/robot/assets/processing.py
```diff
@@ -0,0 +1,336 @@
...
+_RENDERED_URDF_CACHE_ROOT = DERIVED_ASSET_CACHE_ROOT / "rendered_urdfs"
+
+
+def render_urdf(
+    urdf_path: Path | str | os.PathLike[str],
+    package_paths: Mapping[str, Path | str | os.PathLike[str]] | None = None,
+    xacro_args: Mapping[str, str] | None = None,
+    *,
+    package_uri_mode: PackageUriMode = "preserve",
+    removed_joint_names: frozenset[str] = frozenset(),
```
TomCC7: why is this needed

pyproject.toml
TomCC7: Yes. GitPython remains a core dependency because lazy robot-description sources can resolve from normal blueprint and config paths, not only manipulation extras.

dimos/utils/cache.py
TomCC7: Reorganized in a7083676ae: robot checkout discovery and Git protection policy now live in .  only coordinates generic cache locking and removal.

dimos/robot/manipulators/piper/config.py
TomCC7: This is used by  for Piper FK, so I kept it. The generic unused  was removed.

dimos/robot/assets/processing.py
TomCC7: Removed the rendered URDF cache. URDF/Xacro expansion now returns XML in memory on every backend load; only content-addressed converted mesh files remain cached because Drake needs those as filesystem artifacts.

dimos/robot/assets/processing.py
TomCC7: Removed  and the lazy rendered-path layer entirely. Configs retain the source path and Xacro inputs, and each backend loads the expanded XML directly in its own process.

dimos/robot/assets/processing.py
TomCC7: Removed this rendered-asset cache constant together with the rendered/prepared URDF cache. The only derived robot assets now are content-addressed converted meshes.

dimos/robot/assets/git_cache.py
TomCC7: The refresh toggle and result metadata are gone. GitAssetCache.resolve now always fetches and refreshes an existing clean checkout; it skips refresh only when local changes or local-only commits need protection.

dimos/utils/cache.py
TomCC7: Path clarification: robot checkout discovery and Git protection policy now live in dimos/robot/assets/cache.py. dimos/utils/cache.py only coordinates generic cache locking and removal.

dimos/robot/manipulators/piper/config.py
TomCC7: Path clarification: PIPER_FK_MODEL is used by dimos/control/examples/cartesian_ik_jogger.py for Piper FK. The generic unused A750_FK_MODEL was removed.

dimos/robot/assets/processing.py
TomCC7: Name clarification: RenderedRobotDescriptionPath and the lazy rendered-path layer were removed entirely. Configs retain the source path and Xacro inputs, and each backend loads expanded XML directly.

dimos/control/tasks/cartesian_ik_task/pink_control_ik.py
```diff
@@ -106,18 +105,14 @@ def __init__(self, config: PinkControlIKConfig) -> None:
...
+            package_paths=robot.package_paths,
+            xacro_args=robot.xacro_args,
+            package_uri_mode="absolute",
+            additional_fixed_frames=tuple(robot.additional_fixed_frames),
         )
-        if not prepared_path.exists():
-            raise FileNotFoundError(f"prepared Pink control URDF not found: {prepared_path}")
-
-        model = pinocchio.buildModelFromUrdf(str(prepared_path))
+        model = pinocchio.buildModelFromXML(description.xml)
```
TomCC7, line 108: I believe here xml still means urdf...

dimos/robot/assets/cache.py
```diff
@@ -0,0 +1,50 @@
...
+
+"""Robot-asset cache inspection policy."""
+
+import os
+from pathlib import Path
+
+from dimos.robot.assets.git_cache import git_checkout_protection_reason
+
+
+def protected_robot_checkouts(robot_cache_root: Path) -> dict[Path, str]:
```
TomCC7: there's no need to put the file here, would rather put it back to the cache folder, just need to make sure each usecase is well structured within it's own function/section

dimos/robot/assets/xacro.py
TomCC7, line 1: wonder is this ament support file still required since we process xacro directly

dimos/control/tasks/cartesian_ik_task/pink_control_ik.py
TomCC7: Correct: Pinocchio buildModelFromXML parses a URDF XML stream. Commit a4d2d8c0d1 makes that explicit with LoadedUrdf.urdf_xml; MJCF .xml paths now use buildModelFromMJCF instead.

dimos/robot/assets/cache.py
TomCC7: Moved the robot checkout scan back into dimos/utils/cache.py as a dedicated private function and removed dimos/robot/assets/cache.py. Generic cleanup orchestration, robot discovery, recursive removal, and locking remain separate functions.

dimos/robot/assets/xacro.py
TomCC7: Direct Xacro expansion still needs this resolution logic: the active xArm model uses many find xarm_description substitutions, and xacro.process_file delegates those to ament. I moved the helper under robot assets and added a hermetic non-ROS test proving package includes resolve from the supplied package map.

dimos/control/tasks/cartesian_ik_task/pink_control_ik.py
```diff
@@ -106,18 +105,20 @@ def __init__(self, config: PinkControlIKConfig) -> None:
     def build(self) -> _PinkRuntime:
         config = self._config
         robot = config.robot_model
-        prepared_path = Path(
-            prepare_urdf_for_drake(
-                robot.model_path,
+        model_path = robot.model_path.resolve()
+        if model_path.suffix.lower() == ".xml":
```
TomCC7: do we really have a mjcf usage path? I think we need to handle mjcf/urdf specification from upper level rather than consumer level

dimos/robot/assets/processing.py
```diff
@@ -0,0 +1,182 @@
...
+def normalize_package_paths(
+    package_paths: Mapping[str, Path | str | os.PathLike[str]],
+) -> dict[str, Path]:
+    return {
+        package_name: Path(os.fspath(package_path)).resolve()
+        for package_name, package_path in package_paths.items()
+    }
+
+
+def _expand_xacro(
```
TomCC7: why is this not in xacro.py then, need cleaner responsibility split over files

dimos/control/tasks/cartesian_ik_task/pink_control_ik.py
TomCC7: There is no production MJCF-backed RobotModelConfig. Commit fa0b921237 makes that boundary explicit: RobotModelConfig.model_path is replaced by the URDF/Xacro-only urdf_path field, config construction rejects other formats, and the Drake/Pink consumers no longer inspect suffixes or call MJCF loaders. MuJoCo simulation assets and standalone generic model APIs retain their separate MJCF paths.

dimos/robot/assets/processing.py
TomCC7: Moved in fa0b921237. processing.py now only dispatches a Xacro source to xacro.expand_xacro; the intermediate _expand_xacro wrapper is gone. xacro.py owns expansion, the optional-dependency error, ament registration, and the non-ROS find fallback.

dimos/manipulation/planning/spec/config.py
```diff
@@ -59,13 +61,14 @@ class RobotModelConfig(ModuleConfig):
...
     name: str
-    model_path: Path
+    urdf_path: Path
     srdf_path: Path | None = None
     base_pose: PoseStamped = Field(default_factory=PoseStamped)
     joint_names: list[str]
     base_link: str = "base_link"
     planning_groups: list[PlanningGroupDefinition] = Field(default_factory=list)
     package_paths: dict[str, Path] = Field(default_factory=dict)
+    additional_fixed_frames: list[FixedFrameDefinition] = Field(default_factory=list)
```
TomCC7: don't like this design too intrusive. we need a unified way to programically modify a robot model after loading from file...

TomCC7: Reworked in e8c542a30c. RobotModelConfig no longer has the special-purpose additional_fixed_frames field. It now exposes an ordered urdf_processors pipeline of pickle-safe LoadedUrdf -> LoadedUrdf callables, applied after Xacro/package resolution and before every backend. A1Z uses AddFixedFrame as one processor. I also evaluated yourdfpy as a canonical model, but its serializer dropped xArm gazebo, transmission, and ros2_control elements, so this seam preserves the expanded URDF and unknown tags.

TomCC7: Reworked this again in a9d6238fc2. The processor abstraction is gone entirely: RobotModelConfig now owns a lazy, immutable RobotModel, model-level edits use with_fixed_frame(), and adapters only materialize the finished model with load(). Xacro expansion and package URI resolution stay internal to that model boundary, with no rendered-URDF files or downstream processor callbacks. I also consolidated the tests around public behavior, worker pickling, laziness, and backend consumption. Full fast suite: 4,064 passed.

docs/coding-agents/code-quality-rules.md
```diff
@@ -18,7 +18,7 @@ Rules dimos code is expected to follow. They address recurring issues found in c
 * Specify only what differs from defaults. Don't restate defaults like `tick_rate=100.0`, `publish_joint_state=True`, or default topics (`/cmd_vel`, `/odom`).
 * `.transports({...})` applies to all matching modules, so define a remap once, not twice across sub-blueprints.
 * No lambdas -- they can't be pickled to worker processes. Use named functions.
-* Do no work at import time: no subprocesses, viewers, model parsing, or network. In particular don't call `get_data(...)` (it blocks import until the download finishes) -- use `LfsPath` (resolved at access time) or build the config in `start`/`build`. Any process you start must be managed (shut down when not needed).
+* Do no work at import time: no subprocesses, viewers, model parsing, or network. In particular don't call `get_data(...)` (it blocks import until the download finishes) -- use lazy path adapters such as `RobotDescriptionSource` paths for upstream robot descriptions or `LfsPath` for intentionally vendored assets, or build the config in `start`/`build`. Any process you start must be managed (shut down when not needed).
```
TomCC7: revert this is robot model specific rules...

docs/usage/cli.md
```diff
@@ -318,6 +318,20 @@ dimos cache clean
...
+signal that cleanup was incomplete. Pass `--force` to delete those checkouts as
+well:
+
+```bash
+dimos cache clean --force
+```
+
+`--force` does not bypass the active-blueprint safety check or the confirmation
+prompt. Combine it with `--yes` only when intentional data loss is acceptable.
+
```
TomCC7:
```suggestion
```

docs/coding-agents/code-quality-rules.md
TomCC7: Reverted in f0d8a60e30. The general import-time rule no longer includes robot-model-specific guidance.

docs/usage/cli.md
TomCC7: Removed this section in f0d8a60e30.

dimos/robot/manipulators/a750/test_config.py
mustafab0, line 1: I think we can skip these config test files. 

They are all testing different stuff for different files. 

We could just make a more generic test file that tests all config files if necessary. But seems like we can skip

dimos/robot/manipulators/openarm/config.py
mustafab0, line 34: Are we using openarm_description from Lfs becuase openarm does not have an official repo that maintains this?

mustafab0 (dismissed):

mustafab0 (dismissed):

dimos/robot/manipulators/openarm/config.py
TomCC7: this is because we are still on main branch which uses single arm openarm models... I'll update accordingly in my new openarm support pr

mustafab0 (approved):

Issue #1932, opened 2026-04-29
memory2 - audio ingress


leshy:
capture all audio input, (on go2 maybe for a start) store in memory2
(figure out proper messages for this - what does ros use? do we need to design our own? audio is a bit complex with encodings, chunks etc)

PR #2507, opened 2026-06-16
feat(audio): add AudioModule for issue #1932

GuoZhuoRan:
Adds mic audio capture and chunked publishing as AudioStamped on an Out stream, mirroring CameraModule. Validated on macOS Apple Silicon at 50 Hz / 20 ms frames with both synthetic (sine tone) and real mic sources.

- dimos/msgs/audio_msgs/AudioStamped.py: Python overlay wrapping foxglove_msgs.RawAudio for LCM encode/decode, with from_pcm() and to_numpy() helpers. Flags that builtin_interfaces.Time (not std_msgs.Header) is the wire type, so frame_id is not preserved.
- dimos/hardware/sensors/audio/module.py: AudioModule(Module) with AudioConfig(ModuleConfig), async def main() lifecycle, @rpc start/stop, @skill record_clip.
- examples/audio/validate_audio_module.py: LCM round-trip assert + live stream rate/timestamp validation.

## Problem

Closes DIM-XXX


dimos/msgs/audio_msgs/AudioStamped.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+class AudioStamped(Timestamped):
+    """Stamped audio chunk with PCM payload.
+
+    Carries one frame of raw PCM audio together with a std_msgs.Header and
+    audio metadata.  Serialises to/from foxglove_msgs.RawAudio on the wire.
+    """
+
+    msg_name = "foxglove_msgs.RawAudio"  # wire type used for LCM
```
leshy, line 52: we don't use foxglove, where does this come from?

dimos/msgs/audio_msgs/AudioStamped.py
```diff
@@ -0,0 +1,168 @@
...
+        return arr
+
+    # ------------------------------------------------------------------
+    # LCM serialisation
+    # ------------------------------------------------------------------
+
+    def lcm_encode(self) -> bytes:
+        """Encode to foxglove_msgs.RawAudio wire bytes.
+
+        NOTE: frame_id and seq from self.header are NOT preserved (the wire
```
leshy: ros2 header has no seq

why not preserve frame_id?

dimos/msgs/audio_msgs/AudioStamped.py
GuoZhuoRan: Foxglove is not a new dependency; it's already mirrored into dimos_lcm, and RawAudio is the only audio type in there, so I reused it. Left a note that it's a stand-in pending a native Header-bearing type.

dimos/msgs/audio_msgs/AudioStamped.py
GuoZhuoRan: My comment is wrong; I will fix this issue. Frame_id does exist, but based on the RawAudio format, it only carries a timestamp; there is no frame_id on the wire type to put in. Preserving a frame_id means adding a header-bearing audio type to dimos-lcm, and we can discuss it today

PR #2525, opened 2026-06-18
krishna/Booster_K1_robot_integration

KrishnaH96:
## Add Booster K1 robot integration

A dimos adapter for the Booster K1 humanoid base: connection plus runnable teleop and coordinator blueprints. Validated on hardware. Solves issue: DIM-583

**BoosterRPCConnection** (robot-agnostic, reusable for the T1) wraps booster-rpc:
- Non-blocking coalescing sender at 30 Hz with a 0.5 s dead-man stop, bridging the SDK's blocking gRPC under the coordinator's 100 Hz tick.
- Camera JPEG stream decoded onto an Observable. All long-running work runs as coroutines the Module spawns on its own loop, the wrapper owns no threads.
- Mode transitions are requested once, confirmed via get_mode(), then given a settle period for the physical motion. Transport and vendor failures return False instead of raising.

**K1Connection**: camera and velocity control as dimos streams and RPCs.
- The operator arms the robot to WALKING with the Booster app before launch, the same convention as the G1. start() only verifies the mode and never commands transitions.
- The standup RPC and stand skill perform the full arming sequence for deliberate use.
- The walk skill rejects non-finite input, clamps velocity and duration to a validated envelope, and confirms delivery of both a movement command and the final stop before reporting success.
- Camera intrinsics are placeholders with a startup warning. Measured calibration is a follow-up.

**Blueprints:** booster-k1-basic, booster-k1-keyboard-teleop, booster-k1-coordinator, booster-k1-coordinator-keyboard-teleop. Plus the booster extra and blueprint registration.

**Run:**
```bash
uv sync --extra booster
# arm the robot to WALKING with the Booster app, then
dimos --robot-ip <ip> --viewer rerun run booster-k1-coordinator-keyboard-teleop
```

**Validated on hardware (K1 on gantry):** WASD teleop + camera-in-rerun on both the direct and coordinator paths, smooth under the 100 Hz tick.


dimos/robot/booster/k1/connection.py
```diff
@@ -0,0 +1,344 @@
...
+    """Low-level wrapper around booster-rpc; the Module never touches the SDK directly.
+
+    booster-rpc's ``move`` is a synchronous gRPC call (~58/sec ceiling), so calling it
+    from a high-rate publisher (the 100 Hz coordinator) backs up. ``move()`` is therefore
+    non-blocking: it records the latest command, and ``_sender_loop`` issues the actual
+    gRPC call at ``send_hz``, always sending the latest value (stale commands dropped).
+    """
+
+    cmd_vel_timeout = 0.5  # dead-man: send zero if no new command within this window (s)
+    send_hz = 30.0  # command rate to the robot, kept under the ~58/sec move ceiling
```
mustafab0: Magic numers should be defined as constants at the top

dimos/robot/booster/k1/connection.py
```diff
@@ -0,0 +1,344 @@
...
+
+
+class BoosterRPCConnection:
+    """Low-level wrapper around booster-rpc; the Module never touches the SDK directly.
+
+    booster-rpc's ``move`` is a synchronous gRPC call (~58/sec ceiling), so calling it
+    from a high-rate publisher (the 100 Hz coordinator) backs up. ``move()`` is therefore
+    non-blocking: it records the latest command, and ``_sender_loop`` issues the actual
+    gRPC call at ``send_hz``, always sending the latest value (stale commands dropped).
+    """
```
mustafab0: There are 2 `connection.py` in the unitree folder. 

One specific just for the webRTC connection similar to the `BoosterRPCConnection` you have here.

There is another `connection.py` in the go2/g1 folder which are Connection Module implementation for the robot.

We should refactor this file to separate out the `BoosterRPCConnection` and the `K1Connection` module.

This helps us use the `booset/connection.py` for both K1 and the T1 robot both

mustafab0 (changes requested):

dimos/robot/booster/k1/connection.py
```diff
@@ -0,0 +1,332 @@
...
+            y: Left/right velocity (m/s)
+            yaw: Rotational velocity (rad/s)
+            duration: How long to move (seconds); 0 = continuous until the next command
+        """
+        twist = Twist(linear=Vector3(x, y, 0.0), angular=Vector3(0.0, 0.0, yaw))
+        if not self.move(twist, duration=duration):
+            return "Failed to move."
+        if duration > 0:
+            return f"Moved at velocity=({x}, {y}, {yaw}) for {duration}s then stopped."
+        return f"Moving at velocity=({x}, {y}, {yaw}) continuously; send a stop command to halt."
```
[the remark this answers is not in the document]
KrishnaH96: Fixed in f803a66 — walk() now requires a positive duration and returns "moved for Ns then stopped" semantics; the continuous-motion claim is gone.

dimos/robot/booster/k1/connection.py
```diff
@@ -0,0 +1,332 @@
...
+            with self._lock:
+                self._conn.change_mode(RobotMode.PREPARE)
+            logger.info("K1 mode -> PREPARE")
+            time.sleep(3)
+        with self._lock:
+            self._conn.change_mode(RobotMode.WALKING)
+        logger.info("K1 mode -> WALKING")
+        time.sleep(3)
+        with self._lock:
+            return bool(self._conn.get_mode() == RobotMode.WALKING)
```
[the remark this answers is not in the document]
KrishnaH96: Fixed in f803a66 — standup() now refuses any mode outside {WALKING, DAMPING, PREPARE} with a warning and returns False instead of forcing WALKING. The PREPARE→WALKING path was exercised on hardware during re-validation.

dimos/robot/booster/k1/connection.py
```diff
@@ -0,0 +1,332 @@
...
+
+
+def make_connection(ip: str, cfg: GlobalConfig) -> BoosterRPCConnection:
+    # Real hardware only; add sim/replay branches (keyed off cfg) here when they exist.
+    return BoosterRPCConnection(ip)
+
+
+def _camera_info_static() -> CameraInfo:
+    # TODO: replace with measured K1 camera intrinsics (these are placeholders).
+    fx, fy, cx, cy = (400.0, 400.0, 272.0, 153.0)
```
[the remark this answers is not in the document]
KrishnaH96: Addressed in f803a66 with a startup WARNING so operators know projections are inaccurate. Measured calibration is still a TODO — placeholders stay the default until we calibrate the K1 camera.

dimos/robot/booster/k1/connection.py
KrishnaH96: Done in 3be5426 — camera intrinsics are constants at the top of k1/connection.py; transport tunables (SEND_HZ, CMD_VEL_TIMEOUT_S, MODE_TRANSITION_SETTLE_S) at the top of booster_rpc.py.

dimos/robot/booster/k1/connection.py
KrishnaH96: Done in 3be5426 — BoosterRPCConnection moved to robot-agnostic booster/booster_rpc.py so a T1 module can reuse it; K1Connection stays in k1/connection.py. Named it booster_rpc.py (for the transport) mirroring unitree_webrtc.py, rather than connection.py.

dimos/robot/booster/k1/connection.py
```diff
@@ -0,0 +1,196 @@
...
+
+        logger.warning(
+            "K1 camera intrinsics are placeholders; 3D projection/perception will be "
+            "inaccurate until replaced with a measured calibration."
+        )
+        self._camera_info_thread = Thread(target=self._publish_camera_info, daemon=True)
+        self._camera_info_thread.start()
+
+        # Arm the robot so it accepts velocity commands.
+        self.standup()
```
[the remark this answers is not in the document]
KrishnaH96: Fixed in fd425ff — start() now checks the result and logs a clear error that velocity commands will be dropped, with recovery instructions. Deliberately keeps the module (and camera) running rather than failing startup: the camera is exactly what you want alive to diagnose why the robot didn't arm.

dimos/robot/booster/k1/connection.py
```diff
@@ -0,0 +1,201 @@
...
+        self._camera_info_thread = Thread(target=self._publish_camera_info, daemon=True)
+        self._camera_info_thread.start()
+
+        # Arm the robot so it accepts velocity commands. On failure keep the module
+        # (and camera) running for diagnosis, but say loudly that moves will be dropped.
+        if not self.standup():
+            logger.error(
+                "K1 did not reach WALKING on start; velocity commands will be dropped. "
+                "Resolve the robot's mode, then call the `standup` RPC (or `stand` skill)."
+            )
```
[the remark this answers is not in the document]
KrishnaH96: Deliberate — the robot can drop out of WALKING at any time mid-session (we hit exactly this during hardware validation: fault → PREPARE while teleop was live), so "running but not armed" is a state the module must tolerate regardless. Failing startup wouldn't make that state unreachable, just unrecoverable without a process restart. Keeping the module up preserves the recovery path: camera to see the robot's state, standup RPC to re-arm once cleared. The condition is loudly logged at start and every rejected move is logged by the sender.

dimos/robot/booster/booster_rpc.py
```diff
@@ -0,0 +1,186 @@
...
+        now = time.monotonic()
+        with self._cmd_lock:
+            self._latest = (twist.linear.x, twist.linear.y, twist.angular.z)
+            self._deadline = now + (duration if duration > 0 else self.cmd_vel_timeout)
+        if duration > 0:
+            # Discrete "move for N seconds then stop" (the walk skill): block, then expire.
+            time.sleep(duration)
+            with self._cmd_lock:
+                self._latest = (0.0, 0.0, 0.0)
+                self._deadline = time.monotonic()
```
mustafab0: This is not necessary. This makes the move command blocking for `N seconds` 

We don't want this to block a continuous loop to block, or it may block camera streaming and other commands coming to the robot. 

This logic can live at a higher level. not needed here. Please remove

mustafab0 (changes requested):

dimos/robot/booster/booster_rpc.py
```diff
@@ -0,0 +1,186 @@
...
+            with self._lock:
+                self._conn.change_mode(RobotMode.PREPARE)
+            logger.info("Booster mode -> PREPARE")
+            time.sleep(MODE_TRANSITION_SETTLE_S)
+        with self._lock:
+            self._conn.change_mode(RobotMode.WALKING)
+        logger.info("Booster mode -> WALKING")
+        time.sleep(MODE_TRANSITION_SETTLE_S)
+        with self._lock:
+            return bool(self._conn.get_mode() == RobotMode.WALKING)
```
mustafab0: This method should be broken in 2 methods 

`_arm()` and `standup`

one method should ont do so many things like stepping state machine

dimos/robot/booster/k1/blueprints/basic/booster_k1_coordinator.py
```diff
@@ -0,0 +1,90 @@
...
+            tasks=[
+                TaskConfig(
+                    name="vel_booster_k1",
+                    type="velocity",
+                    joint_names=_k1_joints,
+                    priority=10,
+                ),
+            ],
+        ),
+    )
```
mustafab0, line 68: 👍

pyproject.toml
```diff
@@ -258,6 +258,12 @@ unitree-dds = [
     "mcap>=1.2.0",  # decode Go2 DDS mcap recordings (go2 dds store)
 ]
 
+booster = [
```
paul-nechifor, line 272: Need to add to `all`

dimos/robot/booster/k1/blueprints/basic/booster_k1_basic.py
```diff
@@ -0,0 +1,84 @@
...
+from dimos.core.coordination.blueprints import TransportSpec, autoconnect
+from dimos.core.global_config import global_config
+from dimos.core.stream import Transport
+from dimos.core.transport import pSHMTransport
+from dimos.msgs.sensor_msgs.Image import Image
+from dimos.robot.booster.k1.connection import K1Connection
+from dimos.visualization.vis_module import vis_module
+
+# High-bandwidth camera frames go over shared memory (esp. needed on macOS UDP).
+_mac_transports: dict[tuple[str, type], TransportSpec | Transport[Any]] = {
```
paul-nechifor: Is this still the case now that we default to Zenoh on Mac?

dimos/robot/booster/k1/blueprints/basic/booster_k1_basic.py
```diff
@@ -0,0 +1,84 @@
...
+    ("color_image", Image): pSHMTransport(
+        "color_image", default_capacity=DEFAULT_CAPACITY_COLOR_IMAGE
+    ),
+}
+
+_transports_base = (
```
paul-nechifor: I think this was copied from `unitree_g1_primitive_no_nav.py` before we had merged the Zenoh PR. It no longer exists there now that we use Zenoh on Mac.

dimos/robot/booster/k1/blueprints/basic/booster_k1_coordinator.py
```diff
@@ -0,0 +1,90 @@
...
+            (WebsocketVisModule, "tele_cmd_vel", "cmd_vel"),
+            (RerunWebSocketServer, "tele_cmd_vel", "cmd_vel"),
+        ]
+    )
+    .transports(
+        {
+            ("cmd_vel", Twist): LCMTransport("/cmd_vel", Twist),
+            ("twist_command", Twist): LCMTransport("/cmd_vel", Twist),
+            ("k1_cmd_vel", Twist): LCMTransport("/booster_k1/cmd_vel", Twist),
+            ("joint_state", JointState): LCMTransport("/coordinator/joint_state", JointState),
```
paul-nechifor: What does this refer to? I thought I had converted all of them to `coordinator_joint_state`.

dimos/robot/booster/booster_rpc.py
KrishnaH96: Removed in febe0118b - the transport's move() no longer takes a duration and never blocks; it's just latest-command + dead-man now. The timed-move logic lives in the walk() skill, which refreshes the command inside the dead-man window and sends an explicit stop.

dimos/robot/booster/booster_rpc.py
KrishnaH96: Done in b23c662fd - standup() keeps only the mode decision; _arm() steps the DAMPING → PREPARE → WALKING transitions.

pyproject.toml
KrishnaH96: Added in 839800650.

dimos/robot/booster/k1/blueprints/basic/booster_k1_basic.py
KrishnaH96: You're right - removed in 839800650. It was copied from the pre-Zenoh g1 blueprint; with Zenoh the default on Mac the SHM workaround is obsolete, so the platform branch and _mac_transports are gone.

dimos/robot/booster/k1/blueprints/basic/booster_k1_coordinator.py
KrishnaH96: Dead wiring from before the coordinator_joint_state rename - nothing publishes under that name. Removed in 839800650.

dimos/robot/booster/test_booster_rpc.py
```diff
@@ -0,0 +1,127 @@
...
+    c._sender_stop.set()
+    if c._sender_thread is not None:
+        c._sender_thread.join(timeout=1.0)
+
+
+def _run_sender(c: BoosterRPCConnection) -> None:
+    """Start only the fixed-rate sender thread (skip the asyncio video loop)."""
+    c._sender_stop.clear()
+    c._sender_thread = threading.Thread(target=c._sender_loop, daemon=True)
+    c._sender_thread.start()
```
paul-nechifor: Don't start threads like this. If the test fails they never get stopped. Threads should be started in a fixture so you can handle shutdown or with try-finally.

But in this case, _run_sender and _stop_sender basically do what BoosterRPCConnection should be doing.

Test code should be exercising source code. If you're duplicating source code in the tests and testing that, how do you know the source code works?

dimos/robot/booster/test_booster_rpc.py
```diff
@@ -0,0 +1,127 @@
...
+
+@pytest.fixture
+def conn():
+    """A BoosterRPCConnection with the gRPC SDK patched out (`_conn` is a mock)."""
+    with patch("dimos.robot.booster.booster_rpc.BoosterConnection"):
+        c = BoosterRPCConnection(ip="mock")
+    yield c
+    c._sender_stop.set()
+    if c._sender_thread is not None:
+        c._sender_thread.join(timeout=1.0)
```
paul-nechifor: This is the same as _stop_sender. But neither place should have that because this is all code copied from BoosterRPCConnection.stop. You should be calling that, not duplicating the stop code here.

dimos/robot/booster/booster_rpc.py
```diff
@@ -0,0 +1,183 @@
...
+    booster-rpc's ``move`` is a synchronous gRPC call with a ~58/sec ceiling, so a
+    high-rate publisher (the 100 Hz coordinator) would back it up. ``move()`` is
+    therefore non-blocking: it records the latest command, and ``_sender_loop`` issues
+    the gRPC call at ``send_hz``, always sending the latest value (stale ones dropped).
+    """
+
+    def __init__(self, ip: str) -> None:
+        self._conn = BoosterConnection(ip=ip)
+        self._lock = Lock()  # serialize gRPC calls to the connection
+        self._loop = asyncio.new_event_loop()
```
paul-nechifor: Why create a new loop? You have one in K1Connection. Use that one.

dimos/robot/booster/booster_rpc.py
```diff
@@ -0,0 +1,183 @@
...
+        self._sender_thread: Thread | None = None
+        self._sender_stop = Event()
+        self.cmd_vel_timeout = CMD_VEL_TIMEOUT_S
+        self.send_hz = SEND_HZ
+
+    def start(self) -> None:
+        self._thread = Thread(target=self._run_loop, daemon=True)
+        self._thread.start()
+        self._sender_stop.clear()
+        self._sender_thread = Thread(target=self._sender_loop, daemon=True)
```
paul-nechifor: Don't create another thread for the sender loop. Just run the whole thing in the K1Connection loop (`K1Connection._loop`, pass it here and use it, same as the other comment).

dimos/robot/booster/booster_rpc.py
```diff
@@ -0,0 +1,183 @@
...
+from dimos.msgs.geometry_msgs.Twist import Twist
+from dimos.msgs.sensor_msgs.Image import Image, ImageFormat
+from dimos.utils.logging_config import setup_logger
+from dimos.utils.reactive import backpressure
+
+logger = setup_logger()
+
+SEND_HZ = 30.0  # gRPC send rate to the robot, kept under booster-rpc's ~58/sec move ceiling
+CMD_VEL_TIMEOUT_S = 0.5  # dead-man: send one zero if no new command arrives within this window
+MODE_TRANSITION_SETTLE_S = 3.0  # settle time after each DAMPING/PREPARE/WALKING change
```
paul-nechifor: How did you derive this number? Is there a way to check in the API if the transition has happened. It would be better to check than to just assume that a transition has happened after a certain number of seconds.

dimos/robot/booster/k1/connection.py
```diff
@@ -0,0 +1,207 @@
...
+    def rerun_views(cls):  # type: ignore[no-untyped-def]
+        """Rerun view blueprint for the K1 camera."""
+        return [
+            rrb.Spatial2DView(name="Camera", origin="world/robot/camera/rgb"),
+        ]
+
+    def __init__(self, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self._stop_event = Event()
+        self.hw = make_connection(self.config.ip, self.config.g)
```
paul-nechifor: Should start with `_`, but is `hw` really the best name? Maybe call it `self._connection`?

dimos/robot/booster/k1/connection.py
```diff
@@ -0,0 +1,207 @@
...
+    @rpc
+    def start(self) -> None:
+        super().start()
+        self.hw.start()
+
+        def on_image(image: Image) -> None:  # publish AND cache for observe()
+            self.color_image.publish(image)
+            self._latest_frame = image
+
+        self.register_disposable(self.hw.camera_stream().subscribe(on_image))
```
paul-nechifor: Much of this would be simpler if it used async modules (see in docs/).

Here, you're using an observable which runs:

```python
        self._video_future = asyncio.run_coroutine_threadsafe(
            self._conn.stream_video(on_jpeg), self._loop
        )
```

If you used async modules, you could do something like:

```python
self._video_future = self.spawn(self.hw.run_camera())
```

And use this in `hw`:

```python
    async def run_camera(self) -> None:
        def on_jpeg(jpeg: bytes) -> None:
            arr = cv2.imdecode(np.frombuffer(jpeg, dtype=np.uint8), cv2.IMREAD_COLOR)
            if arr is None:
                return
            self._frames.on_next(
                Image.from_numpy(arr, format=ImageFormat.BGR, frame_id="camera_optical")
            )

        await self._conn.stream_video(on_jpeg)
```

dimos/robot/booster/k1/blueprints/basic/booster_k1_basic.py
KrishnaH96: Right, removed in 8398006. Obsolete now that Zenoh is the Mac default.

dimos/robot/booster/test_booster_rpc.py
KrishnaH96: Done in ddccb58f. The helpers are gone. Tests start the production run_sender() as an asyncio task from a fixture and tear down through the production stop() in a finally block, so a failing test still cleans up.

dimos/robot/booster/test_booster_rpc.py
KrishnaH96: Removed in ddccb58f. Teardown calls the real stop().

dimos/robot/booster/booster_rpc.py
KrishnaH96: Removed in b66a9fb7. The connection owns no event loop and no threads. It exposes run_sender() and run_camera() coroutines and the Module spawns them on its own loop.

dimos/robot/booster/booster_rpc.py
KrishnaH96: Done in b66a9fb7. The sender is a coroutine on the module loop. Each blocking gRPC send goes through asyncio.to_thread so a stalled call cannot freeze the camera sharing the loop.

dimos/robot/booster/booster_rpc.py
KrishnaH96: This constant is gone. Startup no longer commands mode transitions at all: the operator arms the robot with the Booster app before launch, the same convention as the G1, and start() only verifies the mode. The standup RPC that remains checks each transition with get_mode() and then waits 5 seconds, which is the 3 to 4 seconds the physical motion was observed to take on hardware after the mode flag flips, plus margin, since the API reports the mode change but not motion completion.

dimos/robot/booster/k1/connection.py
KrishnaH96: Renamed to self._connection in b66a9fb7.

dimos/robot/booster/k1/connection.py
KrishnaH96: Done in b66a9fb7, exactly this shape. run_camera() lives on the connection and start() spawns it with self.spawn(), which also surfaces stream exceptions the discarded run_coroutine_threadsafe future used to swallow. That already paid off on hardware: a camera websocket drop that would have died silently was logged.

dimos/robot/booster/k1/connection.py
```diff
@@ -0,0 +1,210 @@
...
+            return "Specify a positive duration (seconds); compute it from the distance and speed."
+        twist = Twist(linear=Vector3(x, y, 0.0), angular=Vector3(0.0, 0.0, yaw))
+        deadline = time.monotonic() + duration
+        while time.monotonic() < deadline:
+            self.move(twist)
+            time.sleep(CMD_REFRESH_S)
+        self.move(Twist(linear=Vector3(0.0, 0.0, 0.0), angular=Vector3(0.0, 0.0, 0.0)))
+        if self._connection.send_failed:
+            return "The robot rejected the move commands; check that it is armed (mode WALKING)."
+        return f"Moved at velocity=({x}, {y}, {yaw}) for {duration}s then stopped."
```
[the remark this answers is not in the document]
KrishnaH96: Fixed in 9abe7a84 and 8946763a. walk() detects rejected sends mid-loop, owns its final zero in a finally block, confirms delivery of that specific zero, and separately requires at least one accepted nonzero send before claiming the robot moved. It also rejects non-finite inputs and clamps velocity and duration to a validated envelope. The public skill is tested end to end through a real K1Connection with the transport mocked.

dimos/robot/booster/booster_rpc.py
```diff
@@ -0,0 +1,189 @@
...
+            # The robot rejects moves outside a locomotion mode ("Failed to move: code = 100").
+            self._send_failed = True
+            logger.warning("Booster move failed: %s: %s", type(e).__name__, e)
+
+    def standup(self) -> bool:
+        """Arm the robot for walking; no-op if already WALKING.
+
+        Refuses modes outside {WALKING, DAMPING, PREPARE} rather than forcing an unsafe transition.
+        """
+        mode = self._get_mode()
```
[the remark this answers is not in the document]
KrishnaH96: Fixed in 9abe7a84. get_mode, change_mode, and sit contain both grpc.RpcError and the SDK's RuntimeError on OPERATION_FAIL, returning False. start() keeps its camera-active state through any vendor failure. Both failure types are covered by tests.

dimos/robot/booster/booster_rpc.py
```diff
@@ -0,0 +1,227 @@
...
+        finally:
+            self._send_count += 1
+
+    def confirm_stop(self, timeout: float = 1.0) -> bool:
+        """Block until the queued zero command has been transmitted. False if rejected or unsent."""
+        target = self._send_count + 2
+        deadline = time.monotonic() + timeout
+        while time.monotonic() < deadline:
+            if self._send_count >= target:
+                return not self._send_failed
```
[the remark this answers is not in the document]
KrishnaH96: Fixed in 9abe7a84. confirm_stop() confirms only a transmitted, accepted zero. Sends of any other command never satisfy it. A regression test simulates a concurrent publisher overwriting the stop and asserts it is reported unconfirmed.

dimos/robot/booster/k1/connection.py
```diff
@@ -0,0 +1,255 @@
...
+        duration = min(duration, WALK_MAX_DURATION_S)
+        twist = Twist(linear=Vector3(x, y, 0.0), angular=Vector3(0.0, 0.0, yaw))
+        zero = Twist(linear=Vector3(0.0, 0.0, 0.0), angular=Vector3(0.0, 0.0, 0.0))
+        deadline = time.monotonic() + duration
+        rejected = False
+        try:
+            while time.monotonic() < deadline:
+                self.move(twist)
+                time.sleep(CMD_REFRESH_S)
+                if self._connection.send_failed:
```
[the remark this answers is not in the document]
KrishnaH96: Fixed in 8946763a. walk() records a send baseline and reports motion only if a nonzero command was transmitted and accepted after it. A sender stalled through the whole walk reports that no movement command was delivered. Tested at transport and skill level.

dimos/robot/booster/booster_rpc.py
```diff
@@ -0,0 +1,263 @@
...
+    @property
+    def send_count(self) -> int:
+        with self._cmd_lock:
+            return self._send_count
+
+    def motion_delivered_since(self, baseline: int) -> bool:
+        """True if a nonzero command was transmitted and accepted after `baseline`."""
+        with self._cmd_lock:
+            return self._last_motion_ok > baseline
+
```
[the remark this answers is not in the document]
KrishnaH96: Concurrent-writer arbitration is the ControlCoordinator's job, not the adapter's. walk() is documented as sole-writer, and its tested guarantees are exact: command submitted, a movement delivered, stop delivered.

dimos/robot/booster/test_booster_rpc.py
```diff
@@ -0,0 +1,303 @@
...
+        with patch("dimos.robot.booster.k1.connection.make_connection", return_value=conn):
+            k1 = K1Connection(ip="mock")
+        request.addfinalizer(k1._close_module)
+        return k1, conn
+
+    return build
+
+
+class TestWalkSkill:
+    pass
```
paul-nechifor:
```suggestion
```

dimos/robot/booster/k1/connection.py
```diff
@@ -0,0 +1,259 @@
...
+        """
+        if not all(math.isfinite(v) for v in (x, y, yaw, duration)):
+            return "All arguments must be finite numbers."
+        if duration <= 0:
+            return "Specify a positive duration (seconds). Compute it from the distance and speed."
+        x, y = _clamp_linear(x, y, WALK_MAX_LINEAR_MPS)
+        yaw = _clamp(yaw, WALK_MAX_YAW_RPS)
+        duration = min(duration, WALK_MAX_DURATION_S)
+        twist = Twist(linear=Vector3(x, y, 0.0), angular=Vector3(0.0, 0.0, yaw))
+        zero = Twist(linear=Vector3(0.0, 0.0, 0.0), angular=Vector3(0.0, 0.0, 0.0))
```
paul-nechifor:
```suggestion
        zero = Twist.zero()
```

dimos/robot/booster/booster_rpc.py
```diff
@@ -0,0 +1,263 @@
...
+                with self._cmd_lock:
+                    vx, vy, vyaw = self._latest
+                    active = time.monotonic() <= self._deadline
+                if active:
+                    await asyncio.to_thread(self._send, vx, vy, vyaw)
+                elif was_active:
+                    # one dead-man stop on active->idle, then go quiet
+                    await asyncio.to_thread(self._send, 0.0, 0.0, 0.0)
+                was_active = active
+                await asyncio.sleep(period)
```
paul-nechifor: Technically, you need to account for the time it took to execute one iteration of the loop.

dimos/robot/booster/k1/connection.py
```diff
@@ -0,0 +1,259 @@
...
+            logger.error(
+                "K1 is not in WALKING mode. Velocity commands will be dropped. "
+                "Arm it with the Booster app, or call the `standup` RPC (or `stand` "
+                "skill) once the robot is clear."
+            )
+        logger.info("K1Connection started (ip=%s)", self.config.ip)
+
+    @rpc
+    def stop(self) -> None:
+        for future in (self._camera_future, self._camera_info_future):
```
paul-nechifor: Don't you need to cancel _sender_future too?

dimos/robot/booster/k1/connection.py
```diff
@@ -0,0 +1,259 @@
...
+    ip: str = Field(default_factory=lambda m: m["g"].robot_ip)
+
+
+def make_connection(ip: str, cfg: GlobalConfig) -> BoosterRPCConnection:
+    # cfg reserved for future sim/replay backends.
+    return BoosterRPCConnection(ip)
+
+
+def _camera_info_static() -> CameraInfo:
+    return CameraInfo(
```
paul-nechifor: Please remove the duplication by using CameraInfo.from_intrinsics or CameraInfo.from_fov.

dimos/robot/booster/k1/connection.py
```diff
@@ -0,0 +1,259 @@
...
+
+        # The operator arms the robot before launch, the G1 convention. start()
+        # never commands mode transitions.
+        if not self._connection.is_armed():
+            logger.error(
+                "K1 is not in WALKING mode. Velocity commands will be dropped. "
+                "Arm it with the Booster app, or call the `standup` RPC (or `stand` "
+                "skill) once the robot is clear."
+            )
+        logger.info("K1Connection started (ip=%s)", self.config.ip)
```
paul-nechifor:
```suggestion
```

dimos/robot/booster/booster_rpc.py
KrishnaH96: Fixed in 468218dc. The sender schedules against a tick deadline, so send time no longer stretches the period, and it resyncs after an overrun instead of burst catching up.

dimos/robot/booster/k1/connection.py
KrishnaH96: You are right, done in 468218dc. All three futures are cancelled uniformly. The transport stop still sends its own explicit final zero, so the stop delivery does not depend on the sender draining.

dimos/robot/booster/k1/connection.py
KrishnaH96: Done in 468218dc, from_intrinsics. Verified it produces identical K, P, and D values to the removed matrices.

paul-nechifor (dismissed):

dimos/robot/booster/booster_rpc.py
```diff
@@ -0,0 +1,265 @@
...
+    def _get_mode(self) -> RobotMode | None:
+        try:
+            with self._lock:
+                mode: RobotMode = self._conn.get_mode()
+            return mode
+        except SDK_ERRORS as e:
+            logger.warning("Booster get_mode failed: %s", e)
+            return None
+
+    def sit(self) -> bool:
```
mustafab0: Why is this not called lie down but sit?

dimos/robot/booster/k1/connection.py
```diff
@@ -0,0 +1,251 @@
...
+        return x, y
+    scale = limit / norm
+    return x * scale, y * scale
+
+
+class ConnectionConfig(ModuleConfig):
+    ip: str = Field(default_factory=lambda m: m["g"].robot_ip)
+
+
+def make_connection(ip: str, cfg: GlobalConfig) -> BoosterRPCConnection:
```
mustafab0: We are retiring global config. IP should be in module config

dimos/robot/booster/booster_rpc.py
KrishnaH96: Renamed in 09c1b0c. The transport and RPC are liedown, matching the SDK call and the Go2, and the agent skill is lie_down. sit is gone.

dimos/robot/booster/k1/connection.py
KrishnaH96: Done in 09c1b0c. make_connection takes only the ip from module config, the GlobalConfig parameter is gone.

mustafab0 (dismissed):

paul-nechifor (approved):

dimos/robot/booster/k1/connection.py
```diff
@@ -0,0 +1,250 @@
...
+            self.move(zero)
+        stopped = self._connection.confirm_stop()
+        moved = self._connection.motion_delivered_since(baseline)
+        if rejected:
+            return "The robot rejected the move commands. Check that it is armed (mode WALKING)."
+        if not moved:
+            return "Could not confirm any movement command was delivered. The robot may not have moved."
+        if not stopped:
+            return "Sent movement commands but could not confirm the stop. Verify the robot halted."
+        return f"Moved at velocity=({x}, {y}, {yaw}) for {duration}s then stopped."
```
mustafab0, line 228:
```suggestion
     @skill
    def walk(self, x: float, y: float = 0.0, yaw: float = 0.0) -> str:
        x, y = _clamp_linear(x, y, WALK_MAX_LINEAR_MPS)
        yaw = _clamp(yaw, WALK_MAX_YAW_RPS)
        if not self.move(Twist(linear=Vector3(x, y, 0.0), angular=Vector3(0.0, 0.0, yaw))):
            return "The robot rejected the move command. Check that it is armed (mode WALKING)."
        return f"Now moving at velocity=({x}, {y}, {yaw})."
```

The walk skill is a blocking skill and the connection module should not be doing anything blocking. it is safety critical. The agent cant block the whole connection for 30s.

dimos/robot/booster/booster_rpc.py
```diff
@@ -0,0 +1,265 @@
...
+            )
+
+        await self._conn.stream_video(on_jpeg)
+
+    def move(self, twist: Twist) -> bool:
+        # DimOS Twist (SI, body frame: +x fwd, +y left, +z yaw CCW) -> booster (vx, vy, vyaw).
+        with self._cmd_lock:
+            self._latest = (twist.linear.x, twist.linear.y, twist.angular.z)
+            self._deadline = time.monotonic() + self.cmd_vel_timeout
+        return True
```
mustafab0, line 118: We don't need this to return a bool. Better to just return None here


```suggestion
    def move(self, twist: Twist) -> None:
        # DimOS Twist (SI, body frame: +x fwd, +y left, +z yaw CCW) -> booster (vx, vy, vyaw).
        with self._cmd_lock:
            self._latest = (twist.linear.x, twist.linear.y, twist.angular.z)
            self._deadline = time.monotonic() + self.cmd_vel_timeout
```

dimos/robot/booster/k1/connection.py
mustafab0: An RPC can block for a while, because we can be using RPC calls for robot FSM changes.

dimos/robot/booster/booster_rpc.py
```diff
@@ -0,0 +1,265 @@
...
+
+        def on_jpeg(jpeg: bytes) -> None:
+            arr = cv2.imdecode(np.frombuffer(jpeg, dtype=np.uint8), cv2.IMREAD_COLOR)
+            if arr is None:
+                return
+            self._frames.on_next(
+                Image.from_numpy(arr, format=ImageFormat.BGR, frame_id="camera_optical")
+            )
+
+        await self._conn.stream_video(on_jpeg)
```
mustafab0, line 111: This should retry establishing video stream if video stream breaks for any reason

mustafab0 (changes requested):

PR #2527, opened 2026-06-18
Add Transport.spec

Dreamsorcerer:
This avoids instantiating Transports in the Blueprint.transports() method. Instead we pass a .spec(), and instantiation happens later. This allows config etc. to be passed in at construction time cleanly.

### DIFF
--- a/dimos/control/blueprints/basic.py
+++ b/dimos/control/blueprints/basic.py
@@ -58,7 +58,7 @@ coordinator_basic = ControlCoordinator.blueprint(
     joint_state_frame_id="coordinator",
 ).transports(
     {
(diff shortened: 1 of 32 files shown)


dimos/control/blueprints/basic.py
```diff
@@ -58,7 +58,7 @@ def _mujoco_if_sim(sim_path: str, dof: int) -> tuple[Blueprint, ...]:
     joint_state_frame_id="coordinator",
 ).transports(
     {
-        ("joint_state", JointState): LCMTransport("/coordinator/joint_state", JointState),
+        ("joint_state", JointState): LCMTransport.spec("/coordinator/joint_state", JointState),
```
paul-nechifor, line 61: I think you need to rebase. I've removed these `.transports`.

PR #2528, opened 2026-06-18
WIP chore: package robot description assets

TomCC7:
## Summary
- ship supported robot descriptions from `dimos/robot/descriptions` instead of Git LFS archives
- add `robot_description_path()` and update built-in catalogs/adapters/examples/docs to use package paths
- update package metadata and OpenSpec artifacts for packaged robot descriptions

## Wheel size
- Built wheel: `dist/dimos-0.0.12-cp312-cp312-linux_x86_64.whl`
- Size: 50,845,144 bytes (48.49 MiB)

## Notes
- branch push required `lfs.allowincompletepush=true` because unrelated historical LFS objects already missing upstream are not present locally; final tree removes only the four robot-description LFS archives.


dimos/robot/descriptions/xarm_description/urdf/xarm6/xarm6.urdf
```diff
@@ -0,0 +1,408 @@
+<?xml version="1.0" ?>
+<!-- =================================================================================== -->
+<!-- |    This document was autogenerated by xacro from /home/ruthwik/Documents/dimos/data/xarm_description/urdf/xarm_device.urdf.xacro | -->
```
paul-nechifor, line 3: If it was generated, why do we need to commit it? Also don't include user info like this.

TomCC7: hahaha these are unzipped directly from the tgz in lfs. I think that was to make loading easier (bypassing xacro generation). but sure we should clean up. maybe first you can verify if the overall direction is ok?

PR #2532, opened 2026-06-19
Add Python unit test skill

TomCC7:
## Summary
- Add a model-invoked Python unit test skill at `.agents/skills/python-unit-tests/SKILL.md`.
- Guide agents toward hermetic pytest tests: behavior-focused assertions, fixture cleanup and reuse, stable mocks, scoped validation, and the Paul reviewer self-check.
- Add `.slim/deepwork/` to `.gitignore` so local deepwork state stays out of commits.

## Testing
- Not run (skill/documentation-only change).
- Validated the skill frontmatter shape locally.

### DIFF
--- a/.agents/skills/python-unit-tests/SKILL.md
+++ b/.agents/skills/python-unit-tests/SKILL.md
@@ -1,11 +1,11 @@
 ---
-name: dimos-unit-testing
-description: DimOS unit tests: Use when writing, fixing, or reviewing pytest tests, fixtures, mocks, or test PR feedback.
+name: python-unit-tests
+description: Python unit tests: Use when writing, fixing, or reviewing pytest tests, fixtures, mocks, or test PR feedback.
 ---
 
-# DimOS Unit Testing
+# Python Unit Tests
 
-Use this skill before adding, changing, or reviewing DimOS Python tests. The goal is **hermetic** tests: behavior-focused, deterministic, isolated, and cheap to run.
+Use this skill before adding, changing, or reviewing Python unit tests. The goal is **hermetic** tests: behavior-focused, deterministic, isolated, and cheap to run.
 
 Consult these only when the branch needs more context:
 
--- a/.agents/skills/python-unit-tests/SKILL.md
+++ b/.agents/skills/python-unit-tests/SKILL.md
@@ -90,7 +90,7 @@ Do not run mypy for test-only edits. Run `uv run mypy` when production source ty
 
 ## Final self-check
 
-Before stopping, assume a strict reviewer will read every changed test and comment on each avoidable mistake.
+Before stopping, summon Paul: assume he will review every changed test and comment on each avoidable mistake.
 
 - Is the test next to the code it tests?
 - Does it prove behavior with real assertions?
--- a/.agents/skills/python-unit-tests/SKILL.md
+++ b/.agents/skills/python-unit-tests/SKILL.md
@@ -17,10 +17,10 @@ Consult these only when the branch needs more context:
 ## Steps
 
 1. **Scope the behavior.** Identify the code under test, the caller-visible outcome, and the smallest test file next to it: `dimos/core/foo.py` gets `dimos/core/test_foo.py`. Completion: every new or changed test has a named behavior target and lives beside the code it covers.
-2. **Build a hermetic setup.** Use module-level imports, Arrange-Act-Assert structure, fixtures for shared or resource-owning setup, and context managers for one-test local resources. Completion: setup owns all cleanup, restores global state, and contains no fixed sleeps.
+2. **Build a hermetic setup.** Use module-level imports, Arrange-Act-Assert structure, fixtures for shared or resource-owning setup, and context managers for one-test local resources. Completion: setup owns all cleanup, restores global state, reuses existing fixtures when they match, and contains no fixed sleeps.
 3. **Assert the contract.** Use small examples and exact expected values. Completion: every test has an unconditional assertion that proves behavior a caller depends on.
 4. **Mock the boundary.** Mock only slow, nondeterministic, or external boundaries. Completion: patches use `mocker.patch`, `mocker.patch.object`, or `monkeypatch`; no direct method assignment or `__new__` fixture shells remain.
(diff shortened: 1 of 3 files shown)


.agents/skills/python-unit-tests/SKILL.md
```diff
@@ -0,0 +1,118 @@
...
+- Prefer `assert result == expected` over shape-only checks.
+- Do not add no-value tests that only prove a dataclass stored constructor arguments or that a default equals itself.
+
+## Fixtures and cleanup
+
+- Use `yield` fixtures so teardown runs even when assertions fail.
+- Use `tmp_path` for temporary files and directories.
+- Clean up modules, stores, servers, transports, subscriptions, sessions, threads, subprocesses, and global state.
+- Do not put `stop()`, `close()`, or cleanup calls at the end of a test body after assertions; an earlier failure skips them.
+- If a resource supports a context manager and the setup is local to one test, use `with`.
```
paul-nechifor, line 41:
```suggestion
- Prefer to do cleanup in fixtures. That is, handle resource setup and tear down in the same place. That ensures a resource is released even when the test fails.
- If a resource supports a context manager and the setup is local to one test, you can use `with`.
- In rare cases it might be useful to do the setup and tear down all in the test, but you should always use try-finally in that case so that resources are released even on failure.
```

.agents/skills/python-unit-tests/SKILL.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Assertions
+
+- Every test needs a meaningful assertion.
```
paul-nechifor, line 46:
```suggestion
- Every test needs a meaningful assertion.
- Do not over-assert. Assert the things which are relevant to test. Do not assert things which are purely incidental to what is being tested.
```

.agents/skills/python-unit-tests/SKILL.md
```diff
@@ -0,0 +1,118 @@
+---
+name: python-unit-tests
+description: Use when writing, fixing, or reviewing Python pytest unit tests, fixtures, mocks, or test PR feedback.
+---
```
paul-nechifor, line 4: I'd mention something about code deduplication.

When adding a fixture of some kind make sure it doesn't already exist in the codebase. If it does, try to extract it to a common place and reuse it. But don't force it if the resemblance is only partial.

.agents/skills/python-unit-tests/SKILL.md
```diff
@@ -0,0 +1,118 @@
...
+- Prefer `mocker.patch(...)` or `mocker.patch.object(...)` for patches and call assertions.
+- Use `monkeypatch` for environment variables, paths, and module attributes that should be restored automatically.
+- Prefer real lightweight value objects over mocks for dataclasses and simple message objects.
+- Do not replace methods by direct assignment when a patch or spy gives clearer assertions and automatic cleanup.
+- Do not build fake modules out of custom classes when standard pytest mocking can express the same behavior.
+- Do not construct objects with `__new__` and fill in fields by hand. Construct normally and patch the one side effect you need to avoid.
+
+For call assertions, use the standard pattern: patch the target with `mocker.patch.object(...)`, run the behavior, then assert with `assert_called_once_with(...)` or another precise mock assertion.
+
+## Types in tests
```
paul-nechifor: We don't currently run mypy on test files. Personally, I don't write type annotations at all for test files, and `# type: ignore` isn't needed at all (doesn't even need to be fixed).

.agents/skills/python-unit-tests/SKILL.md
```diff
@@ -0,0 +1,118 @@
...
+- A rare `# type: ignore[...]` is acceptable only when narrow, justified, and caused by something mypy cannot know, such as an untyped third-party library or decorator-generated attribute.
+- Avoid `Any` unless the value can genuinely be anything and a narrower type would be dishonest.
+
+## Global state
+
+- Use `monkeypatch.setenv`, `monkeypatch.delenv`, or fixtures instead of direct `os.environ[...]` mutation.
+- Restore `global_config` or other shared state after changing it.
+- A shared fixture must leave the system in the same state for the next test.
+
+## Anti-patterns
```
paul-nechifor: All of this is already mentioned... Are you sure it's needed?

.agents/skills/python-unit-tests/SKILL.md
```diff
@@ -0,0 +1,118 @@
...
+
+```bash
+./bin/pytest-fast
+```
+
+Run `uv run mypy` when source typing changed or before a broader quality gate. Run `pre-commit run --all-files` for broad/autofix/final review work, not after every small unit-test edit.
+
+## Final self-check
+
+Before stopping, summon Paul: assume he will review every changed test and comment on each avoidable mistake.
```
paul-nechifor: Legend has it, you just have to say Paul three times and I will be summoned. 🪄

But I don't think agents need to know this. :)

TomCC7: might also gives some pressure to agents so it works harder lol

paul-nechifor (dismissed):

.agents/skills/python-unit-tests/SKILL.md
paul-nechifor: But it will also probably confuse it... Might was well remove it. :)

.agents/skills/python-unit-tests/SKILL.md
```diff
@@ -0,0 +1,101 @@
...
+
+```bash
+./bin/pytest-fast
+```
+
+Do not run mypy for test-only edits. Run `uv run mypy` when production source typing changed or before a broader quality gate. Run `pre-commit run --all-files` for broad/autofix/final review work, not after every small unit-test edit.
+
+## Final self-check
+
+Before stopping, summon Paul: assume he will review every changed test and comment on each avoidable mistake.
```
TomCC7:
```suggestion
```

.agents/skills/python-unit-tests/SKILL.md
```diff
@@ -0,0 +1,100 @@
...
+Broader local check:
+
+```bash
+./bin/pytest-fast
+```
+
+Do not run mypy for test-only edits. Run `uv run mypy` when production source typing changed or before a broader quality gate. Run `pre-commit run --all-files` for broad/autofix/final review work, not after every small unit-test edit.
+
+## Final self-check
+
```
TomCC7, line 92:
```suggestion
```

paul-nechifor (approved):

PR #2566, opened 2026-06-23
docs: reorganize Go2 platform docs + map framing default

jeff-hykin:
Move premap stuff to go2 since its go2 specific. Touch up go2 docs.


dimos/robot/unitree/go2/blueprints/smart/unitree_go2.py
```diff
@@ -58,6 +59,17 @@ class Go2Memory(Recorder):
     odom: In[PoseStamped]
     config: Go2MemoryConfig
 
+    _last_odom_pose: Pose | None = None
+
+    @pose_setter_for("odom")
+    def _odom_pose(self, msg: PoseStamped) -> Pose | None:
+        self._last_odom_pose = msg
```
paul-nechifor, line 66: This is why I'm not fond of too much magic. My understanding is that `_odom_pose` is never called directly, instead `@pose_setter_for` registers the method to be used in a subscription, so `_odom_pose` and `_lidar_pose` can be called from different threads meaning `self._last_odom_pose` is not safe to use without a lock.

If we used async we wouldn't have to worry so much about concurrent read/writes.

Also, what does `odom alwyas wins the race` mean? How can you guarantee that? Is it because odom runs at a higher frequency?

paul-nechifor (approved):

dimos/robot/unitree/go2/blueprints/smart/unitree_go2.py
jeff-hykin: Thats a great point. By coincidence I'm the author of pose_setter_for, and I agree it should register async callback to avoid races. 

Theres 3 places pose_setter_for is used so I'll make a small PR that changes all three.

PR #2589, opened 2026-06-24
fix: double start bug

paul-nechifor:
## Problem

In certain tests (`test_dimos_skills`) `lcm.start` could be called twice.

## Solution

Make `LCMService.start` idempotent.

### DIFF
--- a/dimos/protocol/service/lcmservice.py
+++ b/dimos/protocol/service/lcmservice.py
@@ -65,6 +65,7 @@ class LCMService(Service):
     _stop_event: threading.Event
     _loop_running: threading.Event
     _l_lock: threading.Lock
+    _start_lock: threading.Lock
     _thread: threading.Thread | None
     _call_thread_pool: ThreadPoolExecutor | None = None
     _call_thread_pool_lock: threading.RLock = threading.RLock()
@@ -79,6 +80,7 @@ class LCMService(Service):
             self.l = lcm_mod.LCM(self.config.url) if self.config.url else lcm_mod.LCM()
 
         self._l_lock = threading.Lock()
+        self._start_lock = threading.Lock()
         self._stop_event = threading.Event()
         self._loop_running = threading.Event()
         self._thread = None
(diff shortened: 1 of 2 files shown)


dimos/protocol/service/lcmservice.py
```diff
@@ -109,6 +109,9 @@ def __setstate__(self, state) -> None:  # type: ignore[no-untyped-def]
         self._call_thread_pool_lock = threading.RLock()
 
     def start(self) -> None:
+        if self._thread is not None and self._thread.is_alive():
+            return
```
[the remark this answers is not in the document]
paul-nechifor: Fixed.

dimos/protocol/service/lcmservice.py
```diff
@@ -105,24 +108,29 @@ def __setstate__(self, state) -> None:  # type: ignore[no-untyped-def]
         self._loop_running = threading.Event()
         self._thread = None
```
spomichter, line 109: nitpick but `_thread` set to None here and never set bacl to None in stop(), but the dead thread should suffice for the `is_alive()` check

paul-nechifor: Done.

dimos/protocol/service/lcmservice.py
```diff
@@ -143,40 +151,43 @@ def _lcm_loop(self) -> None:
...
-                self._call_thread_pool.shutdown(wait=not is_pool_thread)
-                self._call_thread_pool = None
+            if self._thread is not None:
+                # Only join if we're not the LCM thread (avoid "cannot join current thread")
+                if threading.current_thread() != self._thread:
+                    self._thread.join(timeout=DEFAULT_THREAD_JOIN_TIMEOUT)
+                    if self._thread.is_alive():
+                        logger.warning("LCM thread did not stop cleanly within timeout")
+
+            self._thread = None
```
mustafab0: This will set `_thread` to None even if `if self._thread.is_alive():` is true, right?

Shouldn't we only put it to None if we successfully verify if it is dead

spomichter (approved):

PR #2594, opened 2026-06-25
Introduce runtime scene packages for MuJoCo and Rerun

Nabla7:
## Problem

DimOS has no clean runtime concept for a simulated environment.

Until now, a "scene" was usually a loose set of backend-specific files: MuJoCo XML, meshes, browser assets, object metadata, and local path assumptions. That makes scenes hard to reuse across simulation, visualization, navigation, and future renderers.

This PR introduces **scene packages**: cooked environments with one metadata file and a stable runtime contract.

## Scene Packages

A scene package is a directory with `scene.meta.json` plus backend-specific artifacts.

The metadata describes what the package contains and where each consumer should look. A package can include:

- MuJoCo XML for physics
- optional MuJoCo `.mjb` binaries for faster loading of large scenes
- visual assets for Rerun or browser viewers
- future visual assets such as Gaussian splats
- static environment geometry
- dynamic or separately spawned entities
- object prototypes reused by many instances
- coordinate-frame and scale alignment metadata

The core rule is: runtime code resolves a scene package once, then asks the package for the artifact it needs. Consumers should not hardcode scene-specific folders.

## Runtime API

Scene inputs are resolved through the catalog/loader:

```python
from dimos.simulation.scenes.catalog import resolve_scene_package

package = resolve_scene_package(global_config.scene)
```

Supported inputs:

- `None` / `--scene none`
- named packages such as `office` or `supermarket`
- path to `scene.meta.json`
- path to a scene package directory
- path to a precompiled MuJoCo `.mjb` in consumers that support direct binary loading

Once resolved, consumers use the package object instead of knowing folder layout:

```python
if package is not None:
    mujoco_xml = package.mujoco_scene_path
    rerun_visual = package.browser_visual_path("rerun")
    browser_collision = package.browser_collision_path
    objects_json = package.objects_path
    entities = package.entities
    alignment = package.alignment
```

For example:

- MuJoCo loads `package.mujoco_scene_path`, or a direct `.mjb` path when a large scene has been precompiled.
- Rerun asks for `package.browser_visual_path("rerun")`.
- Future browser renderers can ask for their own target-specific assets.
- Future splat renderers can be added as another artifact target without changing every consumer.

## What This PR Adds

Adds the runtime scene package contract and resolver.

Adds MuJoCo loading for scene packages, including XML scenes, optional `.mjb` loading, entity metadata, and precomposed G1 demo scenes for large environments.

Adds Rerun support for scene package visuals, G1 URDF robot visualization, raycast lidar, map/costmap/path visualization, and command velocity wiring from the Rerun WASD controls.

Uses `unitree-g1-groot-wbc` as the first full integration path. The same blueprint can now run with no scene, the office scene, or the supermarket scene.

Ships the required runtime data through LFS:

- `scene_packages.tar.gz`: `dimos_office`, `supermarket`
- `g1_urdf.tar.gz`: G1 URDF and meshes for Rerun visualization

## Why This Touches Multiple Systems

The reusable piece is the scene package contract.

The extra plumbing is here because the first consumer is a real robot stack, not a metadata-only test. To prove the package works, G1 needs to spawn in the scene, see it with lidar, build a map, plan through it, show the scene and robot in Rerun, reset cleanly, and respond to velocity commands.

## Non-Goals

This PR does not add the scene cooking pipeline.

It also does not solve every large-scene problem. Large scenes may need `.mjb` artifacts for startup time, and many dynamic objects remain a MuJoCo scaling issue. The package format is designed to support richer dynamic assets, reused prototypes, browser assets, and future render targets such as splats, but this PR focuses on runtime loading and the first end-to-end G1 demo.

## How To Test

No scene:

```bash
dimos \
  --simulation mujoco \
  --scene none \
  --viewer rerun \
  --rerun-open native \
  --n-workers 12 \
  run unitree-g1-groot-wbc
```

Office:

```bash
dimos \
  --simulation mujoco \
  --scene office \
  --viewer rerun \
  --rerun-open native \
  --n-workers 12 \
  run unitree-g1-groot-wbc
```

Supermarket:

```bash
dimos \
  --simulation mujoco \
  --scene supermarket \
  --viewer rerun \
  --rerun-open native \
  --n-workers 12 \
  run unitree-g1-groot-wbc
```

Prefer headless MuJoCo with Rerun native for normal testing. `mujocosimmodule.headless=false` opens the MuJoCo viewer but can run much slower.

## Review Guide

Start with:

- `dimos/simulation/scene_assets/spec.py`
- `dimos/simulation/scenes/catalog.py`

Then review the consumers:

- MuJoCo scene loading
- Rerun scene and robot visualization
- G1 Groot blueprint integration
- LFS package updates

## Checks

Focused checks run locally across the branch included `ruff`, `py_compile`, mypy on touched Rerun scene code, commit hooks, LFS checks, and runtime smoke testing through the G1 MuJoCo/Rerun commands above.


dimos/visualization/rerun/urdf_robot.py
```diff
@@ -0,0 +1,218 @@
...
+            path = f"{parent_path}/{_rerun_path_part(node)}"
+            matrix = np.asarray(edge_data.get("matrix", np.eye(4)), dtype=float)
+            entities.append((path, _matrix_to_rerun_transform(rr, matrix)))
+            entities.append((path, _mesh_to_rerun(rr, mesh)))
+
+        return entities
+
+    def _load_robot(self) -> Any:
+        if self._robot is None:
+            from yourdfpy import URDF  # type: ignore[import-not-found]
```
paul-nechifor: `# type: ignore` is almost always a hint that you're doing something wrong. In this case I assume the issue is that you're importing `yourdfpy` even though you're not declaring it in `pyproject.toml`. I assume it's a dependency of something else. You should depend on it directly.

Edit: same for the other place you're using it.

dimos/visualization/rerun/scene_package.py
```diff
@@ -0,0 +1,112 @@
...
+        y_to_z = ((1.0, 0.0, 0.0), (0.0, 0.0, -1.0), (0.0, 1.0, 0.0))
+        matrix = _matmul(matrix, y_to_z)
+    return [list(row) for row in matrix]
+
+
+def _matmul(
+    a: Sequence[Sequence[float]],
+    b: Sequence[Sequence[float]],
+) -> list[list[float]]:
+    return [[sum(a[i][k] * b[k][j] for k in range(3)) for j in range(3)] for i in range(3)]
```
paul-nechifor:
```suggestion
    matrix = Rotation.from_euler("ZYX", alignment.rotation_zyx_deg, degrees=True).as_matrix()
    if alignment.y_up:
        matrix = matrix @ np.array([[1.0, 0.0, 0.0], [0.0, 0.0, -1.0], [0.0, 1.0, 0.0]])
    return matrix.tolist()
```

No need to rotate by hand.

Also need to import numpy and `from scipy.spatial.transform import Rotation`

dimos/control/tasks/g1_groot_wbc_task/test_g1_groot_wbc_task.py
```diff
@@ -278,6 +284,32 @@ def test_dry_run_suppresses_output_but_keeps_policy_hot(
     assert np.any(task._obs_buf != 0.0)
 
 
+def test_reset_runtime_state_clears_policy_state_and_rearms(
+    task: G1GrootWBCTask, joints_29: list[str]
+) -> None:
+    task.start()
```
paul-nechifor, line 301: Don't you need to call `stop()` too? Best to call it in the fixture.

dimos/hardware/whole_body/registry.py
```diff
@@ -71,14 +71,33 @@ def available(self) -> list[str]:
...
         import dimos.hardware.whole_body as hw_pkg
+        import dimos.simulation.adapters.whole_body as sim_pkg
 
         self._discover_in("dimos.hardware.whole_body", hw_pkg.__path__[0], max_depth=2)
+        self._discover_simulation_adapters(
+            "dimos.simulation.adapters.whole_body",
+            sim_pkg.__path__[0],
+        )
+
+    def _discover_simulation_adapters(self, pkg_path: str, dir_path: str) -> None:
```
paul-nechifor, line 89: I don't like this pattern of scanning the file system, importing anything that's found and trying to match purely on the module having a `register` attribute defined. It seems haphazardous.

The alternative isn't necessarily listing everything manually. We have the same issue for blueprints, but it's handled a bit differently. Every blueprint is listed in `dimos/robot/all_blueprints.py`. But the list is not created by hand, it's generated by `dimos/robot/test_all_blueprints_generation.py`. The test also fails if a new blueprint is added but not included in `all_blueprints.py`.

I think this is a better pattern because it gets rid of having to do things at startup. You can also add a more precise check, like verifying the signature instead of just the name `register`.

dimos/core/global_config.py
```diff
@@ -58,6 +58,7 @@ class GlobalConfig(BaseSettings):
     mujoco_global_map_from_pointcloud: str | None = None
     mujoco_start_pos: str = "-1.0, 1.0"
     mujoco_steps_per_frame: int = 7
+    scene: str | None = None
```
paul-nechifor: I'd make this a little more specific. For example `mujoco_room` does this specifically for the `mujoco` simulator and `dimsim_scene` is the `dimsim` equivalent.

dimos/robot/unitree/g1/blueprints/basic/unitree_g1_groot_wbc.py
```diff
@@ -94,43 +157,132 @@
...
             "imu_accel",
         ),
         require_imu=True,
     )
 
+    def _legacy_mujoco_backend() -> Any:
+        return MujocoSimModule.blueprint(
+            address=_MJCF_PATH,
+            headless=True,
+            dof=29,
```
paul-nechifor: Best to consolidate this under a single constant.

At this point I know how many motors G1 has better than I know my age:

```
control/tasks/g1_groot_wbc_task/test_g1_groot_wbc_task.py:183:    q = np.zeros(29, dtype=np.float32)
control/tasks/g1_groot_wbc_task/test_g1_groot_wbc_task.py:184:    dq = np.ones(29, dtype=np.float32)
control/tasks/g1_groot_wbc_task/g1_groot_wbc_task.py:57:# The 29 DDS motor names plus the per-joint kp/kd values used with the
control/tasks/g1_groot_wbc_task/g1_groot_wbc_task.py:134:# Default joint angles for all 29 G1 joints. The policy treats these as
control/tasks/g1_groot_wbc_task/g1_groot_wbc_task.py:171:_NUM_MOTORS = 29
control/tasks/g1_groot_wbc_task/g1_groot_wbc_task.py:204:        all_joint_names: All 29 coordinator joint names in DDS order
control/tasks/g1_groot_wbc_task/g1_groot_wbc_task.py:206:            the observation, which feeds all 29 joint states.
control/tasks/g1_groot_wbc_task/g1_groot_wbc_task.py:207:        default_positions_29: Default joint angles for all 29 joints
control/tasks/g1_groot_wbc_task/g1_groot_wbc_task.py:295:                f"(full 29-DOF G1), got {len(config.all_joint_names)}"
control/tasks/g1_groot_wbc_task/g1_groot_wbc_task.py:381:        """Pull current q/dq for the full 29 from ``CoordinatorState``.
control/tasks/g1_groot_wbc_task/g1_groot_wbc_task.py:383:        Updates last-known-good caches and returns True iff the full 29
control/components.py:193:    """Create joint names for a 29-DOF humanoid.
control/components.py:201:        List of 29 joint names like ["g1/left_hip_pitch", ..., "g1/right_wrist_yaw"]
hardware/whole_body/transport/adapter.py:45:        dof: int = 29,
simulation/adapters/whole_body/g1.py:49:_NUM_MOTORS = 29
simulation/engines/mujoco_shm.py:44:# (Unitree G1: 29) push higher.  32 leaves headroom while keeping all
robot/unitree/g1/wholebody_connection.py:18:motor_command (In[MotorCommandArray]). 29 motors, ordering from
robot/unitree/g1/wholebody_connection.py:53:_NUM_MOTORS = 29
robot/unitree/g1/wholebody_connection.py:54:_NUM_MOTOR_SLOTS = 35  # G1 hg LowCmd has 35 slots; only 29 are used
robot/unitree/g1/wholebody_connection.py:55:# Default mode_machine for the 29-DOF G1 gear variant. Users can override
robot/unitree/g1/wholebody_connection.py:98:        positions: Motor positions for the 29 controllable G1 motors.
robot/unitree/g1/wholebody_connection.py:99:        velocities: Motor velocities for the 29 controllable G1 motors.
robot/unitree/g1/wholebody_connection.py:100:        efforts: Estimated motor torques for the 29 controllable G1 motors.
robot/unitree/g1/assets/g1_29dof.xml:3:    Robot-only 29-DOF Unitree G1 MJCF adapted from NVIDIA
robot/unitree/g1/blueprints/basic/unitree_g1_groot_wbc.py:178:            dof=29,
robot/unitree/g1/blueprints/basic/unitree_g1_groot_wbc.py:196:                    dof=29,
robot/unitree/g1/blueprints/basic/unitree_g1_groot_wbc.py:217:                    dof=29,
robot/unitree/g1/blueprints/basic/unitree_g1_groot_wbc.py:232:                dof=29,
```

dimos/simulation/engines/test_mujoco_sim_module.py
```diff
@@ -63,3 +106,232 @@ def signal_ready(self, *, num_joints: int) -> None:
...
+
+    engine = MujocoEngine(config_path=robot_xml, headless=True, model=model)
+    assert engine.model is model
+    assert engine.root_qpos_adr == 0
+    assert any(name.endswith("hinge") for name in engine.joint_names)
+
+
+@pytest.mark.mujoco
+def test_compose_model_reuses_entity_mesh_assets(tmp_path: Path) -> None:
+    mujoco = pytest.importorskip("mujoco")
```
paul-nechifor:
```suggestion
```
Don't skip tests.

dimos/hardware/whole_body/registry.py
mustafab0: We are using this pattern everywhere because of me. 🫢 

It started with the sim engine. I implemented it in the control coordinator and I guess why we are doing similar for registering here.

I was not aware of a better way to do this. Will create a ticket to remove this pattern from the codebase.

dimos/robot/unitree/g1/blueprints/basic/unitree_g1_groot_wbc.py
```diff
@@ -94,43 +157,132 @@
...
+            return _legacy_mujoco_backend(), _MJCF_PATH
+        if package.mujoco_scene_path is None:
+            raise ValueError(f"scene package has no MuJoCo scene artifact: {package.metadata_path}")
+
+        composed_scene = _precomposed_g1_scene(package.package_dir)
+        if composed_scene is not None:
+            return (
+                MujocoSimModule.blueprint(
+                    address=composed_scene,
+                    headless=True,
```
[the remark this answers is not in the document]
Nabla7: Addressed. The G1 blueprint now looks up the composed MuJoCo binary through the scene package metadata by key, robot, and entity policy instead of checking the package directory name. If the metadata declares a binary but the file is missing, we now fail with a clear error instead of silently taking a slow path.

dimos/visualization/rerun/urdf_robot.py
Nabla7: We now declare yourdfpy in the visualization extra, but exclude Linux aarch64 because its embreex dependency does not have a usable wheel there. The import is lazy so non-Rerun/URDF paths do not require it.

dimos/visualization/rerun/scene_package.py
Nabla7: Applied this change. Oversight on my part, LLMs tend to do this.

dimos/control/tasks/g1_groot_wbc_task/test_g1_groot_wbc_task.py
Nabla7: Good point. The task fixtures now own teardown and call `stop()` in `finally`, so tests that start the task clean it up even if an assertion fails.

dimos/core/global_config.py
Nabla7: I renamed this to `scene_package` to make the intent more explicit. It is not MuJoCo-specific: the same package can be consumed by MuJoCo, Rerun, browser collision, and future scene consumers.

dimos/robot/unitree/g1/blueprints/basic/unitree_g1_groot_wbc.py
Nabla7: Handled in the new blueprint wiring with `_G1_NUM_MOTORS = len(g1_joints)`. Opened #2668 for the broader G1 cleanup outside this PR.

dimos/simulation/engines/test_mujoco_sim_module.py
Nabla7: Removed the importskip.

dimos/simulation/engines/mujoco_engine.py
```diff
@@ -312,21 +460,140 @@ def _render_cameras(self, now: float, cam_renderers: dict[str, _CameraRendererSt
...
             loop_start = time.time()
+            reset_done_event = None
+            if self._reset_requested.is_set():
+                with self._lock:
+                    self._reset_requested.clear()
+                    self._reset_unlocked()
+                    reset_done_event = self._reset_done_event
+                    self._reset_done_event = None
+                if reset_done_event is not None:
+                    reset_done_event.set()
```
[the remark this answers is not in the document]
Nabla7: Reset requests still coalesce with the latest pose winning, but `wait=True` callers now all get released when that reset is applied. Added a regression test for this.

dimos/simulation/scenes/catalog.py
```diff
@@ -0,0 +1,112 @@
...
+        known = ", ".join(sorted(_PACKAGE_DIRS))
+        raise ValueError(f"unknown scene '{scene_text}'. Known scenes: {known}")
+
+    if name == DEFAULT_SCENE:
+        return _resolve_dimos_office()
+
+    metadata_path = _scene_package_dir() / _PACKAGE_DIRS[name] / "scene.meta.json"
+    if not metadata_path.exists():
+        raise FileNotFoundError(f"scene package '{name}' is not cooked yet: {metadata_path}")
+    return load_scene_package(metadata_path)
```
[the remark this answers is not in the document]
Nabla7: This is now handled in `_resolve_named_scene_package`: all named scene packages are loaded through the same path and validated with `_validate_package_artifacts(...)` before being returned.

paul-nechifor (approved):

PR #2595, opened 2026-06-25
Feat/webrtc transport with Sam config merged on top 

spomichter:
## Problem

Closes DIM-XXX


pyproject.toml
```diff
@@ -303,6 +303,12 @@ dds = [
     "cyclonedds>=0.10.5",
 ]
 
+webrtc = [
+    # WebRTC DataChannel pubsub (Cloudflare Realtime SFU)
+    "aiortc>=1.14.0",
```
paul-nechifor, line 308: You have added just two packages, but `uv.lock` has thousands of lines different.

I think you should reset `uv.lock` to main, and run `uv lock` again. I assume you must have upgraded all the packages producing such a large change.

PR #2642, opened 2026-06-28
feat: move tool tests

paul-nechifor:
## Problem

`tool` tests are not real tests but they interfere with `pytest`. You always have to specify `-m 'not tool'` in one place or another.

Closes DIM-1055

## Solution

* Rename `test_` files which contain tool tests to `tool_`. That way they re not picked up by pytest.
* You can still run them manually. E.g.: `uv run pytest -s dimos/memory2/tool_module.py`


Dreamsorcerer:
Yes please. I feel like most of these are used to create something, like a file? Seems like they shouldn't be written as tests at all to me. Would make more sense to add a simple CLI interface into each file under a `if __name__ == "__main__":` gate.

paul-nechifor:
> Yes please. I feel like most of these are used to create something, like a file? Seems like they shouldn't be written as tests at all to me. Would make more sense to add a simple CLI interface into each file under a `if __name__ == "__main__":` gate.

I think for some pytest is required because they use fixtures. That's why I just renamed them.

Dreamsorcerer:
> > Yes please. I feel like most of these are used to create something, like a file? Seems like they shouldn't be written as tests at all to me. Would make more sense to add a simple CLI interface into each file under a `if __name__ == "__main__":` gate.
> 
> I think for some pytest is required because they use fixtures. That's why I just renamed them.

Yeah, but the fixtures I saw looked like they were just for that one function, in which case it can just be converted to a regular function call. Bigger refactor though, I'm not saying to do it in this PR.

leshy (approved):

PR #2650, opened 2026-06-29
WIP: importor skip

paul-nechifor:
## Problem

Closes DIM-XXX

### DIFF
--- a/dimos/codebase_checks/test_no_importorskip.py
+++ b/dimos/codebase_checks/test_no_importorskip.py
@@ -1,0 +1,97 @@
+# Copyright 2026 Dimensional Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
+from dimos.constants import DIMOS_PROJECT_ROOT
+
+PATTERN = "importorskip"
+
+IGNORED_DIRS = {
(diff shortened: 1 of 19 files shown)


dimos/codebase_checks/test_no_importorskip.py
Dreamsorcerer, line 1: I would recommend against this. This will make it more difficult for downstream to test dimos. Instead, I suggest paying more attention to codecov (the CI checks should all be in place now).

For release, we added the error-on-skips to ensure we do test everything before a release (should be an automated step soon), so any damage caused by an importorskip is short term.

As prior art, we use this in several places in aiohttp for similar reasons:
https://github.com/search?q=repo%3Aaio-libs%2Faiohttp%20importorskip&type=code

I also suspect that people will see the importskip being banned in future, and then just use the less preferred try/except pattern to achieve the same result..

paul-nechifor: @Dreamsorcerer 

The issue I'm trying to prevent here is having tests which do not run. People are already too giddy about adding importor skip for everything. I'm not sure how else to prevent dead tests.

Usually, you can prevent this by enforcing 100% coverage of tests, but I don't think we can do that since we have so many different types of runners.

Do you have ideas about how to make sure we don't write tests which are never ran?

Dreamsorcerer: I suggested this before, but if tests were in a separate tests/ directory (like almost all open source Python projects), then we can potentially gate the tests/ directory on 100% coverage.

Multiple runners is not a problem, as codecov gives us a merged coverage report, so we'd gate on that.

On aiohttp, we just use maintainer discipline to check the codecov results, rather than hard gating it (the one case where you legitimately won't have 100% coverage is for xfail tests).

As mentioned, we also have the error on skip in the release tests (I'm only 19 tests away from having that merged), so anything that did slip through would get fixed before each release, meaning it's only temporary.

paul-nechifor: I've added bin/pytest-all which uses `-m '' --error-for-skips` so I guess I can at least check myself that everything runs. :)

I'll close this PR.

PR #2653, opened 2026-06-29
fix(whole-body): discover sim adapters in registry

KrishnaH96:
## What

In `dimos/hardware/whole_body/registry.py`, made `discover()` also scan the sim adapter root (`dimos/simulation/adapters/whole_body/`) via a new `_discover_flat()` helper — so flat `<name>.py` adapter modules get imported and registered. Graceful `ImportError` skip so hardware-only installs are unaffected. (~20 lines, `registry.py` only.)

## Why

`discover()` only scanned `dimos/hardware/whole_body/`, so the sim adapter `sim_mujoco_g1` was never registered — `dimos --simulation mujoco run unitree-g1-groot-wbc` crashed with `KeyError: Unknown whole-body adapter: sim_mujoco_g1`. The registry's own docstring already says it should scan both roots, so this makes the code match the documented intent. (Same fix Pim intended in the closed part-3 PR #2225, isolated as a surgical change.)

## How tested

- **Repro → fixed:** before, the command crashed at adapter lookup; after, it runs end-to-end — `SimMujocoG1WholeBodyAdapter connected`, GR00T policy `armed`, `TickLoop started at 50.0Hz`, robot pose rendering in the viewer.
- **No regression:** hardware adapters still register (the `transport_lcm`/`transport_ros` path is untouched and ordered first).
- **Gates green:** `ruff check` + `ruff format --check` pass; `mypy` clean on the file.


KrishnaH96:
Superseded by #2749 (manifest-based adapter discovery). Confirmed on current main: sim_mujoco_g1 registers natively and the groot-wbc sim loads clean. Closing.

PR #2665, opened 2026-06-30
Initial  xArm6 WorldBelief perception stack

jhengyilin:
# WorldBelief live object identity for xArm6 manipulation

```mermaid
flowchart TB
    camera([RealSense RGB-D camera])

    detector["YOLO-E prompt detector<br/><i>prompt mode, tracker disabled in xArm blueprint</i>"]
    osr["ObjectSceneRegistration<br/><i>RGB-D objects + CLIP/DINO crop embeddings</i>"]
    wb["WorldBelief<br/><i>stable IDs, support windows, re-acquisition</i>"]
    rerun["Rerun<br/><i>annotated image + 3D workspace</i>"]
    manip["PickAndPlaceModule<br/><i>reads present objects</i>"]
    recorder["XArm6WorldBeliefRecorder<br/><i>records replay-critical streams</i>"]

    subgraph m2 ["memory2"]
        direction TB
        subgraph session ["per-run recording DB"]
            direction LR
            color[(color_image)]
            depth[(depth_image)]
            info[(camera_info)]
            det2d[(detections_2d)]
            det3d[(detections_3d)]
            audit[(worldbelief_audit)]
        end
        subgraph history ["worldbelief_history.db"]
            direction LR
            object_events[(object evidence)]
            semantic_vec[(semantic vectors)]
            visual_vec[(visual vectors)]
        end
    end

    camera --> detector
    detector --> osr
    camera --> osr
    osr -->|"frame objects"| wb
    osr -->|"annotated image + current-frame pointcloud"| rerun
    wb -->|"present objects"| manip
    wb -->|"detections_3d + audit"| rerun
    wb ==> object_events
    wb ==> semantic_vec
    wb ==> visual_vec
    object_events -. "rehydrate compact identity state on restart" .-> wb
    camera ==> recorder
    osr ==> recorder
    wb ==> recorder
    recorder ==> color
    recorder ==> depth
    recorder ==> info
    recorder ==> det2d
    recorder ==> det3d
    recorder ==> audit

    classDef stream fill:#fef3c7,stroke:#d97706,stroke-width:2px
    classDef module fill:#dbeafe,stroke:#2563eb,stroke-width:2px
    classDef memory fill:#dcfce7,stroke:#16a34a,stroke-width:2px
    classDef external fill:#f3f4f6,stroke:#6b7280,stroke-width:1px
    class color,depth,info,det2d,det3d,audit,object_events,semantic_vec,visual_vec stream
    class detector,osr,wb,rerun,manip,recorder module
    class m2,session,history memory
    class camera external
```

## What this unlocks

This PR adds a live object identity layer for the xArm6 perception/manipulation stack. The detector still sees frame-local masks; WorldBelief turns those observations into stable workspace objects that pick/place and visualization can trust.

| Capability | What changed | Why it matters |
|---|---|---|
| Stable live object IDs | `WorldBelief` associates current 3D observations against maintained identity state using geometry, labels, CLIP, DINO, support windows, and re-acquisition policy. | A can should keep its identity while it moves, disappears briefly, or is seen again after camera motion. |
| Manipulation-facing present set | OSR publishes WorldBelief `present_objects` to the `objects` port and `Detection3DArray`. | Pick/place consumes a filtered workspace state, not every noisy frame detection. |
| Cross-session identity seed | WorldBelief writes compact object evidence and vector evidence into a stable Memory2 history DB, then rehydrates maintained state on restart. | The next process starts with prior identity evidence instead of a blank identity table. |
| Cleaner Rerun view | Blueprint opens Rerun with annotated image on the left and 3D workspace on the right. Current-frame pointclouds are used for visual blobs. | The display reflects both live camera evidence and trusted WorldBelief state without stale pointcloud trails. |
| Replay/debug evidence | A dedicated recorder writes a fresh per-run Memory2 DB for color/depth/camera info/detections/audit streams. | We can inspect what the stack saw and what WorldBelief decided without mixing sessions. |

## How it works - walkthrough

| t | Event | What happens |
|---|---|---|
| 0 | Blueprint starts | The xArm6 WorldBelief blueprint wires RealSense, YOLO-E, OSR, WorldBelief, Rerun, recorder, and pick/place. The stable history DB is opened if configured. |
| 1 | Camera sees objects | YOLO-E produces prompt-mode masks. OSR uses color, depth, camera info, and TF to build 3D `Object` observations. |
| 2 | Appearance evidence is attached | OSR crops each object and attaches CLIP semantic embeddings and DINO visual embeddings. |
| 3 | WorldBelief updates | The identity engine matches observations to existing IDs or creates candidates. Objects become `present` only after enough recent support. |
| 4 | Manipulation reads state | `present_objects` are published to the `objects` port and `detections_3d`. Pick/place sees stable workspace objects rather than raw detector churn. |
| 5 | Rerun displays state | The annotated image uses current-frame identity assignments, while the 3D view receives trusted boxes plus current-frame colored pointcloud visualization. |
| 6 | Object leaves view | Frustum and camera-motion handling keep a hidden identity available for re-acquisition instead of immediately deleting or publishing stale visual blobs. |
| 7 | Object returns | WorldBelief can reuse the prior ID when geometry and appearance evidence are strong enough. If evidence is ambiguous, creating a new ID is safer than forcing a wrong merge. |
| 8 | Process restarts | Compact WorldBelief evidence is rehydrated from Memory2 history so the identity table does not always start from scratch. |

## Runtime object model

The PR deliberately separates four related concepts:

| Layer | Meaning | Used by |
|---|---|---|
| Raw detections | YOLO-E masks/classes from the current image. | OSR object construction. |
| Frame objects | Current-frame 3D objects after WorldBelief assigns IDs. | Annotated image and audit. |
| Present objects | Objects with enough recent support to be trusted as present. | `objects`, `detections_3d`, pick/place, 3D boxes. |
| Maintained objects | Remembered identities kept internally for hidden re-acquisition/history. | WorldBelief lifecycle and Memory2 history. |

This split is why the Rerun pointcloud can stay visually clean while manipulation still receives stable present objects.

## Why this design

- **WorldBelief is live task memory.** Memory2 stores durable evidence, but the robot still needs a current materialized workspace state with deterministic identity policy.
- **Raw detector output is not enough for manipulation.** Prompt-mode detection can flicker, split, or merge frame to frame. WorldBelief adds support windows, lifecycle state, and appearance-aware association before publishing objects to pick/place.
- **Memory2 remains the durable layer.** WorldBelief writes object and vector evidence into Memory2 history and can rehydrate from it, but this PR does not claim to ship the final natural-language Memory2 query workflow.
- **Detector tracking is intentionally not the identity source.** The xArm6 blueprint disables YOLO tracking so stable IDs come from the robot-side identity model instead of detector-local track state.
- **Visualization is separated from manipulation state.** Current-frame pointcloud blobs are for visual clarity; `present_objects` and `detections_3d` remain the manipulation-facing outputs.

## Main files

| Area | Files |
|---|---|
| Product blueprint | `dimos/robot/manipulators/xarm/blueprints/worldbelief.py`, `dimos/robot/all_blueprints.py`, `dimos/manipulation/blueprints.py` |
| OSR integration | `dimos/perception/object_scene_registration.py` |
| Identity engine | `dimos/perception/detection/world_belief.py`, `identity_association.py`, `identity_features.py` |
| Durable history | `dimos/perception/detection/world_belief_history.py` |
| Object representation | `dimos/perception/detection/type/detection3d/object.py` |
| Embeddings/detector support | `dimos/models/embedding/dino.py`, `clip.py`, `mobileclip.py`, `dimos/perception/detection/detectors/yoloe.py` |
| Recording | `dimos/robot/manipulators/xarm/worldbelief_recorder.py`, `dimos/memory2/module.py` |

## Known limits / follow-ups

- Exact same-location swaps between visually similar cans are still a hard identity case. If the physical objects exchange positions perfectly, stronger appearance mismatch policy may need a follow-up threshold split.
- CLIP + DINO improve identity evidence but add latency. If annotated-image smoothness becomes the priority, the next step is a fast visual overlay before embedding completion while keeping trusted WorldBelief outputs after embeddings.
- Runtime detector prompt updates currently replace the active prompt list. Append-to-default prompt workflow is a follow-up.
- Memory2 vector evidence is recorded for future search/recall work, but this PR does not add the final agent query API.


jhengyilin:
# xArm6 WorldBelief — pulled perception with cross-session memory

Pivot to **pulled perception**: nothing runs until asked; when asked, the answer covers the *complete* recorded interval since the last question — no observation holes.

```mermaid
flowchart LR
    CAM["camera just records<br/>color + depth + pose"] --> REC[("session<br/>recording")]
    REC -- "scan (on demand)" --> SS["SceneScanner<br/>frames → 3D sightings"]
    SS --> WB["WorldBelief<br/>sightings → stable identities"]
    WB --> OUT["reply: name · id · xyz<br/>· trust · basis"]
    WB <--> MEM[("long-term memory<br/>survives restarts")]
    MEM -- "recall (on demand)" --> RC["&quot;where did I last see X?&quot;"]
```

## How a scan actually works 

**The camera only records.** RealSense color + depth frames and the camera's world pose stream into the session's recording `.db`. No GPU touches them at record time.

**Frame selection** — fast-forward while nothing moves, slow down when something does. The recorded camera poses set the pace: ~1 frame/s while still, 7.5/s while moving. And if the *set of detected objects* changes between two kept frames (a hand swapped something while the camera was still), the scanner bisects the gap and looks between them. Net cost: 11–18 % of frames ever see the detector, with no blind spots.


## The worldbelief identity core

The detector has no memory — it never says "*that same* mug". `WorldBelief` adds it. It keeps one **track** per believed object: a stable `object_id`, the last observed position, and a small **gallery** of DINO embeddings (views of the object from different angles). Every incoming detection must be *explained*, and the model always prefers the cheapest explanation — reusing an existing identity over creating one.

```mermaid
flowchart TB
    S["detection arrives:<br/>label + 3D center + DINO embedding"] --> Q1{"Q1 · inside an active track's<br/>match radius?<br/>min(0.18 + 0.15·dt, 0.40) m"}
    Q1 -- "yes — label agrees (or cos ≥ 0.8 override),<br/>no appearance contradiction (cos ≥ 0.5)" --> T1["update that track<br/><i>basis: tracked</i>"]
    Q1 -- no --> Q2{"Q2 · inside an established track's<br/>0.13 m noise envelope?"}
    Q2 -- yes --> T2["claimed by that track —<br/>mint suppressed (envelope default)"]
    Q2 -- no --> Q3{"Q3 · gallery match to an absent or<br/>stored identity?<br/>cos ≥ 0.5, margin ≥ 0.05, same label"}
    Q3 -- yes --> T3["identity revived<br/><i>basis: reacquired / restored</i>"]
    Q3 -- no --> Q4["Q4 · mint a new track —<br/>established after 3 frames ∧ 1.5 s"]

    P["established track goes unseen"] -- "depth ray passes THROUGH<br/>its center — 2 votes" --> GONE["state: absent<br/>(gallery kept)"]
    GONE -. "revivable via Q3" .-> Q3
```

**Q1 — Is it inside an active track's match radius?** Every track predicts the object near its last observed position. The radius starts at 0.18 m and grows with time-unobserved (`min(0.18 + 0.15·dt, 0.40) m`) — deliberately *above* the measured sensor noise, since even a perfectly static object appears to jump by up to 0.126 m between lifts. Two embedding checks guard the match: a detection under a *different label* only matches on a strong appearance override (cos ≥ 0.8 — the same object seen 180° apart scores ~0.94, different objects max ~0.33), and a detection with the *right* label but a clearly foreign embedding (cos < 0.5 against the track's gallery) is vetoed — detector labels lie. A match updates the track: `basis: tracked`.

**Q2 — Is it inside an established track's 0.13 m noise envelope?** Then that track claims it, by default. Two rigid objects can't co-locate: a detection landing within the envelope of an established object *is* that object unless stronger evidence says otherwise, and minting there is suppressed (the mint-guard). This envelope-as-null-hypothesis rule kills most identity-theft bugs.

**Q3 — Does its embedding match an absent or stored identity's gallery?** No active track claims it — so compare the detection's DINO embedding against the galleries of every absent identity, from this session or any earlier one. The best match must cos ≥ 0.5, beat by a margin ≥ 0.05 (a near-tie between two candidates means the evidence can't tell them apart — refuse to guess), carry the same label, and resemble the absent identity *more than any co-located neighbour*. Success revives the identity: `basis: reacquired` (same session) or `restored` (previous session).

**Q4 — Mint a new track.** Only when Q1–Q3 all fail is a new `object_id` minted — and the track stays unestablished until seen **3 frames across ≥ 1.5 s**. Hand blurs and one-frame glitches never establish, so they never reach a scan reply.

**Absence is evidence-only.** A track is never dropped for being unseen — it goes `state: absent` only when the depth ray *passes through* its center (the sensor ranges the table behind it) on **2 separate frames**. Panning away is not absence. The gallery is kept, so an absent identity stays revivable via Q3.

**Galleries are the long-term memory.** Each established sighting adds sufficiently-novel views to the track's gallery in `worldbelief_history.db` (megabytes — recordings stay the heavy, per-file-deletable archive). Restart everything tomorrow, scan once: Q3 runs against the stored galleries and the mug returns with its old id.

**`recall` reuses the same skepticism.** "Where did I last see X" searches whole-frame CLIP vectors across *all* indexed sessions — but CLIP similarity is nearly flat within one scene (measured top-30 spread: 0.259→0.248), so its ranking only *proposes*. Each candidate moment is verified by a small detection scan of its source recording; the first confirmed moment is the answer, and `where_object` is either detector-verified or honestly `None`.

Where each piece lives in the code:

| Concept | In the code |
|---|---|
| a track / the belief table | `_Track` / `WorldBelief` (`world_belief.py`) |
| embedding gallery (RAM + persistent) | `_Track.gallery` / `_GalleryStore` → `worldbelief_history.db` |
| Q1 — match radius, label override, appearance veto | `_match_frame()` |
| Q2 — 0.13 m envelope + mint-guard | `_envelope_neighbours()`, checked in `_match_frame()` / `_acquire()` |
| Q3 — gallery match with floor + margin | `_reacquire_candidate()` / `_acquire()` |
| Q4 — mint + establishment (3 frames ∧ 1.5 s) | `_insert()` / `_hit()` |
| absence (2 depth votes, gallery kept) | `_vote_absence()` + `absence.py` |
| viewpoint-diversity trust (`tentative` → `confirmed`) | `trust_of()` — 2 distinct viewpoints |

## How to Test

```bash
dimos --xarm6-ip <ROBOT_IP> run xarm6-worldbelief
# in a second terminal:
dimos mcp call scan -a prompt='["mug","coke can"]'
dimos mcp call recall -a text="mug"
```

Move an object by hand between scans → same id, `basis: reacquired`. Remove it → drops from the reply (2-vote absence). Restart the process and scan → same ids, `basis: restored`. `recall` returns when/where it was last seen — including from a previous session's recording.

dimos/perception/absence.py
```diff
@@ -0,0 +1,112 @@
...
+    The depth at the projected point is the **median of a ``(2*patch_px+1)`` square** of
+    valid (finite, >0) readings — real depth sensors have holes and speckle, so a single
+    pixel can flip the verdict. Returns one of OUT_OF_VIEW / ABSENT / OCCLUDED / PRESENT.
+
+    ``free_margin_m`` need only clear sensor noise + center-estimation error, NOT the
+    distance to the background: an opaque object's depth reading is its NEAR surface, so
+    a present object always measures ``d <= expected + noise`` — any reading beyond that
+    means the near surface is gone. On a real tabletop the support surface sits only
+    4-13 cm behind a can's believed center (measured on the arm), so a generous margin
+    silently classifies every removed object as PRESENT and resets its absence votes.
```
mustafab0: Please dont add llm verbose for basic things. Both these can be written as 1 line descriptions

dimos/perception/absence.py
```diff
@@ -0,0 +1,112 @@
...
+      its center) still counts as "surface present" and resets absence votes.
+    - A ``border_guard_px`` band (default ``patch_px + 3``) around the image border is
+      OUT_OF_VIEW: mid-pan, cm-level pose error can place the believed center fractionally
+      inside the border while the object is fully outside — no evidence either way.
+    """
+    cam_from_world = np.linalg.inv(np.asarray(world_from_camera, dtype=np.float64))
+    pw = np.array(
+        [float(center_world.x), float(center_world.y), float(center_world.z), 1.0],
+        dtype=np.float64,
+    )
```
mustafab0: Do NOT use handbuilt matrices. We have Transform message types with great utils and methods for transformation.

This is not a good way to program in robotics. Please use Transform types

dimos/perception/absence.py
```diff
@@ -0,0 +1,112 @@
...
+      a believed center sitting 1 px outside the silhouette must not read the background
+      median and kill an object that is plainly in frame.
+    - ``near_extent_m`` extends the PRESENT band toward the camera by the object's own
+      half-depth, so a deep object's near surface (e.g. a bottle read 5 cm in front of
+      its center) still counts as "surface present" and resets absence votes.
+    - A ``border_guard_px`` band (default ``patch_px + 3``) around the image border is
+      OUT_OF_VIEW: mid-pan, cm-level pose error can place the believed center fractionally
+      inside the border while the object is fully outside — no evidence either way.
+    """
+    cam_from_world = np.linalg.inv(np.asarray(world_from_camera, dtype=np.float64))
```
mustafab0: This should simply be `world_from_camera.inverse()`

dimos/perception/detection/world_belief.py
```diff
@@ -0,0 +1,752 @@
...
+BASIS_RESTORED = "restored"  # past-session identity restored from the gallery store
+
+Viewpoint = tuple["NDArray[np.float64]", "NDArray[np.float64]"]  # (camera position, forward)
+
+_CANDIDATE_TTL_S = 5.0  # never-established junk is dropped after this long unseen
+_MAX_ABSENT = 128  # geometric death never times out, so cap retained corpses
+_POSITION_WINDOW = 9  # running-median window for the entity center
+
+
+def camera_viewpoint(world_from_camera: NDArray[np.float64]) -> Viewpoint:
```
mustafab0: Should be a Pose object. not a NDArray

dimos/perception/detection/world_belief.py
```diff
@@ -0,0 +1,752 @@
...
+        self._reacq_margin = 0.05
+        self._min_frames = min_frames
+        self._min_span = min_span_s
+        self._absent_threshold = absent_threshold
+        self._min_viewpoints = min_viewpoints
+        self._viewpoint_angle_deg = 10.0
+        self._min_baseline_m = 0.05
+        self._gallery_novelty = 0.9
+        self._gallery_max = 8
+        self._tracks: dict[str, _Track] = {}
```
mustafab0, line 260: Why is this not a module?

mustafab0, line 260: All these should be MOduleConfig type

dimos/perception/detection/world_belief.py
```diff
@@ -0,0 +1,752 @@
...
+    def __init__(self, obj: Object, ts: float, viewpoint: Viewpoint | None, basis: str) -> None:
+        self.obj = obj
+        self.label_counts: Counter[str] = Counter([obj.name])
+        self.first_ts = ts
+        self.last_seen = ts
+        self.n_frames = 1
+        # Association anchors on the LAST raw observed position (it follows a carried
+        # object frame-to-frame); the REPORTED center is the running median (stable).
+        self.last_pos: NDArray[np.float64] = obj.center.to_numpy()
+        self.positions: list[NDArray[np.float64]] = [obj.center.to_numpy()]
```
mustafab0: These should be PosStamped objects

mustafab0 (changes requested):

dimos/robot/manipulators/xarm/worldbelief_recorder.py
```diff
@@ -0,0 +1,300 @@
...
+        suffix += 1
+    return candidate
+
+
+class XArm6WorldBeliefRecorderConfig(RecorderConfig):
+    db_path: str | Path = _RECORDING_BASE_PATH
+    default_frame_id: str = "world"
+
+
+class XArm6WorldBeliefRecorder(Recorder):
```
mustafab0: These modules should have nothing to do with Xarm6 specifically. 

The WorldBelief recorder and And also the WorldBelief module should have absolutely 0 hardware specific dependency.

Please rename

mustafab0 (changes requested):

dimos/perception/detection/world_belief.py
```diff
@@ -0,0 +1,752 @@
...
+
+
+class _Track:
+    """One believed entity: the folded Object plus its identity/lifecycle state."""
+
+    __slots__ = (
+        "absent_votes", "basis", "established", "first_ts", "gallery", "label_counts",
+        "last_pos", "last_seen", "latest_emb", "n_frames", "obj", "positions",
+        "sizes", "state", "viewpoints",
+    )
```
TomCC7: is there a reason not to use dataclass here? should be easier to declare all the fields. also I think it would be better to separate a track's real property (size, position, etc) from the field used for identification/deduplication (as some sort of metadata)

dimos/manipulation/blueprints.py
```diff
@@ -31,3 +31,6 @@
 from dimos.robot.manipulators.xarm.blueprints.simulation import (
     xarm_perception_sim as xarm_perception_sim,
 )
+from dimos.robot.manipulators.xarm.blueprints.worldbelief import (
+    xarm6_worldbelief as xarm6_worldbelief,
+)
```
TomCC7:
```suggestion
```

dimos/perception/scene_scan.py
```diff
@@ -0,0 +1,453 @@
...
+    ) -> None:
+        self._detector = detector
+        self._target_frame = target_frame
+        self._text_prompts = list(text_prompts) if text_prompts else None
+        self._embed = embed
+        self._visual_embedder = visual_embedder
+        # 0.6: lower (0.4) proved noisy on the arm — sub-0.5 junk kept entering the
+        # lifecycle; every benchmark prompt (incl. "green can" at 0.89-0.97) clears 0.6.
+        self._detector_conf = detector_conf
+        self._belief = WorldBelief(
```
TomCC7: world belief should own and manage scanner result. ownership mismatched. also this file is imported in recorder module so recorder handles everything in this case

TomCC7:
@jhengyilin @mustafab0 (@leshy pls audit) I think it's fine to make world-belief a dedicated module as it makes lifecycle and api boundary clean, concurrent reads with single write is not a problem for sqlite as read operates on the snapshot. My only concern about having to specify db path for mutliple modules can be relieved in the future by providing some pre-launch config injection (I'm thinking of using some timestamp based default location and make sure two modules sync on this path)

dimos/perception/absence.py
jhengyilin: Addressed. 
the vector is replaced with `rotation.rotate_vector(center) + translation`.

dimos/perception/absence.py
jhengyilin: Addressed. 
replaced with `cam_from_world = world_from_camera.inverse()`

dimos/perception/detection/world_belief.py
jhengyilin: Addressed. 
replaced with `camera_viewpoint(pose: Pose) -> Viewpoint`

dimos/perception/detection/world_belief.py
jhengyilin: Addressed. 
With a cleaner split of recorder module and worldbelief module in the updated commit

jhengyilin: Addressed. 
Now it lives in `WorldBeliefConfig` on top of the worldbelief module class

dimos/perception/detection/world_belief.py
jhengyilin: Addressed. 
Went with `PointStamped`: the anchor is a pure position (the identity core
never estimates an orientation, so `PoseStamped` would carry a meaningless identity quaternion.
Happy to switch to `PoseStamped` if you'd rather standardize on it.

dimos/robot/manipulators/xarm/worldbelief_recorder.py
jhengyilin: Addressed.

`WorldBeliefRecorder` + `WorldBeliefModule` in `dimos/perception/`

dimos/perception/detection/world_belief.py
jhengyilin: Adressed.

`_Track` is a `@dataclass(slots=True)` with fields declared in three labeled
groups (physical state / identity evidence / lifecycle)

dimos/manipulation/blueprints.py
jhengyilin: Removed.

This is just for legacy compatability.

dimos/perception/scene_scan.py
jhengyilin: Addressed.

`SceneScanner` is now stateless
worldbelief is owned by `WorldBeliefModule`, and the
recorder no longer hosts any query logic.

jhengyilin:
# xArm6 WorldBelief — pulled perception with cross-session memory

Pivot to **pulled perception**
```mermaid
flowchart TB
    CAM["camera<br/>color + depth + pose"] --> REC["WorldBeliefRecorder<br/>"]
    REC -- writes --> RDB[("session recording<br/>per-session .db")]
    RDB -- "scan / recall<br/>(on demand)" --> WB["WorldBeliefModule<br/>@scan · @recall<br/>SceneScanner + WorldBelief"]
    MEM[("long-term memory<br/>worldbelief_history.db<br/>survives restarts")] <--> WB
    WB --> OUT["reply: name · id · xyz<br/>· trust · basis"]
    WB -- "objects · detections_3d · pointcloud" --> CONS["PickAndPlace / Rerun"]
```

## Two modules

The stack is split as 2:

- **`WorldBeliefRecorder`** (record plane) only writes: raw color/depth/camera_info/joint_state + tf into a per-session, timestamped recording db. 
- **`WorldBeliefModule`** (query plane) owns everything intelligent: the warm `WorldBelief` fold, the scan/CLIP/DINO models, and the `scan`/`recall` MCP skills. It reads the recording db on demand (W writes the long-term history db. 


## How a scan actually works — follow one call

**The camera only records.** RealSense color + depth frames and the camera's world pose stream into the session's recording `.db`. No GPU touches them at record time.

**Frame selection** — fast-forward while nothing moves, slow down when something does. The recorded camera poses set the pace: ~1 frame/s while still, 7.5/s while moving. And if the *set of detected objects* changes between two kept frames (a hand swapped something while the camera was still), the scanner bisects the gap and looks between them. Net cost: 11–18 % of frames ever see the detector, with no blind spots.

**Ownership.** `SceneScanner` is a stateless engine (models + frame selection + 2D→3D lift + embeddings); it is handed the module's `WorldBelief` and folds frames *into* it — `scanner.scan(store, belief, …)`. The belief owns all identity/trust/absence state and outlives every scan.

## The worldbelief identity core

The detector has no memory — it never says "*that same* mug". `WorldBelief` adds it. It keeps one **track** per believed object: a stable `object_id`, the last observed position, and a small **gallery** of DINO embeddings (views of the object from different angles). Every incoming detection must be *explained*, and the model always prefers the cheapest explanation — reusing an existing identity over creating one.

```mermaid
flowchart TB
    S["detection arrives:<br/>label + 3D center + DINO embedding"] --> Q1{"Q1 · inside an active track's<br/>match radius?<br/>min(0.18 + 0.15·dt, 0.40) m"}
    Q1 -- "yes — label agrees (or cos ≥ 0.8 override),<br/>no appearance contradiction (cos ≥ 0.5)" --> T1["update that track<br/><i>basis: tracked</i>"]
    Q1 -- no --> Q2{"Q2 · inside an established track's<br/>0.13 m noise envelope?"}
    Q2 -- yes --> T2["claimed by that track —<br/>mint suppressed (envelope default)"]
    Q2 -- no --> Q3{"Q3 · gallery match to an absent or<br/>stored identity?<br/>cos ≥ 0.5, margin ≥ 0.05, same label"}
    Q3 -- yes --> T3["identity revived<br/><i>basis: reacquired / restored</i>"]
    Q3 -- no --> Q4["Q4 · mint a new track —<br/>established after 3 frames ∧ 1.5 s"]

    P["established track goes unseen"] -- "depth ray passes THROUGH<br/>its center — 2 votes" --> GONE["state: absent<br/>(gallery kept)"]
    GONE -. "revivable via Q3" .-> Q3
```

**Q1 — Is it inside an active track's match radius?** Every track predicts the object near its last observed position. The radius starts at 0.18 m and grows with time-unobserved (`min(0.18 + 0.15·dt, 0.40) m`) — deliberately *above* the measured sensor noise, since even a perfectly static object appears to jump by up to 0.126 m between lifts. Two embedding checks guard the match: a detection under a *different label* only matches on a strong appearance override (cos ≥ 0.8 — the same object seen 180° apart scores ~0.94, different objects max ~0.33), and a detection with the *right* label but a clearly foreign embedding (cos < 0.5 against the track's gallery) is vetoed — detector labels lie. A match updates the track: `basis: tracked`.

**Q2 — Is it inside an established track's 0.13 m noise envelope?** Then that track claims it, by default. Two rigid objects can't co-locate: a detection landing within the envelope of an established object *is* that object unless stronger evidence says otherwise, and minting there is suppressed (the mint-guard). This envelope-as-null-hypothesis rule kills most identity-theft bugs.

**Q3 — Does its embedding match an absent or stored identity's gallery?** No active track claims it — so compare the detection's DINO embedding against the galleries of every absent identity, from this session or any earlier one. The best match must cos ≥ 0.5, beat by a margin ≥ 0.05 (a near-tie between two candidates means the evidence can't tell them apart — refuse to guess), carry the same label, and resemble the absent identity *more than any co-located neighbour*. Success revives the identity: `basis: reacquired` (same session) or `restored` (previous session). When identities are compared across sessions, the stored gallery's position/support metadata comes from the most recently **appended** view (wall-clock store time, which survives restarts) — a per-track frame counter resets on restore and would keep electing a stale pre-move position.

**Q4 — Mint a new track.** Only when Q1–Q3 all fail is a new `object_id` minted — and the track stays unestablished until seen **3 frames across ≥ 1.5 s**. Hand blurs and one-frame glitches never establish, so they never reach a scan reply.

**Absence is evidence-only.** A track is never dropped for being unseen — it goes `state: absent` only when the depth ray *passes through* its center (the sensor ranges the table behind it) on **2 separate frames**. Panning away is not absence. The gallery is kept, so an absent identity stays revivable via Q3.

**Galleries are the long-term memory.** Each established sighting adds sufficiently-novel views to the track's gallery in `worldbelief_history.db` (megabytes — recordings stay the heavy, per-file-deletable archive). Restart everything tomorrow, scan once: Q3 runs against the stored galleries and the mug returns with its old id.

**`recall` reuses the same skepticism.** "Where did I last see X" searches whole-frame CLIP vectors across *all* indexed sessions — but CLIP similarity is nearly flat within one scene (measured top-30 spread: 0.259→0.248), so its ranking only *proposes*. Each candidate moment is verified by a small detection scan of its source recording; the first confirmed moment is the answer, and `where_object` is either detector-verified or honestly `None`.

Where each piece lives in the code:

| Concept | In the code |
|---|---|
| record plane (dumb writer + live db path RPC) | `WorldBeliefRecorder` (`dimos/perception/worldbelief_recorder.py`) |
| query plane (`scan`/`recall` skills, owns belief + models) | `WorldBeliefModule` (`dimos/perception/worldbelief_module.py`) |
| stateless scan engine — folds INTO the belief you pass | `SceneScanner.scan(store, belief, …)` (`scene_scan.py`) |
| a track / the belief table | `_Track` (slots dataclass: physical / identity / lifecycle field groups) / `WorldBelief` (`world_belief.py`) |
| every tuning constant (measured values) | `WorldBeliefConfig` — blueprint-overridable |
| embedding gallery (RAM + persistent) | `_Track.gallery` / `_GalleryStore` → `worldbelief_history.db` |
| Q1 — match radius, label override, appearance veto | `_match_frame()` |
| Q2 — 0.13 m envelope + mint-guard | `_envelope_neighbours()`, checked in `_match_frame()` / `_acquire()` |
| Q3 — gallery match with floor + margin | `_reacquire_candidate()` / `_acquire()` |
| Q4 — mint + establishment (3 frames ∧ 1.5 s) | `_insert()` / `_hit()` |
| absence (2 depth votes, gallery kept) | `_vote_absence()` + `absence.py` (Transform-native projection) |
| viewpoint-diversity trust (`tentative` → `confirmed`) | `trust_of()` — 2 distinct viewpoints |



## How to Test

```bash
dimos --xarm6-ip <ROBOT_IP> run xarm6-worldbelief
# in a second terminal:
dimos mcp call scan -a prompt='["mug","coke can"]'
dimos mcp call recall -a text="mug"
```

Move an object by hand between scans → same id, `basis: reacquired`. Remove it → drops from the reply (2-vote absence). Restart the process and scan → same ids, `basis: restored`. `recall` returns when/where it was last seen — including from a previous session's recording. While a scan is running, the recording db keeps growing — recording and scanning live in separate worker processes.

dimos/perception/worldbelief_module.py
```diff
@@ -0,0 +1,307 @@
...
+from dimos.perception.detection.type.detection3d.object import Object
+from dimos.perception.detection.world_belief import WorldBelief, WorldBeliefConfig
+from dimos.perception.worldbelief_recorder import WorldBeliefRecorder
+from dimos.utils.logging_config import setup_logger
+
+if TYPE_CHECKING:
+    from dimos.perception.scene_scan import SceneScanner
+
+_STATE_DIR = STATE_DIR / "worldbelief"
+_HISTORY_PATH = _STATE_DIR / "worldbelief_history.db"
```
TomCC7, line 52: will (or is this intended) to work cross-session?

dimos/perception/detection/world_belief.py
```diff
@@ -0,0 +1,791 @@
...
+        return self.last_pos.ts
+
+    @property
+    def modal_label(self) -> str:
+        return self.label_counts.most_common(1)[0][0]
+
+    def full_gallery(self) -> list[NDArray[np.float32]]:
+        """Diverse views plus the latest — the set the re-acquire gate scores against
+        (the latest view catches a fresh face the novelty gate hasn't kept yet)."""
+        return list(self.gallery) if self.latest_emb is None else [*self.gallery, self.latest_emb]
```
TomCC7, line 155: there's a _Gallery class defined below. Are these related semantically? If no better to have a different name

dimos/perception/worldbelief_module.py
```diff
@@ -0,0 +1,307 @@
...
+        belief = getattr(self, "_belief", None)
+        if belief is not None:
+            try:
+                belief.close()  # hand this session's galleries to a later model on the same path
+            except Exception as e:
+                logger.warning("belief close failed: %s", e)
+        super().stop()
+
+    def _recording_path(self) -> str:
+        if self._recorder is not None:
```
TomCC7: this is stale code, recorder never initialized

TomCC7: also recorder is a module. a module should not own another module

dimos/perception/worldbelief_module.py
```diff
@@ -0,0 +1,307 @@
...
+        import copy as _copy
+
+        from dimos.memory2.store.sqlite import SqliteStore
+        from dimos.perception.detection.type.detection3d.object import (
+            aggregate_pointclouds,
+            to_detection3d_array,
+        )
+
+        rec_path = self._recording_path()
+        with _SCAN_LOCK:  # serialize concurrent scan/recall (shared warm belief)
```
TomCC7: so basically you need this lock everytime you use the scanner variable. inn this case I would suggest to wrap the scanner with context manager (https://book.pythontips.com/en/latest/context_managers.html) so that you can use syntax like
```python
import threading
from contextlib import contextmanager


class SharedData:

    def __init__(self):
        self._lock = threading.Lock()
        self._data = {"count": 0}  # A mutable dict

    @contextmanager
    def access_data(self):
        with self._lock:
            # Code before yield happens on __enter__
            yield self._data
            # Code after yield happens on __exit__


# Usage
shared = SharedData()

with shared.access_data() as data:
    # The lock is safely held right now
    data["count"] += 1
    data["last_updated"] = "now"
# The lock is automatically released here
```

dimos/perception/worldbelief_module.py
```diff
@@ -0,0 +1,307 @@
...
+        moment; ``where_object`` is None when nothing confirms (fields then describe the
+        best unverified hit) or the footage is gone. ``None`` when nothing matches.
+        """
+        from contextlib import nullcontext
+
+        from dimos.memory2.store.sqlite import SqliteStore
+        from dimos.perception.recall import recall as _recall
+
+        rec_path = self._recording_path()
+        with _SCAN_LOCK:  # serialize concurrent scan/recall (shared CLIP model + index watermark)
```
TomCC7: And if this scan_lock is just for guarding clip models, you should make lock scope as small as possible. I think sqlitestore actually allows for concurrent read (at least the case for sqlite itself). Should confine the scope of this lock better.

TomCC7 (commented):
shape is good, left some comments 👍

jhengyilin:
# WorldBelief — pulled perception with cross-session memory

This supersedes the earlier spatial-radius/Q1–Q4 overview. The final implementation uses
global DINO appearance matching for identity. 3D evidence is used for current geometry,
viewpoint trust, and absence—not as a movement-distance identity heuristic.

```mermaid
flowchart TB
    CAM["RealSense<br/>color + depth + camera pose"] --> REC["WorldBeliefRecorder<br/>record plane"]
    REC --> SESSION[("timestamped session DB<br/>raw evidence")]
    SESSION -- "scan / recall<br/>on demand" --> WB["WorldBeliefModule<br/>query plane"]
    HISTORY[("worldbelief_history.db<br/>DINO galleries + CLIP index")] <--> WB
    WB --> REPLY["scan result<br/>name · full ID · xyz · trust · basis"]
    WB --> OBJECTS["objects snapshot"]
    WB --> VIS["detections_3d · pointcloud"]
    CAM --> RERUN["Rerun validation"]
    VIS --> RERUN
    OBJECTS -. "future integration PR" .-> DOWNSTREAM["PickAndPlace / navigation"]
```

## Two modules

- **`WorldBeliefRecorder`** is the record plane. It continuously writes RGB-D, camera info,
  joint state, and resolved pose evidence to a new timestamped SQLite recording for each run.
- **`WorldBeliefModule`** is the query plane. It owns the YOLO-E/DINO/CLIP model instances, the
  mutable `WorldBelief` fold, persistent history, and the `scan`/`recall` skills. Scan and
  recall share one instance lock; published snapshots are copied from mutable belief state.

The Recorder performs no detection. `WorldBeliefModule` warms YOLO-E and DINO at startup so
`scan` is ready; CLIP remains lazy until the first `recall`.

## One `scan` call

1. The xArm blueprint has no built-in vocabulary, so each call supplies a non-empty prompt.
2. The first call reads at most `window` seconds of history. Later calls fold every eligible
   frame after the belief watermark, so `window` does not limit cross-scan catch-up.
3. Pose deltas select coarse stationary frames and denser moving-camera frames. The xArm
   blueprint uses 4 Hz while stationary; moving sampling is 7.5 Hz. First and last eligible
   frames are retained.
4. YOLO-E segments prompted objects, RGB-D and TF lift them into world-frame `Object`s, and
   DINO embeds each accepted crop.
5. Frames fold chronologically into the module-owned belief. If RGB-D/TF evidence cannot reach
   the recording end, the call fails instead of returning a stale success.
6. The module publishes `Object`, `Detection3DArray`, and aggregate point-cloud snapshots after
   releasing the engine lock.


## Identity and long-term memory

```mermaid
flowchart TB
    DET["lifted detection<br/>label + current 3D geometry + DINO"] --> CAND["candidate identities<br/>RAM tracks + retrieved persisted galleries"]
    CAND --> ASSIGN["global one-to-one<br/>DINO cosine assignment"]
    ASSIGN -- "clear winner" --> MATCH["tracked / reacquired / restored"]
    ASSIGN -- "near tie" --> ABSTAIN["abstain<br/>do not guess or mint"]
    ASSIGN -- "no plausible identity" --> NEW["new candidate ID"]
    NEW --> EST["established after<br/>3 frames AND 1.5 s"]
```

- Same-label matches require cosine ≥ 0.5. A label change requires the stronger ≥ 0.8
  appearance override.
- The assignment is global, so two observations cannot claim the same identity in one frame.
  A rival within the 0.05 margin is ambiguous and is not silently converted into a new ID.
- 3D position does not arbitrate identity. A moved object may retain its ID after an arbitrary
  gap when its DINO evidence is distinctive; ambiguous lookalikes remain unresolved.
- Established tracks are `tentative` after one viewpoint and `confirmed` after two sufficiently
  distinct viewpoints. This is identity-evidence status, not manipulation or collision safety.
- Partial or implausibly enlarged observations may refresh the visible center without replacing
  trusted shape geometry.
- An object becomes absent only after two depth frames see through its expected location.
  Occlusion and an out-of-view camera provide no absence vote.
- Established identities persist as bounded, diverse DINO galleries in
  `worldbelief_history.db`. Galleries are namespaced by exact checkpoint and embedding size, so
  incompatible model spaces are not mixed after a configuration change.

Identity basis in the returned object describes the latest evidence:

| Basis | Meaning |
|---|---|
| `new` | newly created candidate |
| `tracked` | matched in the next folded frame |
| `reacquired` | matched to an in-memory identity after a gap or absence |
| `restored` | matched to an identity persisted by an earlier process/session |

## `recall`

`recall(text)` incrementally indexes immutable session frames with CLIP into the same history
database. CLIP ranks candidate moments across sessions; it is not trusted as localization.
Top moments are reopened from their source recordings and checked with a small YOLO-E + DINO
scan. The response includes the recording, timestamp, camera pose, uncalibrated CLIP similarity,
and detector-confirmed `where_object`. If no candidate is confirmed, `where_object` is `None`.

## Configuration

Deployment policy stays in the xArm blueprint; the generic perception modules contain no xArm
paths or vocabulary. The main model and sampling choices are blueprint fields:

```python
WorldBeliefModule.blueprint(
    scan_prompts=[],
    stationary_hz=4.0,
    yoloe_model_name="yoloe-11l-seg.pt",
    dino_model_name="facebook/dinov2-base",
    clip_model_name="openai/clip-vit-base-patch32",
)
```

Persistent gallery and CLIP streams are isolated by checkpoint ID, so changing DINO or CLIP
does not reuse incompatible embeddings.


## How to test

```bash
# terminal 1
uv run dimos --xarm6-ip <ROBOT_IP> run xarm6-worldbelief

# terminal 2: an explicit prompt is required
uv run dimos mcp call scan -a prompt='["coke can", "mug", "yellow can"]'
uv run dimos mcp call recall -a text="mug"
```

Expected checks:

- repeated scans keep a distinctive object's full ID;
- a moved or previously absent object reports `basis: reacquired`;
- after restart, a gallery match reports the same full ID with `basis: restored`;
- two novel camera viewpoints promote an established identity from `tentative` to `confirmed`;
- Rerun shows live RGB/depth plus published 3D detections and point cloud.

dimos/perception/worldbelief_module.py
jhengyilin: Yes, that is intentional. Recordings are separate immutable sessions, while the DINO galleries and CLIP recall index share one persistent history database across sessions.

dimos/perception/detection/world_belief.py
jhengyilin: Addressed.

They were related. The in-memory API is now appearance_views(), and _GalleryStore is the bounded persistent copy used for cross-session re-id.

dimos/perception/worldbelief_module.py
jhengyilin: Addressed.

WorldBeliefModule depends on a narrow WorldBeliefRecorderSpec RPC boundary

dimos/perception/worldbelief_module.py
jhengyilin: Addressed.

 _engine() is now the single context manager for accessing the shared scanner, models, and belief state. The lock is per module instance rather than process-global.

dimos/perception/worldbelief_module.py
jhengyilin: Tightened it. 

Recording-path resolution and SQLite store setup happen outside the engine lock.

dimos/perception/detection/world_belief.py
```diff
@@ -0,0 +1,617 @@
...
+logger = setup_logger()
+
+TENTATIVE = "tentative"  # established identity seen from one camera viewpoint
+CONFIRMED = "confirmed"  # established identity seen from multiple viewpoints; not action safety
+
+# How the entity's identity was last established
+BASIS_NEW = "new"
+BASIS_TRACKED = "tracked"
+BASIS_REACQUIRED = "reacquired"
+BASIS_RESTORED = "restored"  # past-session identity restored from the gallery store
```
mustafab0, line 56: Change Entity identity and status states to Literals

mustafab0: https://github.com/dimensionalOS/dimos/issues/2877

dimos/perception/detection/world_belief.py
```diff
@@ -0,0 +1,617 @@
...
+        for obj in unmatched:
+            self._insert(obj, ts, viewpoint, basis=BASIS_NEW)
+            seen.add(id(self._tracks[obj.object_id]))
+        if camera_transform is not None:
+            self._vote_absence(seen, camera_info, camera_transform, depth_m)
+        self._prune(ts)
+
+    def _match_frame(
+        self, objects: list[Object]
+    ) -> tuple[list[tuple[_Track | str, Object]], list[Object], set[int]]:
```
mustafab0: This should be refactored into smaller functions.

mustafab0 (dismissed):

dimos/perception/detection/world_belief_history.py
```diff
@@ -0,0 +1,375 @@
...
+                ts=a["last"],
+                embedding=a["semantic_emb"],
+                embedding_model=a["semantic_embedding_model"],
+                embedding_device=a["semantic_embedding_device"],
+                embedding_dim=a["semantic_embedding_dim"],
+                visual_embedding=a["visual_emb"],
+                visual_embedding_model=a["visual_embedding_model"],
+                visual_embedding_device=a["visual_embedding_device"],
+                visual_embedding_dim=a["visual_embedding_dim"],
+            )
```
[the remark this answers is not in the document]
jhengyilin: Adressed

Fixed by the gallery-only restore design. We no longer rehydrate old lifecycle timestamps or apply a reacquisition time window. An ID is restored only from a current observation that matches its
persisted DINO gallery.

dimos/perception/detection/world_belief.py
```diff
@@ -0,0 +1,1373 @@
...
+            "input": len(objects),
+            "created": 0,
+            "updated": 0,
+            "matched_track": 0,
+            "matched_distance": 0,
+        }
+        with self._lock:
+            now = self._frame_time(objects, frame_ts)
+            self._now = now
+            scene_established_at_frame_start = len(self._entities) >= max(3, self._min_support)
```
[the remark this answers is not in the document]
jhengyilin: Adressed.

Fixed by separating persistent galleries from live candidates. Persisted entries no longer populate live tracks or affect new-object registration. A new observation follows the same rule.

dimos/perception/detection/world_belief.py
```diff
@@ -0,0 +1,752 @@
...
+        eid = force_id or obj.object_id
+        while force_id is None and eid in self._tracks:  # guard the (rare) id collision
+            eid = uuid.uuid4().hex[:8]
+        obj.object_id = eid  # the entity IS this Object; later folds accumulate into it
+        track = _Track(obj, ts, viewpoint, basis)
+        track.established = self._is_established(track, ts)  # true only for 1-frame/0-span configs
+        track.ingest_appearance(obj, self._gallery_novelty, self._gallery_max)
+        self._tracks[eid] = track
+        if self._gallery is not None and track.established:
+            self._gallery.remember(eid, obj, obj.center.to_numpy(), track.n_frames)
```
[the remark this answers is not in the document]
jhengyilin: Addressed.

Fixed by removing  the stale saved position and support count from identity matching. We now use the saved appearance only to recover the ID, while the object’s position and geometry always come from the latest observation.

dimos/perception/detection/world_belief.py
```diff
@@ -0,0 +1,617 @@
...
+        rows, cols = linear_sum_assignment(scores, maximize=True)
+        best_total = float(scores[rows, cols].sum())
+        matched: list[tuple[_Track | str, Object]] = []
+        used_tracks: set[int] = set()
+        used_objs: set[int] = set()
+        # Do not mint identities from ambiguous appearance evidence.
+        ambiguous_rows = set(np.flatnonzero(np.any(appearance >= self._cfg.reacq_cos, axis=1)))
+        for row, obj in enumerate(objects):
+            if _appearance_vec(obj) is None and obj.name in labels:
+                ambiguous_rows.add(row)
```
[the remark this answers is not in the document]
jhengyilin: Addressed. 

Now the detection remains visible as unresolved when its embedding is missing. We don’t assign or create an identity until DINO evidence becomes available again.

dimos/perception/worldbelief_module.py
```diff
@@ -0,0 +1,396 @@
...
+                        prompt=prompts,
+                    )
+                    if prompt_changed and result.folded_frames == 0:
+                        raise ScanIncompleteError(
+                            "no new frame was available for the changed prompt vocabulary"
+                        )
+                    self._active_scan_prompts = vocabulary
+                    # Out.publish may invoke in-process consumers with the same object.
+                    # Deep snapshots keep those consumers out of mutable belief state.
+                    present = _copy.deepcopy(result.objects)
```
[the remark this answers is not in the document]
jhengyilin: Addressed. 

Unresolved observations still appear in the scan metadata so we don’t hide uncertain evidence, but downstream outputs now contain only established objects. Recall still returns its best matching remembered moment, but it only includes where_object when the detector and identity checks establish the object.

jhengyilin:
# xArm6 WorldBelief — pulled perception with cross-session memory

This supersedes the earlier spatial-radius/Q1–Q4 overview. The final implementation uses
global DINO appearance matching for identity. 3D evidence is used for current geometry,
viewpoint trust, and absence—not as a movement-distance identity heuristic.

```mermaid
flowchart TB
    CAM["RealSense<br/>color + depth + camera pose"] --> REC["WorldBeliefRecorder<br/>record plane"]
    REC --> SESSION[("timestamped session DB<br/>raw evidence")]
    SESSION -- "scan / recall<br/>on demand" --> WB["WorldBeliefModule<br/>query plane"]
    HISTORY[("worldbelief_history.db<br/>DINO galleries + CLIP index")] <--> WB
    WB --> REPLY["scan result<br/>established + unresolved observations"]
    WB --> OBJECTS["established objects snapshot"]
    WB --> VIS["established detections_3d · pointcloud"]
    CAM --> RERUN["Rerun validation"]
    VIS --> RERUN
    OBJECTS -. "future integration PR" .-> DOWNSTREAM["PickAndPlace / navigation"]
```

## Two modules

- **`WorldBeliefRecorder`** is the record plane. It continuously writes RGB-D, camera info,
  joint state, and resolved pose evidence to a new timestamped SQLite recording for each run.
- **`WorldBeliefModule`** is the query plane. It owns the YOLO-E/DINO/CLIP model instances, the
  mutable `WorldBelief` fold, persistent history, and the `scan`/`recall` skills. Scan and
  recall share one instance lock.

The Recorder performs no detection. `WorldBeliefModule` warms YOLO-E and DINO at startup so
`scan` is ready; CLIP remains lazy until the first `recall`. 
##  `scan` 

1. The xArm blueprint has no built-in vocabulary, so each call supplies a non-empty prompt.
2. The first call reads at most `window` seconds of history. Later calls fold every eligible
   frame after the belief watermark, so `window` does not limit cross-scan catch-up.
3. Pose deltas select coarse stationary frames and denser moving-camera frames. The xArm
   blueprint uses 4 Hz while stationary; moving sampling is 7.5 Hz. First and last eligible
   frames are retained.
4. YOLO-E segments prompted objects, RGB-D and TF lift them into configured-target-frame
   `Object`s, and DINO embeds each accepted crop.
5. Frames fold chronologically into the module-owned belief. If RGB-D/TF evidence cannot reach
   the recording end, the call fails instead of returning a stale success.
6. The scan result reports established and unresolved observations. After releasing the engine
   lock, the module publishes established `Object`, `Detection3DArray`, and aggregate
   point-cloud snapshots.


## Identity and long-term memory

```mermaid
flowchart TB
    DET["lifted detection<br/>label + current 3D geometry<br/>+ DINO when available"] --> CAND["candidate identities<br/>RAM tracks + retrieved persisted galleries"]
    CAND --> ASSIGN["global one-to-one<br/>DINO cosine assignment"]
    ASSIGN -- "clear winner" --> MATCH["tracked / reacquired / restored"]
    ASSIGN -- "near tie or missing embedding<br/>for a known label" --> ABSTAIN["retain unresolved observation<br/>do not guess or persist identity"]
    ASSIGN -- "no plausible identity" --> NEW["new candidate ID"]
    NEW --> EST["established after<br/>3 frames AND 1.5 s"]
```

- Same-label matches require cosine ≥ 0.5. A label change requires the stronger ≥ 0.8
  appearance override.
- The assignment is global, so two observations cannot claim the same identity in one frame.
  A rival within the 0.05 margin is ambiguous and is not silently converted into a new ID.
- 3D position does not arbitrate identity. A moved object may retain its ID after an arbitrary
  gap when its DINO evidence is distinctive. Ambiguous or same-label embedding-less detections
  remain visible in scan metadata with `basis: unresolved`, but do not enter identity tracks,
  galleries, or trusted publications.
- Unestablished candidates older than five seconds are excluded before association, so stale
  support cannot establish an expired identity.
- Established tracks are `tentative` after one viewpoint and `confirmed` after two sufficiently
  distinct viewpoints. This is identity-evidence status, not manipulation or collision safety.
- Partial or implausibly enlarged observations may refresh the visible center without replacing
  trusted shape geometry.
- An object becomes absent only after two depth frames see through its expected location.
  Occlusion and an out-of-view camera provide no absence vote.
- Established identities persist as bounded, diverse DINO galleries in
  `worldbelief_history.db`. Galleries are namespaced by exact checkpoint and embedding size, so
  incompatible model spaces are not mixed after a configuration change.

Identity basis in scan results describes the latest evidence:

| Basis | Meaning |
|---|---|
| `new` | newly created candidate |
| `tracked` | matched in the next folded frame |
| `reacquired` | matched to an in-memory identity after a gap or absence |
| `restored` | matched to an identity persisted by an earlier process/session |
| `unresolved` | current physical observation without a defensible durable identity |

## `recall`

`recall(text)` incrementally indexes immutable session frames with CLIP into the same history
database. Frames are processed chronologically and sampled into stable time buckets; the source
watermark advances only after an indexing top-up completes.

CLIP ranks candidate moments across sessions but is not trusted as localization. Distinct top
moments are reopened from their source recordings and checked with YOLO-E + DINO over a small
temporal window. The response includes the recording, timestamp, camera position, uncalibrated
CLIP similarity, and detector-confirmed `where_object`. If no candidate is confirmed, recall
still returns the best CLIP-ranked moment with `where_object: None`. If the index has no matching
moment, recall returns `None`.

## Configuration

Deployment policy stays in the xArm blueprint; the generic perception modules contain no xArm
paths or vocabulary. The main model and sampling choices are blueprint fields:

```python
WorldBeliefModule.blueprint(
    scan_prompts=[],
    stationary_hz=4.0,
    yoloe_model_name="yoloe-11l-seg.pt",
    dino_model_name="facebook/dinov2-base",
    clip_model_name="openai/clip-vit-base-patch32",
)
```

Persistent gallery and CLIP streams are isolated by checkpoint ID, so changing DINO or CLIP
does not reuse incompatible embeddings.


## How to test

```bash
# terminal 1
uv run dimos --xarm6-ip <ROBOT_IP> run xarm6-worldbelief

# terminal 2: an explicit prompt is required
uv run dimos mcp call scan -a prompt='["coke can", "mug", "yellow can"]'
uv run dimos mcp call recall -a text="mug"
```

Expected checks:

- repeated scans keep a distinctive object's full ID;
- unresolved observations remain visible in scan results but stay out of trusted publications;
- a moved or previously absent object reports `basis: reacquired`;
- after restart, a gallery match reports the same full ID with `basis: restored`;
- two novel camera viewpoints promote an established identity from `tentative` to `confirmed`;
- unconfirmed recall still returns the remembered moment with `where_object: None`;
- Rerun shows live RGB/depth plus published 3D detections and point cloud.

mustafab0 (approved):

dimos/perception/detection/world_belief.py
```diff
@@ -0,0 +1,617 @@
...
+                    scores[row, offset] = cos
+                    valid[row, offset] = True
+
+        rows, cols = linear_sum_assignment(scores, maximize=True)
+        best_total = float(scores[rows, cols].sum())
+        matched: list[tuple[_Track | str, Object]] = []
+        used_tracks: set[int] = set()
+        used_objs: set[int] = set()
+        # Do not mint identities from ambiguous appearance evidence.
+        ambiguous_rows = set(np.flatnonzero(np.any(appearance >= self._cfg.reacq_cos, axis=1)))
```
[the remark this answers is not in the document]
jhengyilin: Addressed.

Previously, ambiguous detections could fall between WorldBelief’s restore and new-candidate paths, causing them to disappear from the scan entirely. WorldBelief now keeps them visible in
SceneScan.objects as `unresolved`.

They do not claim a persisted ID, create a new identity, enter the DINO embedding gallery, or get published through belief.present() as trusted state. Each later frame goes through the same matching
pipeline, so conclusive DINO evidence can still resolve the object normally.

PR #2690, opened 2026-07-01
fix: `dimos map global` and `dimos map replay` to use tf

jeff-hykin:
Vibed extraction of some of my existing map/db changes. Will check soon.

## Examples

Rebuild the global map from a recording (registers clouds into `world` via the recording's tf tree, renders april tags if an `april_tags` stream is present):

```bash
dimos map global data/china_office.db --lidar pointlio_lidar --voxel 0.05
```

Plan on the reconstructed map and write a planning .rrd:

```bash
uv run python -m dimos.navigation.nav_3d.mls_planner.utils.plan_rrd data/china_office.db --robot-height 0.4 --lidar-stream pointlio_lidar --odom-stream pointlio_odometry
```


dimos/memory2/store/sqlite.py
```diff
@@ -210,6 +214,23 @@ def list_streams(self) -> list[str]:
...
+        Recordings may reference Python types not present in this checkout;
+        those streams are warned about and omitted rather than crashing the
+        whole summary. Requesting such a stream by name still raises.
+        """
+        lines: list[str] = []
+        for name in self.list_streams():
+            try:
+                stream: Stream[Any] = self.stream(name)
+            except (ImportError, AttributeError, ValueError) as error:
+                logger.warning(f"skipping stream {name!r}: unresolved payload type ({error})")
```
leshy, line 229: why do we have unreslovable type in the database now?

jeff-hykin: If people save custom types (ex: Graph3D on my gsc_pgo branch) it would be nice for the map to still load the db and ignore those unknown types

PR #2692, opened 2026-07-02
docs: clarify native module config migration

dashitongzhi:
## Summary

* Document DimOS module config as the preferred source of truth for native module settings
* Add guidance for migrating native-only CLI flags into typed NativeModuleConfig fields
* Include examples for blueprint and CLI ooptionption overrides while discouraging normal-use extra_args passthroughs

## Test Plan

* Ran `git diff --check`


paul-nechifor:
@dashitongzhi Hi. Thanks for the contribution.

Can you please fix the CI issue? We have a test to make sure we use `dimOS` in docs and not `DimOS`.

PR #2570, opened 2026-06-23
feat: multi level navigation

aclauer:
## Problem

3d navigation and ray tracing integration

Closes DIM-XXX

## Solution

Creates a blueprint using ray tracing and mls planner, plus improvements to other modules. Click and point to navigate to goals.

Ray tracer:
- register point clouds with odometry (breaks on old fastlio recordings!)
- outputs region bounds for robust updates

Planner:
- incremental builds of planner artifacts in separate task so ingest isn't blocked
- path replan on each frame
- string pulled paths with tunable cost metrics
- levers for path safety

Other things:
- add set_motion_mode() to go2 connection, so we can turn off obstacle avoidance and things when instantiating the blueprint
- add a simple path follower. Replace this with better trajectory controllers for better results.

## How to Test

build native modules:
```
cd ~/dimos   && ( cd dimos/hardware/sensors/lidar/pointlio/cpp && nix build .#pointlio_native )   && ( cd dimos/mapping/ray_tracing/rust && nix build path:. )   && ( cd dimos/navigation/nav_3d/mls_planner/rust && cargo build --release )
```

run! (on jetson on go2)
```
dimos --dtop --rerun-host 0.0.0.0 --rerun-open none run unitree-go2-nav-3d
```

other things not tested but would be awesome to try:
1. running with go2 stock lidar
2. running directly on go2 compute

## Contributor License Agreement

- [x] I have read and approved the [CLA](https://github.com/dimensionalOS/dimos/blob/main/CLA.md).

PR #2697, opened 2026-07-02
holonomic trajectory controller

bogwi:
# Add Holonomic Trajectory Controller

This PR ships two modules:

1. Holonomic Trajectory Controller
2. Local Planner

## Holonomic Trajectory Controller, htc

Why write HTC when differential already exists?

`PController` works for a classic diff-drive mental model: point at lookahead, drive forward, rotate in place when badly misaligned. That is simple and robust, which is why it remains the default in `unitree_go2`.

`htc` extends to:

1. **Strafe onto the path** - `HolonomicTrackingController` maps lateral error to body `vy`, so the robot slides back onto the polyline while still moving - on bends and right after a replan nudge.
2. **Keep moving through corners** - HTC tracks the path tangent while translating (default: no pre-align spin), so polylines and MLS hot updates stay continuous instead of stop-spin-go.
3. **Track pose and speed on the path** - each tick HTC closes on a `TrajectoryReferenceSample`: lookahead pose on the path plus body-frame feedforward twist from the path-speed profile. `PController` only steers toward a lookahead XY point at fixed cruise speed.
4. **Speed and commands matched to the path** - `profile_speed_along_polyline` slows for curvature and goal decel; `clamp_holonomic_cmd_vel` slews planar and yaw limits from run profiles. Odometry-derived `kv`/`kw` damp overshoot when feedforward changes. Differential has none of this - scalar speed, no lateral DOF, no geometry caps.

Htc also borrows some functionlity from the excellent "differential" ReplanningAStar and adapts it to its own workflow.

## DanLocalPlanner

Is a local planner module, as it says, and has two features to support MLSPlanner performance coming from kronk-nav.

a) Lock replan, meters - to stabilize the heavy drifting path the global planner supplies. Useful in controlled, constrained environments requiring stable paths over a certain distance. By default it is set to `0.0` but you can experiment locking the path as `DanLocalPlanner.blueprint(lock_replan=meters)` (see `unitree_go2_mls_htc.py`)

b) Path smoothing (borrows directly from the existing stack) - to smooth out the path's jagged edges. As with (a), very useful in a micro environment with many turns.

Eventually both features might migrate to MLSPlanner.

## How to run on bare Go2 native lidar

Very simple:

`dimos --robot-ip <IP> run unitree-go2-mls-htc -o danholonomictc.run_profile=<profile>`

There are three run profiles at the moment, think presets with nominal engineering envelopes . But all you need to know is they cap max speed.

walk - 0.55 m/s<br>trot - 1.0 m/s<br>run_conservative - 1.5 m/s

walk is default so you can just run like this:<br>`dimos --robot-ip <IP> run unitree-go2-mls-htc`

## How to run on Go2 + mod360 + Jetson

Blueprint that works with full kronk-nav dimensionalOS/dimos#2570 will be shared soon. But if you assembled own setup you can just swap DanLocalPlanner and DanHolonomicTC modules into unitree_go2_nav_3d.py instead of BasicPathFollower.

## Next

Must know:

**Run profiles are in m/s but the Go2 driver is not.**

`DanHolonomicTC` (and `MovementManager` before it) publishes `cmd_vel` in real body-frame m/s and rad/s - walk 0.55, trot 1.0, run_conservative 1.5, corner caps, `vy` strafe, all of it. `UnitreeWebRTCConnection.move()` in `dimos/robot/unitree/connection.py` does not send velocities. It maps `twist.linear.x/y` and `twist.angular.z` straight onto `WIRELESS_CONTROLLER` stick fields `lx`/`ly`/`rx`.

**Next PR will be: metric** `cmd_vel` **behind a toggle.**

Sport `Move` on `GO2Connection` gated by a config flag (joystick API vs velocity API). Default stays `WIRELESS_CONTROLLER` so `unitree_go2` and current controller keep today's behavior. `unitree-go2-mls-htc` (and anything else that outputs metric twist) opts into the velocity API. Until that lands, treat on-robot HTC speed as uncalibrated - the control law is in this PR and the proper speed execution is the follow-up.


dimos/navigation/dannav/local_planner/module.py
```diff
@@ -0,0 +1,245 @@
...
+        """Robot arc-length along the committed path, or 0 if it can't be measured."""
+        if self._committed is None or self._robot_xy is None:
+            return 0.0
+        return self._committed.project(self._robot_xy).s_along_path_m
+
+
+def _ends_near(path: Path, goal_xy: NDArray[np.float64], tolerance_m: float) -> bool:
+    """True when ``path``'s last pose is within ``tolerance_m`` of ``goal_xy``."""
+    end = path.poses[-1]
+    return float(math.hypot(end.x - goal_xy[0], end.y - goal_xy[1])) <= tolerance_m
```
bogwi, line 204: SPAM. PoseStamped inherits from Pose. Pose.py, line 113-126,  Pose defines .x, .y, and .z as properties that read self.position.

leshy (approved):
clean, no objections, we will restructure dir arch later `dannav` should go somewhere common

PR #2707, opened 2026-07-03
memory2 tf service

leshy:
we converted recordings to sensor frame, mem now needs a tf service for global mapping


dimos/memory2/tf.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class StreamTF(MultiTBuffer, TFSpec):
+    config: StreamTFConfig
+
+    def __init__(self, stream: Stream[TFMessage] | None = None, **kwargs: Any) -> None:
+        if stream is not None:
+            kwargs["stream"] = stream
+        TFSpec.__init__(self, **kwargs)
+        MultiTBuffer.__init__(self, buffer_size=math.inf)
```
paul-nechifor, line 45: Why infinity? Is it wise to never evict? Was this meant to be `self.config.cache_span?

dimos/memory2/tf.py
```diff
@@ -0,0 +1,110 @@
...
+    def publish(self, *args: Transform) -> None:
+        raise NotImplementedError("StreamTF is a read-only replay service.")
+
+    def publish_static(self, *args: Transform) -> None:
+        raise NotImplementedError("StreamTF is a read-only replay service.")
+
+    def _load(self, lo: float, hi: float) -> None:
+        for obs in self.stream.at((lo + hi) / 2, (hi - lo) / 2):
+            self.receive_transform(*obs.data.transforms)
+        self._covered = (lo, hi)
```
paul-nechifor, line 68: Don't you need to use `self._cv` for modifying `self._covered` too?

dimos/memory2/tf.py
```diff
@@ -0,0 +1,110 @@
...
+            child_frame,
+            time_point,
+            time_tolerance,
+            forward_tolerance=0.0,
+        )
+
+
+if TYPE_CHECKING:
+    # mypy conformance check: StreamTF satisfies the read-side tf protocol.
+    _lookup_impl: TFLookup = cast("StreamTF", None)
```
paul-nechifor, line 111: Is this used anywhere?

dimos/protocol/tf/tf.py
```diff
@@ -423,3 +446,7 @@ class LCMTF(PubSubTF):
 
 
 TF = LCMTF
+
+if TYPE_CHECKING:
+    # mypy conformance checks: the live services satisfy the read-side protocol.
+    _lookup_impls: tuple[type[TFLookup], ...] = (MultiTBuffer, PubSubTF)
```
paul-nechifor, line 452: Is this used anywhere?

paul-nechifor (dismissed):

dimos/protocol/tf/tf.py
leshy: yeah this makes mypy validate my TFLookup protocol conformance

dimos/memory2/tf.py
leshy: it wasn't, this is a bit hacky, this is a smart timeseries buffer that evicts old messages, but in our case we are controlling eviction totally differently, this buffer is manually cleared by the tf service..

abstractions are slightly broken here, tf should actually inherit from some generic timeseries store, that can be backed by a mem2 stream that can conform that timeseries spec, this way TF service would not need changes,

but need to move now and this is computationally efficient enough, and won't be built upon

dimos/memory2/tf.py
leshy: yeah this makes mypy validate my TFLookup protocol conformance

dimos/memory2/tf.py
leshy: yes

paul-nechifor (approved):

PR #2712, opened 2026-07-03
chore: gate backport check on a release-in-flight label

spomichter:
## Problem

The "release in flight" state that drives the backport-intent gate is inferred from the existence of any `backport release/X.Y.Z` label. That overloads those labels: they're the backport-bot trigger *and* the record of which PRs shipped into a release *and* the on/off switch for the gate.

The practical downside is cleanup. To stop the gate blocking every PR after a release wraps, you have to delete the `backport release/X.Y.Z` labels — which also strips them off the merged PRs they tagged, so we lose the record of what was backported. Just hit this cleaning up after 0.0.13 / 0.0.13.post1.

## Solution

Split the switch from the record. A single `release-in-flight` label now toggles enforcement:

- present → PRs to `main` must carry a `backport release/X.Y.Z` label or `backport:skip`
- absent → gate is a no-op

The `backport release/X.Y.Z` labels keep their existing jobs (backport-bot trigger + author intent) and are never deleted, so they persist on merged PRs as the shipped-in-X.Y.Z record. Ending a release is now deleting one throwaway control label instead of the version labels.

Considered two alternatives and went with the label:
- **repo variable** (`RELEASE_IN_FLIGHT`) — cleanest decoupling, but invisible in the PR/labels UI; you can't tell a release is active without opening repo settings.
- **open milestone** — more moving parts and a naming convention to maintain.

The label wins on being one visible, greppable toggle that anyone with triage rights can flip, and it needs no new plumbing since the gate already reads labels.

Runbook updated: create `release-in-flight` when cutting (§1.5), delete it at cleanup only if no other release is still active (§3.2).

## How to Test

n/a — CI/labels workflow change. The gate re-runs on the next `pull_request_target` event; with no `release-in-flight` label present it logs `No 'release-in-flight' label; nothing to enforce` and passes.

### DIFF
--- a/.github/workflows/pr-labels.yml
+++ b/.github/workflows/pr-labels.yml
@@ -6,9 +6,11 @@ name: pr-labels
 # synchronize` clears the stale label on a new push before CI finishes on
 # the new commit.
 #
-# `backport-check`: when a `backport release/X.Y.Z` label exists (release in
-# flight), every PR to main must declare backport intent via that label or
-# `backport:skip`.
+# `backport-check`: while the `release-in-flight` label exists on the repo,
+# every PR to main must declare backport intent via a `backport release/X.Y.Z`
+# label or `backport:skip`. Toggling that one label turns enforcement on and
+# off; the per-version `backport release/X.Y.Z` labels are left in place as a
+# record of what was backported into each release.
 #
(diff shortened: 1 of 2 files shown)


Dreamsorcerer:
This is adding more work, not reducing it. Just need to follow the cleanup instructions after a release and it already works correctly..

PR #2717, opened 2026-07-03
Add a queue to shmrpc

Dreamsorcerer:
I'm not entirely sure if it might be more appropriate to use this everywhere for SHM or not. But, for RPC atleast, this fixes the concurrency tests.


leshy:
> I'm not entirely sure if it might be more appropriate to use this everywhere for SHM or not. But, for RPC atleast, this fixes the concurrency tests.

Likely everywhere, but zenoh supports shm so we'll sunset this transport anyway

leshy (approved):

PR #2725, opened 2026-07-04
feat: blueprint namespaces

### LINKED ISSUE #2491: Blueprint Namespaces proposal for multi-embodiment & teleop & multi-blueprint
Need some idea of Namespaces on a blueprint level that would perform conditional transport routing ONLY to other modules in that given namespace

And allow duplicate modules to be published

Req:

* Namespaces can be created on a blueprint level that would classify all modules under that given namespace. Perhaps implemented:

```python
blueprint = autoconnect(
    ...
).global_config(namespace="robot1")
```

* Modules can be added to Namespaces at runtime
* Transports can be routed between Namespaces via some transport prefix or operator

``` 
.transports({
      ("joint_state", JointState): LCMTransport("{namespace}",/joint_state", JointState), # kind of horrible but something like this
  })
```

OR

```
      ("joint_state", JointState): LCMTransport("state", JointState, namespace="robotB"),  
```

* Should have helpers to do operations on namespaces, trade modules, etc

---
*Synced from* [DIM-1012](https://linear.app/dimensional/issue/DIM-1012)

paul-nechifor:
## Problem

* A coordinator could only run one instance of each module class.
* Stream topics, RPC topics and TF frames were global names, so one blueprint could not control several robots of the same type.

Closes #2491

## Solution

- Add instance name to module instances in the module coordinator, so you can run multiple instances.
- Add `namespace(prefix, *blueprints, expose=...)`. It prefixes instance names, stream topics, RPC topics and TF frames of everything inside. Streams listed in `expose` stay global so shared modules can consume them. Module references resolve inside the namespace first, then outward.
- Per instance config works as `-o robot0/go2connection.ip=10.0.0.5` or env `ROBOT0_GO2CONNECTION__IP=...`.
- Simulation support: with `--simulation` each robot gets its own MuJoCo process and the IPs are ignored. Fixed the hardcoded TF frames in `GO2Connection` (base_link, camera_link, camera_optical) to honor the namespace prefix. The parent world frame stays shared so all robots hang off one tree root.

## How to Test

```bash
ROBOT_IPS=10.0.0.1,10.0.0.2 dimos --simulation run unitree-go2-multi-teleop
```

Two MuJoCo windows and two teleop windows open. Click one teleop window and drive with WASD. Only that robot moves. `dimos topic echo /robot0/odom` shows the namespaced topics. On real robots use their IPs and drop `--simulation`.

### DIFF
--- a/docs/usage/blueprints.md
+++ b/docs/usage/blueprints.md
@@ -257,12 +257,12 @@ blueprint.remappings([
 
 You can use namespaces to control several robot instances.
 
-Use `namespace(prefix, *blueprints, expose=...)` to isolate a set of modules
+Use `.namespace(prefix, expose=...)` on a blueprint to isolate its modules
 under a name prefix so several copies can coexist. Because blueprints are plain
 Python values, a variable-size (and mixed-type) fleet is just a loop:
 
 ```python session=blueprint-ns
-from dimos.core.coordination.blueprints import autoconnect, namespace
+from dimos.core.coordination.blueprints import autoconnect
 from dimos.core.module import Module, ModuleConfig
 from dimos.core.stream import In, Out
 
@@ -281,7 +281,7 @@ robot_ips = ["10.0.0.1", "10.0.0.2"]
 fleet = autoconnect(
     AggregateMapper.blueprint(),   # shared: one instance for the whole fleet
     *[
-        namespace(f"robot{i}", Sensor.blueprint(ip=ip), expose={"pointcloud"})
+        Sensor.blueprint(ip=ip).namespace(f"robot{i}", expose={"pointcloud"})
         for i, ip in enumerate(robot_ips)
     ],
 )
--- a/docs/usage/blueprints.md
+++ b/docs/usage/blueprints.md
@@ -253,6 +253,85 @@ blueprint.remappings([
 })
 ```
 
+## Multi-robot blueprints (namespaces)
+
+You can use namespaces to control several robot instances.
+
+Use `namespace(prefix, *blueprints, expose=...)` to isolate a set of modules
+under a name prefix so several copies can coexist. Because blueprints are plain
+Python values, a variable-size (and mixed-type) fleet is just a loop:
+
+```python session=blueprint-ns
+from dimos.core.coordination.blueprints import autoconnect, namespace
+from dimos.core.module import Module, ModuleConfig
+from dimos.core.stream import In, Out
+
+class SensorConfig(ModuleConfig):
+    ip: str = ""
+
+class Sensor(Module):
+    config: SensorConfig
+    pointcloud: Out[str]
+
+class AggregateMapper(Module):
+    pointcloud: In[str]
+
+robot_ips = ["10.0.0.1", "10.0.0.2"]
+
+fleet = autoconnect(
+    AggregateMapper.blueprint(),   # shared: one instance for the whole fleet
+    *[
+        namespace(f"robot{i}", Sensor.blueprint(ip=ip), expose={"pointcloud"})
+        for i, ip in enumerate(robot_ips)
+    ],
+)
+```
+
+Inside a namespace everything is prefixed:
+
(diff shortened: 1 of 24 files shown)


docs/usage/cli.md
```diff
@@ -48,13 +48,13 @@ dimos [GLOBAL OPTIONS] COMMAND [ARGS]
...
-2. `.env` file → `DIMOS_SIMULATION=true`
-3. Environment variable → `export DIMOS_SIMULATION=true`
-4. Blueprint definition → `.global_config(simulation=True)`
+1. `GlobalConfig` default → `simulation = ""`
+2. `.env` file → `SIMULATION=mujoco`
+3. Environment variable → `export SIMULATION=mujoco`
+4. Blueprint definition → `.global_config(simulation="mujoco")`
 5. CLI flag → `dimos --simulation run ...`
 
-Environment variables and `.env` values must be prefixed with `DIMOS_`.
```
jeff-hykin, line 57: Why'd we change this? Seemed like a good pattern to me

paul-nechifor: This is they way it was initially, but Stash didn't like the long variable names. When I removed the DIMOS_ preifx, I forgot to update this place.

docs/usage/blueprints.md
```diff
@@ -209,6 +209,85 @@ blueprint.remappings([
...
+
+class AggregateMapper(Module):
+    pointcloud: In[str]
+
+robot_ips = ["10.0.0.1", "10.0.0.2"]
+
+fleet = autoconnect(
+    AggregateMapper.blueprint(),   # shared: one instance for the whole fleet
+    *[
+        namespace(f"robot{i}", Sensor.blueprint(ip=ip), expose={"pointcloud"})
```
leshy: why is it 

`namespace("prefix", blueprint)`

and not `blueprint.namespace("prefix")`

as it feels like this breaks the typical API we have so far? with methods modifying blueprints.


also, why isn't aggregatemapper implicit? since we have a passive data structure that's a blueprint, deploy mechanism can auto-include it?

paul-nechifor: Because `namespace` modifies a sequence of blueprints, not just one.

It's the same as `autoconnect`, which takes a sequence of blueprints.

I was actually thinking at one point to remove `autoconnect` and use something like:

```python
Module1.blueprint().join(
    Module2.blueprint(),
    Module3.blueprint(),
)
```

But they annoying part is that you have to pick a blueprint to come first, even though there's no significance to which one goes first.

---

Similarly, I could do:

```python
Module1.blueprint().namespace("prefix", Module2.blueprint(), Module3.blueprint())
```

But I think it kinda looks ugly.

> also, why isn't aggregatemapper implicit? since we have a passive data structure that's a blueprint, deploy mechanism can auto-include it?

Sorry, I'm not sure what you mean here. I'm just using an example module which doesn't have a namespace here. It could be anything.

leshy: > Because namespace modifies a sequence of blueprints, not just one.
> It's the same as autoconnect, which takes a sequence of blueprints.

I'd say clean way would be perhaps - autoconnect before namespacing or namespace in a loop? basically autoconnect bakes you a "one unit" that then you can apply stuff to it

if you have
```python
Module.blueprint().namespace("prefix")
```
multi module namespacing is

```python
autoconnect(blueprint1, blueprint2).namespace(...)
```

Since I can imagine a lot of usage will actually be something like:

```python
unitree_pair = autoconnect(
   global_mapper, 
   go2(ip=...).namespace("go2-1").
   go2(ip=...).namespace("go2-2")
)
```
just like your example is actually:

```python
fleet = autoconnect(
    AggregateMapper.blueprint(),   # shared: one instance for the whole fleet
    *[
        Sensor.blueprint(ip=ip), expose={"pointcloud"}).namespace(f"robot{i}")
```

leshy (approved):

PR #2730, opened 2026-07-06
mem summary cleanup, go2 replay cleanup, progress bars

leshy:
- `dimos mem summary` moved from `dimos map summary`

- mem2 blobstores can tell the total size of a stream

- rich shared cli progressbar compatible with mem2 streams

- dimos mem rerun uses get_data like other mem tooling (previusly it required the full db path)

Go2 `--replay` - resolves `go2_lidar`/`go2_odom` (mid360 recordings, e.g. china_office) with fallback to `lidar`/`odom` (older datasets, e.g. go2_hongkong_office)

### DIFF
--- a/dimos/memory2/utils/progress.py
+++ b/dimos/memory2/utils/progress.py
@@ -18,9 +18,10 @@
 
 Renders a rich progress bar: label, percent, count, data-seconds covered,
 speed relative to wall-clock (``x rt``), and per-frame latency. On completion
-the bar persists as one plain line. All bars share a single live display, so
-sequential or interleaved streams compose; call :meth:`ProgressBar.close` when
-abandoning a stream early (e.g. a bounded time window).
+the live bar is replaced by one plain persisted line. One bar is live at a
+time — starting a new bar finalizes a dangling one (rich allows a single live
+display); call :meth:`ProgressBar.close` when abandoning a stream early
+(e.g. a bounded time window).
 """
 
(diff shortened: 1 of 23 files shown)


dimos/memory2/utils/progress.py
```diff
@@ -12,51 +12,114 @@
...
 
-    def _progress(obs: Observation[Any]) -> None:
-        nonlocal seen, wall_start, last_wall, first_ts
+
+def _close_current() -> None:
+    if _current is not None:
+        _current.close()
+
+
+atexit.register(_close_current)  # restore the cursor if a bar is abandoned
```
paul-nechifor: I don't like the use of globals to do cleanup here. Why just just make `def progress` a context manager and handle cleanup there. No need for globals:

```
with progress(...) as bar:
    bar.advance(...)
```

leshy: Done in 5a432db65: progress() is now a context manager (contextlib.contextmanager yielding the bar); cleanup happens in the finally on exit, so the _current global and the atexit hook are removed. All call sites (replay.py, map.py, memory2/cli/render.py, go2/dds/cli/render.py) use 'with progress(...) as bar'. Added dimos/memory2/utils/test_progress.py covering finalize-on-exit, cleanup-on-exception, early break, and post-completion calls.

paul-nechifor (approved):

PR #2749, opened 2026-07-06
refactor(hardware): lazy __registry__.py manifest discovery for adapter registries

### LINKED ISSUE #2756: Adapter registry refactor
*Synced from* [DIM-1117](https://linear.app/dimensional/issue/DIM-1117) *by mustafa*

mustafab0:
The three hardware adapter registries (manipulators, drive_trains, whole_body) eagerly imported every adapter module at startup to discover them. 

Converge all three onto the ControlTaskRegistry pattern: each adapter package declares its factories in a dependency-free __registry__.py manifest (name -> "module:attr"); the shared LazyAdapterRegistry base resolves the import only on first create(). A missing SDK now fails loudly at create() with the real cause, and the name still shows up in available() (intentional semantics change).

##Reason to not use Blueprint AST scan method:

Blueprint AST scan works great for blueprints, however for `hardware adapters` AST scan can't reliably infer an adapter file just with naming. as duplicate names exist for example both whole body and drive trains have a `transport/adapter.py` 

There is no way of figuring out if its a adapter without importing it. which is what we want to solve.

Also, for tasks and adapters the use case is different. A user would design their own tasks as well as robot adapter and would want to load them which won't be possible with AST scan if the user is using DimOS as a python library

- [x ] I have read and approved the [CLA](https://github.com/dimensionalOS/dimos/blob/main/CLA.md).


dimos/hardware/drive_trains/flowbase/_registry.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2026 Dimensional Inc.
```
paul-nechifor, line 1: Technically, PEP8 recommends against using own dunder names:

Image

But I don't have a good alternative. `__init__.py` isnt a good idea. Maybe `registry_factories.py`? Not sure.

paul-nechifor (dismissed):

dimos/hardware/drive_trains/flowbase/_registry.py
mustafab0: changed name to `_registry.py`

paul-nechifor (approved):

PR #2760, opened 2026-07-07
fix: dimos topic send

paul-nechifor:
`dimos topic send` has been broken for certain messages since I removed `__init__.py` files. 

Closes DIM-1124


leshy (approved):
needs to become proto agnostic now with zenoh

leshy (approved):

PR #2765, opened 2026-07-07
Use mocks for unsupported platform test

Dreamsorcerer:
This allows the test to actually run in CI.


paul-nechifor (approved):

leshy:
I think we are actually deleting this sim?

Dreamsorcerer:
> I think we are actually deleting this sim?

Regardless, it's a small adjustment. If it gets deleted soon, nothing lost.

PR #2766, opened 2026-07-07
test(pubsub): lower flaky high-volume test to 500 messages

leshy:
test_high_volume_messages was flakey


dimos/protocol/pubsub/test_spec.py
Dreamsorcerer, line 387: This test isn't very clear.
We assert that we've received the correct number of messages, but we do that after checking a timeout.
Are the tests failing because they are slow and hitting the timeout, or because they are losing messages?

Dreamsorcerer: If it works in an async context, I'd be tempted to do something like:
```
def callback(...):
    received_messages.append(...)
    if len(received_messages) == num_messages:
        complete_event.set()

...

async with asyncio.timeout(2):
    await complete_event
```

Then we'd just get a timeout error instead, and it'd be obvious that we can just use this as a codspeed benchmark.

spomichter (approved):
Approving since change makes sense but also unsure what the test does and hard coding 5k messages seems arbitrary as we move to more transports.

leshy:
> Approving since change makes sense but also unsure what the test does and hard coding 5k messages seems arbitrary as we move to more transports.

Thinking was for a local transport if you cannot exchange 5k small messages it's not a functional transport

PR #2814, opened 2026-07-08
feat(msgs): typed CompressedImage sensor msg

spomichter:
## Problem
dimos has no first-class compressed image type. jpeg support is bolted onto individual transports (`JpegLcmTransport`, `JpegShmTransport`, nothing for zenoh/webrtc), `JpegShmTransport` loses timestamps on decode, and every new transport needs its own jpeg variant. meanwhile raw images are big — 720p rgb is 2.76MB per frame, ~40MB/s at camera rate. thats fine for shared memory on one host but it kills lcm the moment you leave it: one dropped udp fragment loses the whole frame.

`CompressedImage` is a first-class msg (jpeg/png bytes + timestamp + frame_id, wraps the `dimos_lcm` binding that already existed):

```
CompressedImage.from_image(img, format="jpeg", quality=75, max_width=None)  # encode
ci.decode()                     # -> Image, ts/frame_id survive the round trip
ci.lcm_encode() / lcm_decode()  # typed wire msg on any transport
ci.to_rerun()                   # rr.EncodedImage — viewer renders jpeg with NO pixel decode
```

depth/16-bit images can't go through jpeg — use `format="png"` (lossless, covers GRAY16). float DEPTH raises on both.

## Benchmark report: raw Image vs CompressedImage

(numbers produced with the bench harness in this PR; the codec pin is the bench mechanism, the wins are the type's)

full grid run on this dev box (x86_64, 32 cpus), 6 runs x 10 frames per cell, sequential round-trips. **what "latency" measures**: one `broadcast()` call until the subscriber callback fires, on localhost — so it includes the jpeg encode+decode cpu for the codec path but near-zero wire cost. thats why codec shows *higher* latency here: raw is two memcpys, codec pays ~5ms encode + ~5ms decode of real cpu. localhost is raw's best case — on a real 1Gbps link the raw frame pays ~22ms of transmit time alone vs <1ms for the jpeg, and over webrtc/wan its not close. what you buy for those ms: 8-40x less wire and the difference between frames arriving or not (see drops).

### jpeg codec cpu cost (median of 20, ms)

```
                    raw        jpeg       ratio      encode      decode
480p  q75        0.92MB        49KB         19x         3.0         3.1
720p  q75        2.76MB       142KB         19x         5.1         5.2
1080p q75        6.22MB       315KB         20x         7.6         6.6
720p  q50        2.76MB        69KB         40x         4.7         4.6
720p  q90        2.76MB       345KB          8x         5.0         5.5
```

### round-trip latency, median ms (LOST = zero frames ever arrived)

```
                     480p        720p       1080p
lcm   raw            LOST        LOST         5.5
lcm   codec           6.4        10.3        17.9
zenoh raw             0.7         0.9         6.9
zenoh codec           6.4         9.6        15.2
pshm  raw             0.2         0.6         1.6
pshm  codec           6.2         9.4        15.3
```

### round-trip latency, p95 ms

```
                     480p        720p       1080p
lcm   raw            LOST        LOST         6.5
lcm   codec           7.0        11.8        21.2
zenoh raw             0.7         1.0         8.6
zenoh codec           6.8        11.4        18.8
pshm  raw             0.4         1.4         3.3
pshm  codec           6.8        11.3        17.4
```

### dropped frames (of 60)

```
                     480p        720p       1080p
lcm   raw           60/60       60/60       43/60
lcm   codec          0/60        0/60       37/60
zenoh raw            0/60        0/60        0/60
zenoh codec          0/60        0/60        0/60
pshm  raw            0/60        0/60        0/60
pshm  codec          0/60        0/60        0/60
```

lcm raw 480p/720p lost *every single frame* on this box (untuned kernel udp buffers — `CI=1` skips the lcm autoconf sysctl tuning, so this is what a fresh install sees). note the 1080p codec drops too: 315KB @ q75 is already past what untuned lcm handles reliably — thats what the `quality`/`max_width` knobs are for (q50 = 69KB, zero drops).

### jpeg quality sweep (lcm, 720p)

```
             wire      median         fps       drops
q50          69KB         9.2         109        0/60
q75         142KB        10.3          97        0/60
q90         345KB        13.7          73       40/60
```

### sustained 14Hz 720p camera feed for 8s (the go2 workload)

```
              delivered hz (target 14)
lcm   raw     ▊ 0.8
lcm   codec   ██████████████ 14.0
zenoh raw     ██████████████ 14.0
zenoh codec   ██████████████ 14.0
```

this is the headline: publish a go2-style camera feed at 14Hz over lcm and the raw pipeline delivers **0.8 fps**. the codec pipeline delivers all 112 frames.

### throughput capacity, fps (1 / median latency)

```
                     480p        720p       1080p
lcm   raw            LOST        LOST         183
lcm   codec           156          97          56
zenoh raw            1484        1098         145
zenoh codec           157         105          66
pshm  raw            6239        1765         639
pshm  codec           162         107          65
```

codec capacity is cpu-bound (~1/encode+decode time) and the same on every transport — ~100fps at 720p, 7x a 14Hz camera. pshm raw shows why shm stays the on-host raw escape hatch: 1765fps at 720p, zero encode cost.

end-to-end replay benchmark (full nav stack, jetson-class 4-core profile: raw delivers 6.5 of 14 fps, compressed delivers all of it, with LESS total cpu) is in #2831 with charts.

notes: webrtc isnt in the grid because its loopback provider is an in-memory dict (stack overhead only, no real SCTP — thats P1 of the webrtc backend spec) and the cloudflare path is cred-gated. the vlm/agent blueprints (spatial, temporal-memory, agentic) are the biggest indirect winners — they re-encode jpeg per VLM call today, with option 5 they get wire bytes for free.

follow-ups (not this PR): option 5 migration per #2831 (`GO2Connection` -> `Out[CompressedImage]` + consumers, needs a replay-dataset compat answer since recorded datasets store raw `Image`), migrating `JpegLcmTransport`/`JpegShmTransport` onto CompressedImage, webrtc `ImageDecimation` emitting this type.

## Breaking Changes
None

## How to Test
```
CI=1 uv run pytest dimos/msgs/sensor_msgs/test_CompressedImage.py dimos/protocol/pubsub/benchmark/test_replay_bench.py -s
```
14 tests incl. real lcm+zenoh round-trips (skipped where native libturbojpeg is missing — and ci.yml now installs it so CI cant skip silently). benchmark alone, prints one grid at the end:
```
CI=1 uv run pytest dimos/protocol/pubsub/benchmark/test_replay_bench.py -k benchmark -s
```
e2e replay bench: `python -m dimos.protocol.pubsub.benchmark.tool_replay_bench --blueprint unitree-go2-basic --mode codec --duration 30 --out /tmp/bench` (re-ran post-trim: 473 frames / 33s = 14.3fps, full delivery). full suite + mypy run clean locally.

- [x] I have read and approved the CLA


dimos/memory2/codecs/test_codecs.py
```diff
@@ -134,7 +134,11 @@ def _jpeg_case() -> Case | None:
     from dimos.memory2.store.sqlite import SqliteStore
     from dimos.utils.data import get_data
 
-    db_path = get_data("go2_short.db")
+    try:
+        db_path = get_data("go2_short.db")
+    except RuntimeError:
+        # 84MB LFS pull — refused by the CI LFS guard on standard runners
```
Dreamsorcerer: We either need this for the test, in which case it should be self_hosted, or we don't, in which case don't try and fetch it at all.

dimos/conftest.py
Dreamsorcerer, line 1: I think me and Paul have differing opinions on this. But, I'd use an importorskip and rely on coverage to catch regressions in CI testing. It makes it easier for downstream packagers when testing.

dimos/msgs/sensor_msgs/CompressedImage.py
```diff
@@ -0,0 +1,136 @@
...
+if TYPE_CHECKING:
+    import rerun as rr
+
+    from dimos.msgs.sensor_msgs.Image import Image
+
+CompressionFormat = Literal["jpeg", "png"]
+
+
+@dataclass
+class CompressedImage(Timestamped):
```
leshy, line 35: you invented a message type? you cannot write custom msgs to sensor msgs like this. also likely no need to create a new msg type - see what ros ecosystem uses and respect their message format, I think it's some FoxgloveCompressedImage msg we already have in our types

leshy: I would also recommend to integrate this end-to-end, so add a go2 build that doesn't decompress incoming stream, passes straight into rerun, with some module in the middle able to use image

dimos/msgs/sensor_msgs/CompressedImage.py
```diff
@@ -0,0 +1,136 @@
...
+from dimos.types.timestamped import Timestamped, to_human_readable
+
+# cv2/turbojpeg/Image are imported lazily so this module stays cheap for
+# byte-level consumers that never encode or decode pixels.
+if TYPE_CHECKING:
+    import rerun as rr
+
+    from dimos.msgs.sensor_msgs.Image import Image
+
+CompressionFormat = Literal["jpeg", "png"]
```
leshy: h264 here no?

leshy (dismissed):
This definitely needs h264 support. h265 encoders are almost always in hardware and much cheaper

dimos/msgs/sensor_msgs/CompressedImage.py
TomCC7: ROS has official compressed image type: https://docs.ros2.org/foxy/api/sensor_msgs/msg/CompressedImage.html I believe this impl is mostly aligned, the header field misalignment is exposed by greptile comments below.

TomCC7:
> This definitely needs h264 support. h265 encoders are almost always in hardware and much cheaper

H264/265 are video codec. I don't think the abstraction fits here? Unless every frame is a key frame but it's very inefficient.

dimos/msgs/sensor_msgs/CompressedImage.py
leshy:
```
#string format # Specifies the format of the data
# # Acceptable values:
# # jpeg, png
```

this seems to be only jpeg/png but aparently people abuse and put h264 there since it's string

leshy:
> > This definitely needs h264 support. h265 encoders are almost always in hardware and much cheaper
> 
> H264/265 are video codec. I don't think the abstraction fits here? Unless every frame is a key frame but it's very inefficient.

every frame is a compressed video frame with occasional keyframes (same as any other h264 transport) it's a responsibility of a recevier to reconstruct full images if needed with a small keyframe aware reconstructor

discussion on this here - https://github.com/dimensionalOS/dimos/issues/2831

would be something like `self.compressed_input.pipe(frame_decoder).subscribe( ... you have image here`

also btw if h264 doesn't fit our abstraction - I'd argue the abstraction is wrong since obviously h264 is the correct format to pass video streams in the system (but I think our abstraction is ok)

dimos/msgs/sensor_msgs/CompressedImage.py
paul-nechifor: We also probably want JPEG-XL. It's the best format for 16bit images (which we want for depth images)

dimos/protocol/pubsub/benchmark/test_replay_bench.py
```diff
@@ -0,0 +1,200 @@
...
+import time
+from types import SimpleNamespace
+
+import numpy as np
+import pytest
+
+from dimos.core.transport import LCMTransport, ZenohTransport
+from dimos.msgs.sensor_msgs.CompressedImage import CompressedImage
+from dimos.msgs.sensor_msgs.Image import Image, ImageFormat
+from dimos.protocol.pubsub.benchmark.tool_replay_bench import CompressedCodec
```
paul-nechifor, line 29: I thought we wanted to get rid of CompressedCodec and convert Modules to use the `CompressedImage` message.

spomichter: This is just for the benchmark tests, not used

spomichter:
> This definitely needs h264 support. h265 encoders are almost always in hardware and much cheaper

Coming back to this... lets PR h264 in a separate PR it will be more invovled as would use existing ROS CompressedVideo handling multiple camras that output h264 by default so would be zero-encode needed

This is just for streams of CompressedImage which is needed for depth images for example. (depth is 16 bit)

dimos/msgs/sensor_msgs/CompressedImage.py
spomichter: Will add

dimos/conftest.py
spomichter: siding with paul prefernce here which i guess is that all tests should run in CI and thus hard fail if dep missing. 

should discuss this and in a future PR list all the system deps that would cause issues for some downstream packager and WARN instead of fail

dimos/conftest.py
```diff
@@ -168,6 +183,10 @@ def pytest_collection_modifyitems(config, items):
         "skipif_no_openai": (not os.getenv("OPENAI_API_KEY"), "OPENAI_API_KEY not set"),
         "skipif_no_alibaba": (not os.getenv("ALIBABA_API_KEY"), "ALIBABA_API_KEY not set"),
         "skipif_no_ros": (not _has_ros(), "ROS dependencies are not present"),
+        "skipif_no_turbojpeg": (
```
spomichter, line 238: turbojpeg system dep so will run it in CI and hard fail in CI, but skip if on user machine without turbojpeg installed, and then kind of solves for @Dreamsorcerer concern for downstream packagers since they'll run non-CI tests `CI=0 pytest` and thus turbojpeg shoudlnt block them either

dimos/msgs/sensor_msgs/CompressedImage.py
spomichter: fixed

dimos/memory2/codecs/test_codecs.py
spomichter: i resolved by seperate PR already in main

paul-nechifor (dismissed):

leshy (dismissed):

PR #2829, opened 2026-07-09
fix(config): merge nested module overrides

TomCC7:
## Problem

Nested Pydantic blueprint config objects are replaced by partial dotted CLI override dicts because deployment used a shallow `dict.update`. That drops blueprint-provided nested defaults.

Closes: N/A

## Solution

Recursively merge module deployment kwargs before Python module deployment. Non-mapping overrides still replace the existing value.

## How to Test

```bash
uv run pytest dimos/robot/cli/test_dimos.py -k 'nested_blueprint_config_defaults_survive_cli_override or blueprint_arg_help_nested_config_paths or blueprint_arg_help_uses_nested_backend_defaults'
uv run pytest dimos/core/coordination/test_worker.py -k worker_manager_parallel_deployment
uv run ruff check dimos/core/coordination/worker_manager_python.py dimos/core/coordination/test_worker.py dimos/robot/cli/test_dimos.py
```

### DIFF
--- a/dimos/core/coordination/test_worker.py
+++ b/dimos/core/coordination/test_worker.py
@@ -14,16 +14,14 @@
 
 from typing import TYPE_CHECKING
 
-from pydantic import Field
 import pytest
 
-from dimos.core.coordination.worker_manager_python import WorkerManagerPython, _merge_config_kwargs
+from dimos.core.coordination.worker_manager_python import WorkerManagerPython
 from dimos.core.core import rpc
 from dimos.core.global_config import GlobalConfig, global_config
-from dimos.core.module import Module, ModuleConfig
+from dimos.core.module import Module
 from dimos.core.stream import In, Out
 from dimos.msgs.geometry_msgs.Vector3 import Vector3
-from dimos.protocol.service.spec import BaseConfig
 
 if TYPE_CHECKING:
     from dimos.core.resource_monitor.stats import WorkerStats
@@ -81,15 +79,6 @@ class ThirdModule(Module):
     @rpc
     def get_multiplier(self) -> int:
         return self.multiplier
-
-
-class NestedConfig(BaseConfig):
-    a: int = 0
-    b: int = 2
-
-
-class ModuleWithNestedConfig(ModuleConfig):
-    sub: NestedConfig = Field(default_factory=lambda: NestedConfig(a=10, b=20))
 
 
 class HeavyModule(Module):
@@ -198,17 +187,6 @@ def test_worker_manager_parallel_deployment(create_worker_manager):
     module1.stop()
     module2.stop()
     module3.stop()
-
-
-def test_nested_blueprint_config_defaults_survive_cli_override():
-    blueprint_kwargs = {"sub": NestedConfig(a=10, b=20)}
-    cli_kwargs = {"sub": {"a": "1"}}
-
-    merged = _merge_config_kwargs(blueprint_kwargs, cli_kwargs)
-    config = ModuleWithNestedConfig(**merged)
-
-    assert config.sub.a == 1
-    assert config.sub.b == 20
 
 
 @pytest.mark.skipif_macos_bug
--- a/dimos/core/coordination/test_worker.py
+++ b/dimos/core/coordination/test_worker.py
@@ -14,14 +14,16 @@
 
 from typing import TYPE_CHECKING
 
+from pydantic import Field
 import pytest
 
-from dimos.core.coordination.worker_manager_python import WorkerManagerPython
+from dimos.core.coordination.worker_manager_python import WorkerManagerPython, _merge_config_kwargs
 from dimos.core.core import rpc
 from dimos.core.global_config import GlobalConfig, global_config
-from dimos.core.module import Module
+from dimos.core.module import Module, ModuleConfig
 from dimos.core.stream import In, Out
 from dimos.msgs.geometry_msgs.Vector3 import Vector3
+from dimos.protocol.service.spec import BaseConfig
 
 if TYPE_CHECKING:
     from dimos.core.resource_monitor.stats import WorkerStats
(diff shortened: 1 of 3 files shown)


dimos/core/coordination/test_worker.py
```diff
@@ -189,6 +200,17 @@ def test_worker_manager_parallel_deployment(create_worker_manager):
     module3.stop()
 
 
+def test_nested_blueprint_config_defaults_survive_cli_override():
```
Dreamsorcerer: Existing blueprint tests are in test_dimos.py, can you move the test there and follow the format there (which means classes defined inside the test).

dimos/core/coordination/test_worker.py
```diff
@@ -81,6 +83,15 @@ def get_multiplier(self) -> int:
         return self.multiplier
 
 
+class NestedConfig(BaseConfig):
+    a: int = 0
+    b: int = 2
+
+
+class ModuleWithNestedConfig(ModuleConfig):
+    sub: NestedConfig = Field(default_factory=lambda: NestedConfig(a=10, b=20))
```
Dreamsorcerer: I'm not clear if it makes sense to have a lambda here. Why have defaults on the config, and then replace them with different defaults?

dimos/core/coordination/test_worker.py
Dreamsorcerer: Actually, I'm not too clear on this test.

The test uses _merge_config_kwargs() directly, which doesn't really test that the user flow will work correctly..

dimos/core/coordination/worker_manager_python.py
```diff
@@ -30,6 +32,26 @@
...
+    if not isinstance(override_value, Mapping):
+        return override_value
+
+    if isinstance(base_value, BaseModel):
+        base_value = base_value.model_dump(mode="python")
+
+    if not isinstance(base_value, Mapping):
+        return dict(override_value)
+
+    return _merge_config_kwargs(base_value, override_value)
```
Dreamsorcerer: Should we just expect dicts only and reject BaseModel at the blueprint? Then we don't need to handle BaseModel here and can do a simpler dict merge.

dimos/core/coordination/worker_manager_python.py
```diff
@@ -161,7 +183,11 @@ def deploy_parallel(
             module_class, _, kwargs = specs[i]
             worker = self._select_worker(dedicated=module_class.dedicated_worker)
             worker.reserve_slot()
-            kwargs.update(blueprint_args.get(module_class.name, {}))
+            specs[i] = (
+                specs[i][0],
+                specs[i][1],
+                _merge_config_kwargs(kwargs, blueprint_args.get(module_class.name, {})),
+            )
```
Dreamsorcerer: I feel like we could keep the kwargs mutation, rather than creating a new dict and needing to mutate specs instead.

dimos/core/coordination/test_worker.py
Dreamsorcerer: Yeah, I think actually, this probably makes more sense to have something in test_blueprints.py or similar that validates from the user input down to the worker manager resolving it.

dimos/core/coordination/worker_manager_python.py
TomCC7: Kept BaseModel support here. Blueprint kwargs are currently normal Pydantic config inputs, and arg_help already handles BaseModel defaults, so making blueprint kwargs dict-only would be a broader API change than this fix. I'm already using some nested basemodel config in manipulation module and it would be hard to change that part to dict. (https://github.com/dimensionalOS/dimos/blob/main/dimos/manipulation/manipulation_module.py#L119)

dimos/core/coordination/worker_manager_python.py
TomCC7: Done in ae1a7d8e9. The deploy loop now mutates kwargs in place instead of replacing the specs entry.

dimos/core/coordination/test_worker.py
TomCC7: Done in ae1a7d8e9. I moved the regression into dimos/robot/cli/test_dimos.py with classes defined inside the test, and it now exercises load_config_args plus WorkerManagerPython.deploy_parallel instead of calling the merge helper directly.

dimos/core/coordination/test_worker.py
TomCC7: Removed this test shape in ae1a7d8e9. The new regression uses normal config defaults and a BaseModel instance as the blueprint-level override, which is the behavior the merge needs to preserve.

dimos/core/coordination/worker_manager_python.py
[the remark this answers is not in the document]
TomCC7: Leaving backend-switch semantics out of this small fix. This PR only preserves existing nested blueprint values under partial overrides; discriminator/backend replacement behavior should get its own follow-up change and test.

Dreamsorcerer: I'm not sure I was very clear here. The configs support nested configs, that's fine and needs to still be supported.

What I'm questioning is whether we want to allow them as user inputs. CLI, env, and config files are all dict inputs, it's not possible to have a BaseModel here. The only place a user can create a BaseModel is when creating the blueprint. i.e.

`blueprint = TestModule.blueprint(nested=NestedConfig(mode="manual"))`

I'm wondering if we really want to support this and have the extra logic in merge, or if we just force users to do:

`blueprint = TestModule.blueprint(nested={"mode": "manual"})`

That is the only change that would be required. With either approach, the final result after config loading is complete will be a NestedConfig, and any internal blueprint code will be operating on pydantic models, not dicts. It just means less logic to handle in the merge code.

TomCC7: Thanks, understood. Yeah we should not need to use direct type in blueprint override and it's simpler to handle

Dreamsorcerer (approved):

mustafab0 (approved):

PR #2846, opened 2026-07-10
fix(arduino): give _FakeTransport the channel property _collect_topics reads

leshy:
#2753 switched `NativeModule._collect_topics` to read `transport.channel`; the arduino tests (#1879) were developed against the older `.topic` attribute and merged after, so `_FakeTransport` resolves to an empty topic map — 3 tests red on main, blocking every PR's CI.

One-line fix: mirror `PubSubTransport.channel` on the fake.


leshy:
Folded into #2845.

PR #2860, opened 2026-07-10
Planning Group 5: Add group-aware Viser planning

TomCC7:
## Summary
- add the planning-group-aware Viser manipulation panel while preserving the UI contract from `cc/spec/movegroup@0edb8d3dd`
- add synchronized, group-native multi-robot plan preview with safe cancellation and freshness checks
- replace the robot-scoped Viser adapter with a visualization-local group backend
- rename `dual-xarm6-planner-coordinator-mock-meshcat` to `dual-xarm6-planner-coordinator` and default that blueprint to Viser

## Validation
- 152 required manipulation and Viser tests passed with no skips
- targeted mypy passed for all changed production files
- Ruff, formatting, and `git diff --check` passed
- strict OpenSpec validation passed
- blueprint tests: 4 passed
- generated blueprint registry test: 1 passed

## Notes
- UI parity was reviewed against `0edb8d3dd` only; the prior upstream extraction commit was not used
- UI review was code-level/manual-contract validation, not a live browser or hardware run


dimos/manipulation/manipulation_module.py
```diff
@@ -869,24 +906,41 @@ def plan_to_joint_targets(
         self, joint_targets: Mapping[PlanningGroupID | PlanningGroup, JointState]
     ) -> bool:
         """Plan to joint targets keyed by planning group."""
+        return self._plan_to_joint_targets(joint_targets) is not None
+
+    @rpc
+    def plan_to_joint_targets_with_receipt(
```
TomCC7: why is this rpc required?

dimos/manipulation/manipulation_module.py
```diff
@@ -1326,42 +1346,97 @@ def execute_plan(
...
         """Execute a generated planning-group plan through affected trajectory tasks."""
+        if plan is not None:
+            return self._execute_plan(plan, robot_name=robot_name)
+        reserved_plan = self._reserve_stored_plan()
+        if reserved_plan is None:
+            return False
+        return self._execute_plan(reserved_plan, robot_name=robot_name, execution_reserved=True)
+
+    @rpc
+    def execute_plan_receipt(self, receipt: str) -> bool:
```
TomCC7: why is this required then?

dimos/manipulation/manipulation_module.py
TomCC7: It is not required. I removed the receipt-specific planning RPC and restored plan_to_joint_targets() as the sole public planning-group method returning bool. The panel now uses that existing API; the external plan-replacement race remains the explicitly documented limitation until we introduce a proper plan-token design.

dimos/manipulation/manipulation_module.py
TomCC7: Also not required. I removed execute_plan_receipt() and the receipt state. The panel now calls the existing execute() path. The internal lock-protected one-shot reservation remains so concurrent callers cannot dispatch the same stored plan twice; focused concurrency coverage was added.

dimos/manipulation/visualization/viser/state.py
```diff
@@ -96,11 +97,22 @@ class PanelPlanState:
     target_pose: Pose | None = None
     target_joints: list[float] | None = None
     start_joints_snapshot: list[float] | None = None
```
mustafab0: This seems to be deadcode.

The `PanelPlanState` should maintain a dict of all robots, and their ID, instead of just a single robot

the target_pose would be nice to be dict[PlanningGroupID, targetPose]

dimos/manipulation/visualization/viser/state.py
```diff
@@ -124,15 +136,22 @@ def next_sequence_id(self) -> int:
...
 
     def can_plan(self) -> bool:
         return (
             self.runtime == PanelRuntime.RUNNING
             and self.backend_status == BackendConnectionStatus.READY
-            and self.selected_robot is not None
+            and bool(self.selected_group_ids)
             and self.action_status == ActionStatus.IDLE
             and self.target_status == TargetStatus.FEASIBLE
             and self.manipulation_state in {"IDLE", "COMPLETED", "FAULT"}
```
mustafab0: If manipulation state is "FAULT". this should return false right? Is it intended to plan even at Fault?

dimos/manipulation/planning/world/drake_world.py
```diff
@@ -1184,57 +1201,109 @@ def _set_preview_positions(
...
+
+        Execution timing remains the manipulation module's responsibility.
+        """
+        group_ids = tuple(plan.group_ids)
+        selection = PlanningGroupRegistry(
+            robot_data.config for robot_data in self._robots.values()
+        ).select(group_ids)
+        if not group_ids or len(set(group_ids)) != len(group_ids):
+            raise ValueError("generated plan has empty or duplicate planning groups")
+        if self._meshcat is None or len(plan.path) < 2:
```
mustafab0: Since we are planning to use viser as our visualizer. this meshcat check should just be a generic check if any visualizer is available.

dimos/manipulation/manipulation_module.py
```diff
@@ -1326,42 +1328,87 @@ def execute_plan(
         self, plan: GeneratedPlan | None = None, robot_name: RobotName | None = None
     ) -> bool:
         """Execute a generated planning-group plan through affected trajectory tasks."""
+        if plan is not None:
+            return self._execute_plan(plan, robot_name=robot_name)
```
mustafab0: This does not check if the `ManipulationState` is IDLE or Planning, and may execute while robot is executing previous plan or while the `ManipulationState` is not ready.

dimos/manipulation/manipulation_module.py
```diff
@@ -1326,42 +1328,87 @@ def execute_plan(
...
             robot = self._get_robot(robot_name)
             if robot is None:
-                return self._fail(f"Robot '{robot_name}' not found")
+                return self._fail_execution(f"Robot '{robot_name}' not found", execution_reserved)
             _, _, config, _ = robot
             if not config.coordinator_task_name:
                 logger.error(f"No coordinator_task_name for '{robot_name}'")
-                return self._fail(f"No coordinator_task_name for '{robot_name}'")
+                return self._fail_execution(
+                    f"No coordinator_task_name for '{robot_name}'", execution_reserved
```
mustafab0: This whole method is so bloated 😅

We need to fix this. Not a blocker for this PR. But we should clean things up, as it feels bloated.

mustafab0 (changes requested):

dimos/manipulation/visualization/operator.py
TomCC7, line 1: better move to viz folder

dimos/control/tasks/trajectory_task/trajectory_task.py
```diff
@@ -169,6 +173,61 @@ def on_preempted(self, by_task: str, joints: frozenset[str]) -> None:
...
+            self._clear_active_trajectory()
+
+    def _clear_active_trajectory(self) -> None:
+        """Clear stored trajectory-specific execution state."""
+        self._trajectory = None
+        self._active_joint_names = []
+        self._pending_start = False
+        self._start_time = 0.0
+
+    def _validate_trajectory(self, trajectory: JointTrajectory) -> list[str] | None:
```
TomCC7: too defensive here also I don't think we need the active joint name concept, just execute whatever trajectory provides

dimos/manipulation/manipulation_operator.py
```diff
@@ -0,0 +1,402 @@
...
+    message: str
+    collision_free: bool = False
+    group_ids: tuple[PlanningGroupID, ...] = ()
+    target_joints: JointState | None = None
+    group_diagnostics: Mapping[PlanningGroupID, str] = field(default_factory=dict)
+    group_poses: Mapping[PlanningGroupID, PoseStamped | None] = field(default_factory=dict)
+
+
+@dataclass(frozen=True)
+class PlanSummary:
```
TomCC7: how is plansummary used in vis? maybe we should just return generated plan everytime

dimos/manipulation/manipulation_module.py
TomCC7, line 1: need to justify why there's so many new methods added. to me a lot of them should be safe to remove

mustafab0 (approved):

PR #2865, opened 2026-07-10
ci: skip experimental arduino tests in CI test job

jeff-hykin:
Exclude `dimos/experimental/arduino` from CI


spomichter:
are we skipping all experimental dirs? @jeff-hykin kind of weird just to hardcode an arduino skip

Dreamsorcerer:
Yeah, I don't think we want to do this. Seems to have been a lot of confusion in the chats..

PR #2866, opened 2026-07-10
Build all GC Roots into Cachix

Dreamsorcerer:
And other adjustments to keep runners stable in future.


bin/build-cmu-nav-natives
```diff
@@ -61,6 +61,41 @@ def discover() -> tuple[type[NativeModuleConfig], ...]:
...
 
+# Persistent GC-root dir on the runner's /nix volume — nix's own roots dir, not
+# the git checkout — so `git clean -ffdx` can't delete the roots and
+# nix-collect-garbage never reaps the native closures. Roots are per-store, so
+# each self-hosted runner pins its own /nix here (they can't travel via Cachix,
+# which ships store paths, not roots).
+GCROOTS_DIR = Path("/nix/var/nix/gcroots/dimos-natives")
+
+
+def _pin(root_name: str, out_link: Path) -> str:
```
spomichter, line 72: Assume this logic will be pulled out into generic format but will approve for now since i assume we need to merge to test if this change works

spomichter (dismissed):

bin/build-cmu-nav-natives
Dreamsorcerer: No, testing in the PR works fine.

bin/build-cmu-nav-natives
```diff
@@ -70,24 +105,48 @@ def build_one(cfg_cls: type[NativeModuleConfig]) -> str:
...
+    # Fast path: a previous run on this runner already built and pinned this exact
+    # command. `git clean` wipes the `result` symlink, but the pin under /nix
+    # survives — restore `result` from it so the exe.exists() check below skips
+    # `nix build` entirely. Without this, every run re-evaluates the flake and
+    # re-fetches its multi-GB `-source` closure just for the GC to delete it
+    # again. The `.cmd` sidecar guards correctness: a version bump changes
+    # build_command, misses the pin, and rebuilds.
+    cmd_file = GCROOTS_DIR / f"{name}.cmd"
+    pin = GCROOTS_DIR / name
+    if pin.exists() and cmd_file.is_file() and cmd_file.read_text() == cfg.build_command:
```
jeff-hykin, line 121: I think this is a problem: even if the source code changes I suspect it won't rebuild (because the symlink will exist). 

I'm going to run a quick test right now to see

Dreamsorcerer: OK, looking into this, I think this only breaks pgo. If I'm understanding correctly, I think we need to move pgo to an external model immediately, and use immutable tags, then everything should work correctly.

jeff-hykin:
```py
build_command = "nix build github:jeff-hykin/dimos-module-pgo/v0.1.0 --no-write-lock-file"
```
This Should work

Also Confirmed: we push a new build to cachix on every change (✅) but the tests run using an old binary even after the new one is built (❌). Tested by pushing a couple commits on here and watching CI.

We can merge this anyways and I can push a follow up later. Probably just put at TODO/FIXME in the code so we dont forget.

I'll MIA for probably 24hrs but can push a proper fix later

paul-nechifor (approved):

PR #2869, opened 2026-07-11
Module Deployment Proposal

TomCC7:
## Problem

DimOS needs one deployment model for normal Python modules, dependency-isolated packaged Python, native implementations, and remote targets. The existing paths expose pieces of that model, but they do not establish a common coordinator-facing contract or target-side runtime boundary.

## Proposal

This PR centers on `module_deployment_proposal.md`. It defines the intended user-facing API, package convention, concrete deployment topology, worker ownership, control/data-plane split, phased roadmap, and a non-normative reference appendix for the complete design model.

The proposal keeps the coordinator-visible module contract stable while routing external implementations through:

```text
ModuleCoordinator -> WorkerManagerExternal -> ExternalWorker -> runtime
```

Normal Python modules remain on the existing `WorkerManagerPython -> PythonWorker` path.

## First Concrete Step

The implementation is the first local packaged-Python slice of that proposal. It makes the shape executable and reviewable rather than treating the proposal as diagrams alone:

- `ExternalModule` declarations and `DeploymentSpec` planning
- local target-session preparation and a target-side `ExternalWorker`
- serialized launch envelopes and a thin packaged-Python runtime entrypoint
- declared RPC, streams, module references, lifecycle handling, and external-only dependency isolation
- an example whose runtime imports `humanize` only from its isolated package environment

Remote targets, native execution, prepared-plan persistence, and the final public deployment CLI remain later roadmap work.

## How to Test
Run the example manually from the repository root; no `PYTHONPATH` is required:

```bash
uv run python -m dimos.core.deployment.launcher plan examples.external_python_module.deployment:deployment_spec
uv run python -m dimos.core.deployment.launcher prepare examples.external_python_module.deployment:deployment_spec
uv run python -m dimos.core.deployment.launcher run examples.external_python_module.deployment:deployment_spec
```

The final command prints the resolved route, starts the example, and runs until interrupted with Ctrl-C.


dimos/core/deployment/launcher.py
```diff
@@ -0,0 +1,69 @@
...
+    spec = resolve_deployment_ref(ref)
+    plan = plan_deployment(spec)
+    return {
+        "python_modules": [cls.__name__ for cls in plan.python_modules],
+        "external_modules": [env.module_class.__name__ for env in plan.external_modules],
+    }
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description="Temporary local deployment integration launcher")
```
TomCC7: use typer

dimos/core/deployment/models.py
```diff
@@ -0,0 +1,229 @@
...
+    uses_pixi: bool
+    readiness_timeout_s: float = 10.0
+
+    @property
+    def python_dir(self) -> Path:
+        return self.package_root / "python"
+
+
+@dataclass(frozen=True)
+class ModuleLaunchEnvelope:
```
TomCC7: use pydantic to get automatic serde

dimos/core/deployment/models.py
```diff
@@ -0,0 +1,229 @@
...
+    command_prefix: tuple[str, ...]
+
+
+@dataclass(frozen=True)
+class DeploymentSpec:
+    blueprint: Blueprint
+    modules: dict[type[ModuleBase], ModuleDeployment] = field(default_factory=dict)
+
+
+def _required_str(data: Mapping[str, JsonValue], key: str) -> str:
```
TomCC7: so if you use pydantic these are all free

dimos/core/deployment/ref.py
```diff
@@ -0,0 +1,40 @@
...
+        )
+    try:
+        module = __import__(module_name, fromlist=[variable_name])
+    except Exception as exc:
+        raise ValueError(
+            f"Could not import deployment reference module {module_name!r}: {exc}"
+        ) from exc
+    if not hasattr(module, variable_name):
+        raise ValueError(f"Deployment reference variable {variable_name!r} was not found")
+    obj = getattr(module, variable_name)
```
TomCC7, line 35: there's no typechecking here

dimos/core/coordination/module_coordinator.py
```diff
@@ -309,9 +325,19 @@ def build(
         _run_configurators(blueprint)
         _check_requirements(blueprint)
         _verify_no_name_conflicts(blueprint)
+        if deployment_plan is None:
+            reject_external_modules_without_deployment_spec(
```
TomCC7: we can just pass a blueprint into it so no need to parse stuff outside

dimos/core/coordination/module_coordinator.py
```diff
@@ -309,9 +325,19 @@ def build(
...
+            )
 
         logger.info("Starting the modules")
         coordinator = cls(g=global_config)
+        if deployment_plan is not None:
+            external_manager = cast(
+                "WorkerManagerExternal",
+                coordinator._managers[WorkerManagerExternal.deployment_identifier],
+            )
+            external_manager.configure_plan(deployment_plan)
```
TomCC7, line 342: this is essentially impl leaking, better way to do this only through workerspec?

PR #2874, opened 2026-07-11
Fixes/mustafa/go2 controller benchmark

leshy:
moved utils, mypy fix, merged main


mustafab0 (approved):

dimos/control/tasks/velocity_tracking_pid.py
```diff
@@ -0,0 +1,164 @@
...
+    max_integral: float = 0.5  # anti-windup clamp
+    output_min: float = -1.0
+    output_max: float = 1.0
+
+
+@dataclass
+class VelocityTrackingConfig:
+    """Configuration for all three velocity channels."""
+
+    vx: VelocityPIDConfig = None  # type: ignore[assignment]
```
paul-nechifor, line 50: Can't do this.

leshy: @mustafab0 plz fix so we can merge

dimos/robot/unitree/go2/blueprints/basic/unitree_go2_rpp_benchmark.py
```diff
@@ -0,0 +1,80 @@
...
+from dimos.core.transport import LCMTransport
+from dimos.msgs.geometry_msgs.PoseStamped import PoseStamped
+from dimos.robot.unitree.go2.blueprints.basic.unitree_go2_rpp_controller import (
+    unitree_go2_rpp_controller,
+)
+
+# The Benchmarker's ``odom`` In must read the same topic the controller emits leg
+# odom on (/go2/odom). path/speed/cmd_vel/gate already share names+topics with
+# the controller blueprint, so they wire up by the controller's transports.
+_BENCHMARK_TRANSPORTS: dict[tuple[str, type], TransportSpec | Transport[Any]] = {
```
paul-nechifor, line 65: It seems unnecessary to extract this as `_BENCHMARK_TRANSPORTS` when it could have been inline. Also, why have the override at all? You only need to rename the topic to `/go2/odom` because you set it to `/go2/odom` in the other place. If you leave both places as the default `/odom` you don't need to override in blueprints.

I think both comments in this file are unnecessary. This blueprint should be 7 lines long.

dimos/robot/unitree/go2/blueprints/basic/unitree_go2_rpp_controller.py
```diff
@@ -0,0 +1,145 @@
...
+        [
+            (GO2Connection, "cmd_vel", "go2_cmd_vel"),
+            (GO2Connection, "odom", "go2_odom"),
+        ]
+    )
+    .transports(
+        {
+            # External command interface: teleop publishes on /cmd_vel; the
+            # coordinator's twist_command reads it and echoes cmd_vel back.
+            ("cmd_vel", Twist): LCMTransport("/cmd_vel", Twist),
```
paul-nechifor, line 127: This is the default. Many others are also defaults.

dimos/robot/unitree/go2/blueprints/basic/unitree_go2_rpp_controller.py
```diff
@@ -0,0 +1,145 @@
...
+            ("go2_cmd_vel", Twist): LCMTransport("/go2/cmd_vel", Twist),
+            ("go2_odom", PoseStamped): LCMTransport("/go2/odom", PoseStamped),
+            # Controller IN: planned path + target speed (from any source).
+            ("path", Path): LCMTransport("/path", Path),
+            ("speed", Float32): LCMTransport("/speed", Float32),
+            # Operator gate (teleop -> benchmark) to pace runs.
+            ("gate", Int8): LCMTransport("/benchmark/gate", Int8),
+            # Aggregated joint state for observability (positions = [x,y,yaw]).
+            ("joint_state", JointState): LCMTransport("/coordinator/joint_state", JointState),
+            ("coordinator_joint_state", JointState): LCMTransport(
```
paul-nechifor, line 139: In https://github.com/dimensionalOS/dimos/pull/2515 I removed all of these.

There should be a very good reason to use `.transports`. It should not be the norm. The norm should be to rely on the defaults.

dimos/msgs/std_msgs/Float32.py
```diff
@@ -0,0 +1,30 @@
...
+# limitations under the License.
+
+"""Float32 message type."""
+
+from typing import ClassVar
+
+from dimos_lcm.std_msgs import Float32 as LCMFloat32
+
+
+class Float32(LCMFloat32):  # type: ignore[misc]
```
paul-nechifor, line 23: Don't ignore.

dimos/control/coordinator.py
```diff
@@ -768,6 +866,12 @@ def stop(self) -> None:
         if self._buttons_unsub:
             self._buttons_unsub()
             self._buttons_unsub = None
+        if self._path_unsub:
```
paul-nechifor, line 869: This has become incredibly long. Why not use register_disposable?

dimos/control/coordinator.py
```diff
@@ -743,6 +811,36 @@ def start(self) -> None:
...
+        if has_speed_task:
+            try:
+                self._speed_unsub = self.speed.subscribe(self._on_speed)
+                logger.info("Subscribed to speed for speed-capable tasks")
+            except Exception:
+                logger.warning(
+                    "Speed-capable task configured but could not subscribe to speed. "
+                    "Set transport via blueprint."
+                )
+
```
paul-nechifor, line 843: * This is quite hard to read.
* The entire function follows the if-try-except pattern. Surely that should be extracted as a function if it's repeated so many times. One of the reasons our PRs are so long is because of so much duplicate code.
* So many of the `logger.info` lines are unnecessary. I don't see why anyone would care to read that. We're just spamming logs.
* Subscribing should be reliable. If a module fails to subscribe to its required streams, I don't see how it can continue on. I don't think `try-except-logger.warning` makes sense. It would be simpler to remove the try-except entirely. If you're having issues with subscribing, that's a bug that should be reported and fixed, not hidden away with `try-except-pass`.

dimos/control/benchmarking/benchmark.py
```diff
@@ -0,0 +1,523 @@
+# Copyright 2025-2026 Dimensional Inc.
```
paul-nechifor, line 1: Some autofixes here: https://github.com/dimensionalOS/dimos/pull/2881/changes

PR #2882, opened 2026-07-13
Control task commands

mustafab0:
## Problem

RPC based tasks used a single @rpc task_invoke method, without any checking if task actually does support the rpc commands

Closes DIM-XXX

## Solution

Activates the commands half of the task binding card. A task type lists the command names it accepts; the method's own signature is the argument schema (`inspect.signature().bind()`), so there's no second schema to maintain and no new files. `task_invoke` keeps its exact wire signature.

- [x ] I have read and approved the [CLA](https://github.com/dimensionalOS/dimos/blob/main/CLA.md).

### DIFF
--- a/dimos/control/coordinator.py
+++ b/dimos/control/coordinator.py
@@ -26,6 +26,7 @@ Features:
 """
 
 from dataclasses import dataclass, field
+import inspect
 import threading
 import time
 from typing import TYPE_CHECKING, Any
@@ -182,6 +183,10 @@ class ControlCoordinator(Module):
         # Card-declared stream routes: stream name -> (task, handler, routing).
         # Guarded by _task_lock; entries are added/pruned with their task.
         self._routes: dict[str, list[tuple[ControlTask, str, Routing]]] = {}
+
+        # Card-declared command names per task, keyed by task name.
+        # Guarded by _task_lock; added/pruned with their task.
+        self._task_commands: dict[TaskName, frozenset[str]] = {}
 
         # Tick loop (created on start)
         self._tick_loop: TickLoop | None = None
@@ -417,6 +422,9 @@ class ControlCoordinator(Module):
             self._tasks[task.name] = task
             if task_type is not None:
                 self._register_routes(task, task_type)
+                self._task_commands[task.name] = self._commands_for(task_type)
+            else:
+                self._task_commands[task.name] = frozenset()
             logger.info(f"Added task {task.name}")
         self._sync_stream_subscriptions()
         return True
@@ -430,6 +438,7 @@ class ControlCoordinator(Module):
                 return False
             for entries in self._routes.values():
                 entries[:] = [entry for entry in entries if entry[0] is not task]
+            self._task_commands.pop(task_name, None)
             logger.info(f"Removed task {task_name}")
         self._sync_stream_subscriptions()
         return True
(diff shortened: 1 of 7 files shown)


dimos/control/coordinator.py
```diff
@@ -735,6 +819,7 @@ def stop(self) -> None:
             self._stream_unsubs.clear()
         with self._task_lock:
             self._routes.clear()
+            self._task_commands.clear()
```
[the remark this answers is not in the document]
mustafab0: Dropped both clears. Routes and commands are derived from _tasks

dimos/control/coordinator.py
```diff
@@ -728,13 +812,12 @@ def stop(self) -> None:
         """Stop the coordinator."""
         logger.info("Stopping ControlCoordinator...")
 
-        # Unsubscribe from streaming commands and drop the route table
+        # Route/command tables are kept: they track _tasks, which survives stop(),
+        # and add_task() skips known names so a restart would never rebuild them.
```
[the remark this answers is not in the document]
mustafab0: This will be addressed part of the declarative adapter topic definition

dimos/control/test_coordinator_commands.py
```diff
@@ -0,0 +1,325 @@
...
+    JointCommandOutput,
+    ResourceClaim,
+)
+from dimos.msgs.trajectory_msgs.JointTrajectory import JointTrajectory
+from dimos.msgs.trajectory_msgs.TrajectoryPoint import TrajectoryPoint
+
+ARM_JOINTS = frozenset({"arm/joint1", "arm/joint2"})
+
+
+class CommandRecordingTask(BaseControlTask):
```
paul-nechifor, line 49: It feels like I've seen these implementation classes in tests before.

If ControlTask is a protocol, perhaps BaseControlTask should implement common methods and subclasses should override.

For example, how many ways are there to implemement `def name`? I've looked at every single one, and all of them (except for one test one) implement it as:

```
    @property
    def name(self) -> str:
        return self._name
```

That suggests it should be part of BaseControlTask, and probably many other methods should be as well.

But generally, if the setup for a test is long, that suggests the API is bad.

dimos/control/test_coordinator_commands.py
```diff
@@ -0,0 +1,325 @@
...
+        return self._name
+
+    def claim(self) -> ResourceClaim:
+        return ResourceClaim(joints=self._joints)
+
+    def is_active(self) -> bool:
+        return False
+
+    def compute(self, state: CoordinatorState) -> JointCommandOutput | None:
+        _ = state
```
paul-nechifor: You'll be happy to know I've made a PR to disallow unused `_` variables: https://github.com/dimensionalOS/dimos/pull/2928/changes 

In the mean while, can you please remove these from here?

dimos/control/coordinator.py
```diff
@@ -616,25 +631,94 @@ def reset_runtime_state(self, reactivate: bool | None = None) -> dict[str, bool]
...
                 kwargs["t_now"] = time.perf_counter()
 
+            if method in self._task_commands.get(task_name, frozenset()):
+                return self._invoke_declared(task, task_name, method, kwargs)
+
+            logger.warning(
+                f"undeclared task_invoke {task_name}.{method} — declare it in TASK_EXPOSES"
+            )
+            if not hasattr(task, method):
+                logger.warning(f"Task {task_name} has no method {method}")
```
paul-nechifor: Not having a method sounds like a catastrophic error to me. Can the system actually work when it tries to call a method which doesn't exist?

mustafab0: Added AttributeError instead of just a warning

dimos/control/test_coordinator_commands.py
mustafab0: removed `_` from test

dimos/control/test_coordinator_commands.py
mustafab0: agreed — and once the card routing fully lands, the stream-listener no-ops (on_buttons etc.) can come out of BaseControlTask and I will add the common methods to the BaseControlTask

mustafab0: https://github.com/dimensionalOS/dimos/issues/2942

paul-nechifor (approved):

PR #2807, opened 2026-07-08
Branch: Feat/ivan/m20  Optimized the adaptation of simple_nav for M20

MeloLong:
1. Costmap Optimization

(1) Introduced local ground height estimation to suppress the influence of elevated structures on costmap generation.

A local ground height is first estimated. If a grid cell only contains points that are significantly higher than the estimated ground, those points are treated as ceiling or elevated structures and ignored, instead of being directly used to generate the traversal cost. This prevents elevated points from being mistakenly classified as ground bumps or obstacles, thereby avoiding the incorrect removal of traversable areas.

(2) Relaxed the ground continuity constraint to improve robustness under sparse point cloud observations.

The original method required a grid cell and all four of its neighboring cells (front, back, left, and right) to have valid observations before slope and traversability costs could be computed reliably. Under sparse point clouds or local ground gaps, this strict requirement caused many valid ground cells to be marked as unknown. We therefore relaxed the criterion from complete neighborhood observation to a sufficient number of valid neighboring observations, reducing the spread of unknown regions caused by sparse measurements and improving the completeness of traversable areas.

2. Trajectory Controller Optimization

Resolved the M20 velocity command response issue by enforcing the robot's minimum velocity response thresholds in the simple_nav trajectory planner. The minimum command limits are:

Linear velocity (x): ≥ 0.2 m/s

Linear velocity (y): ≥ 0.3 m/s

Angular velocity (yaw): ≥ 0.6 rad/s

When linear and angular motions are commanded simultaneously: yaw ≥ 0.2 rad/s

### **How to run:**

Tips: m20 is a IP of m20 robot, please replace it.

`DIMOS_ZENOH_CONNECT=tcp/m20_ip python -m dimos.robot.deeprobotics.m20.zenoh_lcm_bridge`
`dimos run m20-simple-nav`

PR #2885, opened 2026-07-13
Whale Dynamics M20 Integration

MeloLong:
# M20 Support

## Overview

The purpose of this PR is the same as **PR #2807**: to add support for the M20.

The difference is that this PR targets `main` instead of `feat/ivan/m20`. As a result, it also includes the additional `ray_tracing` compatibility changes required for the current `main` branch.

`ReplanningAStarPlanner` has been optimized at the source-code level. Additionally, a new module parameter configuration mechanism has been introduced, allowing parameters to be configured through the config file. This feature has been initially applied in `m20_simple_nav`.

### `ray_tracing` Compatibility

The current implementation of `ray_tracing` on the `main` branch only supports:

- Point clouds published in the `body` frame
- Odometry published in the `map` frame

This PR extends `ray_tracing` to also support:

- Point clouds published in the `map` frame
- Odometry published in the `map` frame

This is compatible with the output of the M20's built-in SLAM module.

## Build and Run

See `dimos/robot/deeprobotics/m20/nav/README.md` for build and run instructions.


dimos/utils/cli/spy/spy.py
```diff
@@ -0,0 +1,147 @@
...
+        kwargs: dict[str, object] = {}
+        if connect:
+            kwargs["connect"] = list(connect)
+        if iface:
+            kwargs["multicast_iface"] = iface
+        return ZenohPubSubBase(**kwargs), Topic(key or "**")
+    return LCMPubSubBase(), Topic(re.compile(key) if key else re.compile(".*"))
+
+
+class Spy:
```
leshy: this is already in main, can remove spy utils from this PR

dimos/robot/deeprobotics/m20/nav/readme.md
```diff
@@ -0,0 +1,207 @@
...
+
+# Table of Contents
+
+- [Prerequisites](#prerequisites)
+- [Installation](#installation)
+  - [1. Switch to the DimOS workspace](#1-switch-to-the-dimos-workspace)
+  - [2. Register all blueprints](#2-register-all-blueprints)
+  - [3. First Launch](#3-first-launch)
+- [Running the System](#running-the-system)
+  - [Terminal 1: Start the Zenoh Bridge](#terminal-1-start-the-zenoh-bridge)
```
leshy: dimos now supports zenoh directly now so there is no need for zenoh bridge, this bridge can be removed from the codebase entirely
`dimos --transport=zenoh run unitree-go2`

(probably should add --zenoh-connect to global config, if m20 node autodiscovery doesn't work)

dimos/robot/deeprobotics/m20/nav/readme.md
```diff
@@ -0,0 +1,207 @@
...
+
+```bash
+cd ~/workplace/dimos
+source .venv/bin/activate
+```
+
+Launch the navigation system:
+
+```bash
+dimos run m20-simple-nav
```
leshy: how to specify robot ip? or are we relying on zenoh autodiscovery?

MeloLong: I have updated the PR by adding `--zenoh-connect` to the global configuration, documenting how to obtain the robot IP address, and removing `spy.py`.

leshy:
You need to pass tests, `./bin/pytest`  locally or can check CI here

leshy:
please resolve code comments that have been addressed

dimos/robot/deeprobotics/m20/readme.md
```diff
@@ -0,0 +1,205 @@
...
+
+Before getting started, make sure that:
+
+- The DimOS environment has been installed successfully.
+- You are working from the DimOS workspace.
+- The robot is powered on and connected to the same network as your workstation.
+
+---
+
+# Installation
```
leshy, line 32: You should write main m20 readme, no need for separate nav readme for now.

Readme should include steps for deploying drdds zenoh bridge on the robot. Zenoh router if needed (check systemd services, I added those)

This readme will be read by a fresh person on our end integrating a new M20 as validation

dimos/mapping/ray_tracing/rust/src/main.rs
```diff
@@ -236,6 +233,27 @@ fn nearest_pose(
     }
 }
 
+fn register_points(
+    points: &[(f32, f32, f32)],
+    translation: Vector3<f32>,
+    rotation: UnitQuaternion<f32>,
+    registered_clouds: bool,
```
aclauer, line 240: This is a style thing, but I feel like a function called `register_points` should unconditionally register the points. The branch for if the cloud is registered or not should be done before, then we don't even call this function if we don't need to.

dimos/mapping/ray_tracing/rust/src/main.rs
```diff
@@ -404,6 +422,33 @@ mod tests {
         assert_eq!(poses.back().unwrap().0, (POSE_BUFFER_LEN + 9) as f64);
     }
 
+    #[test]
+    fn register_points_leaves_registered_clouds_unchanged() {
```
aclauer, line 426: going off my previous comment, then you don't need a test like this

dimos/mapping/ray_tracing/rust/src/voxel_ray_tracer.rs
```diff
@@ -31,6 +31,8 @@ pub struct Config {
     pub max_range: f32,
     #[validate(range(min = 1))]
     pub ray_subsample: u32,
+    /// Input cloud points are already expressed in the odometry/world frame.
```
aclauer, line 34: This comment explains what one specific value of the config means, not what the config itself it. Should be something like "Whether input point clouds have been world registered or not"

dimos/mapping/ray_tracing/rust/src/python.rs
```diff
@@ -106,6 +106,7 @@ impl VoxelRayMapper {
             max_health,
             graze_cos,
             support_min,
+            registered_clouds: true,
```
aclauer, line 109: We should align the defaults on configs across the module and FFI. For example, here registered_clouds is true, but in the native module `module.py` it is set to false. I'd prefer defaulting to false.

dimos/navigation/nav_3d/mls_planner/utils/plan_rrd.py
```diff
@@ -288,6 +288,12 @@ def main(
     align_tol: float = typer.Option(0.05, "--align-tol", help="Lidar/odom alignment tolerance (s)"),
     voxel_size: float = typer.Option(0.08, "--voxel-size", help="Voxel edge length (m)"),
     max_range: float = typer.Option(30.0, "--max-range", help="Max ray cast distance (m)"),
+    registered_clouds: bool = typer.Option(
+        True,
+        "--registered-clouds/--no-registered-clouds",
+        help="Clouds are already world-registered (fastlio). Use --no-registered-clouds for "
```
aclauer, line 376: Fastlio does not world register the point clouds, they are in sensor frame when they are published by fastlio, right?

aclauer (commented):
left some comments on the Rust stuff. I'm not familiar with the m20, but are we sure we can't get point clouds that are in sensor frame?

If there isn't a way, then I'm fine adding a config flag to the voxel mapper, but the defaults should be consistent (or even better we don't specify a default so the user has to know what they want).

.gitignore
```diff
@@ -47,7 +44,7 @@ package-lock.json
 # Ignore build artifacts
 dist/
 build/
-
+data/
```
paul-nechifor: Why did you change this?

paul-nechifor:
Please update the PR title and description to reflect that the PR does.

> Merged PR https://github.com/dimensionalOS/dimos/pull/2807 into main

This is not true. That PR is still open.

dimos/robot/deeprobotics/m20/connection.py
```diff
@@ -0,0 +1,485 @@
...
+            logger.warning("M20 abnormal status: %s", errors)
+        if self.on_report is not None and isinstance(asdu, dict):
+            self.on_report(int(asdu.get("Type", 0)), int(asdu.get("Command", 0)), items)
+
+
+# --- dimos module -------------------------------------------------------------
+
+
+class M20Config(ModuleConfig):
+    ip: str = Field(default_factory=lambda m: m["g"].robot_ip or "10.21.31.103")
```
paul-nechifor: Do not hard code IPs

dimos/robot/deeprobotics/m20/connection.py
```diff
@@ -0,0 +1,485 @@
...
+        super().stop()
+
+    def _bringup(self) -> None:
+        """Regular mode -> stand -> standard pose -> basic gait (drivable)."""
+        link = self._link
+        assert link is not None
+        link.send(*_USAGE_MODE, {"Mode": _MODE_REGULAR})
+        time.sleep(0.5)
+        link.send(*_MOTION_STATE, {"MotionParam": _STATE_STAND})
+        time.sleep(3.0)
```
paul-nechifor, line 370: Is there to way to check if the state has changed/is ready? It would be better if you can avoid sleeps

dimos/robot/deeprobotics/m20/connection.py
```diff
@@ -0,0 +1,485 @@
...
+    def move(self, twist: Twist, duration: float = 0.0) -> bool:
+        """Set teleop velocity. linear.x = forward (m/s), linear.y = left (m/s),
+        angular.z = yaw (rad/s). Held until the next command; the control loop
+        re-sends at ``cmd_rate_hz`` and zeroes it after ``cmd_timeout``.
+
+        ``duration`` is accepted for parity with other dimos connections but is
+        unused — hold behaviour is handled by the deadman in the control loop.
+        """
+        self._cmd = twist
+        self._cmd_ts = time.time()
```
paul-nechifor: This is changed in one thread and accessed in another. You need a lock.

dimos/robot/deeprobotics/m20/connection.py
```diff
@@ -0,0 +1,485 @@
...
+    def _control_loop(self) -> None:
+        # Re-send the latest command at cmd_rate_hz (manual recommends 20 Hz);
+        # the deadman zeroes it if cmd_vel goes stale, so the robot stops if we do.
+        period = 1.0 / self.config.cmd_rate_hz
+        cfg = self.config
+        while self._running:
+            twist = _apply_deadman(self._cmd, time.time() - self._cmd_ts, cfg.cmd_timeout)
+            axes = _axes_from_twist(twist, cfg.max_linear, cfg.max_angular)
+            self._link.send(*_MOVE, axes)  # type: ignore[union-attr]
+            time.sleep(period)
```
paul-nechifor: The sleep needs to account for the time it took to run one loop in order to maintain the frequency.

dimos/robot/deeprobotics/m20/connection.py
```diff
@@ -0,0 +1,485 @@
...
+
+    def _control_loop(self) -> None:
+        # Re-send the latest command at cmd_rate_hz (manual recommends 20 Hz);
+        # the deadman zeroes it if cmd_vel goes stale, so the robot stops if we do.
+        period = 1.0 / self.config.cmd_rate_hz
+        cfg = self.config
+        while self._running:
+            twist = _apply_deadman(self._cmd, time.time() - self._cmd_ts, cfg.cmd_timeout)
+            axes = _axes_from_twist(twist, cfg.max_linear, cfg.max_angular)
+            self._link.send(*_MOVE, axes)  # type: ignore[union-attr]
```
paul-nechifor: Please do not ignore

dimos/robot/deeprobotics/m20/connection.py
```diff
@@ -0,0 +1,485 @@
...
+            port.publish(img)
+        cap.release()
+
+    @skill
+    def observe(self) -> Image | None:
+        """Return the latest front-camera frame from the M20, or None if none yet."""
+        return self._latest_video_frame
+
+
+def deploy(dimos: ModuleCoordinator, ip: str = "10.21.31.103", prefix: str = "") -> ModuleProxy:
```
paul-nechifor: We don't have deploy functions anymore.

dimos/mapping/ray_tracing/rust/src/main.rs
MeloLong: The M20 has a built-in SLAM module that only publishes point clouds in the `map` frame rather than the sensor frame.

I'll separate the responsibilities so that `register_points` only performs point registration. The decision of whether registration is needed will be made beforehand using a boolean configuration flag, allowing the voxel mapper to support both sensor-frame and map-frame point clouds.

dimos/navigation/nav_3d/mls_planner/utils/plan_rrd.py
MeloLong: You're right. The original Fast-LIO publishes point clouds that are already registered in the `map` frame. However, the Fast-LIO integration in DimOS publishes point clouds in the `body` frame instead.

I'll follow your first suggestion and make the voxel mapper support both cases by separating the registration logic and adding a configuration option, so it can also handle the `map`-frame point clouds published by the M20.

.gitignore
MeloLong: i will delete it

dimos/robot/deeprobotics/m20/dds/cpp/main.cpp
```diff
@@ -0,0 +1,242 @@
+// Copyright 2026 Dimensional Inc.
```
leshy: why is this implementation here in cpp/ dir, and then we also have onboard/drdds-zenoh-bridge dir? seems like these files belong at the same place or are duplicated?

dimos/robot/deeprobotics/m20/readme.md
MeloLong: I have already updated readme.md

dimos/robot/deeprobotics/m20/connection.py
MeloLong: delete this hard code IPs

dimos/robot/deeprobotics/m20/connection.py
MeloLong: add a lock

dimos/robot/deeprobotics/m20/connection.py
MeloLong: Modified it in _control_loop function.

paul-nechifor:

PR #2923, opened 2026-07-13
feat: costmapper NativeModule

ghost:
## Problem

This establishes the integration boundary for a Rust-native cost mapper while preserving compatibility with DimOS streams, configuration, and LCM message types.

## Solution

- Added a `NativeModule` wrapper with typed `PointCloud2` inputs and an `OccupancyGrid` output.
- Added configuration serialization from the existing Python occupancy configuration to the Rust subprocess.
- Scaffolded a Rust crate that produces:
- A standalone DimOS native-module executable.
- A PyO3 extension for calling the cost mapper directly from Python.
- Integrated the DimOS Rust module SDK and generated LCM message types.
- Added a Python baseline benchmark covering latency, throughput, CPU usage, and
  memory usage.

The native cost-mapping algorithm, safe-radius processing, and runtime stream handlers remain scaffolded with `todo!()` and are outside the completed scope of this submission.
  
## How to Test

1. Verify that the Rust crate compiles: `cargo check --locked --manifest-path dimos/mapping/native_costmapper/rust/Cargo.toml`

2. Run the Python performance baseline: `uv run python demo_benchmark_python.py --profile`


dimos/mapping/native_costmapper/rust/src/costmapper.rs
```diff
@@ -0,0 +1,1083 @@
...
+}
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(default, deny_unknown_fields)]
+struct SimpleOccupancyConfig {
+    resolution: f64,
+    frame_id: Option<String>,
+    min_height: f64,
+    max_height: f64,
+    closing_iterations: i64,
```
aclauer: does this actually get used anywhere?

dimos/mapping/native_costmapper/rust/src/costmapper.rs
```diff
@@ -0,0 +1,1083 @@
...
+    let period = (length * 2) as isize;
+    let reflected = index.rem_euclid(period);
+    if reflected >= length as isize {
+        (period - 1 - reflected) as usize
+    } else {
+        reflected as usize
+    }
+}
+
+fn dilate_free_space(data: &mut [i8], width: usize, height: usize, radius: usize) {
```
aclauer: all these functions like gaussian convoluions, morphological operations, and transposing have existing crates that could be used. It's fine to handwrite functions, but is there a reason to not use an existing crate?

there's actually no right answer, i'm just curious what you think because I tend to reuse a crate when possible rather than handwrite

dimos/mapping/native_costmapper/rust/src/costmapper.rs
```diff
@@ -0,0 +1,1083 @@
...
+        return Err(CostmapError::new("height limits must be finite"));
+    }
+    Ok(())
+}
+
+pub fn calculate_costmap(
+    message: &PointCloud2,
+    config: &Config,
+) -> Result<OccupancyGrid, CostmapError> {
+    let algorithm = parse_algorithm(config)?;
```
aclauer: this looks like it's getting called on every point cloud, so that's checking the algorithm and validating the config every frame. why not do that on startup and cache which algo we're using we it doesn't do extra work?

dimos/mapping/native_costmapper/rust/src/costmapper.rs
```diff
@@ -0,0 +1,1083 @@
...
+        &mut vertical_a,
+        &mut vertical_b,
+        height,
+        width,
+        &kernel,
+        radius,
+    );
+    (
+        transpose(&vertical_a, height, width),
+        transpose(&vertical_b, height, width),
```
aclauer: how much faster is it to do this convolve -> transpose -> convolve --> transpose vs doing row convolve -> column convolve?

curious what the overhead from allocating and transposing is and if it beats the cache row jumping I assume you're trying to avoid here

clearbenchmark.sh
```diff
@@ -0,0 +1,40 @@
...
+printf 'Delete all benchmark results in %s? [y/N] ' "${RESULTS_DIR}"
+read -r response
+if [[ ! "${response}" =~ ^[Yy]$ ]]; then
+    echo "Cancelled."
+    exit 0
+fi
+
+rm -rf -- "${RESULTS_DIR}"
+mkdir -p "${RESULTS_DIR}"
+echo "Benchmark results cleared."
```
aclauer: i know this is all testing, but these scripts shouldn't be in the root dir

dimos/mapping/native_costmapper/rust/src/costmapper.rs
```diff
@@ -0,0 +1,1083 @@
...
+    resolution: f64,
+    frame_id: Option<String>,
+    can_pass_under: f64,
+    can_climb: f64,
+    ignore_noise: f64,
+    smoothing: f64,
+}
+
+impl Default for HeightCostConfig {
+    fn default() -> Self {
```
aclauer: since the native modules are never spawned independently (for now at least), it's best to not have default values in Rust at all and just validate the ones coming in through the parent python process. that way we don't need to maintain multiple places for default values.

dimos/mapping/native_costmapper/rust/src/costmapper.rs
aclauer: i think it might not actually even be used in python lol

dimos/mapping/native_costmapper/rust/src/costmapper.rs
ghost: thinking about it now, i agree we should just use an existing module. i just have a habit of asking codex to write these functions because it helps when optimizing later on especially for cuda but i think that would be premature optimization at this point

dimos/mapping/native_costmapper/rust/src/costmapper.rs
ghost: agreed. this now happens once during startup

dimos/mapping/native_costmapper/rust/src/costmapper.rs
ghost: that was codex overthinking, removed while going through the generated code. gaussian filtering now uses imageproc::separable_filter_equal

clearbenchmark.sh
ghost: removed

dimos/mapping/native_costmapper/rust/src/costmapper.rs
ghost: removed t. all algorithm fields are now required during deserialization, populated from the python-provided configuration

dimos/mapping/native_costmapper/rust/src/costmapper.rs
ghost: yeah just kept for compatibility,

swordfishgt:
run_costmapper_blueprint

ran it 5 times each and profiled on CostMap vs NativeCostMap. CPU load is indeed lower after the initial setup spike, but dtop on mac does not give memory stats so cannot test for that on my end

dimos/mapping/native_costmapper/rust/src/costmapper.rs
```diff
@@ -0,0 +1,1103 @@
...
+        }
+    })?;
+
+    Ok(make_grid(
+        message,
+        &bounds,
+        config.frame_id.as_deref(),
+        data,
+    ))
+}
```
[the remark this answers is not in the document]
ghost: python removed this in f8b048997, restoring should be a separate pr for both backends

PR #2928, opened 2026-07-13
fix: disallow unused _ variables

paul-nechifor:
What it says on the tin.


Dreamsorcerer:
Isn't there a ruff rule for this?

paul-nechifor:
> Isn't there a ruff rule for this?

No, checked. @Dreamsorcerer

mustafab0 (approved):

PR #2929, opened 2026-07-14
[FEATURE]:Add Huber loss and TEASER++

aswingururaj:
## Problem

This PR enhances the Pose Graph Optimization (PGO) backend by replacing standard unweighted/quadratic (L2) factor tracking with a robust estimation architecture. While the front-end local tracking is accurate enough to generate visually acceptable maps under normal conditions, standard least-squares optimization remains vulnerable to local minima, planar slippage, and high-drift loop closure snaps.

## Solution

To address this structural risk, this PR introduces Huber Loss M-estimation to handle factor tension gracefully and integrates TEASER++ global alignment as a preprocessing step to refine raw loop closure candidate initialization guesses before local ICP refinement.

Swapped Quadratic Loss for Huber Loss: Wrapped all incoming loop closure BetweenFactorPose3 constraints in a Huber robust kernel (Huber.Create(1.345)). This limits the influence of outlying loop closures and prevents the graph from warping healthy submaps to satisfy noisy constraints.

Integrated TEASER++ Global Initialization: Upgraded the _icp routine to decouple rotation and translation calculation using Translation Invariant Measurements (TIMs), Graduated Non-Convexity (GNC-TLS), and 1D Consensus voting. This ensures local point-to-plane ICP is initialized near the global optimum rather than falling into local minima.

Integrated Telemetry Tracking: Added real-time tracking metrics to stream current graph residuals (chi2) and maximum geometric keyframe displacement (max_graph_jump_meters) directly to stdout.

Dual-panel diagnostic evaluation comparing a standard least-squares PGO baseline (Red) against the upgraded robust PGO configuration featuring TEASER++ and Huber Loss (Green).
pgo_performance_comparison

1. Left Panel: Trajectory Structural Tearing (Lower is Better)
The Baseline Spike (Event 29): Without improvements, standard local ICP gets trapped far from the global optimum. When the loop forces a closure, the optimizer executes a violent 1.32-meter physical snap to tie the ends together.

The Improved Mitigation (Event 29): With TEASER++ preprocessing providing a globally verified initialization guess, the maximum trajectory deformation is capped safely at 0.76 meters, distributing the alignment tension smoothly across historical keyframes.

2. Right Panel: Optimization Residual Profile
The Red Line (False Precision): Standard quadratic optimization forces its math score down to near-zero by bending and warping the physical trajectory to satisfy noisy inputs. Its low error score is an artifact of map distortion.

The Green Line (Robust Rigidity): The higher, climbing chi2 profile is the mathematically correct behavior of the Huber Loss M-estimator. Instead of allowing noisy constraints to warp the map, the Huber kernel caps their penalty force linearly. The keyframes remain rigid and geometrically accurate, while unavoidable real-world sensor noise accumulates safely within the residual gauge.

## How to Test

`python3 -m dimos.robot.cli.dimos map global go2_hongkong_office --export`

Note: Scripts to generate the plots are not attached here

### DIFF
--- a/dimos/mapping/loop_closure/pgo.py
+++ b/dimos/mapping/loop_closure/pgo.py
@@ -762,7 +762,7 @@ def _icp(
     dist_B = np.linalg.norm(B_tims, axis=0)
     inline_mask = np.abs(dist_A - dist_B) < max_dist
 
-    if np.sum(inline_mask) < min_inliers:
+    if np.sum(inline_mask) < min_inliers - 1:
         return Transform.identity(), float("inf")
 
     A_tims_f = A_tims[:, inline_mask]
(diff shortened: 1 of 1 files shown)


paul-nechifor:
Closed for inactivity. Can be reopened later.

PR #2939, opened 2026-07-14
feat: Galaxea A1Z arm integration (adapter, macOS CAN transport, coordinator blueprint)

jetsonearth:
## Problem

dimos has no support for the Galaxea A1Z, a 6-DOF CAN-bus arm (with G1Z gripper) we need for upcoming manipulation work. Two gaps:

- No `ManipulatorAdapter` for the vendor's A1Z SDK, so the arm can't participate in the planning stack.
- Every existing CAN arm integration assumes Linux SocketCAN. There is no way to develop against a CAN arm from macOS, which is what much of the team carries.

## Solution

Three commits, one per component:

- **`GalaxeaA1ZAdapter`** (`dimos/hardware/manipulators/galaxea_a1z/adapter.py`) - implements the `ManipulatorAdapter` protocol over the vendor a1z SDK (CAN, MIT PD + gravity comp @ 250 Hz). POSITION (min-jerk planned moves) and SERVO_POSITION (streaming) modes, latching soft e-stop, FK cartesian reads, G1Z gripper support
  - Startup is a *verified zero-force sequence*: motors enable at kp=0 and every motor must report fresh, in-limit feedback before hold gains engage at the measured pose. The vendor's stock `start()` position-holds a single 50 ms read and snaps the arm to zero when feedback arrives late (reproduced on hardware). `safe_start=False` restores stock behavior.
  - Shutdown re-sends disable frames to all motors including the gripper (the vendor sends the gripper's exactly once; a lost frame leaves it energized and unsupervised - reproduced on hardware).
- **`GsUsbMacBus`** (`gs_usb_bus.py`) - a python-can `BusABC` driving the bundled HHS USB-CANFD adapter over libusb, no SocketCAN required. Auto-selected on macOS (`transport="auto"`); Linux keeps socketcan. Though on MacOS it's likely unstable.
- **`coordinator-galaxea-a1z`** blueprint & robot config. Defaults to the stable arm-only configuration; see follow-ups for why the gripper ships dormant.

**Hardware validation** (physical A1Z & the G1Z gripper, driven from an M4 Pro MacBook): 250 Hz control loop sustained 30 s (p99 cycle 5.2 ms, 0/7500 over the SDK's 12.5 ms watchdog), coordinator boot + client-submitted multi-joint trajectories over LCM RPC.

**Known issues / follow-ups**

- Gripper defaults off: the vendor's gripper SDK branch ships a G1Z gravity model that mismatched our unit's mounting - it pushes the arm during the zero-force startup window, and disabling gravity comp as a workaround leaves the vendor's soft e-stop unable to catch the arm. Re-enable per-deployment after gravity/zero-point calibration (vendor `tools/set_zero.py`).
- CAN feedback staleness creeps up over long coordinator sessions on macOS (warnings ~50 ms; SDK e-stops at 200 ms); suspected worker-process thread contention, needs profiling.
- Generic (not arm-specific): trajectory tasks report COMPLETED even if hardware e-stops mid-trajectory; `CoordinatorClient` joint inference miscounts when a `<hw>/gripper` joint exists (7 vs 6).

## How to Test

No hardware:
```
pytest dimos/hardware/manipulators/galaxea_a1z/ -q   # 22 tests, fake SDK, no arm needed
```

With an A1Z on macOS (vendor SDK `pip install`ed, arm powered, USB-CANFD connected):
```
dimos --transport lcm run coordinator-galaxea-a1z    # boots, safe-start, holds pose; I used LCM becasue my VPN would have conflict with Zenoh, but should be an easy fix
python -m dimos.manipulation.control.coordinator_client   # submit a small trajectory to traj_arm
```


dimos/hardware/manipulators/galaxea_a1z/gs_usb_bus.py
TomCC7, line 1: python-can has gs_usb implementation, so there's no need to reimplement here. https://python-can.readthedocs.io/en/stable/interfaces/gs_usb.html

dimos/hardware/manipulators/galaxea_a1z/adapter.py
```diff
@@ -0,0 +1,660 @@
...
+        dof: int = 6,
+        *,
+        gravity_comp_factor: float = 1.0,
+        zero_gravity: bool = False,
+        control_freq_hz: int = 250,
+        urdf_path: str | None = None,
+        gripper: bool = False,
+        gripper_max_torque: float = 2.0,
+        gripper_max_opening_m: float = _GRIPPER_MAX_OPENING_M,
+        transport: str = "auto",
```
TomCC7, line 129: transport is an existing concept in dimos which is a very differnt thing, better change to a different name

dimos/hardware/manipulators/galaxea_a1z/adapter.py
```diff
@@ -0,0 +1,660 @@
...
+        address: str = "can0",
+        dof: int = 6,
+        *,
+        gravity_comp_factor: float = 1.0,
+        zero_gravity: bool = False,
+        control_freq_hz: int = 250,
+        urdf_path: str | None = None,
+        gripper: bool = False,
+        gripper_max_torque: float = 2.0,
+        gripper_max_opening_m: float = _GRIPPER_MAX_OPENING_M,
```
TomCC7, line 128: should create a config class for this instead of spamming init args

dimos/hardware/manipulators/galaxea_a1z/adapter.py
```diff
@@ -0,0 +1,660 @@
...
+    - POSITION: minimum-jerk planned move (SDK move_joints, runs in a
+      background thread so the call does not block)
+    - SERVO_POSITION: high-frequency joint position streaming
+      (SDK command_joint_pos)
+    """
+
+    def __init__(
+        self,
+        address: str = "can0",
+        dof: int = 6,
```
TomCC7, line 120: remove this dof, it should be hardcoded since a1z won't have other number of dofs

dimos/hardware/manipulators/galaxea_a1z/adapter.py
```diff
@@ -0,0 +1,660 @@
...
+                print(f"Galaxea A1Z planned move failed: {e}")
+            finally:
+                self._move_lock.release()
+
+        self._move_thread = threading.Thread(target=_move, name="a1z_planned_move", daemon=True)
+        self._move_thread.start()
+        return True
+
+    def write_joint_velocities(self, velocities: list[float]) -> bool:
+        """Not supported - the a1z SDK has no velocity command API."""
```
mustafab0, line 364: Sad if true

mustafab0 (commented):
Nice

dimos/hardware/manipulators/galaxea_a1z/adapter.py
mustafab0: This is what the robot/config file is for

jetsonearth:
Superseded by #2966 - same branch, now pushed directly to dimensionalOS/dimos after getting write access (review comments from here will be addressed there).

PR #2959, opened 2026-07-15
Control coordinator refactor to add declarative task

mustafab0:
## Problem

The `ControlCoordinator` hardcoded which task consumes which input stream and which methods are RPC-invokable. Adding a task meant editing the coordinator, and stream allowlists drifted from the tasks they mirrored.

Issue: #2958
Closes DIM-1231

---

## Solution

Tasks now declare their own I/O in `_registry.py` manifests as import-free string cards. `ControlTaskRegistry` parses them into typed `StreamBinding`/`TaskBindings` (`routing.py`), and the coordinator builds its route table and command allowlist from the registry at task-registration time. 

`task_invoke` only dispatches methods a task declares via `TASK_EXPOSES`. Twist, cartesian, velocity, teleop, servo, and g1-groot tasks migrated to cards.

---

## Breaking Changes

None

---

## How to Test

1. `.venv/bin/python -m pytest dimos/control -k "routing or coordinator_commands or registry"`
2. Register a task with a binding card; confirm the coordinator routes its stream without coordinator edits.
3. Call `task_invoke` on an undeclared method — expect a clear "declare it in TASK_EXPOSES" error.


dimos/control/coordinator.py
```diff
@@ -409,18 +425,92 @@ def add_task(self, task: ControlTask) -> bool:
                 logger.warning(f"Task {task.name} already registered")
                 return False
             self._tasks[task.name] = task
+            if task_type is not None:
+                self._register_routes(task, task_type)
+                self._task_commands[task.name] = self._commands_for(task_type)
+            else:
+                self._task_commands[task.name] = frozenset()
```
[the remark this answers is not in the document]
mustafab0: Not a concern. intended design

dimos/control/coordinator.py
```diff
@@ -535,88 +589,152 @@ def _on_twist_command(self, msg: Twist) -> None:
...
+
+            if not hasattr(task, method):
+                raise AttributeError(
+                    f"task_invoke({task_name!r}, {method!r}): task has no such method; "
+                    f"declared commands: {sorted(self._task_commands.get(task_name, frozenset()))}"
+                )
+            logger.warning(
+                f"undeclared task_invoke {task_name}.{method} — declare it in TASK_EXPOSES"
+            )
             return getattr(task, method)(**kwargs)
```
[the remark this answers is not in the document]
mustafab0: deliberate choice for backwards compatibility

dimos/control/tasks/test_registry.py
```diff
@@ -70,3 +76,251 @@ def test_declared_task_factory_paths_resolve() -> None:
         factory = getattr(module, attr, None)
         assert factory is not None, f"{name}: {module_name!r} has no attribute {attr!r}"
         assert callable(factory), f"{name}: {factory_path!r} is not callable"
+
+
+def _manifest_modules() -> list[ModuleType]:
+    pkg = importlib.import_module("dimos.control.tasks")
+    modules = []
+    for root in pkg.__path__:
+        for child in sorted(Path(root).iterdir()):
```
spomichter: Flagging this may break depending on install method

dimos/control/tasks/test_registry.py
```diff
@@ -70,3 +76,251 @@ def test_declared_task_factory_paths_resolve() -> None:
...
+            if not child.is_dir() or child.name.startswith(("_", ".")):
+                continue
+            if not (child / "_registry.py").exists():
```
spomichter: And this whole block looks like bad practice

spomichter (dismissed):

dimos/control/tasks/test_registry.py
mustafab0: Verified this empirically, made claude resolve what pip wheel packages. 

If someone pip installs dimos. This will not break

dimos/control/tasks/test_registry.py
mustafab0: I cleaned this up, kept the enumeration because it's an intentional independent check, but tidied up the rest

dimos/control/_control_test_helpers.py
```diff
@@ -0,0 +1,74 @@
...
+    def __init__(self, name: str, joints: frozenset[str] = frozenset()) -> None:
+        self._name = name
+        self._joints = frozenset(joints)
+        self.cartesian_calls: list[tuple[Any, float]] = []
+        self.ee_twist_calls: list[tuple[Any, float]] = []
+        self.buttons_calls: list[Any] = []
+
+    @property
+    def name(self) -> str:
+        return self._name
```
paul-nechifor: I mentioned before, but you shouldn't have to implement this method everywhere. It's always the same. It should be in the base. Same for other methods.

dimos/control/coordinator.py
```diff
@@ -440,81 +530,45 @@ def get_active_tasks(self) -> list[str]:
...
-                logger.warning("EEF twist command for unknown task", task_name=task_name)
+            entries = self._routes.get(stream)
+            if not entries:
                 return
-            task.on_ee_twist_command(msg, t_now)
 
-    def _on_twist_command(self, msg: Twist) -> None:
-        """Convert Twist → virtual joint velocities and route via _on_joint_command.
+            claimable: set[str] | None = None
+            frame_id = getattr(msg, "frame_id", "")
```
paul-nechifor, line 555: The previous code was using `msg.frame_id`, why was it changed to `getattr(msg, "frame_id", "")`?

dimos/control/_control_test_helpers.py
mustafab0: fixed

dimos/control/coordinator.py
mustafab0: Previously each the messages were processed by their separate methods depending on message type but now the messages are dispatched by a common dispatcher.

This is a temporary solution because control coordinator has a fixed number of input ports.

My following PRs will remove this when I have separate In ports for each task this wouldn't be necessary

paul-nechifor (approved):

PR #2970, opened 2026-07-15
WIP: feat: web project setup

paul-nechifor:
## Problem

Closes DIM-XXX

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -230,6 +230,32 @@ jobs:
       #   working-directory: docs
       #   run: npm run validate
 
+  web:
+    timeout-minutes: 15
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read  # For checkout
+
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Extract Deno version
+        id: deno-version
+        run: echo "version=$(grep -oP 'DENO_VERSION = "\K[^"]+' dimos/utils/deno.py)" >> "$GITHUB_OUTPUT"
+
+      - uses: denoland/setup-deno@v2
+        with:
+          deno-version: ${{ steps.deno-version.outputs.version }}
+          cache: true
+
+      - name: Deno fmt, lint, check, test
+        working-directory: web
+        run: |
+          deno fmt --check
+          deno lint
+          deno task check
+          deno task test
+
   tests:
     timeout-minutes: 20
     strategy:
@@ -729,6 +755,7 @@ jobs:
     - rust
     - md-babel
     - docs-validate
+    - web
     - tests
     - cmu-nav-natives
     - self-hosted-tests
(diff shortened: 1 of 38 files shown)


.github/workflows/ci.yml
```diff
@@ -230,6 +230,39 @@ jobs:
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read  # For checkout
+
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: denoland/setup-deno@v2
+        with:
+          deno-version: v2.6.10  # keep in sync with dimos/utils/deno.py DENO_VERSION
```
Dreamsorcerer: Add a step to extract the version as a variable?

.github/workflows/ci.yml
```diff
@@ -230,6 +230,39 @@ jobs:
...
+          cache: true
+
+      - name: Deno fmt, lint, check, test
+        working-directory: web
+        run: |
+          deno fmt --check
+          deno lint
+          deno task check
+          deno task test
```
Dreamsorcerer, line 257: Should really be separate steps.

dimos/utils/deno.py
```diff
@@ -0,0 +1,88 @@
...
+        return which
+
+    exe_name = "deno.exe" if platform.system() == "Windows" else "deno"
+    deno_dir = STATE_DIR / "deno" / DENO_VERSION
+    deno_path = deno_dir / exe_name
+    if deno_path.exists():
+        return str(deno_path)
+
+    triple = _deno_triple()
+    url = f"https://github.com/denoland/deno/releases/download/{DENO_VERSION}/deno-{triple}.zip"
```
Dreamsorcerer, line 46: Can we include a check or something that the release is immutable, in case the project ever disables immutable releases in future?

dimos/web/relay_bridge/_wt_session.py
Dreamsorcerer, line 1: This is probably a really bad idea. Rolling your own HTTP code from a quic abstraction is likely to cause problems. We should really look at adding quic support to aiohttp instead...

From what I can see, I think the actual TLS connection stack is a minimal implementation for testing purposes, not something that should be used in production.

.github/workflows/ci.yml
paul-nechifor: Done.

.github/workflows/ci.yml
paul-nechifor: But it seems fine this way... Why add more lines?

dimos/utils/deno.py
paul-nechifor: You mean add the hash here and check it?

.github/workflows/ci.yml
Dreamsorcerer: It improves debugability, and is less likely to run into weird edge cases. For example, if the lint step fails, it probably won't fail the workflow if run on a windows runner. Best practice is to keep one step == one command.

dimos/utils/deno.py
Dreamsorcerer: I mean the GH releases are tagged as immutable. So, the simple option (assuming it's trivial to get via API) is to simply check that the release is immutable. In that case, the hash can't change anyway, and we save ourselves the extra complexity of managing hashes.

PR #2981, opened 2026-07-15
declared_streams hook for config-derived module streams

mustafab0:
## Problem

A module's stream ports could only come from class annotations (`x: In[T]`). Tasks that need ports whose shape depends on config (per-task inputs, per-robot outputs) had no way to declare them.

## Solution

Adds `ModuleBase.declared_streams(cls, kwargs) -> Sequence[StreamRef]`, a pure hook (default `()`) consulted at the two existing annotation-derivation sites — `BlueprintAtom.create` (host) and `Module.__init__` (worker). Both get the same kwargs, so host, worker, and restart agree on the ports. No-op for every existing module.

---

## Breaking Changes

None

---

## How to Test

1. `uv run pytest dimos/core/coordination/test_declared_streams.py`
2. See the end-to-end test: a zero-annotation consumer receives a message on a config-derived port.
3. `uv run pytest dimos/core dimos/control` — existing suites unchanged.

- [ x] I have read and approved the [CLA](https://github.com/dimensionalOS/dimos/blob/main/CLA.md).

### DIFF
--- a/dimos/core/coordination/blueprints.py
+++ b/dimos/core/coordination/blueprints.py
@@ -19,7 +19,7 @@ import operator
 import sys
 import types as types_mod
 from types import MappingProxyType
-from typing import TYPE_CHECKING, Any, Literal, Union, get_args, get_origin, get_type_hints
+from typing import TYPE_CHECKING, Any, Union, get_args, get_origin, get_type_hints
 
 from pydantic import BaseModel, create_model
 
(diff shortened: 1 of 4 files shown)


leshy:
> A module's stream ports could only come from class annotations (`x: In[T]`). Tasks that need ports whose shape depends on config (per-task inputs, per-robot outputs) had no way to declare them.

I don't really understand this, can you elaborate on what the issue is?

edit, found it https://discord.com/channels/1341146487186391173/1470305322437578887/1527171916249698354

mustafab0:
As discussed with @leshy and @paul-nechifor .

This hook may not be necessary for declarative streams on the control coordinator. As subclassing is the canonical way of adding custom I/O with proper type checking.

Bloating the API does not provide as much benefit and could cause more issues than it solves.

PR #2983, opened 2026-07-15
Fix/ivan/3d planner voxel colors

leshy:
slightly nicer colors IMHO but sloppy so your call


dimos/navigation/nav_3d/mls_planner/utils/plan_rrd.py
```diff
@@ -124,20 +138,53 @@ def _log_odometry(
...
+        & (rel[:, 2] <= crop.above)
+    )
+    local = rel[keep]
+    if local.size == 0:
+        rr.log("world/local/voxel_map", rr.Points3D([]))
+        return
+    # Its own turbo: spread over the crop's own height range, not the building's,
+    # so a 1 m band of floor still reads as a full gradient.
+    z = local[:, 2]
+    class_ids = ((z - z.min()) / (z.max() - z.min() + 1e-8) * 255).astype(np.uint8)
```
spomichter, line 186: As a nit this same hight--> class id conversion used twice

spomichter, line 186: Also im pretty sure this will cause color flicker as the robot walks and your cropped area height changes but the height of the voxels dont chagne , but can just see and fix later not worth testing this as small and cosmetic

spomichter (approved):

dimos/navigation/nav_3d/mls_planner/utils/plan_rrd.py
leshy: no, it's good

leshy: turbo is calculated twide depending on the scale, for local we have a diff scale

aclauer (approved):

PR #2985, opened 2026-07-15
feat(core): add external Python modules

TomCC7:
## Problem

DimOS Python modules cannot use their own local dependency environment while preserving normal Blueprint, RPC, stream, module-reference, and restart behavior.

Closes DIM-1222

## Solution

Add `ExternalPythonModule` declarations backed by a sibling uv runtime project and a private `external-python` worker manager. The declaration owns the RPC contract; the runtime imports that contract directly from the packaged example, without `PYTHONPATH` or a separate contract distribution.

The implementation adds lifecycle diagnostics, process-group cleanup, restart-safe Blueprint overrides, a runnable example external package, and real end-to-end coverage.

## How to Test

```bash
uv run python examples/external_python_module/deployment.py
uv run pytest dimos/core/test_external_python_end_to_end.py -q
```


dimos/core/coordination/external_python_worker.py
```diff
@@ -0,0 +1,231 @@
...
+
+from dimos.core.external_python_module import ExternalPythonModule
+
+
+class ExternalPythonRuntime:
+    """Resolve, prepare, and own one external Python process."""
+
+    startup_timeout = 30.0
+    shutdown_timeout = 5.0
+    output_limit = 64 * 1024
```
TomCC7, line 39: what is this for

dimos/core/coordination/external_python_worker.py
```diff
@@ -0,0 +1,231 @@
...
+
+    def __init__(
+        self,
+        declaration: type[ExternalPythonModule],
+        global_config: object,
+        kwargs: dict[str, object],
+    ) -> None:
+        self.declaration = declaration
+        self.global_config = global_config
```
TomCC7: where is this variable used??

dimos/core/coordination/external_python_worker.py
```diff
@@ -0,0 +1,231 @@
...
+        self.kwargs = kwargs
```
TomCC7: what is kwargs here

examples/external_python_module/python/example_external/runtime.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def _publish_doubled(self, message: Int32) -> None:
+        self.doubled.publish(Int32(message.data * self._multiplier))
+
+    @rpc
+    def get_multiplier(self) -> int:
+        return self._multiplier
+
+    @skill
+    def set_multiplier(self, multiplier: int) -> str:
```
TomCC7, line 41: this is not used, try to use it?

TomCC7: Now exercised by the runnable example and real E2E: it starts at configured 3, sets to 5 through set_multiplier(), then verifies restart resets to 3.

dimos/core/coordination/external_python_worker.py
TomCC7: Added an inline comment: startup_timeout waits for READY; shutdown_timeout is the SIGTERM grace period before SIGKILL; output_limit retains the final 64 KiB of each stdout/stderr stream for failure diagnostics.

dimos/core/coordination/external_python_worker.py
TomCC7: Renamed this to constructor_kwargs. These are normal declaration Blueprint arguments, for example ExampleExternal.blueprint(initial_multiplier=3), then serialized to the child constructor along with injected global config.

dimos/core/coordination/external_python_worker.py
TomCC7: Removed the unused runtime global_config state. The manager still injects g into constructor_kwargs because Module configuration needs it.

pyproject.toml
```diff
@@ -22,7 +22,7 @@ include-package-data = false
 
 [tool.setuptools.packages.find]
 where = ["."]
-include = ["dimos*"]
+include = ["dimos*", "examples*"]
```
paul-nechifor: Does this include the examples dir in the package published to PyPI?

examples/external_python_module/python/.gitignore
```diff
@@ -0,0 +1,2 @@
+uv.lock
```
paul-nechifor, line 1: Why ignore lock files?

pyproject.toml
TomCC7: yes, should revert

examples/external_python_module/python/.gitignore
TomCC7: should we keep them? I feel like there's no need to keep locks for example purpose

TomCC7:
close in lieu of #3478

PR #2993, opened 2026-07-16
ci: skip code test jobs on docs-only PRs

spomichter:
A docs typo currently runs the full suite: ~115 min of runner time (8-way `tests` matrix, rust, self-hosted Linux + macOS, large-tests) to check a banner string. #2984 is the example.

A `changes` job answers "did anything outside `docs/` move?"; `rust`, `tests` and `self-hosted-*` are gated on it. `lint`, `docs-validate` and `md-babel` still run on docs — md-babel executes the code blocks *in* the docs.

**Why not `paths-ignore` on the trigger:** `ci-complete` is the required check. A workflow that never runs never reports it, so docs PRs would hang forever on "waiting for required check". Gating jobs keeps `ci-complete` reporting; `rust`/`tests` join the `allowed-skips` it already has.

Fails open — non-PR events (push, merge_group) always get `code: true`. Fork guards on `self-hosted-*` are extended with `&&`, not replaced. `cmu-nav-natives` is left alone (already cache-gated, self-skips on docs).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,21 @@ env:
   UV_NO_SYNC: "1"
 
 jobs:
+  changes:
+    # Is anything outside docs/ touched? Docs-only PRs skip the code jobs.
+    # Non-PR events (push, merge_group) always get 'true' — they run everything.
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read  # For paths-filter to list the PR's files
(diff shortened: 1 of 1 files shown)


Dreamsorcerer:
> A docs typo currently runs the full suite: ~115 min of runner time

Actual blocking time is about 25 mins though (i.e. the slowest job). I've been hoping to avoid adding in this complexity.

PR #3042, opened 2026-07-18
feat(web): PR 1 - Deno

paul-nechifor:
* Starting the foundation of the web project (just the shared stuff).
* This PR just handles the Deno setup in `web`.
* I've moved ensure_deno outside DimSim since DimSim isn't the only thing using Deno now.
* I've set up the CI for web.

Part of https://linear.app/dimensional/issue/DIM-1163/webtransport-foundation

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,6 +55,18 @@ jobs:
         run: uv sync --only-group lint --frozen
       - name: Mypy
         run: uv run mypy
+
+      # The deno-fmt pre-commit hook needs deno on PATH.
+      - name: Extract Deno version
+        id: deno-version
+        run: echo "version=$(grep -oP 'DENO_VERSION = "\K[^"]+' dimos/utils/deno.py)" >> "$GITHUB_OUTPUT"
+
+      - name: Install Deno
+        uses: denoland/setup-deno@v2
+        with:
+          deno-version: ${{ steps.deno-version.outputs.version }}
+          cache: true
+
       - name: Run pre-commit
         uses: pre-commit/action@v3.0.1
 
@@ -229,6 +241,42 @@ jobs:
       # - name: Validate Mintlify docs
       #   working-directory: docs
       #   run: npm run validate
+
+  web:
+    timeout-minutes: 15
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read  # For checkout
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Extract Deno version
+        id: deno-version
+        run: echo "version=$(grep -oP 'DENO_VERSION = "\K[^"]+' dimos/utils/deno.py)" >> "$GITHUB_OUTPUT"
+
+      - name: Install Deno
+        uses: denoland/setup-deno@v2
+        with:
+          deno-version: ${{ steps.deno-version.outputs.version }}
+          cache: true
+
+      - name: Deno fmt
+        working-directory: web
+        run: deno fmt --check
+
+      - name: Deno lint
+        working-directory: web
+        run: deno lint
+
+      - name: Deno check
+        working-directory: web
+        run: deno task check
+
+      - name: Deno test
+        working-directory: web
+        run: deno task test
 
   tests:
     timeout-minutes: 20
@@ -733,6 +781,7 @@ jobs:
     - rust
     - md-babel
     - docs-validate
+    - web
     - tests
     - cmu-nav-natives
     - self-hosted-tests
--- a/dimos/utils/deno.py
+++ b/dimos/utils/deno.py
@@ -1,0 +1,103 @@
+# Copyright 2026 Dimensional Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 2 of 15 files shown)


.github/workflows/ci.yml
```diff
@@ -55,6 +55,17 @@ jobs:
         run: uv sync --only-group lint --frozen
       - name: Mypy
         run: uv run mypy
+
+      # The deno-fmt pre-commit hook needs deno on PATH.
+      - name: Extract Deno version
+        id: deno-version
+        run: echo "version=$(grep -oP 'DENO_VERSION = "\K[^"]+' dimos/utils/deno.py)" >> "$GITHUB_OUTPUT"
```
Dreamsorcerer, line 62: This is probably fine, but are we unable to extract the version by directly importing via python (give that uv has already been setup)? Like `python -c 'from dimos.utils.deno import DENO_VERSION; print("version=" + repr(DENO_VERSION))` or something.

.github/workflows/ci.yml
```diff
@@ -230,6 +241,32 @@ jobs:
...
+      contents: read  # For checkout
+
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Extract Deno version
+        id: deno-version
+        run: echo "version=$(grep -oP 'DENO_VERSION = "\K[^"]+' dimos/utils/deno.py)" >> "$GITHUB_OUTPUT"
+
+      - uses: denoland/setup-deno@v2
```
Dreamsorcerer: For consistency, I've been using name on all steps. I think this is best practice.

.github/workflows/ci.yml
```diff
@@ -230,6 +241,32 @@ jobs:
...
+          deno-version: ${{ steps.deno-version.outputs.version }}
+          cache: true
+
+      - name: Deno fmt, lint, check, test
+        working-directory: web
+        run: |
+          deno fmt --check
+          deno lint
+          deno task check
+          deno task test
```
Dreamsorcerer: As mentioned before, best practice is to use 4 separate steps.
Clearer debugging in the logs, less likely to break in subtle ways (e.g. on Windows runner) etc.

dimos/utils/deno.py
```diff
@@ -0,0 +1,88 @@
...
+        return which
+
+    exe_name = "deno.exe" if platform.system() == "Windows" else "deno"
+    deno_dir = STATE_DIR / "deno" / DENO_VERSION
+    deno_path = deno_dir / exe_name
+    if deno_path.exists():
+        return str(deno_path)
+
+    triple = _deno_triple()
+    url = f"https://github.com/denoland/deno/releases/download/{DENO_VERSION}/deno-{triple}.zip"
```
Dreamsorcerer, line 47: As mentioned before, if we can add a check that the release is labelled as immutable, that would help protect us from supply chain attacks.

dimos/utils/deno.py
```diff
@@ -0,0 +1,88 @@
...
+        if machine in ("x86_64", "amd64"):
+            return "x86_64-apple-darwin"
+        if machine in ("arm64", "aarch64"):
+            return "aarch64-apple-darwin"
+    elif system == "Windows" and machine in ("amd64", "x86_64"):
+        return "x86_64-pc-windows-msvc"
+    raise RuntimeError(
+        f"Unsupported platform for deno auto-install: {system} {machine}. "
+        "Install deno manually from https://deno.com/"
+    )
```
Dreamsorcerer, line 103: Probably cleaner with a match statement?

.github/workflows/ci.yml
paul-nechifor: Tried this, but it's not worth it because you have to install the python dependencies or move the imports inside functions.

> give that uv has already been setup

Not quite, above only has `uv sync --only-group lint --frozen`

.github/workflows/ci.yml
paul-nechifor: Done.

.github/workflows/ci.yml
paul-nechifor: Done.

dimos/utils/deno.py
paul-nechifor: Done.

dimos/utils/deno.py
paul-nechifor: Done.

web/shared/protocol.ts
```diff
@@ -0,0 +1,272 @@
...
+  push(chunk: Uint8Array): Msg[] {
+    const merged = new Uint8Array(this.#buf.length + chunk.length);
+    merged.set(this.#buf, 0);
+    merged.set(chunk, this.#buf.length);
+    this.#buf = merged;
+    const msgs: Msg[] = [];
+    while (this.#buf.length >= 4) {
+      const len = new DataView(this.#buf.buffer, this.#buf.byteOffset).getUint32(0, true);
+      if (len > MAX_HEADER_LEN) throw new Error(`control frame too large: ${len}`);
+      if (this.#buf.length < 4 + len) break;
```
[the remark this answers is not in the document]
paul-nechifor: Fixed.

leshy (approved):

Dreamsorcerer (approved):
I've not really looked at the TS, but rest looks good.

PR #3043, opened 2026-07-18
feat(web): PR 2 - Relay

paul-nechifor:
* Added the code for an initial relay, which handles connections from the robot (DimOS) and the frontend (cockpit).
* This is just a start. I definitely want to break up `web/relay/server.ts` into smaller more cohesive parts.
* Added a cockpit stub in a single file: `web/relay/static/debug.html`. This will be replaced.

### DIFF
--- a/web/README.md
+++ b/web/README.md
@@ -1,13 +1,14 @@
 # DimOS web
 
 Deno workspace for the robot web stack. `shared/` holds the wire protocol and its golden vectors;
-the WebTransport relay (`relay/`, with `deno task dev`), its Python mirror + client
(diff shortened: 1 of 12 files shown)


Dreamsorcerer:
This is perfect example of where we could have used a stacked PR.

leshy:
Code looks clean in both PRs (https://github.com/dimensionalOS/dimos/pull/3044/) but just to understand the architectural approach - why do we have `(zenoh) -> python -> (webtransport) -> deno -> (webtransport) -> clients`

why not either connect deno directly to zenoh (idk if possible) or serve clients directly from python? (hoping to get rid of the middle guy)

leshy (approved):
had an arch chat in private

PR #3044, opened 2026-07-18
feat(web): PR 3 - Python client

paul-nechifor:


dimos/web/relay_bridge/_wt_session.py
Dreamsorcerer, line 1: From previous PR, this is probably a bad idea. Should probably focus on adding this to aiohttp instead.

aioquic is designed as a foundation for HTTP libraries/frameworks to implement on top of. I get the impression that the TLS implementation in aioquic is meant for trivial testing/debugging purposes, not for production use.

dimos/web/relay_bridge/protocol.py
```diff
@@ -0,0 +1,323 @@
...
+
+
+@dataclass
+class Stop:
+    t: ClassVar[str] = "stop"
+    seq: int
+    ts: float
+
+
+Msg = Union[Hello, Welcome, Ping, Pong, Error, Twist, Stop]
```
Dreamsorcerer: Think we should be able to use `|` here..

dimos/web/relay_bridge/protocol.py
```diff
@@ -0,0 +1,323 @@
...
+    cls.t: cls for cls in (Hello, Welcome, Ping, Pong, Error, Twist, Stop)
+}
+
+# Runtime field validation, derived from the dataclass annotations (mirrors
+# MSG_FIELDS in protocol.ts): "string" is a JSON string, "number" any JSON
+# number except booleans. An unmapped annotation fails loudly at import.
+_KIND_BY_ANNOTATION = {"int": "number", "float": "number", "str": "string", "Role": "string"}
+_MSG_FIELD_KINDS: dict[str, dict[str, str]] = {
+    # cast: `from __future__ import annotations` makes f.type always a str.
+    t: {f.name: _KIND_BY_ANNOTATION[cast("str", f.type)] for f in fields(cls)}
```
Dreamsorcerer: No, you can't do this. `from __future__ import annotations` is a backward-compatibility stop-gap that will be removed in a future release. The correct behaviour that all future releases will follow is the default (without the import) in Python 3.14+. This doesn't return strings all the time (which would be lame...).

dimos/web/relay_bridge/protocol.py
```diff
@@ -0,0 +1,323 @@
...
+    for t, cls in _MSG_TYPES.items()
+}
+
+
+def _is_kind(value: Any, kind: str) -> bool:
```
Dreamsorcerer: This feels really awkward. Can we not just typecheck the entire message with pydantic's TypeAdapter (think we've already done this elsewhere in the codebase..)?

dimos/web/relay_bridge/protocol.py
```diff
@@ -0,0 +1,323 @@
...
+    header: FrameHeader
+    payload: bytes
+
+
+def _dump_json(obj: dict[str, Any]) -> bytes:
+    # Canonical form shared with JSON.stringify: compact separators, raw UTF-8.
+    return json.dumps(obj, separators=(",", ":"), ensure_ascii=False).encode()
+
+
+def _msg_to_dict(msg: Msg) -> dict[str, Any]:
```
Dreamsorcerer: Is there a problem with dataclass.asdict()?

dimos/web/relay_bridge/protocol.py
```diff
@@ -0,0 +1,323 @@
...
+def msg_from_dict(data: dict[str, Any]) -> Msg:
+    t = data.get("t")
+    if not isinstance(t, str) or t not in _MSG_TYPES:
+        raise ProtocolError(f"unknown message type: {t!r}")
+    cls = _MSG_TYPES[t]
+    for name, kind in _MSG_FIELD_KINDS[t].items():
+        if not _is_kind(data.get(name), kind):
+            raise ProtocolError(f"invalid {t} message: field {name!r} is not a {kind}")
+    msg: Msg = cls(**{f.name: data[f.name] for f in fields(cls)})
+    return msg
```
Dreamsorcerer: With pydantic, I think this could literally just be:
```
return MsgTA.validate_python(data)
```

dimos/web/relay_bridge/protocol.py
```diff
@@ -0,0 +1,323 @@
...
+                raise ProtocolError(f"control frame too large: {length}")
+            if len(self._buf) < 4 + length:
+                break
+            body_bytes = self._buf[4 : 4 + length]
+            del self._buf[: 4 + length]
+            try:
+                body = json.loads(body_bytes.decode())
+                if not isinstance(body, dict):
+                    raise ProtocolError("control frame is not a JSON object")
+                msgs.append(msg_from_dict(body))
```
Dreamsorcerer: Can introduce the TypeAdapter from here too, and have pydantic handle the JSON conversion too.

dimos/web/relay_bridge/protocol.py
```diff
@@ -0,0 +1,323 @@
...
+    if (
+        not isinstance(ch, str)
+        or not _is_kind(seq, "number")
+        or not _is_kind(ts, "number")
+        or delivery not in ("latest", "reliable")
+        or ("meta" in body and not isinstance(meta, dict))
+    ):
+        raise ProtocolError(f"invalid data frame header: {body!r}")
+    # cast: _is_kind proved seq/ts are numbers, which is as precise as JSON
+    # gets (no int/float split on the wire, mirroring protocol.ts).
```
Dreamsorcerer: TypeAdapter or a TypeIs annotation should make this type check correctly.
In general, avoid cast() if at all possible. If you can't type check correctly, an assert is usually better.

dimos/web/relay_bridge/protocol.py
Dreamsorcerer, line 1: Is there any possibility of an existing library being able to handle this. This is exactly the kind of code that commonly has security vulnerabilities.

dimos/web/relay_bridge/_wt_session.py
paul-nechifor: * But we can't wait for `aiohttp` to add it.
* `aioquic` looks stable and it's used by several other big projects.
* If not `aioquic`, what current alternatives are there?

dimos/web/relay_bridge/protocol.py
paul-nechifor: I don't know. It would have to work across languages, and we might move to Rust anyway later.

Do you have suggestions?

dimos/web/relay_bridge/_wt_session.py
Dreamsorcerer: > `aioquic` looks stable and it's used by several other big projects.

Like what? If I ask Claude, it seems to confirm what I just said:

> aioquic (the Python QUIC + HTTP/3 stack by aiortc) is designed to be embedded into other libraries, so its dependents are mostly other Python networking projects. The ones named in aioquic's own README/PyPI:      
>                                 
> - dnspython — DNS over QUIC (DoQ) support                                                                                                                                                                            
> - Hypercorn — ASGI server, uses it for HTTP/3                                                                                                                                                                        
> - mitmproxy — intercepting proxy with QUIC/HTTP/3 support (via a dedicated fork, aioquic-mitmproxy)                                                                                                                  
> - Web Platform Tests (WPT) — W3C's cross-browser test suite

paul-nechifor: Yes, those are the ones. And it fits dimos perfectly because dimos is also a library. :)

You haven't answered what the alternative is.

dimos/web/relay_bridge/protocol.py
paul-nechifor: Done.

dimos/web/relay_bridge/protocol.py
paul-nechifor: Done.

dimos/web/relay_bridge/protocol.py
paul-nechifor: Done.

dimos/web/relay_bridge/protocol.py
paul-nechifor: Not needed since switched to Pydantic's model_dump_json

dimos/web/relay_bridge/protocol.py
paul-nechifor: Done.

dimos/web/relay_bridge/protocol.py
paul-nechifor: Done.

dimos/web/relay_bridge/protocol.py
paul-nechifor: Done.

leshy (dismissed):
Sam is right about the security issue, but also paul has lots on his plate now, so I'd like to let him merge and we should revisit web dir discovery

dimos/web/relay_bridge/_wt_session.py
Dreamsorcerer: > Yes, those are the ones. And it fits dimos perfectly because dimos is also a library. :)

No, the key word in both Claude and my response, is _networking_ library. These libraries _provide_ the networking, they don't use aioquic's code, which as I said is presumably only meant for demo/testing purposes.

If we can prioritise my time on aiohttp, I could start looking at adding aioquic. Maybe we use this as a stopgap now, but we shouldn't really be using in full production.

dimos/web/relay_bridge/protocol.py
Dreamsorcerer: If I'm reading correctly, we're literally creating our own binary level protocol? Surely there's an established protocol we could use?

paul-nechifor: We're not creating our own protocol. We're just using WebTransport which is built on QUIC. `aioquic` implements WebTransport for Python. There are alternative libraries, like pywebtransport. But that only has 20 stars on GitHub compared to 2000 for aioquic.

Yes, we do send binary messages. But that's not implementing our own protocol. We're just using WebTransport. If sending binary messages is implementing our own protocol, then sending a POST request to a REST endpoint with an image is also "creating our own binary level protocol".

dimos/web/relay_bridge/_wt_session.py
paul-nechifor: > only meant for demo/testing purposes

Does `aioquic` say that?

But I don't know what you want me to do? Not implement this?

mustafab0 (approved):

dimos/web/relay_bridge/protocol.py
Dreamsorcerer: I mean, WebTransport is the transport layer protocol, to transfer the data, but then we're creating our own messaging protocol on top of that, if I'm understanding correctly. The code looks like a parser for that protocol. Whereas POSTing an image (unless you do something weird) would just leverage existing HTTP and image parsers (which have both likely been through far more rigorous security testing).

A safer custom protocol would be one built on JSON, for example. An existing binary protocol would be protobuf.

dimos/web/relay_bridge/_wt_session.py
Dreamsorcerer: > Does `aioquic` say that?

It's pretty vague, it just says "minimal TLS 1.3 implementation". Minimal doesn't suggest robust to me.. They don't really highlight the feature beyond that, so it's clearly not a feature they care about that much..

> But I don't know what you want me to do? Not implement this?

If we need it now, I guess merge and create a ticket to come back to it. If we want to depend on this for important features (especially anything web based), then we should prioritise some work on aiohttp.

PR #3047, opened 2026-07-18
feat: rewrite go2 voxel + cost mapper as Rust native modules (#2820)

kumarganduri:
## Contribution path

- [x] Linked issue or discussion: #2820 (DIM-1153)

## Problem

The unitree-go2 mapping stack (`VoxelGridMapper` + `CostMapper`) runs as Python
modules. Profiling them on the `go2_short` replay (60 s, 7.7 Hz lidar, ~23k
points/frame) shows per-frame cost growing with map size: `add_frame` went from
8.9 ms to 22 ms within one minute of mapping. On longer runs this trends 
toward the 130 ms frame budget. The costmap path also pays a numba JIT stall 
on the first costmap after boot: ~180 ms with a cold cache, ~90 ms once numba's 
on-disk cache (@njit(cache=True)) is warm.

## Solution

Behavior-preserving Rust ports of both modules on the `dimos-module` framework
(same ports, message types, config knobs), swappable via the new
`unitree-go2-rust-mapping` blueprint. Same algorithms, including deliberately
keeping the naive O(map) column-carve so the comparison is apples-to-apples --
an XY-indexed carve is a documented follow-up.

Correctness: golden tests replay the recorded frames through both
implementations -- voxel key sets match exactly and costmaps are bit-identical
at all five checkpoints (final: 143k voxels, 280x455 grid).

Benchmarks on the identical replayed frames (CPU-only, Apple silicon; details
and caveats in `scripts/benchmarks/README.md`):

| stage | Python mean | Rust mean | speedup |
|---|---|---|---|
| add_frame | 14.9 ms | 0.73 ms | 20x |
| emit | 5.8 ms | 0.29 ms | 20x |
| costmap | 3.2 ms | 2.2 ms | 1.5x (numba/scipy is already native) |

End-to-end on the live replay: lidar->map latency 31 -> 3.2 ms mean (p95
58 -> 5.1 ms); worst costmap hop 125 ms (JIT) -> 11 ms; ~840 MB less peak RSS.

Key decisions:
- cargo build for now (`cargo build --release`); a nix flake matching the other
  native modules is a follow-up (needs a nix machine for the cargoHash).
- `block_count`/`device` (Open3D/CUDA-specific) intentionally dropped; on CUDA
  machines the Python mapper can use GPU — this port is CPU-only. The GPU
  comparison is still open: `scripts/benchmarks/run_all.py` runs Python-CUDA,
  Python-CPU and Rust on one machine and prints the table, so it can be
  settled on any CUDA box.
- Only `height_cost` is ported (the go2 default); other OCCUPANCY_ALGOS remain
  Python-only and the config rejects them.
- `frame_id` is exposed as `map_frame_id` (reserved on NativeModuleConfig).

## How to Test

    cd dimos/mapping/rust && cargo build --release && cargo test --release
    uv run dimos --replay run unitree-go2-rust-mapping

    # full offline comparison, one table out
    uv run python scripts/benchmarks/run_all.py

Golden data + full benchmark reproduction: `scripts/benchmarks/README.md`.
Replay-tested **only**; not tested on robot hardware.

## AI assistance

Done with the help of Claude Code (Fable 5).

## Checklist

- [x] This PR is scoped to one issue or clearly stated problem.
- [x] I ran the relevant checks (`uv run pytest`, pre-commit) for the files I changed.
- [x] I have reviewed and understood every line in this PR.
- [x] I disclosed AI assistance above.


leshy:
> block_count/device (Open3D/CUDA-specific) intentionally dropped; on CUDA
> machines the Python mapper can use GPU — this port is CPU-only.

Hi, we should compare speed with py side on GPU as intended, this is required for us to make a decision when and what to use

kumarganduri:
> > > block_count/device (Open3D/CUDA-specific) intentionally dropped; on CUDA
> > > machines the Python mapper can use GPU — this port is CPU-only.
> > 
> > 
> > Hi, we should compare speed with py side on GPU as intended, this is required for us to make a decision when and what to use


Hello,

Agreed, that's the number that matters for a deploy decision.

To be explicit about what's in the PR today: I benchmarked on an Apple-silicon
Mac, where Open3D falls back to `CPU:0`, so the table is CPU-vs-CPU and the
voxel-mapper speedups are against the fallback path, not the intended CUDA one.
Bolting a GPU column onto those numbers wouldn't be valid either, since it'd be
a different machine.

So instead I'll make the benchmark produce all three configs in one command on
one machine - Python `CUDA:0`, Python `CPU:0`, and Rust - each recording the
device it actually resolved to. That gives you the deployment comparison, plus
how much the GPU is really buying, with no hardware confounds. (The scripts
currently hardcode `CUDA:0`, so this also fixes the CPU-only-contributor issue
Greptile flagged.)

One part is hardware-independent either way: the costmap has no GPU path in
Python - `height_cost_occupancy` is numba + scipy on CPU - so only
`add_frame`/`emit` are open questions.

I don't have CUDA hardware on my side, so the run itself would need to come from
yours, a GPU box or a self-hosted runner. I'll wire it as a single command
against the go2_short replay; if you can run it, paste the output and I'll
update the results table and write up the when-to-use-which guidance.

PR #3116, opened 2026-07-21
Setup Sphinx

Dreamsorcerer:
Moved the existing docs to old-docs/ to allow us to migrate them over, so just ignore all those changes.

This just sets up the index page, with the start of an API reference (just whatever is referenced by the index page currently).

The rest of the docs can then be migrated pretty easily, one section at a time (to reduce conflict hell).


.github/workflows/ci.yml
```diff
@@ -212,24 +212,44 @@ jobs:
...
+      #   working-directory: docs-old
       #   run: npm run validate
 
+  docs:
+    timeout-minutes: 15
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read  # For checkout
+    steps:
+      - uses: actions/checkout@v7
```
```suggestion
      - name: Checkout
        uses: actions/checkout@v7
```

paul-nechifor:
I guess not needed anymore now that we use mkdocs

PR #3117, opened 2026-07-21
feat: add agent-triggered Go2 episode recording

yusufjarada:
## Contribution path

## Problem

DimOS can already persist robot streams and export episodes to LeRobot/HDF5, but an agent cannot mark labeled training episodes at runtime. Existing collection workflows depend on teleop buttons.

## Solution

- Expose `start_recording`, `stop_recording`, and `discard_recording` directly from `EpisodeMonitorModule`.
- Use one runtime `EpisodeCommand` enum for start/save/discard/toggle commands and normalize RPC input once.
- Keep task-label updates and state transitions under one lock.
- Preserve the current task label across repeated takes until a new label is supplied or the session counters are reset.
- Make save/discard while idle explicit no-ops.
- Add an opt-in `unitree-go2-agentic-record` blueprint.
- Record Go2 camera, odometry, velocity commands, episode status, and TF into a timestamped memory2 session DB.
- Preserve the default `unitree-go2-agentic` blueprint unchanged.

`cmd_vel` is produced by the existing `MovementManager` in the transitive Go2 agentic blueprint. A blueprint invariant test verifies that producer and recorder input are both present.

## How to Test

```bash
dimos --simulation --viewer none run unitree-go2-agentic-record
```

In another terminal:

```bash
dimos mcp call start_recording --arg task_label=walk-test
dimos mcp call relative_move --arg forward=0.2
dimos mcp call stop_recording
# Omit task_label to record another take of walk-test
dimos mcp call start_recording
```

Verification completed locally:

- `pytest dimos/learning/collection dimos/learning/dataprep/test_core.py dimos/robot/test_all_blueprints_generation.py -q` — 51 passed
- Ruff format and lint checks passed for all changed Python files
- `go2_recorder.py` focused coverage — 100%
- MuJoCo MCP smoke test completed two start → move → save takes; both persisted the same task label and the DB contained camera, odom, cmd_vel, status, and TF records with no runtime errors

## AI assistance

Claude Code with Opus 4.8 produced the initial design and implementation. OpenAI Codex with GPT-5 reviewed maintainer feedback, revised the implementation, added regression tests, resolved upstream merges, and ran verification. I reviewed and understand the resulting changes.

## Checklist


dimos/agents/system_prompt.py
```diff
@@ -45,6 +45,14 @@
 - `where_am_i` gives your current street/area and nearby landmarks
 - `map_query` finds places on the OSM map by description and returns coordinates
 
+## Episode Recording
```
TomCC7: we inject skill prompt into agent context automatically, so this might not be necessary here.

dimos/learning/collection/episode_monitor.py
```diff
@@ -110,6 +113,35 @@ def reset_counters(self) -> EpisodeStatus:
...
+        physical controller.
+
+        Args:
+            event: one of "start", "save", or "discard".
+            task_label: description of the demonstrated task; applied on "start".
+
+        Returns:
+            A human-readable status string (also safe to surface to the agent).
+        """
+        if event not in ("start", "save", "discard"):
```
TomCC7: better using enum/literal

dimos/agents/skills/episode_recording.py
```diff
@@ -0,0 +1,73 @@
...
+path. The recorded episode is later exported to LeRobot/HDF5 with ``dimos dataprep``.
+"""
+
+from dimos.agents.annotation import skill
+from dimos.core.core import rpc
+from dimos.core.module import Module
+from dimos.learning.collection.episode_monitor_spec import EpisodeControlSpec
+
+
+class EpisodeRecordingSkillContainer(Module):
```
TomCC7: seems fine to just add these functions to episode monitor directly

TomCC7 (changes requested):
Hi, would you fill in the PR description so we have a better understanding of what solution you are providing here? Also it would be good if you can cleanup the modified files a bit (superpower, logs, etc)

MUJOCO_LOG.TXT
```diff
@@ -0,0 +1,18 @@
+Wed Jul 15 12:43:01 2026
```
paul-nechifor: This should not be committed.

paul-nechifor:
Please fill in the PR description.

dimos/agents/skills/episode_recording.py
yusufjarada: Done. The three recording skills now live directly on EpisodeMonitorModule; the separate skill container and RPC spec were removed.

dimos/learning/collection/episode_monitor.py
yusufjarada: Updated the public event parameter to EpisodeAction, a Literal restricted to start, save, and discard. The internal EpisodeCommand still includes toggle for button input.

dimos/agents/system_prompt.py
yusufjarada: Removed the system-prompt addition. The skill docstrings provide the injected tool guidance.

dimos/learning/collection/episode_monitor.py
```diff
@@ -110,6 +113,35 @@ def reset_counters(self) -> EpisodeStatus:
...
+            with self._lock:
+                self._task_label = task_label
+        self._transition(cast("EpisodeCommand", event), time.time())
+        with self._lock:
+            saved, discarded = self._saved, self._discarded
+        if event == "start":
+            return f"Recording started (task: {task_label})." if task_label else "Recording started."
+        if event == "save":
+            return f"Episode saved. Total saved this session: {saved}."
+        return f"Episode discarded. Total discarded this session: {discarded}."
```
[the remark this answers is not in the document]
yusufjarada: Fixed. Save and discard while idle now return a clear no-op response and emit no EpisodeStatus event. Both paths have regression coverage.

MUJOCO_LOG.TXT
yusufjarada: Removed MUJOCO_LOG.TXT and the other unrelated local artifacts from the branch.

dimos/learning/collection/episode_monitor.py
```diff
@@ -86,6 +86,9 @@ def __init__(self, **kwargs: Any) -> None:
         self._saved: int = 0
         self._discarded: int = 0
```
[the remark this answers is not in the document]
yusufjarada: Fixed. reset_counters now restores the configured default task label. A regression test starts a labeled take, resets, and verifies that the next unlabeled take does not inherit the stale label.

dimos/learning/collection/go2_recorder.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class Go2CollectionRecorder(Recorder):
+    """Records Go2 camera + pose + action + episode-boundary streams for training."""
+
+    config: Go2CollectionRecorderConfig
+
+    color_image: In[Image]  # observation (camera)
+    odom: In[PoseStamped]  # observation (pose)
+    cmd_vel: In[Twist]  # action channel
```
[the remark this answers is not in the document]
yusufjarada: Verified that the transitive Go2 blueprint includes MovementManager.cmd_vel as Out[Twist]; GO2Connection is the consumer, not the only endpoint. I added a blueprint invariant test covering the MovementManager producer and recorder input. The MuJoCo smoke test recorded 5 cmd_vel rows during a 0.2 m move.

dimos/learning/collection/episode_monitor.py
```diff
@@ -110,6 +113,35 @@ def reset_counters(self) -> EpisodeStatus:
...
+
+        Returns:
+            A human-readable status string (also safe to surface to the agent).
+        """
+        if event not in ("start", "save", "discard"):
+            return f"error: unknown episode event '{event}'; use start, save, or discard."
+        if event == "start" and task_label is not None:
+            with self._lock:
+                self._task_label = task_label
+        self._transition(cast("EpisodeCommand", event), time.time())
```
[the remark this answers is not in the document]
yusufjarada: Fixed. set_episode now passes the requested label into _transition, and the label assignment plus state mutation happen under the same lock. A concurrent-start regression test verifies that alpha and beta each retain their own emitted label.

yusufjarada:
Addressed the requested changes and pushed the revision:

- filled in the PR description and AI disclosure
- removed logs, planning notes, and unrelated local files
- moved recording tools directly onto `EpisodeMonitorModule`
- fixed the label race, idle save/discard responses, and reset label leakage
- added typed episode actions and regression coverage
- verified the existing `MovementManager.cmd_vel` producer with a blueprint test and a successful MuJoCo recording
- merged current `main` and resolved the generated blueprint conflict

Local verification: 47 collection/dataprep tests passed, generated blueprint check passed, Ruff passed, and the simulated MCP start → move → save flow produced all expected DB streams without runtime errors.

The remaining Backport check requires a maintainer to add `backport:skip` (the fork account cannot add labels). Ready for another review when convenient.

dimos/learning/collection/episode_monitor.py
```diff
@@ -38,8 +39,9 @@
 
 logger = setup_logger()
 
-# A button/keyboard press requests one of these; `toggle` resolves to
-# `start`/`save` based on the current state, so it never reaches the output.
+EpisodeAction: TypeAlias = Literal["start", "save", "discard"]
+# A button/keyboard press can also request `toggle`, which resolves to
+# `start`/`save` based on the current state and never reaches the output.
 EpisodeCommand: TypeAlias = Literal["start", "save", "discard", "toggle"]
```
TomCC7: These two definitions are nearly the same. Can you elaborate what's the difference between an action and a command? And is it possible to unify it?

dimos/learning/collection/episode_monitor.py
TomCC7: I don't think you are using it here. You defined the enum but this `if` statement still use the bare tuple
```suggestion
        if event not in EpisodeAction:
```

dimos/learning/collection/episode_monitor.py
```diff
@@ -107,9 +112,60 @@ def reset_counters(self) -> EpisodeStatus:
...
+        status = self._transition(event, time.time(), task_label)
+        if status is None:
+            return f"No episode is currently recording; nothing to {event}."
+        if event == "start":
+            return (
+                f"Recording started (task: {task_label})." if task_label else "Recording started."
+            )
+        if event == "save":
+            return f"Episode saved. Total saved this session: {status.episodes_saved}."
+        return f"Episode discarded. Total discarded this session: {status.episodes_discarded}."
```
TomCC7: better using a match statement or just a simple dict here as these three cases are same level, and using any as fallback looks weird.

dimos/learning/collection/episode_monitor.py
```diff
@@ -153,18 +214,28 @@ def _transition(self, event: EpisodeCommand, ts: float) -> None:
...
+                self._discarded += 1
                 self._state = "idle"
             # Snapshot under the mutation's lock so the event matches the state.
             status = self._snapshot(event, ts)
-        self._emit(status)
+            # A terminated take's label must not leak into the next one. Reset here
+            # (inside _transition) so it happens regardless of entry point — button,
+            # keyboard, or the set_episode RPC.
+            if event in ("save", "discard"):
+                self._task_label = self.config.default_task_label
```
TomCC7: I'm not sure what's the settings here. Generally during data collection we do the same task multiple times instead of doing different tasks right? Let me know what you think.

TomCC7 (changes requested):

dimos/learning/collection/episode_monitor.py
yusufjarada: Unified them into one runtime EpisodeCommand StrEnum with start, save, discard, and toggle values. The separate EpisodeAction alias is removed.

dimos/learning/collection/episode_monitor.py
yusufjarada: The previous EpisodeAction was a typing.Literal alias, so membership against it would not work at runtime. I replaced both aliases with a real EpisodeCommand StrEnum and normalize/validate the incoming value once with EpisodeCommand(event).

dimos/learning/collection/episode_monitor.py
yusufjarada: Changed the response dispatch to an explicit match over the resolved status event, with separate start, save, and discard cases.

dimos/learning/collection/episode_monitor.py
yusufjarada: Updated this to session-level task semantics. Save and discard now preserve the active task label, and start_recording accepts an optional label so repeated takes reuse it. Supplying a new label changes tasks; reset_counters restores the configured default. The MuJoCo smoke test persisted two takes with the same label.

yusufjarada:
Pushed the follow-up review fixes in `e0005448d`:

- unified action/command types into one runtime `EpisodeCommand` enum
- normalized RPC inputs against that enum
- changed response handling to `match`
- preserved task labels across repeated takes and made the skill label optional
- added direct coverage for the odometry pose setter, bringing `go2_recorder.py` to 100% focused coverage
- synced the branch with current `main`

Local verification: 51 tests passed, Ruff passed, blueprint generation passed, and a two-take MuJoCo MCP run persisted the same label across both episodes with all expected streams. Ready for re-review.

dimos/learning/collection/episode_monitor.py
```diff
@@ -139,32 +214,41 @@ def _on_keyboard(self, msg: KeyPress) -> None:
...
+            elif event is EpisodeCommand.DISCARD:
+                if self._state != "recording":
+                    return None
+                self._discarded += 1
                 self._state = "idle"
             # Snapshot under the mutation's lock so the event matches the state.
-            status = self._snapshot(event, ts)
-        self._emit(status)
+            status = self._snapshot(event.value, ts)
+        return self._emit(status)
```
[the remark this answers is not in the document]
yusufjarada: Fixed in 550af8880. Episode state mutation/snapshotting still uses the state lock, while a separate transition lock now serializes the complete snapshot-and-publish sequence. This preserves publication order without holding the state lock during subscriber I/O. Added a deterministic concurrent-transition regression test that reproduced the alpha/beta inversion before the fix. Focused suite: 52 passed; episode monitor suite: 19 passed; Ruff and diff checks pass.

PR #3120, opened 2026-07-21
feat(hosted-teleop): auto-select operator cockpit from robot kind

ruthwikdasyam:
## Contribution path

- [ ] Linked issue or discussion: DIM-XXX / #XXX / URL

## Problem

The hosted operator UI required a manual Go2/Arm toggle before connecting - nothing on the robot side told the broker which cockpit to open, so an operator could connect an arm and land on the Go2 view.

## Solution

Each hosted command module declares its cockpit kind and pushes it to the broker at init

- robot_type.py - a `StrEnum` of cockpit kinds (`go2`, `arm`); the value is the wire string. Kept in the app layer, not the transport (`BrokerConfig` stays a plain `str`), so the generic provider doesn't depend on this enum.

- CommandModule declare `ROBOT_TYPE` and call `BrokerProvider.set_robot_type()` at init.

- the session POST sends the declared kind; no silent default (raises if undeclared), and re-declaring a *conflicting* kind raises (two command modules in one process would open the wrong cockpit).

## How to Test

    dimos run teleop-hosted-go2-transport

Connect from the operator dashboard — the arm cockpit opens without the manual toggle. (`teleop-hosted-xarm7` opens the arm cockpit the same way.)

## AI assistance

Claude Code (Fable 5 / Opus 4.8) - used for research, refactoring, review

## Checklist


dimos/teleop/hosted/test_arm_command.py
```diff
@@ -276,3 +278,12 @@ def test_robot_state_reports_estop_and_engage(module: ArmCommandModule) -> None:
     payload = json.loads(module.robot_state.publish.call_args.args[0])
     assert payload["estopped"] is True
     assert payload["engaged"] == {"left": False, "right": False}
+
+
+def test_init_declares_robot_type_to_broker(module: ArmCommandModule) -> None:
+    """__init__ pushes ROBOT_TYPE to the shared broker provider, which sends it
+    in the session-create POST so the operator dashboard opens the arm cockpit."""
+    from dimos.protocol.pubsub.impl.webrtc.providers.broker import BrokerProvider
```
paul-nechifor: already imported.

dimos/protocol/pubsub/impl/webrtc/providers/broker.py
```diff
@@ -95,6 +95,35 @@ class BrokerProvider(AsyncProviderBase):
...
 
+    # Robot kind for the create-session POST, declared by the command module at
+    # init (one session + one kind per process). No default: _require_robot_type
+    # raises rather than guess.
+    _robot_type: ClassVar[str | None] = None
+
+    @classmethod
+    def set_robot_type(cls, robot_type: str) -> None:
+        """Declare the robot kind sent to the broker at session create, so the
+        operator UI auto-selects the cockpit. Called by the hosted command module
```
paul-nechifor: Wait, you're calling it the cockpit too? I picked that name we don't confuse the two systems 😅  (hence why I renamed the broker to "relay")

dimos/teleop/hosted/arm_command.py
```diff
@@ -64,6 +69,7 @@ class ArmCommandModule(ArmTeleopModule):
 
     def __init__(self, **kwargs: Any) -> None:
         super().__init__(**kwargs)
+        BrokerProvider.set_robot_type(self.ROBOT_TYPE)
```
paul-nechifor: Hm, I don't like how initializing this class modifies a global in `BrokerProvider`.

dimos/protocol/pubsub/impl/webrtc/providers/broker.py
```diff
@@ -95,6 +95,35 @@ class BrokerProvider(AsyncProviderBase):
     INBOUND_CHANNELS = ("cmd_unreliable", "state_reliable")
     OUTBOUND_CHANNELS = ("state_reliable_back", "map_unreliable")
 
+    # Robot kind for the create-session POST, declared by the command module at
+    # init (one session + one kind per process). No default: _require_robot_type
+    # raises rather than guess.
+    _robot_type: ClassVar[str | None] = None
```
paul-nechifor: I don't like using class variables since they're essentially namespaced globals. They might be justifiable in normal Python, but with the way we use processes for Python modules it will probably lead to issues.

paul-nechifor (approved):

PR #3123, opened 2026-07-21
fix(perception): use pos_x/pos_y/pos_z metadata keys in spatial location queries

hvent90:
## Contribution path

## Problem

`SpatialVectorDB.query_by_location()` and `get_all_locations()` filter stored frames by metadata keys `"x"`/`"y"`/`"z"`, but frames are stored with keys `pos_x`/`pos_y`/`pos_z` (see the metadata dict built in `SpatialMemory._process_frame`). The keys never match, so both methods always return empty results against real spatial-memory data.

## Solution

Read `pos_x`/`pos_y`/`pos_z` in both methods.

Added `dimos/perception/test_spatial_vector_db.py`: stores frames in an in-memory `SpatialVectorDB` using the exact metadata shape `_process_frame` writes, and asserts location queries find them. The test fails against the old
code and runs in the default (fast) CI suite.

## How to Test

`uv run pytest dimos/perception/test_spatial_vector_db.py`

Also verified against real recorded data:

First I populated the chromadb: `dimos --replay --replay-db go2_china_office run unitree-go2-spatial --disable perceive-loop-skill` (`--disable` because the blueprint's PerceiveLoopSkill requires an agent module it doesn't include — separate issue). Let it log `Stored frame at position (…)` for ~60s, then stop it.

Then, I ran this script:
```sh
uv run python - <<'EOF'
import chromadb
from dimos.perception.spatial_vector_db import SpatialVectorDB

client = chromadb.PersistentClient(path="assets/output/memory/spatial_memory/chromadb_data")
db = SpatialVectorDB(collection_name="spatial_memory", chroma_client=client)
frames = db.image_collection.get(include=["metadatas"])
print(len(frames["ids"]), "frames stored")          # non-zero — the data exists
print("sample metadata:", frames["metadatas"][0])   # keys are pos_x/pos_y/pos_z
print("get_all_locations:", db.get_all_locations()) # [] — the bug
print("query (infinite radius):", db.query_by_location(0, 0, radius=1e9))  # still []
EOF
```

Running it on `main` produced empty query results. Running it on this feature branch produced results.

## AI assistance

My associate (Fable 5) discovered the issue, proposed a fix, and implemented it. I then reviewed the output and manually ran the verification against the chromadb as outlined in the `How to Test` section.

### DIFF
--- a/dimos/perception/spatial_vector_db.py
+++ b/dimos/perception/spatial_vector_db.py
@@ -177,8 +177,8 @@ class SpatialVectorDB:
         filtered_results = {"ids": [], "metadatas": [], "distances": []}  # type: ignore[var-annotated]
 
         for i, metadata in enumerate(results["metadatas"]):  # type: ignore[arg-type]
-            item_x = metadata.get("x")
-            item_y = metadata.get("y")
+            item_x = metadata.get("pos_x")
+            item_y = metadata.get("pos_y")
 
             if item_x is not None and item_y is not None:
                 distance = np.sqrt((x - item_x) ** 2 + (y - item_y) ** 2)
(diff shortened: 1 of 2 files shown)


paul-nechifor:
Hi @hvent90 . Thank you for the contribution. However, it looks like `query_by_location` is dead code. Could you instead please delete that method and the `query_by_location` method which uses it.

paul-nechifor:
Closed in favor of https://github.com/dimensionalOS/dimos/pull/3460

PR #3128, opened 2026-07-21
feat: ensure that tests have 100% coverage (i.e. no dead code in tests)

paul-nechifor:
Closes DIM-1318

## Problem

We need to make sure that all tests are executed.If they're not, they're dead code and should be deleted.

## Solution

* Check that we have 100% coverage on test files.
* Exclude tests which can't run (marked with skipif_in_ci and mujoco)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -129,11 +129,25 @@ jobs:
           maturin develop -m dimos/mapping/ray_tracing/rust/Cargo.toml
           maturin develop -m dimos/navigation/nav_3d/mls_planner/rust/Cargo.toml
           python -c "import dimos_voxel_ray_tracing, dimos_mls_planner"
-          python -m pytest -c /dev/null --rootdir . --noconftest --import-mode=importlib -p no:cacheprovider -v \
+          # Under coverage so the test-file-coverage gate can prove these
+          # files run: they importorskip themselves out of every other job.
+          python -m coverage run -m pytest -c /dev/null --rootdir . --noconftest --import-mode=importlib -p no:cacheprovider -v \
             dimos/mapping/ray_tracing/test_voxel_map.py \
             dimos/mapping/ray_tracing/test_transformer.py \
             dimos/navigation/nav_3d/mls_planner/test_mls_planner.py \
             dimos/navigation/nav_3d/mls_planner/test_transformer.py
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -710,6 +745,37 @@ jobs:
         run: |
           df -h
 
+  test-file-coverage:
+    # Detects tests that silently never run: every test file must be fully
+    # line-covered by the union of all CI pytest jobs' coverage data.
+    # No `if:` needed: plain `needs` skips this job whenever an input job
+    # is skipped (fork PRs skip self-hosted-tests) or failed/cancelled.
+    needs: [tests, self-hosted-tests, rust]
```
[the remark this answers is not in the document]
paul-nechifor: Fixed.

.github/workflows/ci.yml
```diff
@@ -129,11 +129,25 @@ jobs:
           maturin develop -m dimos/mapping/ray_tracing/rust/Cargo.toml
           maturin develop -m dimos/navigation/nav_3d/mls_planner/rust/Cargo.toml
           python -c "import dimos_voxel_ray_tracing, dimos_mls_planner"
-          python -m pytest -c /dev/null --rootdir . --noconftest --import-mode=importlib -p no:cacheprovider -v \
+          # Under coverage so the test-file-coverage gate can prove these
+          # files run: they importorskip themselves out of every other job.
+          python -m coverage run -m pytest -c /dev/null --rootdir . --noconftest --import-mode=importlib -p no:cacheprovider -v \
```
Dreamsorcerer, line 134: Why does this skip the conftest in the first place...?

.github/workflows/ci.yml
Dreamsorcerer, line 1: This feels rather messy and convoluted. I'd much rather just have a path split in codecov and just add that as a required status check. No additional code needed. No extra CI steps needed.

Dreamsorcerer: https://docs.codecov.com/docs/codecovyml-reference#component_management

paul-nechifor: Ah, okay. Do you want to do that?

Dreamsorcerer: I'll add a task to look at it (looks like we can use regex or globbing there), though I still think it'd be much cleaner to have a separate tests/ directory like other Python projects.

PR #3129, opened 2026-07-21
feat(manipulation): add OpenYAM Damiao adapter

TomCC7:
## Contribution path

- Linked design discussion: https://github.com/dimensionalOS/dimos/pull/3129
closes DIM-1311

## Problem

DimOS lacked a direct OpenYAM integration for its six Damiao arm motors and gripper. Modeling them as separate manipulator adapters did not match the shared upstream robot lifecycle, shared CAN topology, or normalized gripper interface. The driver also needed gravity compensation without hiding the robot topology in runtime configuration.

## Solution

- Add a generic `DamiaoWholeBodyAdapter` that uses the upstream `Robot`, `Arm`, and `Gripper` interfaces directly. It supports named CAN buses, multiple arm/gripper groups, normalized gripper commands and feedback, one synchronized command tick, and URDF-based gravity compensation.
- Add `OpenYamDamiaoAdapter` as the concrete OpenYAM topology: six arm joints and one gripper on the `openyam` logical bus. The gravity URDF remains lazy until hardware connection.
- Register real and mock whole-body adapters and wire the gripper into the standard OpenYAM coordinator, planner, and keyboard-teleoperation blueprints.
- Expose CAN setup through `dimos can setup` and move the Piper command out of the robot package.
- Add focused behavior tests for dual-arm command routing, gripper normalization, lifecycle rollback, gravity torque composition, adapter registration, CAN CLI execution, and import-time LFS isolation.

The generic `HardwareComponent.address` remains a single-address compatibility field. Damiao whole bodies configure one or more physical CAN interfaces through `DamiaoRuntimeConfig.bus_addresses` instead.

## How to Test

Configure the physical CAN interface, then start the default OpenYAM teleoperation stack:

```bash
dimos can setup can0
dimos run keyboard-teleop-openyam
```

Automated validation:

```bash
uv run pytest dimos/hardware/whole_body dimos/hardware/test_adapter_registries.py dimos/cli/test_can.py
```

The implementation was validated with mock/in-memory CAN and unit tests. Physical OpenYAM motion was not tested by the AI assistant.

## AI assistance

OpenAI Codex with GPT-5 assisted throughout design review, implementation, test development, debugging, and documentation. The author reviewed and approved the design decisions interactively.

## Checklist


dimos/hardware/damiao/__init__.py
TomCC7, line 1: no init file

dimos/hardware/damiao/specs.py
```diff
@@ -0,0 +1,316 @@
...
+@dataclass(frozen=True)
+class DamiaoBusSpec:
+    """Named communication channel for Damiao motors."""
+
+    address: str | Path = "can0"
+    fd: bool = False
+
+
+@dataclass(frozen=True)
+class DamiaoJointGroupSpec:
```
TomCC7: can we use attr instead of dataclass here

dimos/hardware/damiao/runtime.py
```diff
@@ -0,0 +1,399 @@
...
+
+def _load_can_motor_control(
+    *,
+    adapter_type: str,
+    error_type: type[RuntimeError] = DamiaoBindingUnavailableError,
+) -> tuple[Any, Any]:
+    """Lazily load the optional Rust-backed binding and Damiao codec module."""
+
+    try:
+        can_motor_control = importlib.import_module("can_motor_control")
```
TomCC7: use top level import, check how roboplan work

dimos/hardware/damiao/runtime.py
```diff
@@ -0,0 +1,399 @@
...
+        raise error_type(
+            f"The selected '{adapter_type}' adapter requires the Rust-backed "
+            "can-motor-control Python binding in the active environment. On "
+            "supported Linux systems, install dimos[manipulation] before "
+            f"selecting adapter_type='{adapter_type}'."
+        ) from exc
+    return can_motor_control, damiao
+
+
+def _dynamic_attr(value: object, name: str) -> Any:
```
TomCC7: no getattr

dimos/hardware/damiao/specs.py
TomCC7: and specs are bad naming here, should be config etc. spec means protocol in dimos

dimos/hardware/damiao/specs.py
```diff
@@ -0,0 +1,316 @@
...
+    supports_velocity: bool = False
+
+    @property
+    def dof(self) -> int:
+        """Return the number of joints described by this group spec."""
+
+        return len(self.motors)
+
+    @property
+    def joint_names(self) -> tuple[str, ...]:
```
TomCC7: question: how is joint name used in dm_motor_control? if not used why we need them?

dimos/hardware/damiao/specs.py
```diff
@@ -0,0 +1,316 @@
...
+        """Return joint names in command-vector order."""
+
+        return tuple(motor.name for motor in self.motors)
+
+    def validate(self, *, group_name: str, bus_names: set[str] | None = None) -> None:
```
TomCC7: after using attr try to use their default value validator as much as possible

dimos/hardware/damiao/specs.py
```diff
@@ -0,0 +1,316 @@
...
+            bus_ids = ids_by_bus[group.bus_name]
+            for motor in group.motors:
+                if motor.send_id in bus_ids:
+                    raise ValueError(f"duplicate send_id {motor.send_id} on bus {group.bus_name!r}")
+                bus_ids.add(motor.send_id)
+        if len(set(all_joint_names)) != len(all_joint_names):
+            raise ValueError(f"duplicate joint names across DamiaoRobotSpec: {all_joint_names}")
+
+    @classmethod
+    def from_arm_spec(
```
TomCC7: need to simplify all the concept here too many duplication. we basically need two usecase of configs here: config that helps derive a new kind of arm and configs that can be tweaked during runtime stage. better separate these

dimos/hardware/damiao/arm_adapter.py
```diff
@@ -0,0 +1,557 @@
...
+from dimos.hardware.damiao.specs import DamiaoArmSpec, DamiaoRobotSpec
+from dimos.hardware.manipulators.spec import ControlMode, JointLimits, ManipulatorInfo
+from dimos.utils.logging_config import setup_logger
+
+logger = setup_logger()
+
+_CONTROL_MODE_INDEX = {mode: index for index, mode in enumerate(ControlMode)}
+
+
+def _dynamic_attr(value: object, name: str) -> Any:
```
TomCC7: again no attr ever

dimos/cli/can.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+@app.command("setup")
+def setup(
+    interface: str = typer.Argument(..., help="Linux CAN interface name"),
+    bitrate: int = typer.Option(1_000_000, help="Nominal CAN bitrate in bits per second"),
+    txqueuelen: int = typer.Option(1_000, help="Kernel transmit queue length"),
+) -> None:
+    """Configure, bring up, and verify a classic CAN interface."""
+    _positive(bitrate, "bitrate")
```
TomCC7: this should be able to be verified through viser directly

dimos/cli/can.py
```diff
@@ -0,0 +1,90 @@
...
+import os
+import shlex
+import subprocess
+
+import typer
+
+app = typer.Typer(help="Configure and inspect Linux CAN interfaces", no_args_is_help=True)
+
+
+def _run_ip(*args: str, privileged: bool = False) -> subprocess.CompletedProcess[str]:
```
TomCC7, line 28: can you check how piper setup is invoking sudo and maybe unify

TomCC7: and one correction for piper - maybe put piper setup cli stuff also into dimos.cli instead of robot side?

dimos/manipulation/visualization/viser/config.py
```diff
@@ -26,7 +26,7 @@ class ViserVisualizationConfig(BaseModel):
 
     backend: Literal["viser"] = "viser"
     host: str = Field(
-        default="127.0.0.1", validation_alias=AliasChoices("host", "visualization_host")
+        default="0.0.0.0", validation_alias=AliasChoices("host", "visualization_host")
```
TomCC7: defaulting to 0.0.0.0 sounds like a security vulnerability, maybe let's just make sure it's cli overridable

docs/capabilities/manipulation/index.md
```diff
@@ -7,6 +7,11 @@ the default world and native path planner.
 
 ## Quick Start
 
+For the required vendor/approved-bench direction verification before OpenYAM
+planning or teleoperation, see the
+[OpenYAM direction commissioning guide](./openyam_commissioning.md). The
```
TomCC7: remove these

dimos/cli/piper.py
TomCC7, line 1: is that possible to just unify this and can?

dimos/hardware/damiao/config.py
```diff
@@ -0,0 +1,225 @@
...
+            member_validator=attrs.validators.instance_of(DamiaoMotorConfig),
+            iterable_validator=attrs.validators.min_len(1),
+        ),
+    )
+    position_lower: tuple[float, ...] = attrs.field(converter=_to_floats)
+    position_upper: tuple[float, ...] = attrs.field(converter=_to_floats)
+    velocity_max: tuple[float, ...] = attrs.field(converter=_to_floats)
+    kp: tuple[float, ...] = attrs.field(converter=_to_floats)
+    kd: tuple[float, ...] = attrs.field(converter=_to_floats)
+    gravity_torque_limits: tuple[float, ...] | None = attrs.field(
```
TomCC7: remove this config shouldn't need this...

dimos/control/coordinator.py
```diff
@@ -915,6 +915,37 @@ def set_gripper_position(self, hardware_id: str, position: float) -> bool:
                 return False
             return hw.adapter.write_gripper_position(position)
 
+    @rpc
+    def open_gripper(self, hardware_id: str) -> bool:
```
TomCC7: why is this required? previous path does not work?

dimos/control/hardware_interface.py
```diff
@@ -125,14 +125,13 @@ def read_state(self) -> dict[JointName, JointState]:
...
             gripper_pos = self._adapter.read_gripper_position()
-            for gj in self._gripper_joints:
-                result[gj] = JointState(
-                    position=self._physical_to_normalized(gripper_pos)
-                    if gripper_pos is not None
-                    else 0.0,
-                    velocity=0.0,
-                    effort=0.0,
-                )
+            if gripper_pos is not None:
```
TomCC7: what kind of situation will have this?

dimos/hardware/damiao/arm_adapter.py
```diff
@@ -0,0 +1,482 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+from typing import Any
+
+import numpy as np
+import pinocchio  # type: ignore[import-not-found]
```
TomCC7: should not have this ignore

dimos/hardware/damiao/arm_adapter.py
```diff
@@ -0,0 +1,482 @@
...
+    def __init__(
+        self,
+        *,
+        arm_config: DamiaoArmConfig,
+        runtime_config: DamiaoRuntimeConfig | None = None,
+        dof: int | None = None,
+        hardware_id: str = "arm",
+    ) -> None:
+        runtime_config = runtime_config or DamiaoRuntimeConfig()
+        if dof is not None and dof != arm_config.dof:
```
TomCC7: why is this conflicting config necessary?

dimos/hardware/damiao/config.py
```diff
@@ -0,0 +1,272 @@
...
+    attrs.validators.instance_of(int),
+    attrs.validators.ge(1),
+)
+
+
+def _to_floats(values: Sequence[float]) -> tuple[float, ...]:
+    return tuple(float(value) for value in values)
+
+
+def _to_optional_floats(values: Sequence[float] | None) -> tuple[float, ...] | None:
```
TomCC7: don't like these helper functions... does attr have other options?

dimos/hardware/damiao/config.py
```diff
@@ -0,0 +1,272 @@
...
+    )
+    gripper: DamiaoGripperConfig | None = attrs.field(
+        default=None,
+        validator=attrs.validators.optional(attrs.validators.instance_of(DamiaoGripperConfig)),
+    )
+    fd: bool = attrs.field(default=False, validator=attrs.validators.instance_of(bool))
+    supported_control_modes: tuple[ControlMode, ...] = attrs.field(
+        factory=lambda: (
+            ControlMode.POSITION,
+            ControlMode.SERVO_POSITION,
```
TomCC7: how is servo_position used

dimos/hardware/damiao/runtime.py
```diff
@@ -0,0 +1,390 @@
...
+
+def _retry_enobufs(operation: Callable[[], None]) -> None:
+    for delay_s in (*_ENOBUFS_RETRY_DELAYS_S, None):
+        try:
+            operation()
+            return
+        except Exception as exc:
+            if delay_s is None or not _is_enobufs(exc):
+                raise
+            time.sleep(delay_s)
```
TomCC7: the check should not be here. remove that since we correctly setup the interface

dimos/hardware/damiao/runtime.py
```diff
@@ -0,0 +1,390 @@
...
+        arm_config: DamiaoArmConfig,
+        runtime_config: DamiaoRuntimeConfig,
+        adapter_type: str = "damiao",
+    ) -> None:
+        self._arm_config = arm_config
+        self._runtime_config = runtime_config
+        self._adapter_type = adapter_type
+        self._robot: Any | None = None
+        self._arm: Any | None = None
+        self._gripper: Any | None = None
```
TomCC7: Again, these should not be any in any case. also no need to be None in most cases, if it's assigned later just declare type and then fill value later as most of the time it will be fine.

dimos/hardware/damiao/arm_adapter.py
```diff
@@ -0,0 +1,482 @@
...
+        resolved_torque_limits = arm_config.gravity_torque_limits
+        self._gravity_torque_limits = (
+            list(resolved_torque_limits) if resolved_torque_limits else None
+        )
+        if self._gravity_torque_limits is not None:
+            self._validate_length("gravity_torque_limits", self._gravity_torque_limits)
+        self._supported_control_modes = arm_config.supported_control_modes
+        self._control_mode = ControlMode.POSITION
+        self._last_positions: list[float] | None = None
+        self._pin_model: Any | None = None
```
TomCC7: again no any

dimos/hardware/manipulators/openyam_damiao/adapter.py
TomCC7, line 1: seems very weird to me that this file is still so lengthy, should only contain config which is derived from arm_adapter

dimos/robot/manipulators/openyam/config.py
TomCC7, line 1: need to clean up logic here, discuss with me

dimos/hardware/whole_body/mock/adapter.py
```diff
@@ -0,0 +1,76 @@
...
+        address: str | Path | None = None,
+        *,
+        dof: int,
+        initial_positions: list[float] | None = None,
+        hardware_id: str = "whole_body",
+        domain_id: int = 0,
+    ) -> None:
+        del address
+        del hardware_id
+        del domain_id
```
TomCC7: if these are not required why define from the first place

dimos/hardware/whole_body/openyam_damiao/adapter.py
```diff
@@ -0,0 +1,71 @@
...
+            can_motor_control.Robot.builder()
+            .add_bus(
+                "can",
+                can_motor_control.SocketCanBus(self.bus_address("can")),
+                damiao.DamiaoCodec(),
+            )
+            .add_arm("arm", bus="can", motors=arm_motors)
+            .add_gripper(
+                "gripper",
+                bus="can",
```
TomCC7: should hardcode this can name instead of multi reference, also might be good to have a more meaingful name rather than bare can

dimos/hardware/whole_body/damiao/adapter.py
```diff
@@ -0,0 +1,371 @@
...
+
+    arm_joints: dict[str, tuple[str, ...]] = {}
+    gripper_joints: dict[str, str] = {}
+    bus_defaults: dict[str, str] = {}
+    gravity_model_path: Path | None = None
+    gravity_joint_names: tuple[str, ...] = ()
+
+    def __init__(
+        self,
+        address: str | Path | None = None,
```
TomCC7, line 47: how does this address handle dual arm case? is this argument derived from wholebodyadapter?

dimos/cli/can.py
```diff
@@ -0,0 +1,99 @@
...
+    _run_ip("link", "show", "dev", interface)
+    _run_ip("link", "set", "dev", interface, "down", privileged=True)
+    _run_ip(
+        "link", "set", "dev", interface, "type", "can", "bitrate", str(bitrate), privileged=True
+    )
+    _run_ip("link", "set", "dev", interface, "txqueuelen", str(txqueuelen), privileged=True)
+    _run_ip("link", "set", "dev", interface, "up", privileged=True)
+    result = _run_ip("-details", "-statistics", "link", "show", "dev", interface)
+    typer.echo(result.stdout.rstrip())
+    typer.echo(f"Configured {interface}: bitrate={bitrate}, txqueuelen={txqueuelen}")
```
mustafab0, line 100: Can we add a probe command, to check if the can port after setup is actually able to communicate with motors.

Something like a ping.

dimos/cli/can.py
```diff
@@ -0,0 +1,99 @@
...
+    command = ["ip", *args]
+    if privileged and os.geteuid() != 0:
+        command = ["sudo", "--", *command]
+    if privileged:
+        typer.echo(f"Running: {shlex.join(command)}")
+    try:
+        return subprocess.run(
+            command,
+            check=True,
+            capture_output=not privileged,
```
jeff-hykin, line 39: If sudo, print directly to stderr 
If not sudo, go to string, then python prints stderr

^I don't really understand why do that. Not blocking, just a little weird.

dimos/cli/dimos.py
```diff
@@ -173,9 +173,9 @@ def callback(**kwargs) -> None:  # type: ignore[no-untyped-def]
 
 
 main.callback()(create_dynamic_callback())  # type: ignore[no-untyped-call]
-main.add_typer(go2tool_app, name="go2tool")
+hardware_app.add_typer(can_app, name="can")
 main.add_typer(hardware_app, name="hardware")
-main.add_typer(piper_app, name="piper")
+main.add_typer(go2tool_app, name="go2tool")
```
jeff-hykin, line 179: Ivan probably wouldnt like go2tool being under hardware so I'd probably leave that for another PR

dimos/hardware/whole_body/damiao/config.py
```diff
@@ -0,0 +1,47 @@
...
+import attrs
+
+
+@attrs.frozen
+class DamiaoRuntimeConfig:
+    """Deployment values that may vary without changing robot topology."""
+
+    bus_addresses: dict[str, str] = attrs.field(
+        factory=dict,
+        validator=attrs.validators.deep_mapping(
```
jeff-hykin: Is this validators thing typical? Seems kinda verbose. Are pydantic dataclasses and type hints not enough? Or is this arguably to be better cause it's pure python.

dimos/control/tick_loop.py
```diff
@@ -409,7 +409,11 @@ def _write_all_hardware(
             for hw_id, (positions, mode) in hw_commands.items():
                 if hw_id in self._hardware:
                     try:
-                        self._hardware[hw_id].write_command(positions, mode)
+                        accepted = self._hardware[hw_id].write_command(positions, mode)
```
jeff-hykin: If this is a "hot loop", like 500Hz, I would recommend doing this outside of the loop:

```
hardware = self._hardware
```

Python attr access is surprisingly slow:
https://softwarebits.substack.com/p/surprisingly-expensive-python

I don't usually recommend this kind of micro-optimization unless its like a 500hz loop.

jeff-hykin (dismissed):
Image
Theres what I actually looked at ^ 

Image

dimos/cli/can.py
TomCC7: Privileged commands intentionally inherit stdio so sudo can display its prompt and command errors directly in the caller's terminal. Read-only commands capture output because status and the post-setup verification render it through Typer. Added a comment at the subprocess call site to make that distinction explicit.

dimos/cli/dimos.py
TomCC7: Confirmed. go2tool remains a top-level command. Only CAN is registered under hardware, so the command path is dimos hardware can …. The CAN tests now exercise that path through their shared invocation helper without a separate top-level-absence behavior test.

dimos/hardware/whole_body/damiao/config.py
TomCC7: Kept attrs here. DamiaoRuntimeConfig is constructed internally and passed as an opaque value in HardwareComponent.adapter_kwargs; its deployment boundary is multiprocessing pickle, not CLI/config-file JSON or a module schema. This matches the existing A1Z adapter config precedent. The current inline attrs validators retain strict runtime validation.

TomCC7: actually, I found that pydantic supports non-serializable objects as well so I'm creating a new pr to remove all my attr usage...

dimos/control/tick_loop.py
TomCC7: Good suggestion—we decided to take it. Commit cd26ce70c caches self._hardware before the per-command loop and uses the local for both membership and lookup, without changing locking or behavior.

dimos/hardware/whole_body/damiao/config.py
TomCC7: Updated in be40f8b5b to align with #3459. DamiaoRuntimeConfig is now a frozen Pydantic dataclass using ConfigDict(extra=forbid, validate_default=True) and Field/Annotated constraints. Its fields, defaults, strict validation, and pickle behavior are preserved; this PR no longer introduces attrs usage.

docs/capabilities/manipulation/index.md
```diff
@@ -21,10 +21,18 @@ Each blueprint launches the full stack — keyboard UI, mock controller, IK solv
...
 dimos run keyboard-teleop-xarm7   # XArm7 7-DOF
 ```
 
+OpenYAM is exposed as one whole-body device with six angular arm joints and a
+normalized gripper joint. `arm/gripper` uses `0.0` for fully closed and `1.0`
+for fully open; it does not use meters. Hardware activation calibrates both
+mechanical endpoints, so clear the gripper jaws and workspace before startup.
+The gripper has no default startup target and moves only after joint control has
+an explicit target.
+
```
TomCC7:
```suggestion
```

dimos/cli/test_can.py
```diff
@@ -0,0 +1,193 @@
...
+
+    assert result.exit_code == 2
+    assert "x>=1" in result.output
+
+
+def test_setup_nonpositive_bitrate_returns_usage_error() -> None:
+    result = _invoke_can(["setup", "can0", "--bitrate", "0"])
+
+    assert result.exit_code == 2
+    assert "x>=1" in result.output
```
mustafab0: can bitrate minimum should be 10_000 min 
`x>=10000`

1 can pass but is as good as broken

dimos/control/test_control.py
```diff
@@ -261,6 +261,27 @@ class _EEFTwistCoordinator(ControlCoordinator):
...
+        coordinator = make_coordinator()
+        matching_task = RecordingTask("eef")
+        other_task = RecordingTask("other")
+        coordinator._tasks = {"eef": matching_task, "other": other_task}
+        coordinator._routes = {
+            "coordinator_ee_twist_command": [
+                (matching_task, "on_ee_twist_command", Routing.BY_TASK_NAME),
+                (other_task, "on_ee_twist_command", Routing.BY_TASK_NAME),
+            ]
+        }
```
mustafab0: This is now deprecated. You don't need to route `TASK_NAME` anymore

dimos/hardware/whole_body/damiao/adapter.py
```diff
@@ -0,0 +1,383 @@
...
+        self._runtime_config = config
+        self._hardware_id = hardware_id
+        self._connected = False
+        self._active = False
+        self._has_state = False
+        self._robot: can_motor_control.Robot
+        self._arms: dict[str, can_motor_control.Arm]
+        self._grippers: dict[str, can_motor_control.Gripper]
+        self._pin_model: pinocchio.Model
+        self._pin_data: pinocchio.Data
```
mustafab0, line 88: We should consider keeping these IK agnostic. I think it is okay as we primarily work with pinnochio, but this should be something we are able to switch easily. 

Not a blocker.

mustafab0 (dismissed):

dimos/hardware/whole_body/damiao/adapter.py
TomCC7: these are not IK but gravity compensation. if we have more robots that needs similar function in the future we can surely create some shared base

dimos/cli/can.py
TomCC7: A probe would be useful, but we are deferring it rather than reporting unreliable results. Generic CAN has no protocol-independent ping, OpenYAM/Damiao motor topology can vary, and can-motor-control does not expose response freshness or discovery status. A connect/refresh check could therefore pass without proving that a motor replied. We should add this when the motor layer exposes a reliable response signal.

dimos/cli/test_can.py
TomCC7: Fixed in a0db77aaa. The setup command now enforces an inclusive minimum bitrate of 10,000, and the CLI test verifies that 9,999 is rejected.

dimos/control/test_control.py
TomCC7: Fixed in a0db77aaa. Removed the obsolete BY_TASK_NAME test; current per-instance/card routing coverage remains in test_coordinator_routing.py.

pyproject.toml
```diff
@@ -286,11 +286,14 @@ manipulation = [
     "xarm-python-sdk>=1.17.0",
     "a750-control; sys_platform == 'linux' and platform_machine == 'x86_64'",
 
+    "can-motor-control>=0.0.5; sys_platform == 'linux'",
```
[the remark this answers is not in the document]
TomCC7: Fixed in 1a8605c30 by raising both direct constraints to can-motor-control>=0.0.7 and refreshing uv.lock. PyPI 0.0.7 includes cp310-abi3 manylinux_2_34 wheels for both aarch64 and x86_64. I also verified the locked project with an aarch64 manylinux dry-run; it resolves can-motor-control 0.0.7 successfully.

dimos/hardware/whole_body/damiao/adapter.py
```diff
@@ -0,0 +1,383 @@
...
+            self._refresh()
+            return True
+        except Exception:
+            logger.exception(
+                "Damiao whole-body adapter failed to connect",
+                hardware_id=self._hardware_id,
+            )
+            try:
+                if robot.is_connected():
+                    robot.disable()
```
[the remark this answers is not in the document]
TomCC7: Fixed in ccb7686b5. `can-motor-control` 0.0.7 has no public close/disconnect API, and its context-manager exit only disables motors; the published Rust source releases transport resources from `Robot::drop`. The adapter now disables the robot and clears every retained Robot/Arm/Gripper handle on both failed-connect rollback and normal disconnect, allowing the upstream owner to drop. Added regression coverage for group lookup, gravity-model load, and initial-refresh failures; each verifies transport release, successful reconnect, and release on the subsequent disconnect.

mustafab0 (approved):

PR #3130, opened 2026-07-21
feat(control): per-robot JointState output streams

mustafab0:
## Problem

A multi-robot coordinator publishes only one merged `coordinator_joint_state`, so every consumer has to untangle a concatenated joint list to find the arm it cares about.

Issue: #

---

## Solution

Coordinator subclasses can declare `{hardware_id}_joints: Out[JointState]` per robot. The tick loop reuses the read it and publishes outside the hardware lock. 

A missing port fails loudly at `start()` before any adapter connects, and at runtime `add_hardware`. 

`coordinator-dual-mock` is the migrated exemplar; it pins `instance_name="ControlCoordinator"` since a subclass otherwise serves RPCs under its own name (`start()` now warns). 

---

## Breaking Changes

None. Backwards support bool added with `publish_robot_joint_states=True`. Keeps all previous blueprints working.

Will remove this after transisitioning all blueprints to new Control Coordinator

---

## How to Test

1. `uv run pytest dimos/control -m 'not (tool or self_hosted or mujoco or self_hosted_large)'`
2. `dimos run coordinator-dual-mock`
3. Confirm `left_arm_joints` and `right_arm_joints` each carry only their own arm, alongside the unchanged `coordinator_joint_state`.

### DIFF
--- a/dimos/control/test_per_robot_joint_states.py
+++ b/dimos/control/test_per_robot_joint_states.py
@@ -1,0 +1,394 @@
+# Copyright 2026 Dimensional Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tests for opt-in per-robot JointState output streams."""
+
+from __future__ import annotations
+
+from collections.abc import Callable, Iterator
+import threading
+from typing import Any
+from unittest.mock import MagicMock
+
+import pytest
+
+from dimos.control.components import HardwareComponent, HardwareType, make_joints
+import dimos.control.coordinator as coord_mod
+from dimos.control.coordinator import ControlCoordinator, TaskConfig
+from dimos.control.hardware_interface import ConnectedHardware
+from dimos.control.tick_loop import TickLoop
+from dimos.core.stream import In, Out
+from dimos.hardware.manipulators.registry import adapter_registry
+from dimos.hardware.manipulators.spec import ManipulatorAdapter
+from dimos.msgs.sensor_msgs.JointState import JointState
+
+LEFT_JOINTS = make_joints("left_arm", 2)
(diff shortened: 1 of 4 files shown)


dimos/robot/manipulators/test_blueprints.py
```diff
@@ -139,3 +141,13 @@ def test_manipulator_keyboard_blueprint_uses_eef_twist_and_light_keyboard_kwargs
...
+
+
+def test_dual_mock_coordinator_declares_a_port_for_every_arm() -> None:
+    kwargs = _module_kwargs(coordinator_dual_mock, DualMockCoordinator)
+    hints = get_type_hints(DualMockCoordinator)
+
+    assert kwargs["publish_robot_joint_states"] is True
+    assert kwargs["instance_name"] == "ControlCoordinator"
+    for hardware in kwargs["hardware"]:
+        assert get_origin(hints[f"{hardware.hardware_id}_joints"]) is Out
```
paul-nechifor: I don't understand what this test achieves. Asserting config values is not useful, but this isn't even testing a real config. `coordinator_dual_mock` is a test blueprint, right?

I think it would be better to delete the whole file.

dimos/robot/all_blueprints.py
```diff
@@ -181,6 +181,7 @@
     "drone-camera-module": "dimos.robot.drone.camera_module.DroneCameraModule",
     "drone-connection-module": "dimos.robot.drone.connection_module.DroneConnectionModule",
     "drone-tracking-module": "dimos.robot.drone.drone_tracking_module.DroneTrackingModule",
+    "dual-mock-coordinator": "dimos.robot.manipulators.common.mock.DualMockCoordinator",
```
paul-nechifor: This is not a real blueprint, right? It should start with `_` so it's not added here.

dimos/control/test_per_robot_joint_states.py
```diff
@@ -0,0 +1,400 @@
...
+
+def _left() -> HardwareComponent:
+    return _component("left_arm", LEFT_JOINTS, LEFT_POSITIONS)
+
+
+def _right() -> HardwareComponent:
+    return _component("right_arm", RIGHT_JOINTS, RIGHT_POSITIONS)
+
+
+def _wait_for(predicate: Callable[[], bool], timeout: float = 5.0) -> bool:
```
paul-nechifor: We have a million wait functions in the repo. Use the `wait_until` fixture. There's no need to import it. It's defined in the root `conftest.py` so you can access it from any test.

dimos/robot/manipulators/test_blueprints.py
mustafab0: Will undo changes for now

paul-nechifor (approved):

PR #3175, opened 2026-07-24
feat(control): add extension sideloading

TomCC7:
## Problem

External robot packages need a supported way to register ControlCoordinator hardware adapter names and control task types without placing files inside the DimOS source tree. Without this, an outside package can define a blueprint but still cannot use `HardwareComponent(adapter_type=...)` or `TaskConfig(type=...)` for custom hardware/control unless DimOS itself is modified.

Closes DIM-1031

## Solution

Add a public `dimos.control.extensions` facade for explicit extension registration:

- `register_hardware_adapter(HardwareType, adapter_type, factory)` dispatches to the existing manipulator, base, or whole-body adapter registry.
- `register_control_task(task_type, factory_path)` registers lazy control task factories without importing the target module at registration time.
- Hardware and task registries now reject conflicting duplicate registrations while keeping exact same mapping re-registration idempotent.
- Add docs and a runnable no-hardware external package example that logs registration, adapter construction, task creation, task ticks, and adapter writes.

This keeps blueprint usage unchanged: external packages register before coordinator construction, then use normal `HardwareComponent` and `TaskConfig` names.

## How to Test

Manual QA:

```bash
uv run python examples/external_control_extension/demo_external_control.py
```

Expected output includes `[external_test_robot]` logs for:

- registering `BASE/external_test_base`
- registering `external_test_drive`
- constructing and connecting `ExternalTestBaseAdapter`
- creating and ticking `ExternalTestDriveTask`
- writing velocities through the external adapter


TomCC7:
Addressed Greptile's factory-path validation finding in 9f92f2bec. Both control-task and hardware-adapter registration now require exactly one `:` separator, so inputs such as `module:attribute:extra` fail immediately with `ValueError` instead of reaching lazy resolution. Added regression coverage for both registries.\n\nValidation:\n- `uv run pytest dimos/control/tasks/test_registry.py dimos/control/test_extensions.py dimos/hardware/test_adapter_registries.py -q` — 48 passed\n- `uv run mypy` — no issues in 918 source files\n- applicable pre-commit hooks — passed

PR #3178, opened 2026-07-25
feat(cli): add dimos cache clean

TomCC7:
## Contribution path

- Follow-up for robot asset caching in #2505

image

## Problem

DimOS writes regenerable assets to several unrelated platform, state, and temporary directories, but users have no single command to clear them. This becomes more important with #2505, which adds cached robot source checkouts and derived URDFs.

## Solution

Move all DimOS-owned caches beneath the platform-specific `CACHE_DIR`, then add `dimos cache clean` to remove that hierarchy.

The command:

- preserves logs, recordings, datasets, configuration, and third-party caches;
- refuses to clean while a DimOS run is active unless `--force` is supplied;
- preserves robot asset Git checkouts with local changes or local-only commits unless forced;
- displays the cache root and deletion policy, then requires `y/N` confirmation by default;
- supports non-interactive cleanup with `--yes`, independently of `--force`;
- reports filesystem failures and incomplete cleanup with a nonzero exit code; and
- centralizes cache paths so cache writers and the CLI use the same locations.

## How to Test

```bash
dimos cache clean
uv run pytest dimos/utils/test_cache.py dimos/robot/cli/test_dimos.py dimos/robot/cli/test_cli_startup.py dimos/utils/test_ament_prefix.py -q
```

## AI assistance

OpenAI Codex with GPT-5 implemented and tested the change under the author's direction.

## Checklist


dimos/constants.py
```diff
@@ -27,6 +28,12 @@
     STATE_DIR = Path(GLib.get_user_state_dir()) / "dimos"
     CACHE_DIR = Path(GLib.get_user_cache_dir()) / "dimos"
 
+ROBOT_ASSET_CACHE_DIR = CACHE_DIR / "robot_assets"
+DENO_CACHE_DIR = STATE_DIR / "deno"
+DRAKE_URDF_CACHE_DIR = Path(tempfile.gettempdir()) / "dimos_urdf_cache"
+VISER_URDF_CACHE_DIR = Path(tempfile.gettempdir()) / "dimos_viser_urdf_cache"
+AMENT_PREFIX_CACHE_DIR = Path(tempfile.gettempdir()) / "dimos_ament_prefix"
```
TomCC7: all these should just go to cache_dir

TomCC7: Moved every DimOS-owned cache under CACHE_DIR, including Deno (CACHE_DIR/deno), URDF processing, Viser URDFs, and the ament index. cache clean now removes the single CACHE_DIR hierarchy while preserving modified robot Git checkouts.

dimos/utils/ament_prefix.py
```diff
@@ -54,7 +55,7 @@ def _setup_ament_index(package_paths: dict[str, Path]) -> None:
     global _prefix_dir
 
     if _prefix_dir is None:
-        _prefix_dir = Path(tempfile.gettempdir()) / "dimos_ament_prefix"
+        _prefix_dir = AMENT_PREFIX_CACHE_DIR
```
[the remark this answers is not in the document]
TomCC7: Good catch. The in-memory registration cache now skips setup only when the marker file and symlink still exist on disk and the symlink resolves to the expected package. Added a regression test that removes the ament cache between registrations and verifies it is rebuilt.

dimos/robot/cli/dimos.py
```diff
@@ -156,6 +157,93 @@ def callback(**kwargs) -> None:  # type: ignore[no-untyped-def]
 main.add_typer(go2tool_app, name="go2tool")
 main.add_typer(piper_app, name="piper")
 
+cache_app = typer.Typer(help="Manage DimOS caches", no_args_is_help=True)
```
TomCC7: should move to separate file for future extensibility

dimos/utils/ament_prefix.py
```diff
@@ -54,22 +55,28 @@ def _setup_ament_index(package_paths: dict[str, Path]) -> None:
...
-        if _ament_registered.get(pkg_name) == resolved:
+        marker = resource_dir / pkg_name
+        share_link = prefix / "share" / pkg_name
+        if (
+            _ament_registered.get(pkg_name) == resolved
+            and marker.is_file()
+            and share_link.is_symlink()
+            and share_link.resolve() == resolved
+        ):
             continue
```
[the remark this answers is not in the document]
TomCC7: Agreed—the on-disk validation alone could not close this TOCTOU race. `cache clean` now refuses to run whenever a DimOS run is active, including with `--force`. `--force` is narrowed to deleting cached robot Git checkouts with local work, so it can no longer remove the ament prefix from under an active process. Added a CLI regression test for the forced case and updated the documentation.

dimos/robot/cli/dimos.py
TomCC7: Moved the cache command and its Typer app into `dimos/robot/cli/cache.py`, leaving `dimos.py` responsible only for registering the subgroup. I also moved the command tests into `dimos/robot/cli/test_cache.py` so future cache subcommands and their tests have a focused home.

dimos/constants.py
```diff
@@ -27,6 +27,12 @@
     STATE_DIR = Path(GLib.get_user_state_dir()) / "dimos"
     CACHE_DIR = Path(GLib.get_user_cache_dir()) / "dimos"
 
+ROBOT_ASSET_CACHE_DIR = CACHE_DIR / "robot_assets"
+DENO_CACHE_DIR = CACHE_DIR / "deno"
```
paul-nechifor: This is only ever used in dimos/utils/deno.py, right? I don't think it should be here. (Same of others if applies.)

We should probably keep this file short and only for common stuff. Ivan is already upset at me for creating it. 😅

dimos/robot/cli/cache.py
```diff
@@ -0,0 +1,103 @@
...
+
+_CACHE_ENTRY_LABELS = {
+    "ament_prefix": "ament package index",
+    "deno": "downloaded Deno runtime",
+    "robot_assets": "robot source and derived assets",
+    "scene_meshes": "cooked scene meshes",
+    "scene_sources": "normalized scene sources",
+    "urdf": "prepared URDFs and convex hulls",
+    "viser_urdf": "Viser-prepared URDFs",
+}
```
paul-nechifor: Is this really needed? I pretty sure people will forgot to update this so I'd say it's better to not have it at all. Maybe you could just print what's inside it with the size? Maybe something like `du --max-depth 1 <path>`

dimos/constants.py
TomCC7: Agreed. I removed all five child-path constants from `dimos/constants.py`; Deno, Drake, Viser, and ament now derive their paths locally from the shared `CACHE_DIR`. We also decided to defer the robot-assets layout and Git-aware cleanup policy to PR #2505, where the cache producer and its cleanup contract can land together. This PR now removes only the DimOS cache root, and `--force` plus the robot Git inspection code have been removed.

dimos/robot/cli/cache.py
TomCC7: Maybe we should centralize the cache directory registration for each usage? I can introduce this in later pr tho as this carries some priors.

TomCC7: Agreed—the label map would become another registry that cache producers could drift from. I removed it. The preview now lists the actual top-level paths currently present under the DimOS cache root, in sorted order, with no maintained per-entry metadata.

dimos/utils/ament_prefix.py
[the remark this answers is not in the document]
TomCC7: Final follow-up: after simplifying this PR, `--force` and robot-specific cleanup were removed entirely, so cache cleanup can no longer run while a DimOS process is active. I therefore removed the ament stale-registration workaround and its regression test as coverage for a no-longer-reachable forced-cleanup path. The CLI active-run test remains the regression for this safety boundary.

dimos/robot/cli/cache.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+@app.command("clean")
+def clean(
+    yes: bool = typer.Option(
+        False,
+        "--yes",
+        "-y",
+        help="Skip the confirmation prompt",
+    ),
```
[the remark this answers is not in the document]
TomCC7: Good catch. The command now checks the run registry once before displaying the preview and again immediately after confirmation, just before calling the cleanup service. Added a regression test where the first check is inactive and the second returns a newly started run; cleanup is not called and the command exits with the active-run error.

paul-nechifor (dismissed):

dimos/robot/cli/cache.py
```diff
@@ -0,0 +1,84 @@
...
+    else:
+        typer.echo(f"  - {CACHE_DIR} (empty cache root)")
+
+    typer.echo("")
+    if not yes and not typer.confirm("Continue?", default=False):
+        typer.echo("Cache cleanup cancelled.")
+        return
+
+    _refuse_active_run()
+    result = clean_caches()
```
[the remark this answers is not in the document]
TomCC7: Agreed—the second registry check still left a TOCTOU window. I replaced that with an interprocess gate plus per-run usage locks in de2ee322c. `dimos run` now holds a usage lock for its full lifetime; cleanup holds the startup gate while checking usage locks and deleting the cache. Existing runs make cleanup fail, while a run starting during cleanup waits until deletion finishes. Stale usage markers are reclaimed, and the lock protocol has regression coverage.

spomichter (approved):

PR #3182, opened 2026-07-25
fix(go2): override ReplayConnection.stop() to avoid AttributeError on hot-reload

### LINKED ISSUE #1967: PY API - Hot-reload of GO2Connection is often broken
`app.restart(GO2Connection, reload_source=True)` either hangs (timeout) or leaves the underlying connection in a bad state — subsequent calls hit `Exception: Data channel is not open`, `ConnectionRefusedError`, or simply `TimeoutError: result expired`.

Reproduced against a `--replay` daemon (no WebRTC at all): `restart()` timed out at 30s and the module became unreachable for the rest of the session. Daemon log:

```
[err] pubsubrpc.py  Exception in RPC handler for GO2Connection/stop:
  'ReplayConnection' object has no attribute 'stop_timer'
  File ".../dimos/robot/unitree/go2/connection.py", line 275, in stop
    self.connection.stop()
  File ".../dimos/robot/unitree/connection.py", line 129, in stop
    if self.stop_timer:
```

This sometimes works! but mostly not

---
*Synced from* [DIM-838](https://linear.app/dimensional/issue/DIM-838) *by summer*

xiaoyaoqilan:
Fixes #1967.

**Root cause.** `ReplayConnection` intentionally skips `UnitreeWebRTCConnection.__init__()` (that constructor opens a live WebRTC peer via `LegionConnection` + `connect()`), so it never initializes `self.stop_timer`, `self.loop`, or `self.conn`. The inherited `stop()` therefore raises `AttributeError: 'ReplayConnection' object has no attribute 'stop_timer'` as soon as `GO2Connection.stop()` is invoked — which is exactly what `restart()` does during hot-reload, breaking the reload.

**Fix.** Override `stop()` on `ReplayConnection` to clear the deadman timer slot and otherwise do nothing: a replay connection has no live loop/peer to disconnect, and the module's teardown is handled by the framework. This removes the AttributeError that aborted `restart()`.

Note: this fixes the concrete `AttributeError` thrown by `stop()` during hot-reload. If `restart()` still hangs for reasons unrelated to teardown (e.g. the daemon re-registering the module), that is a separate path and worth a follow-up issue.

### DIFF
--- a/dimos/robot/unitree/go2/connection.py
+++ b/dimos/robot/unitree/go2/connection.py
@@ -259,6 +259,15 @@ class ReplayConnection(UnitreeWebRTCConnection, CompositeResource):
         """Fake publish request for testing."""
         return {"status": "ok", "message": "Fake publish"}
 
+    def stop(self) -> None:
+        # A replay connection has no live WebRTC loop or peer to tear down, so
+        # the inherited UnitreeWebRTCConnection.stop() (which touches self.loop
+        # and self.conn) does not apply here. Clear the deadman timer slot and
+        # let the framework handle the rest of the module teardown.
+        # Fixes 'ReplayConnection' object has no attribute 'stop_timer' raised
+        # during GO2Connection hot-reload / restart (#1967).
+        self.stop_timer = None
+
 
 _Config = TypeVar("_Config", bound=ConnectionConfig, default=ConnectionConfig)
 


paul-nechifor:
This does not fix https://github.com/dimensionalOS/dimos/issues/1967 . The issue there is present in with UnitreeWebRTCConnection as well, not just ReplayConnection.

PR #3183, opened 2026-07-25
feat: extract manipulation plan execution manager

TomCC7:
## Contribution path

- Linked issue or discussion: [DIM-1355](https://linear.app/dimensional/issue/DIM-1355/centralize-manipulation-plan-execution-through-one-coordinator)

## Problem

Manipulation execution mixed plan interpretation, per-robot trajectory splitting, coordinator task selection, dispatch, rollback, cancellation, and status polling into `ManipulationModule`.

That design treated a coordinated plan as independently executable robot fragments. It also duplicated trajectory-task policy above the control coordinator, even though the coordinator owns the hardware state and arbitration needed to decide whether a trajectory is safe to start.

## Solution

- Add `PlanExecutionManager` as the boundary between planning and control. It maps one complete `GeneratedPlan` from global planner joint names to coordinator joint names, serializes execute/cancel operations, and makes one coordinator RPC.
- Inject the blueprint's `ControlCoordinator` module reference into `ManipulationModule` and pass that framework-managed RPC proxy directly to `PlanExecutionManager`; no forwarding client or independently owned RPC lifecycle is needed.
- Require every `ManipulationModule` blueprint to include a `ControlCoordinator`. Remove the standalone `xarm6-planner-only` and `dual-xarm6-planner` blueprints in favor of their mock-capable coordinator-backed replacements.
- Give each `ControlCoordinator` exactly one `JointTrajectoryTask`. Multi-arm coordinators configure that task with the union of their controlled joints; single-arm subset plans remain valid.
- Add typed execution and cancellation results. Deterministic validation failures are rejections; RPC failures are `UNCERTAIN` because the remote operation may have occurred.
- Move executable-trajectory validation into `JointTrajectoryTask`, including a start-state check against coordinator-owned hardware positions. Configure `start_position_tolerance` through task parameters.
- Model immutable execution configuration with `attrs`, while retaining Pydantic at the untyped task-parameter boundary.
- Keep direct gripper commands separate from planned trajectory execution.

## Breaking API and behavior changes

| Before | Now | Migration / behavior |
| --- | --- | --- |
| `ManipulationModule.execute(robot_name=...)` | `execute()` | Execution no longer accepts a robot/arm selector. It always dispatches the complete stored plan. |
| `ManipulationModule.execute_plan(plan, robot_name=...)` | `execute_plan(plan)` | Partial execution is expressed by planning for a subset of joints, not by filtering a coordinated plan at execution time. A multi-arm plan is dispatched atomically. |
| `ManipulationModule.get_trajectory_status(robot_name=...)` | Removed | The manipulation layer no longer exposes or polls coordinator task state. Acceptance is reported immediately; callers do not infer completion from a duration sleep. |
| `RobotModelConfig.coordinator_task_name` | Removed | Robot configs no longer route trajectories to named tasks. Configure one trajectory task on the coordinator instead. |
| `CoordinatorClient.execute_trajectory(task_name, trajectory) -> bool` | `execute_trajectory(trajectory) -> TrajectoryExecutionResult` | Remove the task name and handle semantic acceptance/rejection statuses. The interactive shell returns immediately after acceptance. |
| `CoordinatorClient.cancel_trajectory(task_name) -> bool` | `cancel_trajectory() -> TrajectoryCancellationResult` | Remove the task name and handle `CANCELLED`, `ALREADY_STOPPED`, `NO_TRAJECTORY_TASK`, or `UNCERTAIN`. |
| `ManipulationModule` could run without a coordinator | A `ControlCoordinator` module reference is required | Compose manipulation and control in the same blueprint. The framework injects the typed RPC proxy automatically. |
| `xarm6-planner-only` and `dual-xarm6-planner` | Removed | Use `xarm7-planner-coordinator` for a mock-capable single-arm planning stack or `dual-xarm6-planner-coordinator` for dual-arm planning. |
| Multiple `JointTrajectoryTask` instances per coordinator | One task per coordinator | Adding a second trajectory task raises `ValueError`. Multi-arm blueprints must create one task claiming all coordinated joints. |
| `JointTrajectoryTask.execute(trajectory) -> bool` | `execute(trajectory, current_positions) -> TrajectoryExecutionResult` | The coordinator supplies current hardware positions, and the task rejects missing or mismatched start state. |

Generated-plan joint names remain globally scoped as `<robot>/<local_joint>`. `ExecutionTarget` resolves those names to coordinator joints; the caller no longer passes a robot ID during execution.

## How to Test

```bash
uv run pytest dimos/control dimos/manipulation dimos/robot/manipulators/test_blueprints.py -q
```

Current result: 632 passed, 12 deselected.

The suite covers whole-plan mapping, subset trajectories, multi-arm atomic dispatch, execute/cancel serialization, typed coordinator outcomes, start-state rejection, one-task enforcement, module lifecycle cleanup, and updated manipulator blueprints.

Additional validation:

- 12 Drake/self-hosted manipulation integration tests passed.
- 2 self-hosted-large single/dual-arm blueprint and RPC tests passed.
- Generated blueprint registry validation passed.
- Targeted mypy passed for the production execution, manipulation, RPC-shell, and blueprint changes.

## AI assistance

OpenAI Codex with GPT-5 provided substantial assistance with design exploration, implementation, tests, documentation, and review-feedback changes under human direction. The author reviewed and approved the design decisions interactively.

## Checklist


dimos/manipulation/execution_manager.py
```diff
@@ -0,0 +1,526 @@
...
+
+class CoordinatorCancelOutcome(Enum):
+    """Result of asking one coordinator task to cancel."""
+
+    CANCELLED = auto()
+    ALREADY_STOPPED = auto()
+    UNKNOWN = auto()
+
+
+class ExecutionDispatchOutcome(Enum):
```
TomCC7: how is this different from coordinator dispatch outcome can we merge them

dimos/manipulation/execution_manager.py
```diff
@@ -0,0 +1,526 @@
...
+    unresolved_tasks: tuple[TaskName, ...] = ()
+
+
+@dataclass(frozen=True)
+class ExecutionPolicy:
+    """Configuration for planned execution validation."""
+
+    plan_start_tolerance: float = 1e-6
+
+    def __post_init__(self) -> None:
```
TomCC7: use pydantic here

dimos/manipulation/execution_manager.py
```diff
@@ -0,0 +1,526 @@
...
+        if not math.isfinite(self.plan_start_tolerance) or self.plan_start_tolerance < 0:
+            raise ValueError("plan_start_tolerance must be finite and non-negative")
+
+
+@dataclass(frozen=True)
+class ExecutionTarget:
```
TomCC7, line 76: discuss with me about whether using pydantic is good here. it's good for value validation for sure

dimos/manipulation/execution_manager.py
```diff
@@ -0,0 +1,526 @@
...
+    ) -> CoordinatorDispatchOutcome:
+        """Ask a coordinator task to execute a trajectory."""
+        ...
+
+    def cancel(self, task_name: TaskName) -> CoordinatorCancelOutcome:
+        """Ask a coordinator task to stop executing."""
+        ...
+
+
+class CoordinatorExecutionAdapter:
```
TomCC7: maybe should put this into control coordinator's position? Since this looks like a generic client

dimos/manipulation/manipulation_module.py
```diff
@@ -179,6 +178,8 @@ def __init__(self, **kwargs: Any) -> None:
 
         # Coordinator integration (lazy initialized)
         self._coordinator_client: RPCClient | None = None
+        self._execution_manager: PlanExecutionManager | None = None
+        self._execution_manager_client: RPCClient | None = None
```
TomCC7: I guess we should not give execution manager and client a none type at all - none type is only because we need to initilaize in a later lifecycle. we should just declare here and just assign in later init, and there won't be any reference to this variable by the logic of whole system:
```python
self._execution_manager: PlanExecutionManager
```

dimos/manipulation/manipulation_module.py
TomCC7: let's just don't lazy init here. init normally as other classes in this module as we almost always need this

dimos/manipulation/manipulation_module.py
```diff
@@ -1438,61 +1373,50 @@ def _get_coordinator_client(self) -> RPCClient | None:
...
-        Returns:
-            Trajectory with coordinator joint names
-        """
-        if not robot_config.joint_name_mapping:
-            return trajectory  # No translation needed
-
-        # Translate joint names
-        coordinator_names = [
-            robot_config.get_coordinator_joint_name(j) for j in trajectory.joint_names
+    def _get_execution_manager(self) -> PlanExecutionManager | None:
```
TomCC7: again no lazy creation

dimos/manipulation/execution_manager.py
```diff
@@ -0,0 +1,526 @@
...
+            raise ValueError("Execution targets must have unique coordinator task names")
+
+        self._targets = {target.robot_name: target for target in target_items}
+        self._current_joint_state = current_joint_state
+        self._coordinator = coordinator
+        self._policy = policy or ExecutionPolicy()
+
+        self._state_lock = threading.Lock()
+        self._dispatch_lock = threading.Lock()
+        self._cancel_lock = threading.Lock()
```
TomCC7: validate that all three locks are required here and can't be merged

dimos/manipulation/execution_manager.py
```diff
@@ -0,0 +1,526 @@
...
+
+                accepted: list[TaskName] = []
+                for dispatch in prepared:
+                    with self._state_lock:
+                        if self._generation != token or self._cancellation_in_progress:
+                            return self._rejected("Execution was superseded by cancellation")
+                        self._possibly_active_tasks.add(dispatch.task_name)
+
+                    outcome = self._coordinator.execute(dispatch.task_name, dispatch.trajectory)
+                    if outcome is CoordinatorDispatchOutcome.ACCEPTED:
```
TomCC7: should use match

dimos/manipulation/execution_manager.py
```diff
@@ -0,0 +1,526 @@
...
+                self._dispatch_in_progress = False
+
+    def cancel(self) -> CancellationResult:
+        """Cancel every task that might have accepted a generated plan."""
+        with self._cancel_lock:
+            with self._state_lock:
+                self._cancellation_in_progress = True
+                self._generation += 1
+            try:
+                with self._dispatch_lock:
```
TomCC7: so dirty three locks are used together...

dimos/manipulation/execution_manager.py
```diff
@@ -0,0 +1,549 @@
...
+            )
+
+    @classmethod
+    def from_coordinator_mapping(
+        cls,
+        *,
+        robot_name: RobotName,
+        model_joint_names: Sequence[str],
+        coordinator_task_name: TaskName,
+        coordinator_to_model: Mapping[str, str],
```
TomCC7, line 131:
```suggestion
        #TODO: unify coordinator joint name with planner
        coordinator_to_model: Mapping[str, str],
```

dimos/manipulation/execution_manager.py
```diff
@@ -0,0 +1,549 @@
...
+            robot_name=robot_name,
+            model_joint_names=local_names,
+            coordinator_task_name=coordinator_task_name,
+            model_to_coordinator={
+                model_name: reverse.get(model_name, model_name) for model_name in local_names
+            },
+        )
+
+
+class CoordinatorExecutionPort(Protocol):
```
TomCC7: why spec here instead of actual class

dimos/manipulation/execution_manager.py
```diff
@@ -0,0 +1,549 @@
...
+        target_names = [target.robot_name for target in target_items]
+        task_names = [target.coordinator_task_name for target in target_items]
+        if len(set(target_names)) != len(target_names):
+            raise ValueError("Execution targets must have unique robot names")
+        if len(set(task_names)) != len(task_names):
+            raise ValueError("Execution targets must have unique coordinator task names")
+
+        self._targets = {target.robot_name: target for target in target_items}
+        self._current_joint_state = current_joint_state
+        self._coordinator = coordinator
```
TomCC7, line 178: should we use coordinator_client here as naming

dimos/manipulation/execution_manager.py
```diff
@@ -0,0 +1,549 @@
...
+
+
+class PlanExecutionManager:
+    """Validate, dispatch, replace, and cancel complete generated plans."""
+
+    def __init__(
+        self,
+        *,
+        targets: Iterable[ExecutionTarget],
+        current_joint_state: Callable[[], JointState],
```
TomCC7: function should use verb (like `get_current_joint_state`)

dimos/manipulation/execution_manager.py
```diff
@@ -0,0 +1,549 @@
...
+        self._operation_condition = threading.Condition()
+        self._operation_active = False
+        self._cancel_waiters = 0
+        self._possibly_active_tasks: set[TaskName] = set()
+
+    def execute(self, plan: GeneratedPlan) -> ExecutionDispatchResult:
+        """Validate and dispatch one complete generated plan."""
+        prepared = self._prepare_plan(plan)
+        if isinstance(prepared, str):
+            return self._rejected(prepared)
```
TomCC7: raise error + try catch seems more elegant

dimos/manipulation/execution_manager.py
```diff
@@ -0,0 +1,549 @@
...
+                    trajectory=JointTrajectory(
+                        joint_names=coordinator_names,
+                        points=points,
+                        timestamp=trajectory.timestamp,
+                    ),
+                )
+            )
+        return tuple(prepared)
+
+    def _freshness_error(self, plan: GeneratedPlan) -> str | None:
```
TomCC7: this function seems to be doing more than freshness check... summarize what it does, try to simplify logic, and rename if possible

dimos/manipulation/execution_manager.py
```diff
@@ -0,0 +1,549 @@
...
+            replacement = self._cancel_tracked_tasks()
+            if not replacement.safe:
+                return ExecutionDispatchResult(
+                    outcome=ExecutionOutcome.UNCERTAIN,
+                    message="Cannot replace unresolved coordinator tasks",
+                    unresolved_tasks=replacement.unresolved_tasks,
+                )
+
+            freshness_error = self._freshness_error(plan)
+            if freshness_error is not None:
```
TomCC7:
```suggestion
            if self._freshness_error(plan) is not None:
```

dimos/manipulation/execution_manager.py
```diff
@@ -0,0 +1,549 @@
...
+
+        self._targets = {target.robot_name: target for target in target_items}
+        self._current_joint_state = current_joint_state
+        self._coordinator = coordinator
+        self._policy = policy or ExecutionPolicy()
+
+        self._operation_condition = threading.Condition()
+        self._operation_active = False
+        self._cancel_waiters = 0
+        self._possibly_active_tasks: set[TaskName] = set()
```
TomCC7: do we really need to handle multiple jtc tasks? I mean from the whole picture, is that really necessary to define multiple jtc task within control coordinator?

dimos/manipulation/execution_manager.py
```diff
@@ -0,0 +1,549 @@
...
+            finally:
+                self._cancel_waiters -= 1
+        try:
+            return self._cancel_tracked_tasks()
+        finally:
+            with self._operation_condition:
+                self._operation_active = False
+                self._operation_condition.notify_all()
+
+    def _prepare_plan(self, plan: GeneratedPlan) -> tuple[_PreparedDispatch, ...] | str:
```
TomCC7: check if we can simplify this function by remove some check that's not really necessary

dimos/control/tasks/trajectory_task/trajectory_task.py
```diff
@@ -87,6 +129,11 @@ def __init__(self, name: str, config: JointTrajectoryTaskConfig) -> None:
         """
         if not config.joint_names:
             raise ValueError(f"JointTrajectoryTask '{name}' requires at least one joint")
+        if (
+            not math.isfinite(config.start_position_tolerance)
+            or config.start_position_tolerance < 0.0
+        ):
+            raise ValueError("start_position_tolerance must be finite and non-negative")
```
TomCC7: should change this to use attrs as well instead of validation in init

dimos/control/coordinator.py
```diff
@@ -426,6 +435,29 @@ def get_joint_positions(self) -> dict[str, float]:
                     positions[joint_name] = joint_state.position
             return positions
 
+    @rpc
+    def execute_trajectory(self, trajectory: JointTrajectory) -> TrajectoryExecutionResult:
```
TomCC7, line 462: I think the better approach is to keep this rpc within trajectory task, and when we do the actual task compute we provide current state?

dimos/manipulation/execution_manager.py
```diff
@@ -0,0 +1,300 @@
...
+                return CancellationResult(
+                    outcome=CancellationOutcome.UNCERTAIN,
+                    message=f"Coordinator cancel RPC failed: {exc}",
+                )
+
+            outcomes = {
+                TrajectoryCancellationStatus.CANCELLED: CancellationOutcome.CANCELLED,
+                TrajectoryCancellationStatus.ALREADY_STOPPED: (CancellationOutcome.ALREADY_STOPPED),
+                TrajectoryCancellationStatus.NO_TRAJECTORY_TASK: (
+                    CancellationOutcome.NO_TRAJECTORY_TASK
```
TomCC7: what's the point of this conversion vs directly using the same type...

dimos/control/tasks/trajectory_task/trajectory_task.py
TomCC7: Addressed in 25c4d1972. `JointTrajectoryTaskConfig` is now an immutable attrs value type with declarative validators, and the validation was removed from `JointTrajectoryTask.__init__`. Added regression coverage for empty joints and invalid tolerances.

dimos/control/coordinator.py
TomCC7: Discussed and intentionally deferred from this PR. A task cannot safely cache state only from `compute()` because inactive tasks are skipped, so the follow-up needs a properly designed coordinator-owned task runtime/observation lifecycle. This PR keeps the public coordinator trajectory RPCs for now.

dimos/manipulation/execution_manager.py
TomCC7: Addressed in 25c4d1972. The one-to-one cancellation conversion and duplicate manipulation result types are removed; `PlanExecutionManager.cancel()` now returns `TrajectoryCancellationResult` directly. `UNCERTAIN` is synthesized on that same shared result type only for RPC failures.

dimos/robot/manipulators/test_blueprints.py
```diff
@@ -168,9 +174,25 @@ def test_dual_xarm6_planner_coordinator_blueprints_preserve_visualization_backen
...
+        *(f"left_arm/joint{index}" for index in range(1, 7)),
+        *(f"right_arm/joint{index}" for index in range(1, 7)),
+    ]
+
+
+@pytest.mark.parametrize(
+    "blueprint",
+    [openarm_mock_planner_coordinator, openarm_planner_coordinator],
+)
+def test_bimanual_openarm_planners_use_one_trajectory_task(blueprint: Blueprint) -> None:
```
paul-nechifor: These tests don't really do anything? Why not just delete the whole file?

TomCC7: This file contains fast blueprint contract tests, and this specific test caught a real regression in this PR: after `ControlCoordinator` moved to one JTT, both OpenArm planner blueprints still declared two trajectory tasks. That made `openarm-mock-planner-coordinator` fail the self-hosted-large startup test. The assertion now keeps both the mock and hardware blueprints on the required single union-task shape; the runtime test only exercises the mock blueprint, so deleting this would drop coverage of the shipped hardware declaration. I would prefer to keep it.

dimos/manipulation/_test_manipulation_helpers.py
```diff
@@ -14,32 +14,27 @@
...
-        self._coordinator_client = None
-        self.config = MagicMock(planning_timeout=10.0)
+from dimos.manipulation.manipulation_module import ManipulationModule
 
 
 def make_module() -> ManipulationModule:
     """Create a lightweight ManipulationModule harness for behavior tests."""
-    return ManipulationModuleHarness()
+    module = ManipulationModule()
+    module.stop()
```
paul-nechifor: This is a bit odd, no? You're stopping a module that's never been started and using that?

You're calling _initialize_execution after stop, but why not call start, which does that?

dimos/manipulation/control/coordinator_client.py
```diff
@@ -308,36 +305,14 @@ def preview_trajectory(trajectory: JointTrajectory, joint_names: list[str]) -> N
...
-
-    print("\nTimeout waiting for trajectory")
-    return False
+def wait_for_completion(trajectory: JointTrajectory, timeout: float = 60.0) -> bool:
+    """Wait for the expected trajectory duration."""
+    wait_time = trajectory.duration + 0.5
+    if wait_time > timeout:
+        return False
+    time.sleep(wait_time)
+    print("\nTrajectory duration elapsed")
```
paul-nechifor: Don't use print, use logging.

Also, the name kinda doesn't make sense anymore. It doesn't wait for something, it just waits for a fixed amount of time.

Since it operates on JointTrajectory, maybe it should be a method of that class?

dimos/manipulation/control/coordinator_client.py
```diff
@@ -119,22 +125,13 @@ def get_joint_positions(self) -> dict[str, float]:
...
-            return {"state": int(result), "task": task_name}
-        return {}
-
-    def execute_trajectory(self, task_name: str, trajectory: JointTrajectory) -> bool:
-        """Execute a trajectory on a task via task_invoke."""
-        result = self._rpc.task_invoke(task_name, "execute", {"trajectory": trajectory})
-        return bool(result)
+    def execute_trajectory(self, trajectory: JointTrajectory) -> TrajectoryExecutionResult:
+        """Execute through the coordinator's sole trajectory task."""
+        return self._rpc.execute_trajectory(trajectory)  # type: ignore[no-any-return]
```
paul-nechifor: Please do not ignore. (Same for others.)

dimos/manipulation/manipulation_module.py
```diff
@@ -177,8 +172,9 @@ def __init__(self, **kwargs: Any) -> None:
         # Robot-local paths and trajectories are derived from this plan on demand.
         self._last_plan: GeneratedPlan | None = None
 
-        # Coordinator integration (lazy initialized)
-        self._coordinator_client: RPCClient | None = None
+        # Coordinator integration (initialized in start())
+        self._coordinator_client: ControlCoordinatorClient
```
paul-nechifor: This is not how you're supposed to do RPC. See https://github.com/dimensionalOS/dimos/blob/main/docs/usage/blueprints.md#calling-the-methods-of-other-modules

You declare this:

```
class MyModule(Module):
    _control_coordinator: ControlCoordinator
```

And `_control_coordinator` is automatically inserted as an RPC client for ControlCoordinator module.

dimos/control/coordinator_client.py
```diff
@@ -0,0 +1,70 @@
...
+from dimos.control.coordinator import ControlCoordinator
+from dimos.control.tasks.trajectory_task.trajectory_task import (
+    TrajectoryCancellationResult,
+    TrajectoryExecutionResult,
+)
+from dimos.core.rpc_client import RPCClient
+from dimos.msgs.trajectory_msgs.JointTrajectory import JointTrajectory
+
+
+class ControlCoordinatorClient:
```
paul-nechifor: Mentioned in another comment, but this shouldn't be needed.

When you have a class that just calls methods on another class, it's a bit of a sign that something is wrong.

dimos/robot/manipulators/test_blueprints.py
TomCC7: Agreed after separating the two concerns. The invariant is owned by `ControlCoordinator.add_task`, and `TestControlCoordinatorTrajectoryExecution::test_rejects_second_trajectory_task` directly verifies that a second JTT is rejected and not installed. I removed the redundant OpenArm blueprint test and its imports in `4313fd936`; the production blueprint fix remains.

paul-nechifor: > so deleting this would drop coverage of the shipped hardware declaration

What coverage is being dropped? The code here is:

```python
openarm_mock_planner_coordinator = autoconnect(
    planner(
        robots=[
            openarm_model_config("left"),
            openarm_model_config("right"),
        ],
    ),
    coordinator(
        hardware=[mock_left, mock_right],
        tasks=[trajectory_task(mock_left, mock_right)],
    ),
)
```

This is just a top level declaration. It's executed simply by being imported.

dimos/control/coordinator_client.py
TomCC7: Addressed in 1b5d0c62d. I deleted the forwarding ControlCoordinatorClient entirely. ManipulationModule now receives the framework-managed ControlCoordinator module reference and passes it directly to PlanExecutionManager.

dimos/manipulation/manipulation_module.py
TomCC7: Addressed in 1b5d0c62d. ManipulationModule now declares a required _control_coordinator: ControlCoordinator reference, and PlanExecutionManager receives that injected proxy through its constructor. The two planner-only blueprints were removed after confirming with their owner that a coordinator-backed topology is required.

dimos/manipulation/control/coordinator_client.py
TomCC7: Addressed in 1b5d0c62d. The standalone shell still creates a generic RPCClient because it is not a Module, but its dynamic execute/cancel return values now use explicit typed casts; the no-any-return suppressions are gone.

dimos/manipulation/control/coordinator_client.py
TomCC7: Addressed in 1b5d0c62d. I removed wait_for_completion instead of moving a sleep onto JointTrajectory. The shell now reports acceptance and the expected duration, then returns immediately; it no longer claims that a fixed sleep observes completion.

dimos/manipulation/_test_manipulation_helpers.py
TomCC7: Addressed in 1b5d0c62d. The stop-before-use helper is gone. A pytest factory now injects a mock coordinator, starts a normal ManipulationModule, disables only the unwired test port, and guarantees stop during fixture teardown.

paul-nechifor (dismissed):

TomCC7:
@mustafab0 I'll leave this pr for your review, if you think it's good to go can just enable merge

paul-nechifor (approved):

PR #3186, opened 2026-07-25
Resolve remote memory2 replays for Go2

xiaoyaoqilan:
## Summary

- add `dimos-replay://owner/repository/object-id` replay URIs
- download remote memory2 databases on demand
- verify content hashes, validate SQLite integrity, and cache verified recordings locally
- let the existing replay flow select the downloaded database through `DIMOS_REPLAY_DB`
- support the requested `dimos --replay run unitree-go2` workflow

## Motivation

This is PR 4 of the #3158 implementation stack. A recording uploaded in one location can be selected and replayed from another location without manually copying database files.

## Stack dependency

- depends on #3185, #3184, and #3177
- review the commit after `0be0040` for the isolated change

## Validation

- remote URI parsing, download, hash checking, cache reuse, and invalid-database paths are covered by tests
- targeted replay/Go2 tests passed
- `ruff check`, format checks, and targeted mypy passed

PR #3185, opened 2026-07-25
Add indexed memory2 replay uploads

xiaoyaoqilan:
## Summary

- create WAL-consistent SQLite snapshots from memory2 recordings
- validate snapshot integrity before upload
- upload memory2 databases through the replay repository CLI
- maintain an owner/repository dataset index so recordings can be discovered by another developer
- add `dimos mem upload` as the user-facing entry point

## Motivation

This is PR 3 of the #3158 implementation stack. It connects the current local memory2 storage system to the remote repository without requiring users to zip datasets manually.

## Stack dependency

- depends on #3184 and #3177
- review the commit after `c5ba784` for the isolated change

## Validation

- targeted memory2 snapshot, index, and upload tests passed
- SQLite integrity validation is covered
- `ruff check`, format checks, and targeted mypy passed

PR #3184, opened 2026-07-25
Add S3-compatible replay object storage

xiaoyaoqilan:
## Summary

- add an S3-compatible backend for replay artifacts
- support AWS S3, Cloudflare R2, MinIO, and compatible object stores through configurable endpoints and credentials
- preserve content-addressed object IDs and manifest/hash validation
- keep the filesystem backend available for local development and small deployments

## Motivation

This is PR 2 of the #3158 implementation stack. It separates durable cloud object storage from the WebRTC live media path and provides a migration path beyond a single server disk.

## Stack dependency

- depends on #3177
- review the commits after `dca83a5` for the isolated change

## Validation

- targeted object-storage and replay repository tests passed
- `ruff check` and `ruff format --check` passed
- targeted mypy checks passed

PR #3177, opened 2026-07-25
Add WebRTC cross-region and replay repository MVP

xiaoyaoqilan:
## Summary

- add a source-runnable `dimos webrtc-mvp` CLI with `local`, `publish`, and `subscribe` commands
- publish timestamped synthetic frames through a strict H.264 WebRTC path
- report signaling, first-frame, P50/P95 latency, RTP loss, and sequence-gap metrics
- fail explicitly on ICE timeout, missing RTP frames, invalid frame stamps, or unavailable H.264 encoding
- add the foundational replay repository protocol and CLI for raw, manifest-based, and batch upload/download flows
- validate manifests and content hashes on both upload and download
- document local baseline and two-site hosted-SFU testing

## Motivation

This is the transport and repository foundation for #3158. It validates WebRTC/SFU latency and first-frame behavior while providing the smallest usable artifact-sharing path before memory2, remote Go2 replay, object storage, regional selection, and production hardening are layered on top.

## Validation

- targeted pytest suites for WebRTC metrics, replay repository behavior, and repository CLI passed
- `ruff check` and `ruff format --check` passed
- targeted mypy checks passed
- local H.264 WebRTC loopback: 46 decoded frames, 0 invalid timestamps, 0 sequence gaps, 0 packets lost, 82.31 ms media-to-first-frame, 72 ms P50, 87.75 ms P95
- Alibaba Cloud smoke test uploaded and hash-verified a real H.264 MP4 through the repository service

## Follow-up stack

This is PR 1 of the #3158 implementation stack. Follow-ups add S3-compatible storage, memory2 integration, remote `unitree-go2` replay, China/US node selection, and operational hardening.

PR #3187, opened 2026-07-25
Select healthy China and US replay nodes

xiaoyaoqilan:
## Summary

- expose replay node health and region metadata through `/healthz`
- probe configured China and US endpoints concurrently
- select the fastest healthy node for upload/download
- cache the recommendation for one hour, then automatically re-evaluate the network environment
- integrate automatic server selection into the replay repository CLI

## Motivation

This is PR 5 of the #3158 implementation stack. It implements regional auto-detection and hourly dynamic upload recommendations without hard-coding a single China or US endpoint.

## Stack dependency

- depends on #3186, #3185, #3184, and #3177
- review the commit after `aee1f2f` for the isolated change

## Validation

- node health, concurrent probe, latency ordering, failure fallback, and one-hour cache behavior are covered by tests
- `ruff check`, format checks, and targeted mypy passed


xiaoyaoqilan:
Superseded by #3201. The implementation is consolidated into the `dimos/hosted_data` package and unified `dimos data` CLI.

PR #3204, opened 2026-07-27
feat(web): add protocol v2 manifests and reliable streams

paul-nechifor:
- Protocol v2: reliable channels use one persistent stream (Firefox allows about 100 concurrent streams and since Deno sends a late FIN, it breaks)
- Added a new manifest module: `web/shared/manifest.ts` with a Python mirror in `dimos/web/relay_bridge/manifest.py`
- Updated the channel lifecycle. Policies now have `dispose()`, so unsubscribe, re-watch and viewer close drop queued frames and abort the persistent stream instead of leaking it.
- Hardened the Bridge ingress. The receive queue is bounded by bytes as well as frame count, and payloads are capped per encoding using the watched robot's manifest.

### DIFF
--- a/web/relay/server.ts
+++ b/web/relay/server.ts
@@ -4,7 +4,7 @@
 // Session/transport handling lives in session.ts, registration + routing in
 // registry.ts; this file owns the listeners and process-level wiring.
 import { PROTOCOL_VERSION } from "@dimos/shared";
-import { fileURLToPath } from "node:url";
+import { fileURLToPath, pathToFileURL } from "node:url";
 import { makeEphemeralCert } from "./cert.ts";
 import { Registry } from "./registry.ts";
 import { RobotSession, ViewerSession } from "./session.ts";
@@ -20,6 +20,12 @@ export interface RelayOptions {
   host?: string;
   /** Directory served over HTTP. Defaults to ./static next to this module. */
   staticDir?: string;
+  /**
+   * Built Cockpit app (web/cockpit/dist). When set, / serves its index.html
+   * and files resolve here first, with staticDir as the fallback (so
+   * /debug.html keeps working). Without it, / serves the debug page.
+   */
+  cockpitDir?: string;
 }
 
 export interface RelayHandle {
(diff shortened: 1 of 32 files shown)


web/shared/manifest.ts
```diff
@@ -0,0 +1,165 @@
...
+      throw new ManifestError("duplicate_channel_id", `duplicate channel ${spec.ch}`);
+    }
+    chIds.add(spec.ch);
+    if (!boundedId(spec.encoding)) {
+      throw new ManifestError(
+        "invalid_encoding",
+        `encoding must be 1..${MAX_MANIFEST_ID_LEN} chars`,
+      );
+    }
+    if (!Number.isFinite(spec.maxHz) || spec.maxHz <= 0) {
```
mustafab0, line 121: This check occurs on both `manifest.py` and here, and both reject NaN numbers.

But python can parse numbers much bigger than TS handles. 

For some large number bigger than a 64-bit float, will get rejected here but accepted in `manifest.py`

Although in real world no sensor or stream would publish at such a rate. just flagging

web/relay/server.ts
```diff
@@ -40,6 +46,52 @@ const MIME: Record<string, string> = {
   ".png": "image/png",
 };
 
+function resolveDirUrl(dir: string): URL {
+  // Canonical (realPath) so serveFrom compares symlink-free paths (macOS /tmp
+  // is itself a symlink); href must end with "/" so new URL(name, root)
+  // resolves under it.
+  const real = Deno.realPathSync(dir);
+  return pathToFileURL(real.endsWith("/") ? real : real + "/");
```
mustafab0: A typo'd Path silently fails here.

mustafab0 (commented):
Just 2 minor comments. 👍

web/relay/server.ts
paul-nechifor: Fixed.

web/shared/manifest.ts
paul-nechifor: Yeah, it was a bit overzealous to check such things, but they're both the same now.

mustafab0 (approved):

PR #3207, opened 2026-07-27
Shard LCM stream channels across dedicated multicast buses. DIM - 1337

bogwi:
### MacOS

Running any blueprint even a lightweight like this uv run dimos --replay --transport=lcm run unitree-go2 under LCM and on MacOS - makes rerun to stall. That makes impossible to develop in Mac under LCM. 

The changes on this branch fix that.

### Linux

The fix ALSO greatly reduces CPU usage on Linux by 10%! Running the above blueprint on my machine has taken of about ~25% of CPU usage, but now only 15%!

### DIFF
--- a/dimos/protocol/service/lcmservice.py
+++ b/dimos/protocol/service/lcmservice.py
@@ -15,11 +15,14 @@
 from __future__ import annotations
 
 from concurrent.futures import ThreadPoolExecutor
+from ipaddress import IPv4Address
 import os
 import platform
 import threading
 import traceback
 from typing import Any
+from urllib.parse import urlsplit
+import zlib
 
 import lcm as lcm_mod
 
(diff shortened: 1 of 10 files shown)


leshy (changes requested):
This is a good change, had in mind as an LCM escape hatch for a while :D I think we should consider creating a new protocol because this is not an LCM anymore. 

We can call it E-LCM, Efficient LCM, or something like this, in order to keep us LCM compatible with our standard LCM transports. It's viable we'll need to interface with third-party LCM implementations.

dimos/utils/cli/agentspy/demo_agentspy.py
```diff
@@ -26,14 +26,15 @@
...
 
 
 def test_publish_messages() -> None:
     """Publish test messages to verify agentspy is working."""
     print("Starting agent message publisher demo...")
 
-    # Create transport
-    transport = PickleLCM()
+    # Create transport on the same sharded bus agentspy subscribes to
+    transport = PickleLCM(url=lcm_url_for_channel("/agent"))
```
leshy, line 37: You took over decisions on what an URL is, and this is now an protocol level internal variable. users should not be exposed to it at all, URL as a value is effectively killed. We can replace this as a namespace value or something, a number, but it cannot be a URL anymore. (actually yes, "namespace" can effectively be your "port range" - more on this below)

Actual topic setting, a line below, will decide the URL you use, it is not known at the instantiation. This means that one pubsub instance can hold multiple LCM instances per topic on different ports

dimos/protocol/service/lcmservice.py
```diff
@@ -39,6 +42,67 @@
...
+
+    octets = list(address.packed)
+    octets[-1] = ((octets[-1] - 1 + offset) % 254) + 1
+    host = str(IPv4Address(bytes(octets)))
+    port_span = 65535 - 1024 + 1
+    port = ((parsed.port - 1024 + offset) % port_span) + 1024
+    return parsed._replace(netloc=f"{host}:{port}").geturl()
+
+
+_STREAM_BUS_COUNT = max(0, int(os.getenv("LCM_STREAM_BUSES", "16")))
```
leshy, line 71: why so conservative? this is a main knob that gives us efficiency, ideally we want each topic on its own port, I'd actually reserve a port range 50000-60000 or something, Allow user to exclude specific ports or redefine the range, But then just build the deterministic hash function. It takes a topic name and gives you a port number.

dimos/utils/cli/spy/core.py
```diff
@@ -193,17 +193,37 @@ def __init__(self, **lcm_kwargs: Any) -> None:
         # Inline import: an unavailable LCM backend must not break the other
         # spy sources (see default_sources).
         from dimos.protocol.pubsub.impl.lcmpubsub import LCMPubSubBase
+        from dimos.protocol.service.lcmservice import lcm_bus_urls
```
leshy, line 197: If we actually spread each topic per port, spying becomes a problem. I would actually ignore this problem as a first PR, just measure efficiency..

later, since we are implementing this as a separate protocol, we can have a simple meta topic that each instance can use to announce the topic it's on or something similar, can we somehow detect ports in use? and can auto-spy on all of them within our ELCM range

PR #3208, opened 2026-07-27
Fix uv disk persistence

Dreamsorcerer:
It's tricky to test these changes without merging, but the last PR seems to have fixed the redownloading, but still had an issue with uv creating a copy of the entire package set, rather than using it directly. Hopefully this version will fix that part too.


spomichter (approved):

spomichter:
yeah really hard to test just gotta merge and see

PR #3233, opened 2026-07-28
fix(core): resolve remap targets through subclasses

mustafab0:
## Problem

A blueprint can rename a module's port so two modules connect. 

That rename is looked up by class. If a blueprint subclasses the module — which the new
per-robot joint streams need — the lookup misses and the rename never happens. Nothing
errors and nothing warns. You get a G1 that boots fine and ignores every nav and WASD
command.

Issue: #

---

## Solution

If the exact class isn't in the blueprint, look for a subclass of it instead.

Exact match is still tried first. Some modules that get deployed together inherit from
each other (`ObjectDBModule` extends `Detection3DModule`), and there you meant the class
you actually named.

Renaming a module that isn't in the blueprint used to do nothing, quietly. It now raises.

---

## Breaking Changes

None. All 139 importable blueprints resolve identically. `.remappings()` on an absent
module now raises; nothing in-repo does that.

---

## How to Test

1. `pytest dimos/core/coordination/test_module_coordinator.py -k remap`
2. `pytest dimos/core dimos/robot -m 'not (tool or self_hosted or mujoco or self_hosted_large)'`
3. Drop the exact-match preference and `unitree-g1-detection` fails; revert to `is` and the subclass test fails.

### DIFF
--- a/dimos/core/coordination/blueprints.py
+++ b/dimos/core/coordination/blueprints.py
@@ -273,13 +273,26 @@ class Blueprint:
     def _instance_key(self, module: type[ModuleBase] | str) -> str:
         if isinstance(module, str):
             return module
+        # Exact match wins; subclasses are only a fallback. Deployments subclass a
+        # module (e.g. ControlCoordinator) to declare extra ports, and a remapping
+        # named after the base class still means that instance — but sibling modules
+        # also inherit from each other (ObjectDBModule is a Detection3DModule), and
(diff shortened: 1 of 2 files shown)


mustafab0:
After some discussion, it was decided that the check this PR adds is not a problem, as the behaviour would be the same whether using a subclass or a class itself for the module.

This was a fix for a stop-gap migration of the Control coordinator.

PR #3236, opened 2026-07-28
feat(cli): add interactive RPC shell

TomCC7:
## Contribution path

- No tracking issue; the lifecycle rationale and safety boundaries are documented below.

## Problem

DimOS has no general interactive interface for discovering deployed modules and
calling their RPC methods. MCP only exposes skills and requires an `McpServer`,
which makes it a poor fit for low-level debugging.

Attaching during coordinator startup also had a race: discovery published one
ping and waited for that request. If the coordinator subscribed after the
publication, the shell could not recover when the coordinator became ready.

### Shell and introspection APIs

- Add a TTY-only `dimos shell` command backed by IPython.
- Expose human-readable `guide()`, `modules()`, `rpcs()`, and `describe()`
  helpers, plus structured module/RPC metadata through `app`.
- Preserve RPC signatures and docstrings for IPython inspection and support
  exact instance lookup when multiple modules share a class.
- Connect directly to the coordinator; no run-registry entry or `McpServer` is
  required.

These are additive inspection APIs. RPC invocation still uses the existing path
and payload types.

### Lifecycle changes

- **Coordinator service ownership:** `ModuleCoordinator.loop()` now starts the
  coordinator RPC service instead of the two `dimos run` branches. `loop()` is
  the shared entrypoint for foreground, daemon, and direct
  `blueprint.build().loop()` launches, so all three become attachable. Daemon
  startup remains post-fork, shutdown remains in `loop()`'s `finally` block,
  and an `RLock` keeps explicit startup idempotent and serialized with stop.
- **Startup discovery:** `CoordinatorRPC.connect()` retries ping every 250 ms
  within the existing five-second budget. A pub/sub ping sent before the
  coordinator subscribes is lost; retrying the request fixes that race without
  adding an unbounded wait.
- **Timeout cleanup:** `RPCSpec.call_sync()` unsubscribes its local response
  callback on timeout. Retries otherwise leave unreachable callbacks in the
  shared response map. This does not invoke or cancel the remote RPC, and the
  successful-call path is unchanged.

This PR does not change MCP or skill exposure, RPC payload serialization, module
shutdown order, or RPC filtering/confirmation behavior.

## How to Test

Start the test blueprint:

```bash
uv run dimos --viewer none --n-workers 1 run demo-mcp-stress-test
```

In another terminal:

```bash
uv run dimos shell
```

Manual test session:

```text
Attached to run 20260727-221937-demo-mcp-stress-test (demo-mcp-stress-test).

In [1]: guide()
Quick start
  guide()                           Show this guide again
  modules()                         List deployed module instances
  rpcs()                            List every RPC
  rpcs("ModuleName")                Filter RPCs by module
  describe("ModuleName.method")     Show a signature and documentation
  app.ModuleName.method(...)        Invoke an RPC
  app.ModuleName.method?            Inspect an RPC with IPython

In [2]: modules()
┏━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┳━━━━━━┓
┃ Instance         ┃ Class            ┃ RPCs ┃
┡━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━╇━━━━━━┩
│ StressTestModule │ StressTestModule │ 10   │
│ McpServer        │ McpServer        │ 11   │
└──────────────────┴──────────────────┴──────┘

In [3]: rpcs("StressTestModule")
# Printed 10 RPCs with signatures and docstring summaries, including:
# StressTestModule.echo(message: str) -> str
# StressTestModule.ping() -> str

In [4]: describe("StressTestModule.echo")
RPC: StressTestModule.echo
Signature: echo(message: str) -> str

Documentation:
Echo back the given message. Used for latency and connectivity tests.

In [5]: app.StressTestModule.echo("manual shell test")
Out[5]: 'manual shell test'

In [6]: exit
```

The foreground blueprint then stopped cleanly with Ctrl-C.

Automated validation after merging current `main`:

```text
169 passed: shell, cache CLI, porcelain, coordinator lifecycle/discovery,
            and RPC backend regressions
Mypy passed for all 922 source files
Ruff format and lint passed for all relevant Python files
uv lock --check passed
```

## AI assistance

Codex with GPT-5 helped explore the existing coordinator and RPC abstractions,
draft the design, implement the change, write tests and documentation, and run
the automated and manual validation. The author reviewed the design decisions
through an interactive question-and-answer process.

## Checklist


dimos/robot/cli/dimos.py
```diff
@@ -516,6 +471,57 @@ def status() -> None:
...
+        typer.echo(f"Error: {exc}", err=True)
+        raise typer.Exit(1) from exc
+
+    try:
+        typer.echo(_shell_banner())
+        start_ipython(argv=[], user_ns=_shell_namespace(app))  # type: ignore[no-untyped-call]
+    finally:
+        app.stop()
+
+
```
TomCC7: move to dedicated shell command file

dimos/core/coordination/module_coordinator.py
```diff
@@ -632,14 +635,17 @@ def _restart_module(
         return new_proxy
 
     def loop(self) -> None:
-        stop = threading.Event()
+        self.start_rpc_service()
```
TomCC7, line 643: where is this function called originally? why adding here?

dimos/core/coordination/module_coordinator.py
```diff
@@ -632,14 +635,17 @@ def _restart_module(
...
         try:
-            stop.wait()
+            self._wait_until_stopped()
```
TomCC7: should just inline this oneliner

dimos/porcelain/module_source.py
```diff
@@ -24,6 +27,8 @@ class ModuleSource(Protocol):
 
     def list_module_names(self) -> list[str]: ...
 
+    def list_module_descriptors(self) -> list[ModuleDescriptor]: ...
```
TomCC7, line 37: what's the difference between local and remote module source? I think we don';t really have remote module?

dimos/porcelain/dimos.py
```diff
@@ -100,25 +102,110 @@ def restart(self, module_class: type[ModuleBase], *, reload_source: bool = True)
...
+        if module is None:
+            return [rpc_info for module_info in modules for rpc_info in module_info.rpcs]
+
+        instance_name = self._module_instance_name(module)
+        for module_info in modules:
+            if module_info.instance_name == instance_name:
+                return module_info.rpcs
+        raise KeyError(instance_name)
+
+    def describe(self, target: str | Any) -> ModuleInfo | RpcInfo:
```
TomCC7: can you provide more information and kind of pretty print here? like the rpc method's docstring etc

dimos/robot/cli/dimos.py
TomCC7: also can you pretty print the header a bit? also add more instruction on how to use the repl

TomCC7: Moved the command and its tests into dedicated `dimos/robot/cli/shell.py` and `test_shell.py` files.

dimos/core/coordination/module_coordinator.py
TomCC7: The original calls were in both CLI run paths. I removed those duplicates and made `loop()` the single owner of coordinator RPC startup. That also makes direct Python `build().loop()` launches attachable; explicit `start_rpc_service()` remains available for non-blocking/test use.

dimos/core/coordination/module_coordinator.py
TomCC7: Inlined the `threading.Event().wait()` call and updated the lifecycle test.

dimos/porcelain/module_source.py
TomCC7: Local means the coordinator owned in-process by `Dimos.run()`; remote means a client connected to a separately running coordinator through RPC, even when both processes are on the same host. I clarified that distinction in the protocol and implementation docstrings.

dimos/porcelain/dimos.py
TomCC7: Added human-facing helpers: `modules()` prints an instance/class/RPC-count table, `rpcs()` prints qualified names, signatures, and docstring summaries, and `describe()` prints the full signature and documentation. The structured API remains available through `app.list_modules()`, `app.list_rpcs()`, and `app.describe()`.

dimos/robot/cli/dimos.py
TomCC7: Updated the header to reuse the same themed ASCII logo as `humancli` (without the animation), plus attachment details, quick-start examples, IPython inspection syntax, and the RPC safety warning.

dimos/porcelain/dimos.py
TomCC7: should not use any here... just list all possibilities

dimos/porcelain/dimos.py
```diff
@@ -100,25 +102,110 @@ def restart(self, module_class: type[ModuleBase], *, reload_source: bool = True)
...
+            descriptors = source.list_module_descriptors()
+
+        return [_describe_module(descriptor) for descriptor in descriptors]
+
+    def get_module(self, name: str) -> Any:
+        """Return a module proxy by exact instance name or unique class name."""
+        with self._lock:
+            return self._require_source().get_module(name)
+
+    def list_rpcs(self, module: str | Any | None = None) -> list[RpcInfo]:
```
TomCC7: also here, should not add any

dimos/protocol/rpc/spec.py
```diff
@@ -77,6 +77,7 @@ def receive_value(val) -> None:  # type: ignore[no-untyped-def]
 
         unsub_fn = self.call(name, arguments, receive_value)
         if not event.wait(rpc_timeout):
+            unsub_fn()
```
TomCC7, line 82: why is this needed?

dimos/robot/cli/dimos.py
```diff
@@ -419,8 +374,6 @@ def run(
         )
         blueprint = blueprint.disabled_modules(*disabled_classes)
 
-    blueprint = _with_relay_bridge(blueprint)
-
     if show_help:
         print("Blueprint arguments:")
```
TomCC7, line 425: seems like in this branch the loop is not called? can you verify this is safe

paul-nechifor (approved):

PR #3238, opened 2026-07-28
feat: add stateful control task API

TomCC7:
## Contribution path

- Builds on #3183, now merged into `main`

## Problem

Control tasks were originally tick-driven and effectively stateless: `compute()` received the complete `CoordinatorState` needed to produce the next command.

Tasks can now expose RPC commands that run outside `compute()`. For example, `JointTrajectoryTask.execute()` must validate a trajectory against the latest joint state, but it receives no state argument. The previous workaround added trajectory-specific methods to `ControlCoordinator`, making the generic coordinator depend on an optional task and requiring another wrapper for every future stateful task command.

Tasks need a narrow way to fetch the coordinator's latest observation during an RPC command without taking ownership of state collection.

## Solution

- Cache the latest complete, immutable `CoordinatorState` before task filtering.
- Give registered `BaseControlTask` instances transparent access through `self.context.get_state()`.
- Keep `ControlTaskContext` stateless: it exposes a function that reads coordinator-owned state and performs no hardware read or freshness check.
- Tie context access to task registration. Stop/start preserves access; removal or coordinator deletion revokes it; access outside registration raises `RuntimeError`.
- Preserve structural `ControlTask` support without requiring context.
- Route task RPCs through generic `task_invoke(task_name, method, kwargs)`.
- Treat `TASK_EXPOSES` as a strict allowlist, validate arguments, and propagate task results and exceptions unchanged.
- Give `JointTrajectoryTask` the fixed unique name `joint_trajectory`, move start-state validation into the task, and remove trajectory-specific coordinator RPCs and state injection.
- Migrate manipulation execution, coordinator clients, blueprints, documentation, and tests to the generic task-command API.

## Test organization

- Stateful task and registration contracts live in `dimos/control/test_task_context.py`.
- Manipulation execution behavior lives in `test_manipulation_unit.py`; the redundant `test_plan_execution.py` suite is removed.
- Real Drake coverage is named `test_manipulation_drake_integration.py` and keeps one plan-to-dispatch integration boundary instead of duplicating mocked unit behavior.

## Validation

```bash
uv run pytest -q dimos/control dimos/manipulation
uv run mypy \
  dimos/control/task.py \
  dimos/control/coordinator.py \
  dimos/manipulation/manipulation_module.py
uv run ruff check dimos/control dimos/manipulation
git diff --check
```

Local result after rebasing onto `main`: `648 passed, 9 deselected` across
`dimos/control` and `dimos/manipulation`; mypy passed 925 source files.

## AI assistance

Codex with GPT-5 helped design, implement, migrate, and test this change. The author reviewed the design decisions interactively.

## Checklist


dimos/manipulation/test_plan_execution.py
TomCC7, line 1: how is this file different from execution manager can we just merge

dimos/manipulation/test_manipulation_unit.py
TomCC7, line 1: same here how is this different from test manipulation module?

dimos/control/task.py
```diff
@@ -328,12 +351,40 @@ class BaseControlTask(ControlTask):
...
+        """Return the owning coordinator's context.
+
+        A task can be configured before registration, but coordinator services
+        are intentionally unavailable until the task is registered.
+        """
+        if self._context is None:
+            raise RuntimeError(f"Control task {self.name!r} is not bound to a coordinator")
+        return self._context
+
+    def _bind_context(self, context: ControlTaskContext) -> None:
```
TomCC7: can we get a better model than this? I think this can be formalized to some programming pattern?

dimos/control/task.py
```diff
@@ -79,11 +81,17 @@ class JointStateSnapshot:
...
 
-    joint_positions: dict[JointName, float] = field(default_factory=dict)
-    joint_velocities: dict[JointName, float] = field(default_factory=dict)
-    joint_efforts: dict[JointName, float] = field(default_factory=dict)
+    joint_positions: Mapping[JointName, float] = field(default_factory=dict)
+    joint_velocities: Mapping[JointName, float] = field(default_factory=dict)
+    joint_efforts: Mapping[JointName, float] = field(default_factory=dict)
     timestamp: float = 0.0
 
+    def __post_init__(self) -> None:
```
TomCC7: can we use a more elegant solution than this? can check attrs which is added recently

dimos/control/task.py
```diff
@@ -115,10 +123,25 @@ class CoordinatorState:
     """
 
     joints: JointStateSnapshot
-    imu: dict[str, IMUState] = field(default_factory=dict)
+    imu: Mapping[str, IMUState] = field(default_factory=dict)
     t_now: float = 0.0  # Coordinator time (perf_counter) - USE THIS, NOT time.time()!
     dt: float = 0.0  # Time since last tick
 
+    def __post_init__(self) -> None:
```
TomCC7: same as above

dimos/manipulation/test_plan_execution.py
TomCC7: Merged these behavioral cases into `test_manipulation_unit.py` and removed the standalone `test_plan_execution.py`. The execution-manager tests remain focused on the manager boundary.

dimos/control/task.py
TomCC7: Reworked this around a registration-owned lease. The coordinator owns one long-lived stateless context; registering a `BaseControlTask` grants transparent `task.context` access through a weak registration reference. Stop/start preserves the lease, while removal or coordinator deletion revokes it automatically. The lifetime decision is recorded in ADR 0001.

dimos/manipulation/test_manipulation_unit.py
TomCC7: Renamed the real-backend suite to `test_manipulation_drake_integration.py` and narrowed it to Drake integration behavior. Duplicate mocked coordinator-result/state tests now live only in the unit suite; the Drake suite keeps one real plan-to-dispatch boundary test.

dimos/control/task.py
TomCC7: Converted `CoordinatorState` to `@attrs.frozen(slots=False)` as well, with the same immutable mapping converter for `imu`. `ControlTaskContext` is also an attrs frozen class for consistency.

dimos/control/task.py
TomCC7: Converted `JointStateSnapshot` to `@attrs.frozen(slots=False)`. Mapping fields use a shared converter that copies the caller mapping and exposes it through `MappingProxyType`.

Issue #3046, opened 2026-07-18
Stale Unitree movement watchdog can stop a newer command


omarespejel:
#2977 moved the zero-twist publish onto the WebRTC event loop, fixing the cross-thread send.

the watchdog itself still runs in a `threading.Timer`. canceling the timer cannot stop a callback that has already started, so an expired watchdog from the previous move can enqueue zero after a newer command and stop it.

#2643 moves the timer onto the connection loop as part of the broader Unitree connection refactor. if that lands, this case should be covered there. current main still needs the same ordering guarantee.

movement publish, watchdog re-arm/cancel and expiry should share the connection loop. a stale expiry should not affect a newer command.

add a regression test where the first watchdog fires after a replacement command and must not publish zero.

cowsking:
I can pick it up

omarespejel:
@cowsking  are you still working on this? @saisharan0103  that matches the approach I had in mind. I already have a narrow patch locally that moves the watchdog lifecycle onto the connection loop, adds a generation check, and covers the stale-expiry ordering with a deterministic test. Happy to send it if it won’t duplicate existing work

cowsking:
Yes, still on it, planning to push today or tomorrow.

My approach: move the watchdog onto the connection event loop (loop.call_later), with movement publish + re-arm in the same coroutine, and cancel/expiry on the loop as well. Since all four operations are serialized on the loop thread, a cancel is decisive, a stale expiry structurally can't land after a newer command. I ended up not needing a generation token because of that, but happy to compare notes in case the token covers something I'm missing.

I am OK to work on other issues if you have done the code part, let me know your thought, thank you!

omarespejel:
Hey @cowsking, quick check: did you end up opening a PR for this? If not, I still have a deterministic test for the stale expiry and I’m happy to pick it up.

cowsking:
Hi I have opened the PR but feel free to review it, I am also open to add a test for it

omarespejel:
#3243 covers this. The bug still exists on current main. Once refreshed I can test it on Go2 Air.

PR #3243, opened 2026-07-28
fix(unitree): stale movement watchdog can no longer stop a newer command

cowsking:
## Contribution path

- Linked issue or discussion: #3046

## Problem

The auto-stop watchdog runs on a `threading.Timer` thread. `Timer.cancel()` has no effect once the callback has started, so an expired watchdog from a previous `move()` can publish its zero twist after a newer command and halt it. The stale `stop_movement()` also cancels `self.stop_timer`, which by then points at the newer command's timer, so the newer command loses its own watchdog too. The same root cause means a duration move longer than `cmd_vel_timeout` gets a zero twist injected in the middle of its command stream.

## Solution

Move the watchdog onto the connection event loop, as suggested in the issue. `move()` now publishes and re-arms the deadline inside the same coroutine. Expiry and cancellation run on the loop as well. Since the loop serializes publish, re-arm, cancel and expiry, a cancelled deadline can never fire after a replacement command. No locks or generation counter needed. Duration moves re-arm on every publish, so the watchdog no longer interrupts them mid-stream.

## Tests

Four new tests in `test_connection.py`, reusing the existing stubbed driver fixture (no hardware, real event loop). Assertions are made on the recorded wire messages, not internal state.

1. `test_stale_watchdog_expiry_does_not_stop_newer_command`: the regression test requested in the issue. The first expiry callback is gated with a `threading.Event`, so the test holds it in flight while a replacement `move()` is issued from another thread. After releasing the gate, the newer command must be the last message on the wire (the stale zero may only land before it), and the replacement's own watchdog must still fire afterwards.
2. `test_watchdog_zeroes_movement_after_timeout`: baseline. With no follow-up command, the watchdog publishes a zero twist after `cmd_vel_timeout`.
3. `test_stop_movement_disarms_watchdog`: an explicit stop publishes exactly one zero twist and leaves no armed deadline. The deadline check runs on the loop thread, where the handle is owned.
4. `test_duration_move_is_not_zeroed_mid_stream`: a 0.3s duration move with a 0.1s timeout publishes only movement commands until the final stop. This test fails on current main because the old watchdog fires mid-stream.

## How to Test

`uv run pytest dimos/robot/unitree/test_connection.py`

## AI assistance

Claude Code (Fable 5) assisted parts of the implementation and tests. Reviewed and understood line by line.

## Checklist

### DIFF
--- a/dimos/robot/unitree/connection.py
+++ b/dimos/robot/unitree/connection.py
@@ -108,7 +108,9 @@ class UnitreeWebRTCConnection(Resource):
     ) -> None:
         self.ip = ip
         self.mode = mode
-        self.stop_timer: threading.Timer | None = None
+        # Auto-stop watchdog deadline. Armed/cancelled/fired only on self.loop's
+        # thread so a stale expiry can never interleave with a newer command.
+        self._watchdog_handle: asyncio.TimerHandle | None = None
         self.cmd_vel_timeout = 0.2
         self._velocity_api = velocity_api
         self._move_ids = SequentialIds()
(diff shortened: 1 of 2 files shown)


paul-nechifor:
Thanks for the contribution. This looks important. I'll review it soon.

omarespejel:
Hey @cowsking. This is still relevant since the bug remains on current main. Could you refresh it? I can test it on Go2 Air.

PR #3247, opened 2026-07-28
test(memory2): synthetic frames for jpeg codec case — drop 84MB LFS fetch

spomichter:
The jpeg codec round-trip only needs some frames, not real Go2 pixels from the 84MB LFS db — synthetic gradients exercise the same path in ~10ms. Drops the self_hosted mark so the case runs in every CI job.

this test is skipped in CI with LFS which is useless


spomichter:
@greptile

leshy (approved):

Issue #1231, opened 2026-02-11
WIP Task Spec


paul-nechifor:
* I want to introduce a TaskManager and a Task class.

* Task instances will be referenced by their UUID.

* There will be a BaseTask from which custom tasks inherit.

* I will keep returning strings from `@skill`, but it should only be used for tools which return immediately.

* Everything else, will construct a task:

```python
@skill
def start_exploration(self):
    return self.start_task(ExploreTask())
```

* Module.start_task will manage starting the task and give it the thread in which the task will run. The task will run in the same process as the module, but will communicate with the agent when it needs.

* Module.start_task is needed because we will want to encode some logic in there like when a SpeakTask is started, it automatically cancels the previous ones. For some tasks it might be the other way around.

* Task classes declare exclusive_capabilities. E.g. ExploreTask.exclusive_capabilities = ["locomotion"]. No other task can use "locomotion" while one ExploreTask is active. ExploreTask does not reserve "speak" capability because some other task can use it too. However, SpeakTask does reserve "speak" because there shouldn't be two active speak tasks. 

* Tasks should be able to update the agent by themselves in order to give the LLM a chance to intervene. For example. FollowTask can push the message "Lost track for person. trying to reacquire.". If the agent sees it too many times it might ask for a more specific descriptor.

PR #3259, opened 2026-07-29
feat(agents): add Jupyter-backed code policy module

TomCC7:
## Contribution path

- Stacked on #3236

## Problem

DimOS agents can call predefined skills, but they cannot submit a Python program that processes native observations, branches or retries, and composes deployed RPCs as one synchronous policy rollout.

## Solution

- Add a persistent `CodePolicyModule` exposing one synchronous `python_exec(code, timeout_s)` MCP skill.
- Use the standard `jupyter_client` + `ipykernel` stack instead of a custom worker protocol.
- Lazily bootstrap the parent PR's connected `Dimos` app handle and a Memory2 `SqliteStore` attached to the explicitly configured recorder database.
- Preserve the Jupyter namespace across calls and ordinary Python errors.
- On timeout, interrupt first and preserve the namespace when the kernel recovers; restart only an unresponsive or dead kernel.
- Define a private, pure `Recorder` beside the xArm simulation-agent blueprint. It records the enabled joint-state and color-image streams to a stable database path without changing Memory2 or perception.
- Add soft prompt guidance that prefers code execution for observation processing and multi-RPC control flow while retaining direct skills for atomic actions.
- Use standard DimOS structured logs and Jupyter execution identity for submitted source, output, duration, timeouts, interrupts, and restarts.

`ipykernel` and `jupyter-client` are explicit dependencies of the `agents` extra and are imported lazily. The v1 interface is foreground-only, text-only, single-call-at-a-time, and intentionally not a security sandbox. Real-hardware blueprints are unchanged.

## How to Test

Install and run the xArm simulation:

```bash
uv sync --extra agents --extra manipulation
uv run dimos --simulation --viewer none run xarm-perception-sim-agent \
  -o pickandplacemodule.visualization.backend=none
```

From another terminal:

```bash
uv run dimos mcp call python_exec --json-args \
  '{"code":"print(memory.list_streams()); print(app.skills.get_robot_state())"}'
```

Focused automated validation:

```bash
uv run pytest \
  dimos/agents/test_code_policy.py \
  dimos/codebase_checks/test_no_all.py \
  dimos/robot/manipulators/xarm/blueprints/test_agentic.py \
  dimos/robot/test_all_blueprints_generation.py
```

The focused suite passes 19 tests. Ruff, strict mypy over the changed source files, blueprint registry generation, staged pre-commit hooks, and `git diff --check` also pass. A live MCP smoke test verified that `python_exec` can read the latest joint state and color image from the active Recorder database and query bounded observation history. Earlier validation covered DimOS RPC calls, namespace persistence, and interrupt-first timeout recovery.

## AI assistance

OpenAI Codex with GPT-5 was substantially involved in design discussion, implementation, tests, documentation, and validation. The author reviewed and approved the direction interactively.

## Checklist


dimos/agents/code_policy/policy_kernel.py
```diff
@@ -0,0 +1,392 @@
...
+class PolicyKernelConfig(ModuleConfig):
+    output_limit: int = DEFAULT_OUTPUT_LIMIT
+    startup_timeout_s: float = DEFAULT_STARTUP_TIMEOUT_S
+
+
+class PolicyWorkerError(RuntimeError):
+    """The policy subprocess could not complete a protocol operation."""
+
+
+class PolicyWorkerTimeoutError(TimeoutError):
```
TomCC7: no need to add so many special error type

dimos/agents/code_policy/policy_kernel.py
```diff
@@ -0,0 +1,392 @@
...
+    """The policy subprocess did not finish the submitted cell in time."""
+
+
+def resolve_recorder(app: Dimos) -> tuple[str, str]:
```
TomCC7: can just have a config value for the recorder path? really no need to resolve dynamically

dimos/agents/code_policy/policy_kernel.py
```diff
@@ -0,0 +1,392 @@
...
+        return False
+    try:
+        module_path, class_name = info.qualified_path.rsplit(".", 1)
+        candidate = getattr(importlib.import_module(module_path), class_name)
+    except (ImportError, AttributeError, ValueError):
+        return False
+    return isinstance(candidate, type) and issubclass(candidate, Recorder)
+
+
+class KernelProcess:
```
TomCC7: what's the current design shared with the python repl? this seems not the one

TomCC7: and I'm wondering is there some ready to use lib? this seems quite common usecase

dimos/agents/code_policy/policy_kernel.py
```diff
@@ -0,0 +1,392 @@
...
+    def __init__(self, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self._execution_lock = threading.Lock()
+        self._kernel: KernelProcess | None = None
+        self._execution_count = 0
+        self._active_execution: int | None = None
+        self._active_started_at: float | None = None
+        self._kernel_generation = 0
+
+    def __getstate__(self) -> dict[str, Any]:
```
TomCC7: what is this for?

dimos/agents/code_policy/policy_kernel.py
```diff
@@ -0,0 +1,392 @@
...
+                "Policy execution timed out",
+                execution=execution_id,
+                timeout_s=timeout_s,
+            )
+            return (
+                f"Execution {execution_id} timed out after {timeout_s:.1f}s. "
+                "The Python namespace was reset. Remote RPC work may still be running; "
+                "it was not cancelled."
+            )
+        except BaseException as exc:
```
TomCC7: no special handling for the policy worker error?

dimos/agents/code_policy/policy_kernel.py
```diff
@@ -0,0 +1,392 @@
...
+        self._discard_kernel(reason="module stop")
+        super().stop()
+
+    def _ensure_kernel(self) -> KernelProcess:
+        if self._kernel is not None and self._kernel.is_alive():
+            return self._kernel
+        self._discard_kernel(reason="worker unavailable")
+        app = Dimos.connect()
+        try:
+            recorder_name, recording_path = resolve_recorder(app)
```
TomCC7: again if we hardcode recorder we don't need this app

examples/code_policy_xarm_sim.sh
TomCC7, line 1: this file not needed

dimos/agents/code_policy/policy_kernel.py
TomCC7: Addressed in 8fc4a4545. The custom error hierarchy and worker protocol are gone. The module now uses standard Jupyter replies/exceptions and returns a small set of plain-text outcomes.

dimos/agents/code_policy/policy_kernel.py
TomCC7: Addressed in 8fc4a4545. There is no separate PolicyWorker now. Ordinary Python errors preserve the kernel; timeouts interrupt first; only an unresponsive or dead Jupyter kernel is restarted.

dimos/agents/code_policy/policy_kernel.py
TomCC7: Addressed in 8fc4a4545. CodePolicyConfig now takes recording_path directly, and the blueprint passes one explicit stable path to both the recorder and CodePolicyModule. There is no runtime path discovery.

dimos/agents/code_policy/policy_kernel.py
TomCC7: Addressed in 8fc4a4545. The generic Recorder path RPC and dynamic resolver were removed. The blueprint configures the same explicit database path on the recorder and CodePolicyModule.

examples/code_policy_xarm_sim.sh
TomCC7: Addressed in 8fc4a4545. The example shell script was deleted; the documentation now shows the direct blueprint and MCP commands.

dimos/agents/code_policy/policy_kernel.py
TomCC7: Yes; addressed in 8fc4a4545 by using jupyter_client + ipykernel for kernel startup, channels, execution, output capture, interruption, and restart instead of maintaining our own implementation.

dimos/agents/code_policy/policy_kernel.py
TomCC7: Addressed in 8fc4a4545. The custom serialization hooks were removed. Runtime Jupyter objects stay None until the deployed module starts, so normal Module serialization is sufficient.

dimos/agents/code_policy/policy_kernel.py
TomCC7: Addressed in 8fc4a4545. I replaced the custom subprocess protocol with jupyter_client + ipykernel. This keeps the persistent REPL interaction model while relying on the standard kernel implementation; the parent PR still supplies the connected Dimos app API.

dimos/perception/detection/type/detection3d/object.py
```diff
@@ -380,7 +380,7 @@ def aggregate_pointclouds(objects: list[Object]) -> PointCloud2:
         Combined PointCloud2 with all points colored by object (empty if no points).
     """
     if not objects:
-        return PointCloud2(pointcloud=o3d.geometry.PointCloud(), frame_id="", ts=0.0)
```
leshy, line 383: what is this file? this whole file needs to be deleted

dimos/perception/manipulation_policy_recorder.py
```diff
@@ -0,0 +1,58 @@
...
+    WorldBeliefRecorder,
+    WorldBeliefRecorderConfig,
+)
+
+
+class ManipulationPolicyRecorderConfig(WorldBeliefRecorderConfig):
+    timestamp_db_path: bool = False
+
+
+class ManipulationPolicyRecorder(WorldBeliefRecorder):
```
leshy: I think this belongs closer to your blueprints and not in detections or perception

dimos/perception/detection/type/detection3d/object.py
TomCC7: yeah I do feel kind of weird to call something `Object` since we have `object` in python...

dimos/perception/manipulation_policy_recorder.py
TomCC7: Moved the recorder next to the xArm simulation-agent blueprint as a private pure Recorder subclass. It now records only the joint-state and color-image streams that the demo already publishes. I also removed the perception recorder, its tests, and the related perception changes.

TomCC7:
cleaned up, can check again @leshy

TomCC7:
Linear: [DIM-1366 — Add Jupyter-backed code policy interface for agents](https://linear.app/dimensional/issue/DIM-1366/add-jupyter-backed-code-policy-interface-for-agents)

leshy (approved):

dimos/agents/code_policy.py
```diff
@@ -0,0 +1,335 @@
...
+    state = "completed" if status == "ok" else "failed"
+    body = output.rstrip()
+    if not body and status != "ok":
+        body = f"{content.get('ename', 'Error')}: {content.get('evalue', '')}".rstrip()
+    if not body:
+        body = "(completed)"
+    return f"In [{execution_count}] {state} in {duration_s:.2f}s\n\n{body}"
+
+
+code_policy_module = CodePolicyModule.blueprint
```
paul-nechifor, line 335: What is this?

dimos/agents/code_policy.py
```diff
@@ -0,0 +1,335 @@
...
+        self._kernel_manager: Any = None
+        self._kernel_client: Any = None
+
+    @rpc
+    def start(self) -> None:
+        self._execution_lock = threading.Lock()
+        super().start()
+
+    @skill
+    def python_exec(self, code: str, timeout_s: float = MAX_EXECUTION_TIMEOUT_S) -> str:
```
paul-nechifor, line 136: Note that this cannot exceed the RPC timeout, which I think is 30 seconds. You might want to extend that for this call. I think Jeff added a way to extend RPC timeouts for particular methods.

dimos/agents/code_policy.py
```diff
@@ -0,0 +1,335 @@
...
+
+class CodePolicyModule(Module):
+    """Execute trusted agent-authored Python against a running DimOS system."""
+
+    config: CodePolicyConfig
+
+    def __init__(self, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        # Runtime objects remain None until this instance has been deployed into
+        # its worker, so Module serialization needs no custom hooks.
```
paul-nechifor, line 125: I'm not sure what this means... Why not just do:

```
self._execution_lock: threading.Lock = threading.Lock()
```

here?

dimos/agents/code_policy.py
```diff
@@ -0,0 +1,335 @@
...
+        """
+        if not 0 < timeout_s <= MAX_EXECUTION_TIMEOUT_S:
+            return (
+                f"Invalid timeout_s={timeout_s!r}; expected a value in "
+                f"(0, {MAX_EXECUTION_TIMEOUT_S:g}]"
+            )
+
+        lock = self._execution_lock
+        if lock is None:
+            lock = self._execution_lock = threading.Lock()
```
paul-nechifor, line 156: This is quite odd.

If you're mutating `self._execution_lock` from different threads, you need a lock for it... but it's odd to need a lock for a lock.

Of course `self._execution_lock` won't be none in practice because self.start() initializes it. But you can avoid all of this by just initializing it in `__init__`.

dimos/agents/code_policy.py
```diff
@@ -0,0 +1,335 @@
...
+
+app = _Dimos.connect()
+memory = _SqliteStore(path=_os.environ[{_RECORDING_PATH_ENV!r}], must_exist=True)
+memory.start()
+
+del _os, _SqliteStore, _Dimos
+"""
+
+
+class CodePolicyModule(Module):
```
paul-nechifor, line 117: Hmm, the architecture is a bit odd.

You have to run `dimos run ...`, which spawns several processes, one of which runs CodePolicyModule. CodePolicyModule spawns a Jupyter python process which runs `Dimos.connect()` which connects to the `dimos run...` process. A bit circular.

dimos/agents/code_policy.py
```diff
@@ -0,0 +1,335 @@
...
+    """Execute trusted agent-authored Python against a running DimOS system."""
```
paul-nechifor, line 118: I haven't read the full paper, but I think a key part is being able to generate code when an undefined function is called.

> Central to our approach is prompting hierarchical code-gen (recursively defining undefined functions),

Image
Image

TomCC7: The original code as policy paper is in 2022, which is even pre-chatgpt era (1000 years ago in AI lol). So you might want to check this instead: https://research.nvidia.com/labs/gear/aspire/

And there is indeed a naming issue of this module, it's only a small part of the code-as-policy flow.

dimos/agents/code_policy.py
TomCC7: Agree. You actually raised a very good opportunity to do separation here. We can actually just write a script to expose a single execute_python mcp skill to agent so agent won't get confused by all other mcp skills. I'll check if this is better

spomichter:
pretty cool feature

Jerrybery:
Really nice direction — a persistent, agent-authored Python action space is a big unlock beyond one-shot skill calls. I ran the branch on a sim xArm setup (minimal blueprint + an OpenAI-compatible LLM backend) and the core loop works well: one `python_exec` call could command joints via RPC, poll `memory.streams["coordinator_joint_state"]` for convergence, and replan — the ASPIRE-style execution backend is real. A few thoughts from the code-as-policy literature, plus what the hands-on run surfaced.

**On the naming / scope discussion (paul-nechifor's point about hierarchical code-gen):** agreed this module is not full CaP — but I'd argue hierarchical code-gen is not the part of the 2022 paper worth chasing. Two more recent data points reframe where the leverage is:

- **CaP-X (2026)** systematically ablated the CaP stack and found most of the performance comes from the hand-designed primitive scaffolding, not the model — stripping abstraction layers drops closed-model success from 57% to 18%. The implication for this PR: the real asset is the *primitive surface* exposed to the agent (`app`, `memory`, RPC discovery), and this PR gets that right. Naming it after the execution primitive rather than the full CaP flow seems fine.
- **ASPIRE (NVIDIA GEAR, 2026)** — which TomCC7 already linked — shows the large wins come from the *closed loop around execution*, not one-shot synthesis: per-primitive multimodal traces → targeted repair → distilled skill library. Their execution engine alone moved LIBERO-Pro macro success 14% → 62%, and the accumulated skill library transferred zero-shot to long-horizon tasks (31% vs ~4%). From that perspective this module is best framed as the **execution backend** of such a loop — and the loop is currently missing its other half.

**What the loop's other half would look like (v2 candidates, non-blocking):**

1. **Structured per-call traces.** Today the agent gets a bounded text transcript, and a single `python_exec` can contain several RPCs plus a polling loop — when it fails, neither the agent nor the logs can localize *which* call failed. I confirmed the structured log only records `duration_s / execution_count / output` per execution; there is no per-RPC record. ASPIRE's biggest single win was giving the debugging agent structured, per-primitive evidence (return codes + observation snapshots) instead of raw output. A per-RPC trace record — args, result/error, timestamps, memory cursor — would double as eval data, and the existing structured logging feels like the natural place to hang it.

2. **Persistence of verified programs.** The kernel namespace dies with the run, so every run starts from zero — the exact failure mode ASPIRE's skill library solves. A Voyager/ASPIRE-style pattern — save successful programs as named, callable skills retrievable in later sessions — would compound value across runs. memory2 feels like the natural store for it.

3. **Surface the `app` RPC interface to the agent.** Cheapest item on this list, and the one my run actually tripped on: the agent sees the MCP tool schemas but has no visibility into what `app` offers inside the kernel. Mine hallucinated `app.robot.move_to_joints(...)` in a retry branch (real name: `app.ManipulationModule`, different arg shape) — it only survived because the retry path never executed. Injecting the available module names + RPC signatures into the kernel bootstrap preamble or the `python_exec` tool description would eliminate this whole class of errors.

4. **Timeout shape.** Echoing paul-nechifor: `MAX_EXECUTION_TIMEOUT_S = 110` hugs the default RPC timeout, and a long rollout with several blocking RPCs could exceed it. Longer term, an async task handle (submit → poll/notify, along the lines of the Task model in #1231) may fit better than stretching a synchronous call — it also composes with cancellation (#3379).

5. **Trust boundary.** The docs are refreshingly explicit (v1 = sim-only, not a sandbox). When this moves toward real hardware, the existing `CapabilityRegistry` / `uses=` arbitration might be a natural gate for which RPCs submitted code may call (read-only observation vs motion).

Related work worth a look if you haven't: **CodeAct** (Wang et al., ICML 2024, [arXiv:2402.01030](https://arxiv.org/abs/2402.01030)) — evidence that code as the action format outperforms JSON/text tool calls for multi-step control flow in agents, which is exactly the bet this PR makes; and **Graph-as-Policy** (2026) for a contrasting persistent-execution representation.

Two operational notes from my run: (a) while testing I hit an unrelated daemon-mode Zenoh bug that makes the kernel bootstrap fail with "No running DimOS coordinator found" under `--daemon` — filed separately as #3395 with a root-cause analysis; anyone reproducing this branch in daemon mode will hit it. (b) On a headless box the sim module needs `headless=True` (the passive viewer's GLFW init fails without DISPLAY and the adapter then times out waiting for ready) — might be worth a docs line.

Happy to prototype the structured-trace piece against this branch if there's interest.

TomCC7:
closed in lieu of #3378

PR #3263, opened 2026-07-29
feat(cockpit): Video Panel

paul-nechifor:
* Added the VideoPanel which shows the `color_image` published by the bridge (`jpeg.v1` encoded). It also displays FPS and staleness.
* Fixed a but related to the Mujoco process. Its output pipe was never drained and so it would periodically stop publishing data.
* Only encode and send when there's a listener.

Closes DIM-1167


dimos/utils/turbojpeg.py
```diff
@@ -0,0 +1,26 @@
...
+
+from __future__ import annotations
+
+import functools
+
+from turbojpeg import TurboJPEG
+
+
+@functools.cache
+def get_turbojpeg() -> TurboJPEG:
```
leshy, line 24: why :D

paul-nechifor: So it's cached.

mustafab0 (approved):

PR #3268, opened 2026-07-29
feat(teleop): route operator audio to Go2

ruthwikdasyam:
## Problem

Hosted Go2 teleoperation receives video and control data from the operator, but it does not route the operator's microphone audio to the robot speaker. The Go2 Air also lacks speaker hardware, so the integration must avoid repeatedly calling unsupported firmware APIs.

Closes DIM-1243

## Solution

- Add a subscribe-only WebRTC audio transport that shares the existing hosted broker provider.
- Decode operator audio to PCM and bridge it to the Go2 audio-hub megaphone API.
- Resample and mix incoming audio, normalize quiet speech, batch uploads, bound queued latency, and exit megaphone mode after idle or failed uploads.
- Support `auto`, `enabled`, and `disabled` speaker modes so deployments can explicitly handle Go2 Pro and Go2 Air hardware.
- Wire the bridge into the hosted Go2 Cloudflare blueprint without changing the existing video or command session.

## How to Test

`dimos run teleop-hosted-go2-transport -o go2audiobridgemodule.speaker=enabled`
Login at teleop.dimensionalos.com connect to the robot,  Turn Audio ON, and heard at robot's end

## AI assistance

OpenCode with GPT-5.6 was used to review the implementation, address lifecycle cleanup, and verify the change. I reviewed and understands the resulting code.

## Checklist


dimos/teleop/hosted/go2_audio_bridge.py
```diff
@@ -0,0 +1,271 @@
...
+            if not self._megaphone_active:
+                self._request(ENTER_MEGAPHONE)
+                self._megaphone_active = True
+                if self.config.megaphone_enter_delay_sec > 0:
+                    time.sleep(self.config.megaphone_enter_delay_sec)
+            wav_data = self._wav_bytes(pcm)
+            self._upload_wav(wav_data)
+        except Exception:
+            logger.warning("Go2 megaphone upload failed", exc_info=True)
+            self._megaphone_active = False
```
ruthwikdasyam: ⚠️ If `_upload_wav()` fails after `ENTER_MEGAPHONE` succeeded, this clears `_megaphone_active` without sending `EXIT_MEGAPHONE`. From then on `stop()`/idle cleanup will no-op because `_exit_megaphone()` sees `_megaphone_active == False`, so the Go2 firmware can be left in megaphone mode after a partial upload/network failure. Consider attempting `_request(EXIT_MEGAPHONE)` (best-effort) before clearing the flag, or call `_exit_megaphone()` from this exception path while the flag is still true.

ruthwikdasyam (commented):
## Hermes Agent Review

Found 1 issue worth fixing before marking this ready. CI is green, and the audio transport/blueprint wiring otherwise looks coherent.

### ⚠️ Warning
- `dimos/teleop/hosted/go2_audio_bridge.py:170` — failed uploads clear `_megaphone_active` without sending `EXIT_MEGAPHONE`, so the firmware can be left in megaphone mode after a partial upload failure.

### ✅ Looks good
- Broker provider singleton config no longer forks on audio-only options.
- Operator-audio transport is subscribe-only and starts the shared provider lazily.
- Tests cover config override, speaker auto-disable, resampling/mixing, queue drop policy, batching, and response-code handling.

Note: I reviewed via GitHub API/diff because the local clone checkout hit filesystem timeout on this large repo; all GitHub checks are passing.

dimos/core/test_webrtc_audio_transport.py
```diff
@@ -0,0 +1,41 @@
...
+    unsubscribe = transport.subscribe(received.append)
+    audio_callback = provider.set_audio_frame_callback.call_args.args[0]
+    audio_callback(np.array([1, -2, 3], dtype=np.int16).tobytes(), 48000, 1)
+
+    assert len(received) == 1
+    assert received[0].sample_rate == 48000
+    assert received[0].channels == 1
+    np.testing.assert_array_equal(received[0].data, np.array([1, -2, 3], dtype=np.int16))
+
+    unsubscribe()
```
paul-nechifor: Unsubscribe should happen even on test failure.

paul-nechifor (approved):

PR #3272, opened 2026-07-29
unitree-go2 efficiency review

leshy:
Main go2 is not amazing, my laptop thermally throttles eventually and then this explodes to load 50

This branch fixes it:

- reconfigured a bunch of libs to limit number of threads (they normally enumerate cpu N to create a pool, but their imports leaked and would create a pool before forking, which quickly oversubscribes the system, number of total threads in dimos was 1k+

- added a test to forbid non-inline imports of o3d, opencv, rerun (random rerun import alone before forking was 800 threads)

- fixed zenoh autodiscovery blockage, by default running local only

- added zenoh system configurator for (previusly failing) implicit SHM transport

- not important but added a more efficient cpu only voxelizer (4x worse, not 20x worse then cuda, like previously)


dimos/robot/unitree/go2/connection.py
```diff
@@ -500,3 +501,4 @@ def observe(self) -> Image | None:
         Returns None if no frame has been captured yet.
         """
         return self._latest_video_frame
+        return self._latest_video_frame
```
paul-nechifor:
```suggestion
```

Looks like a duplication.

dimos/msgs/sensor_msgs/Image.py
```diff
@@ -679,3 +699,4 @@ def _parse_lcm_encoding(enc: str) -> tuple[ImageFormat, type, int]:
     if enc not in m:
         raise ValueError(f"Unsupported encoding: {enc}")
     return m[enc]
+    return m[enc]
```
paul-nechifor:
```suggestion
```

paul-nechifor (dismissed):

spomichter (dismissed):

paul-nechifor (approved):

PR #3277, opened 2026-07-29
Migrate blob consumers of `coordinaotor_joint_state` topic to per robot joint topic

mustafab0:
## Problem

The control coordinator publishes one message with every robot's joints stitched
together. Anything that wants one robot's joints has to take the whole thing.

That merged message is being retired. Five places still read it, and all five only care
about a single robot.

Issue: #

---

## Solution

Each coordinator now also publishes one stream per robot, and those five readers
subscribe to the one they need.

Two halves per deployment: add the output on the coordinator, point the reader at it.
One commit per reader.

Recorder modules are untouched, so recordings keep their old stream names and DataPrep
configs still work.

The merged message still publishes. Retiring it is a later PR, as is the manipulation
module.

---

## Breaking Changes

None. The go2 controllers keep the same LCM topic, so Foxglove and lcm-spy are unaffected.

---

## How to Test

1. `pytest dimos/control dimos/imitation dimos/perception dimos/robot -m 'not (tool or self_hosted or mujoco or self_hosted_large)'`
2. `dimos --simulation mujoco run unitree-g1-groot-wbc` — G1 still animates in Rerun, nav still drives it.
3. None of the five deployments run in CI. Owner runs on hardware:
   - `learning-collect-quest-piper` — record an episode, DataPrep still builds it
   - `xarm6-worldbelief` — recording still has `coordinator_joint_state`
   - `unitree-g1-coordinator` + `g1_replay.py --dry-run` — picks up current pose
   - `unitree-go2-holonomic-controller` and `-rpp-controller` — tracking unchanged

### DIFF
--- a/dimos/imitation/collection/blueprint.py
+++ b/dimos/imitation/collection/blueprint.py
@@ -49,15 +49,19 @@ def _camera_if_real() -> tuple[Blueprint, ...]:
     return (RealSenseCamera.blueprint(enable_pointcloud=False),)
 
 
-# buttons / color_image / joint_state / status are left to autoconnect — each
-# name is unique across the composed blueprint, so it resolves to a stable
-# /<name> topic shared by producer and recorder.
+# buttons / color_image / status are left to autoconnect — each name is unique
+# across the composed blueprint, so it resolves to a stable /<name> topic shared
+# by producer and recorder. The joint stream is remapped onto the coordinator's
+# per-robot `arm_joints` output; the recorder still writes it to the db under its
+# port name (`coordinator_joint_state`), which is what DataPrep reads.
+_JOINTS_FROM_ARM = [(CollectionRecorder, "coordinator_joint_state", "arm_joints")]
+
 learning_collect_quest_xarm7 = autoconnect(
     teleop_quest_xarm7,
     *_camera_if_real(),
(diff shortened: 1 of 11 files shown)


dimos/imitation/collection/blueprint.py
```diff
@@ -49,20 +49,24 @@ def _camera_if_real() -> tuple[Blueprint, ...]:
...
 
-# buttons / color_image / joint_state / status are left to autoconnect — each
-# name is unique across the composed blueprint, so it resolves to a stable
-# /<name> topic shared by producer and recorder.
+# buttons / color_image / status are left to autoconnect — each name is unique
+# across the composed blueprint, so it resolves to a stable /<name> topic shared
+# by producer and recorder. The joint stream is remapped onto the coordinator's
+# per-robot `arm_joints` output; the recorder still writes it to the db under its
+# port name (`coordinator_joint_state`), which is what DataPrep reads.
+_JOINTS_FROM_ARM = [(CollectionRecorder, "coordinator_joint_state", "arm_joints")]
```
mustafab0, line 57: In a follow up PR I am going to retire the blob `coordinator_joint_state`

and this remapping is going to be deprecated.

paul-nechifor (dismissed):

paul-nechifor (approved):

PR #3292, opened 2026-07-30
fix CI crash log

leshy:


.github/workflows/ci.yml
```diff
@@ -385,6 +392,34 @@ jobs:
...
+          dir="${DIMOS_CRASH_DIR:-${RUNNER_TEMP}/pytest-crash}"
+          # Runs that never crashed leave empty files; drop them so the artifact
+          # is either meaningful or absent.
+          find "$dir" -type f -empty -delete 2>/dev/null || true
+          shopt -s nullglob
+          files=("$dir"/*.log)
+          if [ ${#files[@]} -eq 0 ]; then
+            echo "No crash dumps: this failure was not a hard crash (segfault/abort)."
+            exit 0
+          fi
```
Dreamsorcerer: Do we care if it logs something when there's empty files? We'll only look at it if there's an error anyway.

.github/workflows/ci.yml
```diff
@@ -385,6 +392,34 @@ jobs:
...
+          for f in "${files[@]}"; do
+            echo "::group::$(basename "$f")"
+            cat "$f"
+            echo "::endgroup::"
+          done
+      - name: Upload crash diagnostics
```
Dreamsorcerer: This is just uploading the contents we just logged, right? Why would we need both?

dimos/conftest.py
```diff
@@ -101,7 +104,71 @@ def _is_macos() -> bool:
...
+        return
+    if _crash_log is None:
+        base = os.environ.get("DIMOS_CRASH_DIR") or os.path.join(
+            tempfile.gettempdir(), "pytest-crash"
+        )
+        worker = os.environ.get("PYTEST_XDIST_WORKER", "main")
+        try:
+            directory = pathlib.Path(base)
+            directory.mkdir(parents=True, exist_ok=True)
+            _crash_log = open(directory / f"crash-{worker}-{os.getpid()}.log", "w", buffering=1)
```
Dreamsorcerer:
```suggestion
        log_path = pathlib.Path(base) / f"crash-{worker}-{os.getpid()}.log"
        try:
            log_path.parent.mkdir(parents=True, exist_ok=True)
            _crash_log = log_path.open("w", buffering=1)
```

.github/workflows/ci.yml
leshy: I wonder the same :D

Dreamsorcerer (dismissed):

Dreamsorcerer (dismissed):

dimos/conftest.py
```diff
@@ -101,6 +104,52 @@ def _is_macos() -> bool:
     return platform.system() == "Darwin"
 
 
+# --- crash diagnostics -------------------------------------------------------
```
Dreamsorcerer:
```suggestion
```

Dreamsorcerer (approved):

Issue #3290, opened 2026-07-30
Replay-derived evals for navigate_with_text goal selection


mc856:
### What do you want, and why?

the only behavior-level check on spatial memory today is e2e tests with hand-measured coordinates (`test_dimsim_spatial_memory.py` pins the bed at `(-3.567, -1.332)`). every new question is a manual measurement, and after a prompt or model change there's no repeatable way to tell whether object finding got better or worse.

i built a way to derive the questions and reference positions from a recording the robot already made, and score the goal the shipping `navigate_with_text` selects — execution stubbed:

- **teacher** (offline, once per recording): sample RGB on recorded timestamps, skip near-black frames, open-vocab detect (the in-repo YOLOe), project the nearest lidar sweep through `pose ∘ BASE_TO_OPTICAL` + fisheye intrinsics, reject ground points, take the depth-IQR inliers' map-frame median as the candidate position. a label qualifies only with ≥2 pairwise-independent views (the robot actually moved between them) and exactly one qualifying cluster; qualified labels sitting on the same spot get linked so one physical object can't become several questions.
- names are the weak half of open-vocab on a dim replay, so a committed human-review overlay (verified / renamed / dropped, with reasons and the crops) gates what becomes a question. `go2_bigoffice` yields 26 geometrically qualified labels on 19 locations; review keeps the single-digit subset that survives as askable questions.
- **harness**: fresh coordinator per question, shipping `SpatialMemory` attached read-only, goal + query strings captured from the worker processes, six-state outcome so harness failures can't masquerade as agent failures, one figure per sweep (error distribution per model × system-prompt configuration).

status: implemented and green on a branch — one new directory (`dimos/agents/evals/`), zero existing-file changes, roughly half of it tests; the unit layer runs on the default lane, recording-backed layers are `self_hosted`. related: #3069 (this exercises the in-house agent path), #1913 (the point-placement half of it), #3249 covers the complementary authored-scene closed-loop side in dimsim, #3154 applies the same replay-derived-reference idea to the MLS planner. the success contract mirrors dimsim's `{passed, reason, score}`.

three questions before i open the PR:

1. one PR, or split (offline reference pipeline vs harness+tests)?
2. is `dimos/agents/evals/` the right home? followed the colocation precedent of `nav_3d/evaluator` and `control/benchmarking`
3. sample results in the PR: a keyless pipeline-validation run (recorded model transcript) now with live sweep numbers as a follow-up comment, or hold for live numbers?

[...]

PR #3249, opened 2026-07-28
feat(dimsim): add closed-loop agent evaluations

michaellee8:
## Contribution path

- Linked issue or discussion: #3069

Related to #3069.

## Problem

DimSim workflows could score browser state, but they could not run a
deterministic closed-loop evaluation against a live DimOS agent. There was no
correlated reset → dispatch → scoring lifecycle, authoritative server-physics
reset, bounded infrastructure failure handling, or task-scoped way to validate
an agent declaration.

This also allowed `find-and-go-to-bathtub` to produce false positives: the agent
could claim it found the bathtub without being physically close to it.

## Solution

Add an explicit single-workflow agent mode:

```bash
dimsim eval apartment/find-and-go-to-bathtub --agent
```

The implementation:

- Adds a
  `runEval → evalReady → evalReset → resetAck → agent_send → evalStart → evalResult`
  protocol correlated by `runId`.
- Resets authoritative Rapier position, yaw, velocity, and command state before
  dispatch.
- Dispatches the workflow task exactly once through the existing MCP
  `agent_send` tool.
- Starts scoring only after reset and successful dispatch.
- Adds bounded timeouts and classifies task failures separately from
  infrastructure errors.
- Performs terminal cleanup by stopping exploration, perception watching, and
  navigation and by clearing bridge velocity.
- Adds JSON/JUnit reporting and exit codes for pass, task failure, and
  infrastructure failure.
- Adds apartment bathtub workflows, including an outside-start discovery task.
- Adds a read-only `/agent` output sidecar for the discovery workflow.
- Requires `find-and-go-to-bathtub` to satisfy both:
  - a tool-free assistant response exactly equal to `FOUND_BATHTUB`
  - a final distance of at most 1 metre from the bathtub
- Keeps model credentials and model configuration entirely in DimOS.
- Adds Deno, browser-harness, bridge-physics, reporting, rubric, and Python
  sidecar tests.
- Runs the DimSim Deno suite in CI.

The branch also improves the live DimSim runtime by:

- forwarding and deduplicating odometry, LiDAR, and video through
  `DimSimConnection`
- rejecting low-similarity semantic-navigation matches
- keeping the visual-language model alive across perception-watcher cycles

No shared system prompt is changed. The `FOUND_BATHTUB` instruction belongs only
to the relevant workflow.

## How to Test

Start the live stack:

```bash
dimos --transport lcm --simulation dimsim \
  --dimsim-scene apartment \
  --dimsim-port 8092 \
[...]

Issue #1913, opened 2026-04-24
Investigate Agentic understanding of space


leshy:
if memory2 query returns an occupancygrid (or pointcloud?)  how can we present it to a LLM model in a way in which it understands it?

write an eval, ask basic questions like

## spatial understanding
model will guess - so we need a lot of very specific questions

"how many rooms there are in total"
"how big is that room down the hallway to the right?"
"where is the biggest room in this space"
"what type of space is this"

## point placement

"where should we explore the space next?"
"place a point on a biggest/smallest room"
"place points marking hallways"

some (failed) attempt here of just encoding occupancygrid as a picture for VLM https://github.com/dimensionalOS/dimos/pull/822

PR #3307, opened 2026-07-31
feat(agents): replay-backed spatial goal-selection evals

mc856:
## Contribution path

- Linked issue or discussion: #3290 (proposal issue)
- Related to #3069

## Problem

Behavior-level spatial evaluation today is an e2e test with hand-measured hardcoded coordinates (`dimos/e2e_tests/test_dimsim_spatial_memory.py`). Adding a question means measuring by hand, and after a prompt or model change there's no repeatable way to tell whether object-finding got better or worse.

## Solution

An offline "teacher" derives object reference locations from a recording the robot already made — RGB semantics from an open-vocabulary detector, geometry from time-aligned LiDAR and pose. A pytest harness then scores the goal the shipping `navigate_with_text` path selects, with navigation execution stubbed. Everything lives in one new directory (`dimos/agents/evals/`, no existing file changes), following the landed precedent of evals colocated with the subsystem they test (`dimos/navigation/nav_3d/evaluator/`, `dimos/control/benchmarking/`).

- **`passed` means "the robot ends up where the object is observable."** `navigate_with_text` builds its goal from the retrieved frame's robot pose (`_get_goal_pose_from_result`, `dimos/agents/skills/navigation.py:267`) — a viewpoint, not an object position — so each question's threshold is its own nearest teacher-viewpoint distance + 0.5 m, capped at 3.5 m, and the fine-grained signal lives in the continuous `error_m`. Every record also carries the retrieved frame's pose and match distance (post-hoc, diagnostic only) — the observable that instance-level scoring needs next.
- **The question yield is deliberately strict**: 1040 sampled frames → 26 qualified labels → 3 questions on `go2_bigoffice`; 214 → 14 → 3 on `go2_short`. The generator refuses any question unpassable on the teacher's own evidence and any pair whose pass radius contains another question's viewpoints — in a dense aisle a wrong-object retrieval would otherwise score as a pass. Human review cuts the rest. Full funnels and location-group tables ship in the manifests; per-label verdicts and confusability distances live in the review overlay.
- **References are estimates, not ground truth.** Every detector label passes a committed human-review overlay (`reference/<dataset>/review.json`: verified / renamed / dropped, with reasons), reviewed on rendered crop pairs — the measurement frame plus a sharper neighboring frame for identification. This layer caught real failures: a moving object, and one reference whose point-cloud median had been pulled onto a blank wall by bbox contamination — it survived single-crop review twice and only fell to multi-frame context, which is why the crop tooling renders both frames.
- **One directory per recording.** `reference/<dataset>/` holds refs, manifest, review, questions and crops; the sweep discovers datasets by globbing `*/questions.jsonl`, so adding a recording is adding a directory. The second one (`go2_short`, the repo's default recording) went through the identical protocol, and an independent rerun reproduced its `refs.jsonl` byte-identically.

The eval can't be passed without the memory actually working: goals only come from the shipping retrieval path, "no prediction" is a scored outcome, and broken measurements (harness/timeout/tool errors) are excluded from the rates and reported separately.

### Review evidence

The five publishable identification/measurement pairs, one column per question — top row: the sharpest supported neighboring frame with the derived reference (magenta) and LiDAR inliers (yellow); bottom row: the actual detection frame with the detection box (green) and in-box inliers. Rendered by `teacher.py --crops` from the recordings and composited into one sheet: program output, not a hand- or AI-drawn diagram. The pairs live in this description rather than the tree — `*.jpg` is LFS-routed and the project's LFS endpoint doesn't take anonymous uploads from outside contributors. The sixth question (`go2_short` / office-desk) has no publishable frames: every candidate contains an identifiable person.

image

## Relationship to existing eval work in the tree

`misc/DimSim/evals/` is the repo's documented eval authoring surface for authored sim scenes, and #3249 extends it with closed-loop agent runs. This PR is the complementary boundary: recorded real-sensor data instead of authored scenes, derived references instead of author-placed objects, goal selection instead of closed-loop execution. Result records mirror DimSim's `EvalSuccess` shape (`{passed, reason, score}`) with one semantic difference: DimSim's `score` is a continuous lower-is-better distance, here it's binary and the distance lives in `error_m` — the two shouldn't be aggregated blindly. #2989 (video spatial-relation QA) sits on a different data plane and subject. Conceptually this is the position-question base layer that #1913's "point placement" line would sit on.

## What's in the diff

Offline tools (CLI, run once per recording):

- `teacher.py` — samples RGB on recorded timestamps (skipping near-black frames), runs open-vocabulary detection, projects the nearest LiDAR sweep with `pose ∘ BASE_TO_OPTICAL` + fisheye intrinsics, gates the points (ground rejection, point count, depth IQR, observation envelope), clusters per label, and qualifies a cluster only on views with ≥ 0.3 m of real spatial baseline between them. A label survives only if exactly one of its clusters qualifies; surviving labels on one spot are linked into a `location_group`. Every threshold is a flag echoed into `manifest.json` alongside the funnel, location-group table, camera/extrinsics, detector settings and timing; the same recording and flags reproduce `refs.jsonl` byte-identically on one machine.
- `questions.py` — applies the human-review overlay (silence means dropped); refuses duplicate names, two questions on one location group, unpassable questions, confusable pairs.
- `ingest.py` — fills the SpatialMemory store the eval answers from, backs the saved pickle up read-only.

Harness and scoring:

- `contracts.py` — `QuestionSpec` / `AnswerRecord` / `ScoreResult`, the six-state outcome, `RetrievalRecord` (the post-hoc retrieval observable).
- `testing/modules.py` — observation-only modules: goal-recording navigation stub, a skill container that records the exact `query` strings before delegating to the unwrapped shipping body (a parity test pins its skill metadata to the shipping decorator), fake sensor publishers, a test runner with configurable waits.
- `conftest.py` — one question, one fresh coordinator, always torn down; the shipping `SpatialMemory` attached read-only with per-run disposable artifacts; a store-identity guard fails fast on a missing or empty collection. Deployed tools are navigation + server admin only (`navigate_with_text`, `stop_navigation`, `tag_location`, `agent_send`, `list_modules`, `server_status`) — narrower than a full robot blueprint, which matters when reading the tool-routing numbers.
- `scorer.py` — pure functions; rates divide by agent-attributable outcomes only, broken counts reported alongside; every record carries `run_id` (duplicates within one run are rejected).
- `render.py` — shard directory → one figure (aggregating across runs), hard-failing over the 75 KB limit.

Committed artifacts, one directory per recording under `reference/`: `refs.jsonl` (26 + 14 qualified labels on 19 + 12 locations), `manifest.json`, `review.json`, `questions.jsonl` (3 + 3). The review crops (an identification/measurement pair per question) ship out-of-band: `*.jpg` is LFS-routed and anonymous upload to the project's LFS endpoint returns 401 for outside contributors, so they regenerate locally instead (one `--crops` command, documented in `reference/README.md`) — happy to push the originals as soon as there's a flow for it; the five publishable pairs are attached as one sheet under Review evidence above. One `go2_short` question has no publishable frames at all: every candidate contains an identifiable person.

## How to Test

- CI (this PR): the synthetic unit layer runs by default — 195 tests, ~1 s, no data, no network; two skip while the review crops ship out-of-band. Both self-hosted lanes skip fork PRs, so the recording-backed layers show as skipped here; commands to run them locally:
  - Full chain, no API key (ingests a thin replay slice, replays a frozen transcript, scores and renders):

    ```bash
    uv run pytest dimos/agents/evals/test_spatial_goal_eval.py::test_full_chain_pipeline \
      -m self_hosted -q --no-cov
    ```

  - Live sweep (needs `OPENAI_API_KEY` for the default arms; 8 cases, ~21 min observed). The arms default to `gpt-5.6-luna` + `openai:gpt-5.6-sol`, which is what any reported numbers should be measured on; `DIMOS_EVAL_MODEL_IDS` names your own instead (comma-separated, anything `mcp_client._init_model` resolves — so a machine with no OpenAI credentials can sweep e.g. `ollama:qwen3:8b`). `OPENAI_BASE_URL` is honoured too, but point it somewhere serving a *different* model and the arm label stops matching what answered — name the real model in `DIMOS_EVAL_MODEL_IDS` so the shards stay attributable:

    ```bash
    for DS in go2_bigoffice go2_short; do
      uv run python -m dimos.agents.evals.ingest --dataset $DS \
        --out-dir ~/.local/state/dimos/spatial-eval/$DS
    done
    DIMOS_EVAL_STORE_ROOT=~/.local/state/dimos/spatial-eval \
    DIMOS_EVAL_SHARD_DIR=.ignore.eval-shards \
    uv run pytest dimos/agents/evals/test_spatial_goal_eval.py -m self_hosted -k sweep -q --no-cov
    uv run python -m dimos.agents.evals.render --shards .ignore.eval-shards/go2_bigoffice \
      --out error_distribution.png --threshold-m 3.5   # once per dataset dir
    ```

- Sample results from one live sweep of the default arms (run via an OpenAI-compatible endpoint, after verifying it serves both request surfaces and rejects nonexistent model ids): all 8 configurations routed correctly — n_pred 3/3, no_prediction 0, broken 0 — and the four arms produced identical goals question-for-question, the Responses+reasoning arm included. Per recording, 2/3 questions land inside their thresholds; the two misses (4.54 m against a 2.00 m threshold, 3.99 m against 3.05 m) are identical across arms — a retrieval-layer floor, not a model difference. One case hit the fixture's 180 s answer timeout on its first run and passed a single-case rerun; the broken shard is excluded from the rates and kept aside, which is the scorer's contract doing its job. The dashed line in both panels is the 3.5 m threshold cap — each question's own threshold lives in its record. The figure is `render.py` output rendered from the shard data, composited side by side — program output, not a hand- or AI-drawn diagram.

image

One shard line, as written (the `go2_bigoffice` organization miss — `retrievals` is the post-hoc observable saying where the goal came from):

  ```json
  {"kind": "answer", "data": {"error": null, "goal_x": -3.359025, "goal_y": -4.088011, "goal_yaw": -0.15469104442073522, "model_id": "openai:gpt-5.6-sol", "n_goals": 1, "outcome": "predicted", "prompt_id": "shipping", "prompt_sha256": "f146c8e3e4f7d87d217ce01a80553435ad4804d1bdb32df7df0066a70201a446", "question_id": "go2-bigoffice-organization", "retrievals": [{"distance": 0.7343969941139221, "query": "the stack of storage boxes", "x": -3.359025, "y": -4.088011}], "run_id": "20260801T010525-25561", "tool_invoked": true, "tool_queries": ["the stack of storage boxes"], "wall_time_s": 13.826804374984931}}
  ```

## Non-goals

presence/count questions (needs student-side aggregation tools — the recorded retrieval observable is the first rail for this) · room-level questions (#1913, needs a room representation) · instance-identity pass criteria (needs validation of the retrieval observable shipped here) · LLM judging (needs free-text answer space) · sweep runner / result store / dashboard (needs multi-round trend tracking) · DimSim/sim variant (documented next step) · upstream fixes split out as follow-ups (runner timeout config; `SpatialMemory.stop()` double-save; detector mask letterbox offset).

## Limitations

- Goal coordinates come from `SpatialMemory`'s deterministic CLIP top-1, so a model's influence is confined to whether it calls the tool, what query text it sends, and how often — read the sweep as a tool-routing and query-formulation comparison, not a ranking of spatial ability. In the one live sweep so far (n = 1 per configuration) the four arms produced identical goals question-for-question — the Responses+reasoning arm included — consistent with routing being saturated on these questions and the deterministic retrieval fixing everything downstream; repeated runs under `run_id` are how that reading gets tested.
- Six binary questions across two recordings can't support fine-grained pass-rate comparisons; at this K the instrument's value is the error distribution, the per-question records, and repeated runs (`run_id`). Of the labels review dropped, over half are validity exclusions (dynamic objects, identical instances, confusable neighbors — no unique answer exists), most of the rest are unverifiable (dim or motion-blurred frames), and exactly one was a caught reference error. Lighting is the dominant yield factor across these two recordings — the bright 60 s `go2_short` yielded proportionally more. A larger detector variant (yoloe-11l) was measured and rejected: 26 → 17 / 14 → 8 qualified labels, no meaningful new coverage, 2.2× the detector runtime — the 11s default stands.
- The first seconds of a recording yield nothing — the rolling LiDAR local map is still sparse at recording start (`go2_bigoffice`'s elevator lobby at t+4–8 s: detected, zero supported measurements).
- Camera extrinsics are nominal (fixed forward offset + axis change, no calibrated pitch); dynamic objects are caught by review via location linkage, not by the pipeline.
- First `navigate_with_text` call triggers ChromaDB's default-EF download (~80 MB) for the tagged-location lookup; self-hosted runners need that cached or egress to chroma-onnx-models.s3.amazonaws.com.
- The two live arms use different OpenAI API surfaces (Responses vs chat completions) — the two request paths the shipping client takes; model ID, prompt hash and run id are on every shard line. The baseline arm is the shipping system prompt, so a change to it moves this eval — that's the point.

## AI assistance

Implemented with Claude Code (Fable 5). The review overlay verdicts are mine, from the review evidence (committed crops plus the multi-frame and full-video checks documented in `reference/`).

## Checklist


paul-nechifor:
Closed for inactivity. Can be reopened later.

PR #3315, opened 2026-07-31
refactor(imitation): isolate LeRobot policy runtime

TomCC7:
## Contribution path

- Stacked on #3853.

## Problem

- LeRobot's dependency environment conflicts with the host stack.
- Policy chunks need explicit execution, validation, preemption, and cancellation through the coordinator.

## Solution

- Run LeRobot in an isolated Python project behind the host `LeRobotPolicyModule` contract.
- Own generic rollout behavior here: checkpoint/live-input preflight, ordered action chunks, freshness checks, and explicit start/stop.
- Add named trajectory execution/cancellation, joint-limit validation, accepted-command telemetry, and idle-hold support to the control layer.
- Keep cancellation ordered after in-flight submissions and report uncertain cancellation.
- Publish debounced Quest button edges while retaining continuous grip/trigger inputs.
- Keep robot-specific rollout composition in #3855; collection consumes accepted-command telemetry in #3854.

### API shape

```text
LeRobotPolicyModule
  inputs: color_image: Image, coordinator_joint_state: JointState
          button_pressed: Buttons (optional operator input)
  config: policy_path, task, device, joint_names, fps,
          image_width, image_height, max_observation_age_s,
          trajectory_task_name, rollout_button
  control: PolicyControlSpec → execute_trajectory / cancel_trajectory
```

- `RolloutControlSpec`: `preflight_rollout()`, `start_rollout()`, `stop_rollout()`, `rollout_status()`.
- Preflight loads/validates the checkpoint and current inputs without sending motion.
- This is the existing single-camera LeRobot contract, not a generalized backend/remapping API.

## How to Test

Exercise the isolated runtime contract without robot hardware:

```bash
cd dimos/imitation/policy/lerobot/python
uv run --isolated --locked --group tests --with-editable ../../../../../ python -m pytest
```

- At this layer, 398 affected host tests and 27 isolated-runtime tests passed In the previous cleanup. The final stack also passed 413 control/Quest regression tests and all 36 isolated policy/preparation tests.
- Scoped type checks and commit hooks passed.
- Physical rollout and cancellation remain hardware release checks; mock tests are not safety certification.

- The repository-wide Clippy hook reaches an existing `dim_slam` flake failure resolving `git+file:../../../..`; that unrelated flake is unchanged.

## AI assistance

- Review cleanup updated by OpenAI Codex (agent).

## Checklist


dimos/experimental/robot_policy/lerobot.py
```diff
@@ -0,0 +1,478 @@
...
+
+    def __init__(
+        self,
+        config: LeRobotPolicyModuleConfig,
+        policy: LeRobotPolicyConfig,
+    ) -> None:
+        try:
+            torch = import_module("torch")
+            PreTrainedConfig = import_module("lerobot.configs.policies").PreTrainedConfig
+            policy_factory = import_module("lerobot.policies.factory")
```
TomCC7: should just import from the top level, check how roboplan is doing

dimos/experimental/robot_policy/lerobot.py
```diff
@@ -0,0 +1,478 @@
...
+"""Run a trained LeRobot policy against live DimOS observations."""
+
+from __future__ import annotations
+
+from contextlib import nullcontext
+from copy import copy
+from importlib import import_module
+from threading import Event, RLock, Thread, current_thread
+import time
+from typing import Any, Protocol
```
TomCC7: if you have Any here 99% of the time we need to cleanup

dimos/experimental/robot_policy/lerobot.py
```diff
@@ -0,0 +1,478 @@
...
+class LeRobotPolicyModuleConfig(ModuleConfig):
+    policies: dict[str, LeRobotPolicyConfig] = Field(min_length=1)
+    joint_names: list[str] = Field(min_length=1)
+    fps: float = Field(default=15.0, gt=0)
+    robot_type: str = ""
+    max_observation_age_s: float = Field(default=0.5, gt=0)
+
+
+class _LeRobotBackend:
+    """Lazy LeRobot/torch adapter using the upstream inference pipeline."""
```
TomCC7: should not use lazy adapter here

TomCC7: Removed the lazy adapter entirely: PolicyBackend, _LeRobotBackend, and _load_policy_backend are gone. LeRobot loading and inference now live directly on LeRobotPolicyModule with typed _LoadedPolicy state. Checkpoint loading remains on demand and cached by policy name so module startup stays lightweight.

dimos/experimental/robot_policy/lerobot.py
TomCC7: Switched to guarded top-level Torch and LeRobot imports following the RoboPlan pattern. Selecting this optional module now gives actionable dimos[lerobot] install guidance when the extra is absent, while registry listing remains safe because it stores import strings. The tests fake the upstream module tree only around import.

dimos/experimental/robot_policy/lerobot.py
TomCC7: Removed production Any usage. LeRobot concrete types now come from the installed 0.4.4 package in the lint group with exact-module follow_untyped_imports overrides. Status is a TypedDict, config checks moved to Pydantic validators, and the two upstream-declared untyped processor/observation boundaries are narrowed immediately with documented casts.

dimos/imitation/dataprep/lerobot.py
```diff
@@ -0,0 +1,65 @@
...
+    uv_run_command,
+)
+from dimos.imitation.dataprep.core import DataPrepConfig
+from dimos.imitation.policy.lerobot import module as lerobot_module
+from dimos.utils.cache import cache_usage_guard
+
+
+def lerobot_project() -> Path:
+    """Locate the packaged LeRobot project beside its host contract."""
+    return require_locked_project(Path(lerobot_module.__file__).resolve().parent / "python")
```
TomCC7: can this be a general util function? seems pretty common usecase. like given isolated module find the project?

dimos/imitation/dataprep/lerobot.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def run_lerobot_dataprep(config: DataPrepConfig) -> Path:
```
TomCC7: remind me why we need this wiring file? is this part of dimos cli?

dimos/imitation/policy/lerobot/python/dimos_lerobot/dataprep.py
```diff
@@ -0,0 +1,179 @@
...
+        with suppress(Exception):
+            dataset.clear_episode_buffer()
+        with suppress(Exception):
+            dataset.finalize()
+        raise
+    return Path(dataset.root)
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description="Convert a DimOS recording to LeRobot")
```
TomCC7: use typer

dimos/imitation/collection/blueprint.py
```diff
@@ -68,3 +71,25 @@ def _camera_if_real() -> tuple[Blueprint, ...]:
...
     CollectionRecorder.blueprint(db_path=_session_db("piper")),
 )
+
+
+def _openyam_wrist_camera() -> Webcam:
+    return Webcam(
+        camera_index=0,
+        width=640,
+        height=480,
+        fps=15.0,
```
TomCC7: should be 30

TomCC7: Updated the OpenYam wrist webcam to 30 FPS and added a blueprint assertion for resolution and frame rate in 5a5f831c0.

dimos/imitation/policy/lerobot/python/dimos_lerobot/dataprep.py
TomCC7: Replaced argparse with a typed Typer entry point (typer.run / typer.echo) and validated the real isolated --help path in 5a5f831c0.

dimos/imitation/dataprep/lerobot.py
TomCC7: Yes—this is the thin adapter used by dimos dataprep build. The host CLI serializes DataPrepConfig here and launches the LeRobot writer under the locked isolated environment, so the main DimOS process does not import LeRobot. I kept that boundary but removed its project-discovery duplication in 5a5f831c0.

dimos/imitation/dataprep/lerobot.py
TomCC7: Generalized this as python_native_project(module_class) in core and reused it from both PythonNativeModule.runtime_project and the LeRobot dataprep adapter in 5a5f831c0.

dimos/imitation/policy/lerobot/python/tests/test_dataprep.py
TomCC7, line 1: I think we just put test files in place instead of special test folder

dimos/imitation/policy/lerobot/python/dimos_lerobot/dataprep.py
TomCC7, line 1: why we need the entrypoint here if there's already a bunch of data prep files?

dimos/msgs/control_msgs/TaskPreemption.py
TomCC7, line 1: why is this required can we achieve safe goal without adding this new msg type

dimos/imitation/policy/lerobot/module.py
```diff
@@ -0,0 +1,108 @@
...
+        return self
+
+
+class LeRobotPolicyModule(IsolatedPythonModule):
+    """Convert live image and joint-state observations into joint targets."""
+
+    implementation = "dimos_lerobot.runtime:LeRobotPolicyRuntime"
+    config: LeRobotPolicyModuleConfig
+
+    color_image: In[Image]
```
TomCC7, line 130: wondering if it's possible to declare multi-cam support. this is a hard problem for sure... but essentially, at it's freedom form, we should be able to do anything: https://github.com/dimensionalOS/dimos/pull/3564

dimos/imitation/policy/rollout_controller.py
TomCC7, line 1: how is controller button keymap currently wired and handled? adding a new module just for this seems super weird...

dimos/memory/module.py
```diff
@@ -445,9 +445,14 @@ def _resolve_ts(self, name: str, msg: Any) -> float:
...
-        """Pose to anchor *msg* with. Dispatches to the stream's (async)
-        ``@pose_setter_for`` if one is defined, else falls back to a
-        ``world <- frame_id`` tf lookup."""
+        """Pose to anchor *msg* with.
+
+        Poseless streams skip pose setters and tf resolution. Other streams
+        dispatch to their async ``@pose_setter_for`` when defined, then fall
+        back to a ``world <- frame_id`` tf lookup.
+        """
+        if name in self.config.poseless_streams:
```
TomCC7: I guess they keep the warning here for a reason, so maybe just set a default pose instead of ignoring that?

native/rust/experimental/dimos-memory-recorder/src/decoding.rs
```diff
@@ -141,6 +142,11 @@ fn source_timestamp(payload_type: &str, data: &[u8], reception_ts: f64) -> Resul
                 .context("invalid LCM foxglove_msgs.CompressedVideo")?;
             (message.timestamp.sec, message.timestamp.nanosec)
         }
+        "dimos.msgs.imitation_msgs.EpisodeStatus.EpisodeStatus" => {
+            let message =
+                EpisodeStatus::decode(data).context("invalid LCM imitation_msgs.EpisodeStatus")?;
```
TomCC7: I don't remember the reason why we need a special message type while don't need previously can you remind me

dimos/teleop/quest/quest_extensions.py
```diff
@@ -216,6 +225,12 @@ def _handle_engage(self) -> None:
                     self._engage(hand)
             self._primary_was_pressed[hand] = is_pressed
 
+    def _should_publish(self, hand: Hand) -> bool:
```
TomCC7, line 229: is that better to keep enablement here or controller? I'm thinking

TomCC7:
The combined change has now been split into narrow dependent PRs:

1. #3853 — OpenYAM learning baseline
2. #3856 — isolated LeRobot policy runtime
3. #3854 — native OpenYAM collection
4. #3855 — controlled OpenYAM policy rollout

GitHub does not allow changing the base of a PR already contained in a Stack, and #3315 has unrelated A1Z work above it. This PR remains open as the review-history and A1Z integration anchor; the new PRs carry the decision-clean diffs. Existing threads here are intentionally preserved, including threads that now reference files extracted into later layers.

dimos/codebase_checks/test_no_init_files.py
```diff
@@ -14,10 +14,16 @@
 
 from dimos.constants import DIMOS_PROJECT_ROOT
 
+VIRTUAL_ENVIRONMENT_DIRECTORIES = frozenset({".venv", "venv"})
```
TomCC7: change doesn't seem necessary? I don't see init file?

dimos/core/python_native_environment.py
TomCC7, line 1: where are these coming from? why not use the existing isolated module but need to add extra files?

dimos/codebase_checks/test_no_init_files.py
TomCC7: Removed in 34ff638a52. The nested .venv was only a local, gitignored test artifact; the clean-checkout codebase check passes without an exemption. The isolated test commands now use uv run --isolated instead.

dimos/core/python_native_environment.py
TomCC7: Agreed. Commit 34ff638a52 removes the extra runner files from #3315 and keeps policy environment setup inside the existing IsolatedPythonModule. The shared PythonNativeProject/python_native_call code now first appears in #3854 at 933facffbe, where isolation and synchronous dataprep are both real consumers.

dimos/imitation/policy/lerobot/python/dimos_lerobot/runtime.py
```diff
@@ -0,0 +1,401 @@
...
+            if loaded_policy is not None:
+                self._reset_policy(loaded_policy)
+            with self._lock:
+                self._active = False
+
+    @staticmethod
+    def _reset_policy(loaded_policy: _LoadedPolicy) -> None:
+        cast("_Resettable", loaded_policy.policy).reset()
+        cast("_Resettable", loaded_policy.preprocessor).reset()
+        cast("_Resettable", loaded_policy.postprocessor).reset()
```
TomCC7: will this cast automatically check if there's reset method? if not this cast is useless and better use hasattr checkign

dimos/imitation/policy/lerobot/python/dimos_lerobot/runtime.py
```diff
@@ -0,0 +1,401 @@
...
+        loaded_policy = policy_class.from_pretrained(self.config.policy_path, config=policy_config)
+        preprocessor, postprocessor = make_pre_post_processors(
+            policy_cfg=policy_config,
+            pretrained_path=self.config.policy_path,
+            preprocessor_overrides={"device_processor": {"device": str(device)}},
+        )
+        return _LoadedPolicy(
+            policy=loaded_policy,
+            device=device,
+            preprocessor=cast(
```
TomCC7: I think it should be automatically this type why need casting then

TomCC7: Fixed in 582da9d158. The local PreparedObservation/PolicyBatch approximations and their casts are gone. _LoadedPolicy now uses LeRobot 0.6.0 RobotObservation and PolicyAction directly, so the factory, observation preparation, and select_action types flow through without casts.

dimos/imitation/policy/lerobot/python/dimos_lerobot/runtime.py
TomCC7: Fixed in 582da9d158. The cast only affected static typing and did not check the object. Reset now goes through a callable runtime check before invocation, which also handles LeRobot 0.6.0 exposing an untyped PreTrainedPolicy.reset method without an ignore.

dimos/teleop/webxr/module.py
```diff
@@ -111,6 +112,8 @@ class WebXRTeleopModule(Module):
     left_controller_output: Out[PoseStamped]
     right_controller_output: Out[PoseStamped]
     teleop_buttons: Out[Buttons]
+    button_pressed: Out[Buttons]
+    button_released: Out[Buttons]
```
mustafab0, line 116: This should be a demos utility that any module could use to debounce the buttons. so we only have one source of truth with `teleop_buttons` as raw input.

I will let this go in the interest of time  but let's mark this as a todo

mustafab0 (dismissed):

dimos/teleop/webxr/module.py
mustafab0: @TomCC7

mustafab0: https://github.com/dimensionalOS/dimos/issues/4213

mustafab0 (approved):

PR #3325, opened 2026-08-02
feat(manipulation): add RoboPlan shortcutting and split planner

TomCC7:
## Contribution path

- Linked issue or discussion: N/A

## Problem

RoboPlan produces valid native joint-space RRT paths, but the DimOS adapter did not apply RoboPlan path shortcutting. Planning behavior also lived inside `RoboPlanWorld`, which mixed scene management with motion planning and tied planner configuration to world finalization.

## Solution

Add collision-aware path shortcutting and separate RoboPlan planning from world management.

- Add `RoboPlanPlanner`, initialized with a `RoboPlanWorld` reference and a deep copy of `RoboPlanPlannerConfig`.
- Move joint, selected-joint, Cartesian, shortcutting, validation, and result-conversion behavior into the planner.
- Keep `RoboPlanWorld` responsible for scene state, robots, obstacles, kinematics, and collision queries.
- Make the planner factory return a distinct planner bound to the supplied world.
- Move RoboPlan planner configuration into `planning/planners/roboplan_config.py`.
- Enable shortcutting by default and forward every setting to `PathShortcuttingOptions`.
- Preserve the original valid RRT path when shortcutting raises `RuntimeError`, returns an empty path, or changes either endpoint.
- Preserve existing `PlannerSpec` method signatures and world-identity checks.

The planner uses the bound world scene and lock internally, so callers continue passing the world through the existing `PlannerSpec` API without wiring RoboPlan calls themselves.

## How to test

```bash
.venv/bin/pytest dimos/manipulation/test_roboplan.py dimos/manipulation/test_planning_factory.py dimos/manipulation/planning/planners/test_roboplan_config.py -q
.venv/bin/mypy dimos/manipulation/planning/planners/roboplan_planner.py dimos/manipulation/planning/world/roboplan_world.py dimos/manipulation/planning/factory.py
./bin/pytest-fast
```

Verification completed locally:

- 85 focused RoboPlan, configuration, and factory tests passed.
- 60 dependent manipulation tests passed.
- Targeted mypy passed.
- Full fast suite: 3756 passed, 17 skipped.
- Ruff, formatting, license, lockfile, LFS, doclink, Cargo formatting, and Clippy checks passed.

## AI assistance

OpenAI Codex with GPT-5 assisted substantially with design discussion, implementation, tests, review cleanup, and PR text.

## Checklist


dimos/manipulation/planning/world/roboplan_world.py
```diff
@@ -43,10 +43,21 @@
         "Install the manipulation extra before selecting the roboplan backend."
     ) from exc
 
+if not hasattr(roboplan_core, "PathShortcuttingOptions") or not hasattr(
```
TomCC7: no has attr checking

dimos/manipulation/planning/factory.py
```diff
@@ -162,6 +162,10 @@ def create_planner(
             raise ValueError(_ROBOPLAN_PLANNER_REQUIRES_ROBOPLAN_WORLD)
         if not isinstance(world, PlannerSpec):
             raise ValueError("RoboPlan-native planner requires a RoboPlan world planner object")
+        configure_planner = getattr(world, "configure_planner", None)
+        if not callable(configure_planner):
```
TomCC7: no test config

dimos/manipulation/planning/world/roboplan_world.py
```diff
@@ -1180,6 +1199,55 @@ def _run_native_rrt(
...
+            self._validate_shortcut_path(group, path, shortened)
+        except Exception as exc:
+            logger.warning("RoboPlan path shortcutting failed; using raw path: %s", exc)
+            return path
+        return shortened
+
+    def _validate_shortcut_path(
+        self,
+        group: RoboPlanGroup,
+        original: Any,
```
TomCC7: no any

dimos/manipulation/planning/planners/config.py
mustafab0, line 1: This file is seemingly becoming a RoboPlan config file. Rather than a planner config file.

I think better to make it specific to roboplan in its own folder.

Actually since RoboPlanner is integrated with the RoboPlan world. We could move it there.

Alternatively we remove RoboPlan planner out of the `roboplan_world` and make a directory

```
dimos/manipulation/planner/roboplan/config.py
dimos/manipulation/planner/roboplan/roboplan_planner.py
dimos/manipulation/planner/rrt/config.py
dimos/manipulation/planner/rrt/rrt_planner.py
...
```

mustafab0: You already have safety checks where we check compatibility of planners and viz for specific world implementation.

mustafab0 (changes requested):
Can we change the RoboPlan planner config at runtime.

For example enable disable shortcutting during operation on agent discretion?

TomCC7:
> Can we change the RoboPlan planner config at runtime.
> 
> For example enable disable shortcutting during operation on agent discretion?

that's actually a new feature... which we need to spec out later. I'm thinking of adding a config alternation api in plannerspec and each planner backend can implement it's own config object...

mustafab0 (approved):

PR #3327, opened 2026-08-02
feat: Cockpit blueprints

paul-nechifor:
* Added an API which lets robot blueprints describe how the cockpit should look like for that particular robot.
* Use can use Col/Row and existing panels like Video and Map2D (more to come).

Closes DIM-1169


leshy (approved):

leshy:
love it

PR #3328, opened 2026-08-02
feat: Keyboard Teleop

paul-nechifor:
* Add keyboard teleop panel which drives the robot.
* Uses the same WASD and QE controls.
* This is the first viewer to robot channels.
* Only one cockpit can control at a time. The second one will get an error.

Closes DIM-1168

### DIFF
--- a/dimos/web/relay_bridge/e2e_support.py
+++ b/dimos/web/relay_bridge/e2e_support.py
@@ -24,6 +24,8 @@ from dimos.web.relay_bridge.protocol import (
     Msg,
     Sub,
     Subs,
+    TeleopStart,
+    TeleopStarted,
     Watch,
 )
 from dimos.web.relay_bridge.relay_bridge_module import RelayBridgeModule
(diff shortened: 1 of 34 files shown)


dimos/web/relay_bridge/e2e_support.py
```diff
@@ -68,6 +70,21 @@ async def attach_viewer(
...
+    """Acquire the robot's teleop lease (retrying: the Python viewer's control
+    plane is datagrams, so the teleop_started ack can be lost)."""
+    deadline = time.monotonic() + timeout
+    while True:
+        viewer.send_control(TeleopStart())
+        msg = await next_control(viewer, 0.5)
+        while msg is not None and not isinstance(msg, TeleopStarted):
+            msg = await next_control(viewer, 0.5)
+        if isinstance(msg, TeleopStarted):
+            return
```
mustafab0, line 83: The while loop above already checks this. Any reason to do it twice?

dimos/web/relay_bridge/manifest.py
```diff
@@ -262,6 +262,18 @@ def parse_manifest(data: Any) -> Manifest:
...
+                raise ManifestError(
+                    "invalid_teleop_panel",
+                    f"teleop panel {panel.id} must bind exactly one channel",
+                )
+            cmd = ch_ids[panel.channels[0]]
+            if cmd.encoding != "twist.json.v1" or cmd.delivery != "latest" or cmd.dir != "tx":
+                raise ManifestError(
+                    "invalid_teleop_panel",
+                    f"teleop panel {panel.id} needs a twist.json.v1 latest tx channel",
+                )
```
mustafab0, line 276: Appears that teleop only support twist messages.
This is great for controlling moving base robots (for nav based things), also twist based teleop.

But for manipulation we also stream `PoseStamped` messages as EndEffector positions. Would that be a separate manifest/panel?

mustafab0 (approved):

dimos/web/relay_bridge/e2e_support.py
paul-nechifor: The loop checks two conditions. You still need to check isinstance.

dimos/web/relay_bridge/manifest.py
paul-nechifor: Yes, this is not the final thing. It's still being built up. The PRs are large enough already. I'll add more in subsequent ones. :D

PR #3381, opened 2026-08-05
feat: control coordinator gripper api refactor

jhengyilin:
## Contribution path

- Linked issue or discussion: [DIM-1394](https://linear.app/dimensional/issue/DIM-1394/gripper-api-refactor)

## Problem

Gripper commands bypassed the normal control-task path through coordinator RPCs and adapter-specific scalar APIs. That skipped task ownership and arbitration, spread gripper-specific joint splitting and unit conversion across the generic control API, and constrained the abstraction to a single special-cased joint. It could also apply conversion more than once, such as reducing xArm's native 0–850 command range.

## Solution

- Treat integrated gripper actuators as ordinary entries in each manipulator adapter's ordered joint arrays. `get_dof()`, reads, writes, and limits now cover every adapter-owned joint; adapter-specific arm/gripper splitting stays inside the adapter.
- Add `GripperControlTask` as the dedicated owner of configured gripper joints. It accepts native or normalized targets, resolves limits by joint name, participates in normal arbitration, and exposes normalized task RPCs.
- Remove the coordinator's direct gripper RPC and gripper handling from arm motion tasks. Keyboard and teleop inputs target the gripper task. Trajectory tasks accept any configured hardware-joint subset, while manipulation planning groups naturally emit arm-only trajectories.
- Keep units authoritative at the adapter boundary: xArm uses SDK units (0–850), Piper and A1Z use their physical stroke, and MuJoCo publishes its joint split and range through shared-memory metadata.
- Consolidate regression tests around task behavior, partial trajectory and gripper coexistence, complete joint-array ordering, adapter conversion, blueprint configuration, and MuJoCo readiness metadata.

## How to Test

```bash
uv run pytest dimos/control/tasks/gripper_task dimos/hardware/manipulators/mock dimos/hardware/manipulators/sim/test_shm_adapter.py dimos/hardware/manipulators/xarm/test_adapter.py dimos/robot/manipulators/a1z/blueprints/test_teleop.py dimos/robot/manipulators/piper/blueprints/test_teleop.py dimos/robot/manipulators/xarm/blueprints/test_teleop.py dimos/simulation/engines/test_mujoco_sim_module.py -m 'not (self_hosted or mujoco or self_hosted_large)'
uv run --group lint mypy
```

## AI assistance

OpenAI Codex with GPT-5 was used extensively for implementation, review, test cleanup, and CI debugging. The author reviewed the resulting changes.

## Checklist


dimos/hardware/GRIPPER-SPEC.md
```diff
@@ -0,0 +1,324 @@
...
+
+
+## 1. Principle
+
+> **`write_joint_positions()` is the API for all joints. A gripper joint is just a robot
+> joint of a special type, unified from the consumer's perspective.**
+
+Every requirement below follows from that sentence.
+
+## 2. Why
```
TomCC7: better named title is current state...

dimos/hardware/GRIPPER-SPEC.md
```diff
@@ -0,0 +1,324 @@
...
+**R3.** An adapter needing a distinct SDK call for gripper indices MUST split inside
+`write_joint_positions()`:
+
+```python
+def write_joint_positions(self, positions, velocity=1.0):
+    arm, grip = positions[: self._arm_dof], positions[self._arm_dof :]
+    ok = self._arm.set_servo_angle_j([math.degrees(p) for p in arm], ...) == 0
+    if grip:
+        ok = self._arm.set_gripper_position(grip[0] * _XARM_GRIPPER_MAX_SDK, wait=False) == 0 and ok
+    return ok
```
mustafab0: No I don't like this at all.

mustafab0: We should still control gripper with a seperate `GripperControlTask`.

Gripper JointStates should be routed to the task, and the task can call `self._arm.set_gripper_position`.

your implementation works only for Arms with their specific grippers, like the xarm and its first party gripper. I won't be able to use this to control the elephant grippers

dimos/hardware/GRIPPER-SPEC.md
```diff
@@ -0,0 +1,324 @@
...
+```
+
+**R4.** The adapter MUST learn its gripper joint count at construction (`gripper_dof: int = 0`).
+It MUST NOT infer it from array length.
+
+**R5.** `get_dof()` MUST return the total (arm + gripper). Adapters branching on arm DOF
+internally — e.g. xArm's 6- vs 7-DOF initial pose (`xarm/adapter.py:278-280`) — keep a
+private `_arm_dof`.
```
mustafab0: Unsure about this.

`get_dof()` should be callable by the arm and the gripper independently and we can then add them later if we want to.

In most cases I don't think so.

dimos/hardware/GRIPPER-SPEC.md
```diff
@@ -0,0 +1,324 @@
...
+
+### 3.2 Units — the adapter owns conversion
+
+**R6.** A gripper joint value on `joint_command` and `coordinator_joint_state` is
+normalized: `0.0` fully closed, `1.0` fully open. Out-of-range values are clamped. Arm
+joints are unchanged (radians).
```
mustafab0: This only makes sense for our simple use cases that we discussed. 

`SweepVolumeGrippers` to borrow terminology from `GraspGen`. This defeats our reason to refactor.

I agree with the normalized 0 to 1 range, but it shouldn't be a single scalar. But a list of joint normalized vector. 

For, e.g list of 9 joints for each elephant gripper

dimos/hardware/GRIPPER-SPEC.md
```diff
@@ -0,0 +1,324 @@
...
+**R7.** The **adapter** performs the conversion, in **both** directions. No other layer
+converts. This collapses §2.3's three conversion sites into one.
+
+*Why the adapter.* It is already the unit boundary for every joint —
+`xarm/adapter.py:222-227` converts wire radians into vendor degrees. R7 applies that
+existing rule rather than inventing a second site. Any higher placement would force the
+converting layer to branch on *which joints are grippers*, re-introducing the exact
+special-casing this PR removes. And the range constants already live in the adapters.
+
+**R8.** Each adapter MUST declare its own gripper travel range as a module constant. Two
```
mustafab0: 💯

dimos/hardware/GRIPPER-SPEC.md
```diff
@@ -0,0 +1,324 @@
...
+already do — `piper/adapter.py:44-45` (`GRIPPER_MAX_OPENING_M = 0.08`) and
+`galaxea_a1z/config.py:45-49` (`max_opening_m`).
+
+**R9.** `HardwareComponent.gripper_open_position` / `gripper_closed_position` are removed
+(`components.py:97-98`), together with `_normalized_to_physical` /
+`_physical_to_normalized` (`hardware_interface.py:232-244`).
+
+**R10.** Task-level `gripper_open_pos` / `gripper_closed_pos` are now interpreted as
+normalized. **No task code changes** — the fields and their `0.0` defaults stay as they
```
mustafab0: I think you are constraining the LLM to develop a system based on the current `GripperTask` implementation. 

We should be able to support multi-joint grippers, pneumatic grippers, soft body grippers etc.

dimos/hardware/GRIPPER-SPEC.md
```diff
@@ -0,0 +1,324 @@
...
+**R1.** `write_joint_positions()` and `read_joint_positions()` MUST cover **all** joints
+the component owns, in `HardwareComponent.all_joints` order. Gripper entries occupy the
+trailing indices.
+
+```python
+adapter.write_joint_positions([0.1, -0.4, 0.0, 0.9, 0.0, 1.2, 0.0,  1.0])
+#                              └──────── 7 arm joints, radians ───┘  └ gripper
+```
+
+**R2.** `read_gripper_position()` / `write_gripper_position()` are **retained** on
```
TomCC7: I prefer remove completely, any special reason for keeping this?

dimos/hardware/GRIPPER-SPEC.md
```diff
@@ -0,0 +1,324 @@
...
+> needs position-delta sampling plus per-gripper effort thresholds. That is grasp
+> detection, a design problem in its own right, and out of scope here.
+
+**R27.** `pick_and_place_module`'s three `time.sleep(0.5/1.5/1.0)` calls
+(`:518, :536, :627`) are **retained**. They remain the only completion signal, exactly as
+today. Replacing them requires the grasp detection R26 defers.
+
+### 3.7 Control modes
+
+**R28.** An adapter MUST refuse a mode it cannot honour by returning `False` from
```
mustafab0: I am considering having 2 modes by default.

1) Simple parallel jaw gripper-like sweep actions, even for elephant robot grippers (Ruthwik mentioned yesterday)
2) Joint level control mode. 

Let me know your thoughts

dimos/hardware/GRIPPER-SPEC.md
```diff
@@ -0,0 +1,324 @@
...
+| **New task** | `control/tasks/gripper_task/` | `GripperTask` + `_registry.py` |
+| Coordinator | `coordinator.py:901-931` | Delete 2 RPCs |
+| `ManipulationModule` | `manipulation_module.py` | 2 method bodies + 1 docstring |
+| Robot configs | `xarm`, `piper`, `a1z`, `a750`, `openyam` | Endpoint removal |
+| Blueprints | gripper-bearing blueprints | Add `GripperTask`; `xarm/config.py:63` `0.85`→`1.0` |
+| Message docs | `msgs/sensor_msgs/JointState.py` | Document gripper unit (R12) |
+
+## 7. Non-goals
+
+- A `GripperAdapter` protocol / `HardwareType.GRIPPER`.
```
mustafab0: Isn't this the main goal?

What is the plan to implement. Is it after the major spec change?

mustafab0 (changes requested):

dimos/hardware/GRIPPER-SPEC.md
```diff
@@ -0,0 +1,324 @@
...
+It MUST NOT infer it from array length.
+
+**R5.** `get_dof()` MUST return the total (arm + gripper). Adapters branching on arm DOF
+internally — e.g. xArm's 6- vs 7-DOF initial pose (`xarm/adapter.py:278-280`) — keep a
+private `_arm_dof`.
+
+### 3.2 Units — the adapter owns conversion
+
+**R6.** A gripper joint value on `joint_command` and `coordinator_joint_state` is
+normalized: `0.0` fully closed, `1.0` fully open. Out-of-range values are clamped. Arm
```
TomCC7: we need to support two kinds of api. normalized unit is helpful for teleop usage (so teleop module don't need to know exact value of gripper opening) and exact SI unit is helpful for grasp proposal control.

Here I would suggest to use SI unit on hardware layer, but control task provide interface for normalized values (which is easy since control task have access to the joint limit)

dimos/hardware/GRIPPER-SPEC.md
```diff
@@ -0,0 +1,324 @@
...
+converts. This collapses §2.3's three conversion sites into one.
+
+*Why the adapter.* It is already the unit boundary for every joint —
+`xarm/adapter.py:222-227` converts wire radians into vendor degrees. R7 applies that
+existing rule rather than inventing a second site. Any higher placement would force the
+converting layer to branch on *which joints are grippers*, re-introducing the exact
+special-casing this PR removes. And the range constants already live in the adapters.
+
+**R8.** Each adapter MUST declare its own gripper travel range as a module constant. Two
+already do — `piper/adapter.py:44-45` (`GRIPPER_MAX_OPENING_M = 0.08`) and
```
TomCC7: this should just be the joint position limit as we now unified the representation :)

dimos/hardware/GRIPPER-SPEC.md
TomCC7: this is internal hardware adapter implementation, the upper level representation is not related here

dimos/hardware/GRIPPER-SPEC.md
```diff
@@ -0,0 +1,547 @@
...
+                            |
+                            v
+  arbitration     tick loop - priority, preemption, e-stop
+                  unchanged; grippers now take part for the first time
+                            |
+                            v
+  wrapper         ConnectedHardware - one ordered array, no conversion   [R25]
+ ========================= the only divergence ==============================
+  adapter         the arm's own adapter         |   its own GripperAdapter
+                  joints = arm                  |   joints = gripper
```
TomCC7: I don't understand why we need to semantically separate `joints` and `gripper_joints` here... plus it's weird that `GripperAdapter`'s `gripper_joints` are empty...

dimos/hardware/GRIPPER-SPEC.md
```diff
@@ -0,0 +1,547 @@
...
+
+A gripper comes in two shapes. **Everything above the hardware layer is identical between
+them** — same task, same units, same joint representation, same `write_joint_positions()`.
+Only the adapter and the connection beneath it differ, because that is where the two
+devices genuinely differ.
+
+```
+  PROPOSED ARCHITECTURE
+
+                  FIRST-PARTY (integrated)      |   THIRD-PARTY (standalone)
```
TomCC7: as we discussed, third-party gripper support is a separate pr and should not be in this one. the reason is we don't currently have any 'third party' gripper in the codebase and your abstraction will have no implementation and no way to verify it's working

jhengyilin: yes, breaking this spec into multiple seqential follow-up PRs, the plan is at the end of the markdown

mustafab0: I don't know why we even have this discussion. 

Manipulation should not care whether the gripper is first party or third-party.

dimos/hardware/GRIPPER-SPEC.md
mustafab0: That makes sense

dimos/hardware/GRIPPER-SPEC.md
```diff
@@ -94,231 +197,351 @@ def write_joint_positions(self, positions, velocity=1.0):
...
-**R7.** The **adapter** performs the conversion, in **both** directions. No other layer
-converts. This collapses §2.3's three conversion sites into one.
+```
+PROPOSED — new protocol, does not exist today
+
+MANDATORY
+  connect() / disconnect() / is_connected()
+  get_dof() -> int
+  get_limits() -> JointLimits
+  set_control_mode(mode) -> bool
```
mustafab0: 👍

dimos/hardware/manipulators/galaxea_a1z/adapter.py
```diff
@@ -240,15 +249,22 @@ def get_info(self) -> ManipulatorInfo:
...
 
+    def get_gripper_dof(self) -> int:
+        """1 when a gripper is fitted, else 0."""
+        return self._gripper_dof
+
     def get_limits(self) -> JointLimits:
-        """Get joint limits."""
+        """Arm limits in radians, then the gripper's jaw opening in metres."""
+        gripper = self._config.gripper
+        upper = [gripper.max_opening_m] if self._gripper_dof and gripper else []
```
TomCC7: gripper_upper

dimos/hardware/grippers/xarm_gripper/adapter.py
TomCC7, line 1: should just take place in xarm adapter

dimos/manipulation/manipulation_module.py
```diff
@@ -191,6 +191,11 @@ def __init__(self, **kwargs: Any) -> None:
         # Canonical generated plan for plan/preview/execute workflow.
         # Robot-local paths and trajectories are derived from this plan on demand.
         self._last_plan: GeneratedPlan | None = None
+        # Gripper state, both fed from streams the module already consumes:
+        # the measured position from coordinator_joint_state, the range from
+        # one get_state call per task (ranges are static).
+        self._latest_gripper_position: dict[str, float] = {}
```
TomCC7: should just treat gripper as joint no special handling

dimos/manipulation/manipulation_module.py
```diff
@@ -1642,38 +1653,90 @@ def _get_gripper_hardware_id(self, robot_name: RobotName | None = None) -> str |
...
+        self, method: str, kwargs: dict[str, Any], robot_name: RobotName | None = None
+    ) -> bool:
+        task = self._gripper_task(robot_name)
+        if task is None:
             return False
-        return self._control_coordinator.set_gripper_position(hw_id, position)
+        return bool(self._control_coordinator.task_invoke(task, method, kwargs))
 
     @rpc
     def get_gripper(self, robot_name: RobotName | None = None) -> float | None:
```
TomCC7, line 1540: manip module should not retain gripper range information. gripper control task should provide rpc for fetching current gripper state in normalized range

dimos/manipulation/manipulation_module.py
```diff
@@ -1642,38 +1653,90 @@ def _get_gripper_hardware_id(self, robot_name: RobotName | None = None) -> str |
...
+        """The gripper task addressing this robot, derived not hard-coded.
+
+        Mirrors the joint it claims: task ``arm_gripper`` owns ``arm/gripper``
+        (GRIPPER-SPEC R30).
+        """
         hw_id = self._get_gripper_hardware_id(robot_name)
-        if hw_id is None:
+        return None if hw_id is None else f"{hw_id}_gripper"
+
+    def _gripper_range(self, robot_name: RobotName | None = None) -> tuple[float, float] | None:
```
TomCC7: why we need to special-handle gripper range given that gripper range is also provided as robot model's joint limit

dimos/manipulation/pick_and_place_module.py
```diff
@@ -624,7 +624,7 @@ def _place_with_orientation(
 
         # 3. Release
         logger.info("Releasing object...")
-        self._set_gripper_position(0.85, rname)
+        self._gripper_invoke("set_sweep", {"value": 1.0}, rname)
```
TomCC7: this is api call leakage, why use this when we have manip module defined the set_gripper rpc?

dimos/hardware/grippers/spec.py
TomCC7, line 1: gripper adapter should not stay within this pr

dimos/hardware/manipulators/demo_gripper_units.py
TomCC7, line 1: remove before ready for review

dimos/control/hardware_interface.py
```diff
@@ -47,24 +48,25 @@ class ConnectedHardware:
...
         self._component = component
-        self._arm_joint_names: list[JointName] = list(component.joints)
-        self._gripper_joints: list[JointName] = list(component.gripper_joints)
-        self._joint_names: list[JointName] = component.all_joints
-        self._gripper_open = component.gripper_open_position
-        self._gripper_closed = component.gripper_closed_position
-        if (self._gripper_open is None) != (self._gripper_closed is None):
-            raise ValueError("gripper open/closed positions must be set together")
+        self._joint_names: list[JointName] = list(component.all_joints)
+        # Velocity writes carry arm joints only; positions carry everything.
```
TomCC7: we should not assume that gripper can't have velocity sensor... if a gripper can't carry it it's a capability issue and the adpater's responsibility to provide either place-holder or not populate at all

TomCC7 (dismissed):

dimos/control/blueprints/mobile.py
```diff
@@ -50,7 +50,7 @@ def _mock_twist_base(hw_id: str = "base") -> HardwareComponent:
     return HardwareComponent(
         hardware_id=hw_id,
         hardware_type=HardwareType.BASE,
-        joints=make_twist_base_joints(hw_id),
+        all_joints=make_twist_base_joints(hw_id),
```
TomCC7: rename makes no sense here revert

dimos/control/tasks/gripper_task/gripper_task.py
```diff
@@ -0,0 +1,388 @@
...
+
+    def __init__(
+        self,
+        name: str,
+        config: GripperControlTaskConfig,
+        limits: list[tuple[float, float]],
+    ) -> None:
+        if not config.joint_names:
+            raise ValueError(f"GripperControlTask '{name}' requires at least one joint")
+        if len(limits) != len(config.joint_names):
```
TomCC7, line 79: why is this not in config? or can this be fetched from robot model directly?

dimos/control/tasks/gripper_task/gripper_task.py
```diff
@@ -0,0 +1,388 @@
...
+            raise ValueError(
+                f"GripperControlTask '{name}': got {len(limits)} limit pairs for "
+                f"{len(config.joint_names)} joints"
+            )
+
+        self._name = name
+        self._config = config
+        self._joint_names_list = list(config.joint_names)
+        self._joint_names = frozenset(config.joint_names)
```
TomCC7: how are these two different? need single source of truth

dimos/control/tasks/gripper_task/gripper_task.py
```diff
@@ -0,0 +1,388 @@
...
+    """Configuration for the gripper task.
+
+    Attributes:
+        joint_names: Gripper joints this task owns; must equal one component's
+            gripper_joints, in order.
+        priority: Priority for arbitration.
+        hold_duration: Seconds to keep emitting a target; 0.0 holds forever.
+        reference_pose: Grasp posture per joint, native units. Required for
+            multi-joint grippers; a single jaw uses its closed limit.
+        open_posture: Open end of a sweep; defaults to the upper limits.
```
TomCC7: sweep logic is pre-mature optimization, let's add that when we really have a 3-finger gripper. for now we can just use joint limit...

dimos/control/tasks/gripper_task/gripper_task.py
```diff
@@ -0,0 +1,388 @@
...
+                self._target = None
+
+    def set_position(self, values: list[float], t_now: float | None = None) -> bool:
+        """Set per-joint targets in the adapter's native units."""
+        if len(values) != self._n:
+            logger.warning(
+                "set_position arity mismatch", task=self._name, expected=self._n, got=len(values)
+            )
+            return False
+        clamped = [max(lo, min(hi, v)) for v, (lo, hi) in zip(values, self._limits, strict=True)]
```
TomCC7: should not silently clamp

dimos/control/tasks/gripper_task/gripper_task.py
```diff
@@ -0,0 +1,388 @@
...
+        return self._latch(clamped, t_now)
+
+    def set_normalized(self, values: list[float], t_now: float | None = None) -> bool:
+        """Set per-joint targets as 0.0-1.0 of travel; 0.0 closed, 1.0 open."""
```
TomCC7, line 170: need value range validation

dimos/control/tasks/gripper_task/gripper_task.py
```diff
@@ -0,0 +1,388 @@
...
+            )
+            return False
+        v = max(_CLOSED, min(_OPEN, value))
+        target = [
+            ref + (opn - ref) * v
+            for ref, opn in zip(self._reference_pose, self._open_posture, strict=True)
+        ]
+        return self._latch(target, t_now)
+
+    def set_reference_pose(self, values: list[float]) -> bool:
```
TomCC7: what is this how is this different from set_position

dimos/control/components.py
```diff
@@ -74,45 +79,63 @@ class HardwareComponent:
...
-            uses its parent-native 0.85 endpoint.
-        gripper_closed_position: Adapter-native closed endpoint; typically
-            0.0 for Piper and XArm.
     """
 
     hardware_id: HardwareId
     hardware_type: HardwareType
-    joints: list[JointName] = field(default_factory=list)
+    all_joints: list[JointName] = field(default_factory=list)
+    gripper_dof: int = 0
```
TomCC7: gripper is just special joint should not have special dof count

dimos/control/components.py
```diff
@@ -74,45 +79,63 @@ class HardwareComponent:
...
+            )
+
+    @property
+    def _gripper_start(self) -> int:
+        # len - gripper_dof, never negative slicing: -0 == 0 would report a
+        # gripper-less arm as having no arm joints.
+        return len(self.all_joints) - self.gripper_dof
+
+    @property
+    def arm_joints(self) -> list[JointName]:
```
TomCC7: again why we need special helper function to distinguish arm joints?

dimos/control/tasks/gripper_task/gripper_task.py
```diff
@@ -0,0 +1,321 @@
...
+
+    joint_names: list[str]
+    priority: int = 10
+    hold_duration: float = 0.0
+    hand: str | None = None
+    require_engagement: bool = True
+
+
+class GripperControlTask(BaseControlTask):
+    """Command- and stream-driven owner of a device's gripper joints."""
```
TomCC7:
```suggestion
    """Command- and stream-driven owner of a device's gripper joints.
    TODO: cleanup gripper call wiring
    """
```

dimos/teleop/keyboard/keyboard_teleop_module.py
```diff
@@ -59,19 +59,16 @@
...
 DEFAULT_LINEAR_SPEED = 0.05  # m/s
 DEFAULT_ANGULAR_SPEED = 0.5  # rad/s
-# Normalized gripper command values.
-GRIPPER_OPEN_POSITION = 1.0
-GRIPPER_CLOSED_POSITION = 0.0
-# TODO: Improve gripper handling.
-GRIPPER_JOINT_NAME = "arm/gripper"
+# The keyboard sends open/closed intent; GripperControlTask owns the units.
+GRIPPER_OPEN = True
+GRIPPER_CLOSED = False
```
ruthwikdasyam: As per gripper task, convention is TRUE for Closed and FALSE for open right? It looks otherwise here

TomCC7: make sense

TomCC7: actually I feel the boolean convention too annoying and error-prone, so I removed the whole API completely. caller use the noramlized api directly (0 close 1 open)

dimos/control/tasks/gripper_task/gripper_task.py
```diff
@@ -0,0 +1,323 @@
...
+                    joint_name=name,
+                    value=position,
+                    normalized=value,
+                )
+        return normalized
+
+    def on_gripper_command(self, msg: Bool, t_now: float) -> bool:
+        """Handle an open/closed toggle (True = closed)."""
+        value = _CLOSED if msg.data else _OPEN
+        return self.set_normalized([value] * len(self._joint_names), t_now)
```
ruthwikdasyam: what are `_joint_names` here. we are sending normalized value which means.. we dont care about no of fingers.. we want it to be open or close - this gripper API controls joints/fingers induvidually.
OR 
we add each finger as a joint, and set limits, thus.. controlling each finger between its limits.

At which level is this meant to be operated. or we have this design to support both?

TomCC7: multi-finger gripper api is not handled in this PR (will be handled by next). joint_names is a reserved field for all control coordinator tasks. here we assume only one joint will be binded but we have to keep the shape of the list to align with other tasks.

Nabla7:
pls approve someone

mustafab0 (changes requested):
This PR breaks the Pick and Place blueprint and a few more manipulation blueprints since old blueprints still don't know about the new GripperTask implementation

Please fix 

Need to add 

```
TaskConfig(
    name="arm_gripper",
    type="gripper",
    joint_names=["arm/gripper"],
    priority=20,
),
```

To the following blueprints:
1. xarm/blueprints/basic.py
2. xarm/blueprints/simulation.py
3. a1z/blueprints/basic.py 
4. openyam/blueprints/teleop.py
5. xarm/blueprints/basic.py
6. a750/blueprints/teleop.py

dimos/robot/manipulators/piper/blueprints/basic.py
```diff
@@ -20,14 +20,19 @@
...
 _piper_hw = piper_hardware("arm")
+_piper_model = make_piper_model_config()
 
 coordinator_piper = autoconnect(
     ControlCoordinator.blueprint(
         hardware=[_piper_hw],
         tasks=[joint_trajectory_task(_piper_hw.joints)],
     ),
-    *mujoco_if_sim(PIPER_SIM_PATH, len(_piper_hw.joints)),
+    *mujoco_if_sim(PIPER_SIM_PATH, len(_piper_model.joint_names)),
```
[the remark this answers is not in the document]
mustafab0: Can fix it in follow up PRs. Piper gripper is not a blocker here.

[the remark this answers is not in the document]
TomCC7: I don't even know we have mujoco support for piper...

mustafab0 (dismissed):

mustafab0 (approved):

PR #3351, opened 2026-08-04
Krishna/feat/openarm damiao

KrishnaH96:
## Contribution path

- Linked issue or discussion: #3129 (this PR is stacked on that branch and extends its pattern to OpenArm)

## Problem

OpenArm support predates the whole-body Damiao stack: a hand-rolled CAN driver plus a manipulator-protocol adapter, one instance per bus, grippers present on the bus but never wired, no sim/real switch, and robot-specific gravity compensation code.

## Solution

Stacked on #3129. This ports OpenArm to the generic `DamiaoWholeBodyAdapter`, mirroring the OpenYAM integration:

- `OpenArmDamiaoAdapter`: both v10 arms (2x DM8006, 2x DM4340, 3x DM4310, send ids 0x01..0x07) and both grippers (DM4310 at 0x08) as one whole-body device over two CAN buses (left=can1, right=can0), commanded in one synchronized tick. Gravity compensation uses the bimanual URDF from `openarm_description` (validated in pinocchio: 14 joints, order left1..7 then right1..7, finite torques).
- Config rewritten on the OpenYAM pattern: one 16-joint WHOLE_BODY component, mock/real selection via `global_config.simulation`, hardware-measured MIT gains carried over from the legacy adapter, and per-side planning models now declare an explicit coordinator-to-URDF joint name mapping.
- Blueprints: `coordinator-openarm`, `openarm-planner-coordinator`, `keyboard-teleop-openarm`, and `keyboard-teleop-openarm-planner`. The keyboard jogs the left arm while the right arm's twist task holds pose. The `[` and `]` keys drive both grippers through a single servo task, enabled by a new `KeyboardTeleopConfig.gripper_joint_names` field.
- The legacy driver, adapter, and CAN scripts are removed as superseded. Registry golden sets are updated in both directions. The e2e planning-groups test moves to `openarm-planner-coordinator`, since the test harness passes `--simulation` and now gets the in-memory adapter automatically.

## How to Test

```bash
dimos can setup can0
dimos can setup can1
dimos run keyboard-teleop-openarm
```

Automated validation:

```bash
uv run pytest dimos/hardware/whole_body dimos/hardware/test_adapter_registries.py
uv run pytest dimos/robot/test_all_blueprints.py -k openarm
```

## Hardware validation

Validated on a physical bimanual OpenArm 2.0 (PEAK 2ch adapter, one 1 Mbps
classic CAN bus per arm).

- All 16 motors respond at the declared ids; DM8009 shoulders confirmed.
[...]

PR #3388, opened 2026-08-06
feat: openarm hardware support with dm_control

TomCC7:
Reopens the OpenArm Damiao changes from #3351 after the intentional revert in #3387.

The original PR was merged and therefore cannot be reopened on GitHub. This replacement PR uses the same head branch and targets the same stacked base.

Please carry review context forward from #3351. The branch contains the hardware-validated follow-up fixes that landed before the original merge.

closes DIM-979


dimos/e2e_tests/test_manipulation_planning_groups.py
```diff
@@ -42,7 +42,9 @@
 pytestmark = [pytest.mark.self_hosted_large]
 
 JOINT_STATE_TOPIC = "/coordinator_joint_state#sensor_msgs.JointState"
-BLUEPRINT = "openarm-mock-planner-coordinator"
+# The e2e harness always passes --simulation (DimosCliCall.simulator), so the
+# blueprint's hardware selection resolves to the in-memory whole-body adapter.
+BLUEPRINT = "openarm-planner-coordinator"
 LEFT_GROUP_ID = "left_arm/manipulator"
```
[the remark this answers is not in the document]
TomCC7: Fixed in 262d5a84a5: the E2E now expects openarm/left_manipulator and openarm/right_manipulator, with a registry assertion covering the public IDs. The full self-hosted OpenArm planning-group E2E passes both tests.

dimos/e2e_tests/test_manipulation_planning_groups.py
```diff
@@ -42,7 +42,9 @@
...
 RIGHT_GROUP_ID = "right_arm/manipulator"
```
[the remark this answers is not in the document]
TomCC7: Fixed in 262d5a84a5: the E2E now expects openarm/left_manipulator and openarm/right_manipulator, with a registry assertion covering the public IDs. The full self-hosted OpenArm planning-group E2E passes both tests.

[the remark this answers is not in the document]
TomCC7: Fixed in 262d5a84a5: the E2E now expects openarm/left_manipulator and openarm/right_manipulator, with a registry assertion covering the public IDs. The full self-hosted OpenArm planning-group E2E passes both tests.

dimos/robot/manipulators/openarm/model.py
mustafab0, line 1: Why not part of config?

dimos/robot/manipulators/openarm/config.py
```diff
@@ -12,137 +12,115 @@
...
+
+def openarm_hardware() -> HardwareComponent:
+    """Use mock hardware unless a physical right-arm CAN interface is set."""
+    adapter_type = "openarm_damiao" if global_config.can_port else "mock_whole_body"
+    adapter_kwargs: dict[str, object] = {}
+    if global_config.can_port:
+        adapter_kwargs["runtime_config"] = DamiaoRuntimeConfig(
+            bus_addresses={"right": global_config.can_port},
+            gravity_comp=True,
+        )
```
mustafab0, line 105: Why does runtime config only bind the right arm?

Since bimanual model is supported as the only model, this should bind both sides

mustafab0 (changes requested):

dimos/robot/manipulators/openarm/config.py
mustafab0: change requested for this

dimos/robot/manipulators/openarm/model.py
TomCC7: Addressed in fffad07173: the pinned official description source, bimanual Xacro, and RobotModel now live in openarm/config.py. model.py was removed, and the hardware adapter/tests import the shared model from config.

dimos/robot/manipulators/openarm/config.py
TomCC7: Addressed in fffad07173: OpenArm now requires an explicit --left-can-port/--right-can-port pair for physical hardware and binds both sides in DamiaoRuntimeConfig. Supplying neither keeps the mock default; supplying only one raises a configuration error. The focused OpenArm tests and cumulative restacked OpenArm/OpenYAM suite pass.

mustafab0: > supplying only one raises a configuration error.

should it? I actually dont see a use case for just one arm connection but could be useful in some edge cases to be able to only use 1 arm?

dimos/robot/manipulators/openarm/blueprints/basic.py
```diff
@@ -37,19 +37,37 @@ def _trajectory_task() -> TaskConfig:
...
+class _OpenArmCoordinator(ControlCoordinator):
+    """Select mock or explicitly addressed dual-CAN OpenArm hardware."""
+
+    config: _OpenArmCoordinatorConfig
+
+    def _setup_from_config(self) -> None:
+        self.config.hardware = [
+            openarm_hardware(
+                left_can_port=self.config.left_can_port,
+                right_can_port=self.config.right_can_port,
```
mustafab0: I dont think subclassing here is necessary. The list [hardware] is already designed for this

mustafab0 (dismissed):

dimos/robot/manipulators/openarm/blueprints/basic.py
TomCC7: Addressed in 057f4bcd26. The custom coordinator/config subclasses are removed; both blueprints now pass hardware=[openarm_hardware()] directly to ControlCoordinator. The left/right CAN values are standard global CLI fields, which the CLI resolves before importing and constructing the blueprint.

dimos/robot/manipulators/openarm/config.py
TomCC7: I kept the all-or-none check because OpenArmDamiaoAdapter currently owns one fixed bimanual topology and always constructs both CAN buses, both arms, and both grippers. Treating an omitted side as a default would silently connect an unintended interface. Supporting a true single-arm physical mode would need an explicit single-arm topology/model rather than partial configuration of this bimanual adapter.

mustafab0 (approved):

PR #3391, opened 2026-08-07
cuVSLAM 

jeff-hykin:
### Test

Plug in a realsense (455 or 435if)

```
dimos --viewer rerun run demo-cuvslam-realsense
```

Walk the camera around and watch the path in rerun

Expected performance (stereo with loop closure on no gating/rebasing)

Screenshot 2026-08-10 at 5 45 09 PM

ref DIM-1347


dimos/mapping/cuvslam/src/cuvslam_main.cpp
```diff
@@ -0,0 +1,681 @@
...
+    /// restart is: vision briefly had nothing to hold on to.
+    bool enable_imu;
+    /// Which transform on rig_transforms places the IMU rather than a camera.
+    std::string imu_frame;
+    /// Flattened from the python side's ImuCalibration, which is per-serial and has
+    /// no default -- all zero here means enable_imu is off and none of it is read.
+    double imu_gyro_noise_density;
+    double imu_gyro_random_walk;
+    double imu_accel_noise_density;
+    double imu_accel_random_walk;
```
jeff-hykin, line 114: these need to be removed

dimos/mapping/cuvslam/src/cuvslam_main.cpp
```diff
@@ -0,0 +1,681 @@
...
+}
+
+}  // namespace
+
+struct CuvslamConfig {
+    /// "stereo", "mono" or "rgbd". Mono is accurate only up to an unknown scale, so
+    /// its poses are not metres.
+    std::string camera_mode;
+    /// The stereo pair arrives rectified: D all zero, R identity.
+    bool rectified;
```
jeff-hykin, line 75: needs more clear naming

dimos/mapping/cuvslam/src/cuvslam_main.cpp
```diff
@@ -0,0 +1,681 @@
...
+    std::string imu_frame;
+    /// Flattened from the python side's ImuCalibration, which is per-serial and has
+    /// no default -- all zero here means enable_imu is off and none of it is read.
+    double imu_gyro_noise_density;
+    double imu_gyro_random_walk;
+    double imu_accel_noise_density;
+    double imu_accel_random_walk;
+    /// The rate actually fed. Declaring more than arrives makes cuVSLAM log a drop
+    /// ratio and silently never initialise inertial alignment.
+    double imu_frequency;
```
jeff-hykin, line 116: I want to remove this

dimos/robot/assembly/mid360_realsense_30.py
```diff
@@ -64,54 +63,28 @@
...
 # IMU position in point-cloud (lidar) coordinates, from Livox Mid-360 extrinsics.
 IMU_IN_LIDAR = (0.011, 0.02329, -0.04412)
 
-# The physical mount tree (parent -> child). The gravity-flat "world" helper frame from
-# the offline tooling is omitted here — during recording, world comes from odometry.
+# The rig, hung off the camera's tripod screw. Everything from that screw down into the
+# camera belongs to RealSenseCamera, which reads it off the device: a second publisher
+# here would fight it, and with numbers for whichever model the table was written for.
 FRAMES: list[FrameSpec] = [
-    ("bottom_screw_frame", None, (0.0, 0.0, 0.0), (0.0, 0.0, 0.0)),
```
jeff-hykin, line 100: all this should've been in the camera module (needed by cuVSLAM)

dimos/mapping/cuvslam/flake.nix
```diff
@@ -0,0 +1,131 @@
...
+        lcm = lcm-extended.packages.${system}.lcm;
+        cuda = pkgs.cudaPackages;
+
+        # NVIDIA's official prebuilt C++ SDK. Pinned by hash rather than vendored
+        # into the repo: the NVIDIA Community License permits redistribution, but
+        # a 111 MB binary does not belong in git.
+        #
+        # Version note: both are the cuda12.6 build, chosen to match the pycuvslam
+        # 17.0.0 wheel already in use (same version, same bundled libcudart
+        # 12.6.77) so C++ and Python results are comparable.
```
jeff-hykin: I think this needs to be more generic (not just cuda 12.6)

dimos/hardware/sensors/camera/realsense/camera.py
```diff
@@ -440,7 +815,14 @@ def stop(self) -> None:
 
         self._profile = None
         self._align = None
-        self._color_to_depth_extrinsics = None
+        if self._dropped_frames or self._repeated_frames:
```
jeff-hykin, line 697: repeated frames are a sucky part of the realsense API  https://github.com/realsenseai/librealsense/issues/1686#issuecomment-388606174

dimos/hardware/sensors/camera/realsense/camera.py
```diff
@@ -76,42 +90,94 @@ class RealSenseCamera(DepthCameraHardware, Module, perception.DepthCamera):
     config: RealSenseCameraConfig
     color_image: Out[Image]
     depth_image: Out[Image]
+    infrared_left: Out[Image]
+    infrared_right: Out[Image]
```
jeff-hykin, line 96: (1) Need left/right camera for stereo slam

dimos/hardware/sensors/camera/realsense/camera.py
```diff
@@ -76,42 +90,94 @@ class RealSenseCamera(DepthCameraHardware, Module, perception.DepthCamera):
...
     def _depth_optical_frame(self) -> str:
-        return f"{self.config.camera_name}_depth_optical_frame"
+        return f"{self._name}_depth_optical_frame"
+
+    @property
+    def _infra1_frame(self) -> str:
+        return f"{self._name}_infra1_frame"
+
+    @property
+    def _infra1_optical_frame(self) -> str:
```
jeff-hykin, line 130: (3) frames for stereo cameras to measure space between lenses

dimos/hardware/sensors/camera/realsense/camera.py
```diff
@@ -76,42 +90,94 @@ class RealSenseCamera(DepthCameraHardware, Module, perception.DepthCamera):
     config: RealSenseCameraConfig
     color_image: Out[Image]
     depth_image: Out[Image]
+    infrared_left: Out[Image]
+    infrared_right: Out[Image]
+    imu: Out[Imu]
```
jeff-hykin, line 97: (2) need IMU for VIO

dimos/hardware/sensors/camera/realsense/camera.py
```diff
@@ -76,42 +90,94 @@ class RealSenseCamera(DepthCameraHardware, Module, perception.DepthCamera):
...
+    # A slow start is normal; a minute of nothing is a real fault.
+    MAX_CONSECUTIVE_TIMEOUTS = 60
 
     def __init__(self, *args, **kwargs) -> None:  # type: ignore[no-untyped-def]
         super().__init__(*args, **kwargs)
         self._pipeline: rs.pipeline | None = None
         self._profile: rs.pipeline_profile | None = None
+        self._imu_pipeline: rs.pipeline | None = None
+        # Bounded so a stalled accelerometer cannot grow the queue.
+        self._accel_history: deque[tuple[float, tuple[float, float, float]]] = deque(maxlen=2)
```
jeff-hykin, line 155: accel and gyro don't show up at the same time, interpolate (it matters) to get them synced for an IMU message

dimos/hardware/sensors/camera/realsense/camera.py
```diff
@@ -213,96 +405,288 @@ def _build_camera_info(self) -> None:
...
                 break
-
-            ts = time.time()
+            except RuntimeError:
+                # Raised on timeout as well as on a stop, and a slow first frameset is
+                # normal.
+                if not self._running:
+                    break
+                consecutive_timeouts += 1
+                if consecutive_timeouts == 1 or consecutive_timeouts % 5 == 0:
```
jeff-hykin: prevent quick restart from crashing from camera being busy

dimos/hardware/sensors/camera/realsense/camera.py
```diff
@@ -144,47 +210,174 @@ def start(self) -> None:
...
+        The two are sampled independently and never share a timestamp.
+        """
+        while self._pending_gyro and len(self._accel_history) == 2:
+            (start_ts, start), (end_ts, end) = self._accel_history
+            ts, angular = self._pending_gyro[0]
+            if ts > end_ts:
+                return  # no accelerometer sample past it yet
+            self._pending_gyro.popleft()
+            span = end_ts - start_ts
+            ratio = 0.0 if span <= 0.0 else max(0.0, min(1.0, (ts - start_ts) / span))
```
jeff-hykin, line 344: iterpolating acceleration to match gyro time

dimos/hardware/sensors/camera/realsense/camera.py
```diff
@@ -213,189 +407,282 @@ def _build_camera_info(self) -> None:
...
-        extrinsics: rs.extrinsics,
-        frame_id: str,
-        child_frame_id: str,
-        ts: float,
-    ) -> Transform:
-        rotation_matrix = np.array(extrinsics.rotation).reshape(3, 3)
-        quat = Rotation.from_matrix(rotation_matrix).as_quat()  # [x, y, z, w]
-        return Transform(
-            translation=Vector3(*extrinsics.translation),
-            rotation=Quaternion(quat[0], quat[1], quat[2], quat[3]),
```
jeff-hykin, line 266: This doesn't take into account the reference frame that the realsense is publishing the transform from. E.g. "turn right". Its only ~15mm off for the 435 but it becomes a problem for stereo SLAM. Gotta convert to body, apply, then apply body inverse.

dimos/hardware/sensors/camera/realsense/camera.py
```diff
@@ -213,189 +407,282 @@ def _build_camera_info(self) -> None:
...
-                rotation=self.config.base_transform.rotation,
-                frame_id=self.config.base_frame_id,
-                child_frame_id=self._camera_link,
-                ts=ts,
-            )
-            transforms.append(base_to_camera)
-
-        # camera_link -> camera_depth_frame (identity, depth is at camera_link origin)
-        camera_link_to_depth = Transform(
-            translation=Vector3(0.0, 0.0, 0.0),
```
jeff-hykin, line 346: All of these should be derived from the sensor data itself (and it is now in _mount_edges)

dimos/mapping/cuvslam/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
...
+  ${LCM_INCLUDE_DIRS}
+  ${CUVSLAM_INCLUDE_DIR}
+)
+target_link_libraries(cuvslam_odometry PRIVATE
+  dimos_native
+  ${LCM_LIBRARIES}
+  ${CUVSLAM_LIBRARY}
+)
+target_link_directories(cuvslam_odometry PRIVATE ${LCM_LIBRARY_DIRS})
+if(APPLE)
```
jeff-hykin, line 52: Yep, working on MacOS 😁 CuMetal project

dimos/mapping/cuvslam/cuvslam.py
```diff
@@ -0,0 +1,235 @@
...
+    return _DRIVER_LINK_DIR
+
+
+def driver_cuda_major() -> int:
+    """The CUDA major the installed driver supports, or 0 if there is no driver."""
+    # By absolute path too: a nix-built python ignores ld.so.cache, where the bare
+    # name resolves everywhere else.
+    candidates = ["libcuda.so.1"] + [
+        str(directory / "libcuda.so.1")
+        for directory in (*_DRIVER_ONLY_LIB_DIRS, *_HOST_LIB_DIRS)
```
jeff-hykin, line 99: I hate cuda

dimos/hardware/sensors/camera/realsense/camera.py
```diff
@@ -144,47 +200,170 @@ def start(self) -> None:
...
+
+    def _on_motion_frame(self, frame: rs.frame) -> None:
+        import pyrealsense2 as rs
+
+        motion = frame.as_motion_frame()
+        if not motion:
+            return
+        data = motion.get_motion_data()
+        # Hardware capture time, not host time.
+        ts = motion.get_timestamp() / MILLISECONDS_PER_SECOND
```
TomCC7: I feel like it's better to add a helper function like `ms_to_s` rather than barebone division

dimos/hardware/sensors/camera/realsense/camera.py
TomCC7: interesting that their driver does not handle alignment...

TomCC7: if gyro is faster why not match gyro to accel?

dimos/mapping/cuvslam/src/utils/transform.hpp
```diff
@@ -0,0 +1,85 @@
...
+
+// Quaternion and rigid-transform algebra shared by the cuVSLAM module sources.
+
+#pragma once
+
+#include <algorithm>
+#include <array>
+#include <cmath>
+
+namespace transform_math {
```
TomCC7: are we considering Eigen? Is that hard to package?

dimos/memory2/replay_module.py
TomCC7, line 1: might be the wrong place to put this too many realsense specific logic

dimos/mapping/cuvslam/src/cuvslam_main.cpp
```diff
@@ -0,0 +1,828 @@
...
+    double imu_gyro_noise_density;
+    double imu_gyro_random_walk;
+    double imu_accel_noise_density;
+    double imu_accel_random_walk;
+    /// The rate actually fed. Declaring more than arrives never initialises alignment.
+    double imu_frequency;
+    /// rgbd only: raw depth units per metre. 1000 for sixteen-bit millimetres.
+    double depth_units_per_meter;
+
+    void validate() const {
```
TomCC7: seems like we can validate this from python side?

dimos/mapping/cuvslam/src/cuvslam_main.cpp
```diff
@@ -0,0 +1,828 @@
...
+            }
+            tf_edges_[stamped.child_frame_id] = {stamped.header.frame_id,
+                                                 to_transform(stamped.transform)};
+        }
+        resolve_rig();
+    }
+
+    /// parent_frame -> child_frame through their nearest common ancestor. Not
+    /// time-aware; the mount tree is rigid.
+    std::optional<Transform> tf_lookup(const std::string& parent_frame,
```
TomCC7: pretty weird we don't have this as tf lib function? maybe we just add it?

dimos/mapping/cuvslam/src/cuvslam_main.cpp
```diff
@@ -0,0 +1,828 @@
...
+        return std::nullopt;
+    }
+
+    /// The frame cuVSLAM's rig is expressed in, which need not be the published frame.
+    const std::string& rig_frame() const {
+        return cfg_.rig_frame.empty() ? cfg_.base_frame : cfg_.rig_frame;
+    }
+
+    /// Place every camera against the rig frame, or nothing until they all resolve.
+    void resolve_rig() {
```
TomCC7: this whole function seems to be frame lookup? a bit dirty to me our tf lib should be able to handle frame name based lookup and time synchronization (not the pr's problem I'm just complaining lol

dimos/mapping/cuvslam/src/cuvslam_main.cpp
```diff
@@ -0,0 +1,828 @@
...
+        pending_imu_.push_back(measurement);
+        ++imu_samples_;
+    }
+
+    void on_image(const sensor_msgs::Image& img) {
+        const int index = camera_index(img.header.frame_id);
+        if (index < 0) {
+            ++unplaced_images_;
+            DIMOS_LOG_THROTTLED(logging::Level::Warn, logging::from_secs(10),
+                                "cuvslam dropping image with a frame_id not on the rig",
```
TomCC7: I'm actually quite confused here. All images are just from the same camera with only static transform difference right? We should not really have the case where only some of the camera frame is available

dimos/mapping/cuvslam/src/cuvslam_main.cpp
jeff-hykin: Yeah that's probably better.

dimos/mapping/cuvslam/src/cuvslam_main.cpp
jeff-hykin: Its been on Andrew's backlog for a bit. I feel like if I add a proper C++ tf it probably won't get merged. Maybe I should just do it. 

Normally I just convert everything to rust to use tf, but my cuvslam-rs isnt done yet :shruggie:

dimos/mapping/cuvslam/src/cuvslam_main.cpp
jeff-hykin: CuVSLAM supports multi-camera setups. They do need to have static transforms between each other but my understanding is they can be different models of camera.

But, even on one realsense (4 lenses) we still want to filter: the config picks if I want to run it in stereo mode, or mono, or rgb-d, or just rgb, and that changes which lenses get used.

We could avoid connecting the streams with .remappings but then it means I can't pick mono/stereo/rgb/rgb-d from the config or cli, because remappings is at the blueprint level. We would need a different blueprint per config.

dimos/memory2/replay_module.py
jeff-hykin: Oops, this whole file wasn't supposed to be committed at all.

dimos/mapping/cuvslam/src/utils/transform.hpp
jeff-hykin: That's a good idea. Cuvslam already includes it anyways

dimos/hardware/sensors/camera/realsense/camera.py
jeff-hykin: Most of the error for slam is yaw, so I think its better to keep gryo as-is (full accuracy) and interpolate linear motion. 

Also yeah its kinda crazy they dont do this for us.

jeff-hykin: Even more crazy Intel doesnt sync the IMU and camera clock. They both have hardware timing but they're different clocks! Realsense converts them to system time independently.

dimos/mapping/cuvslam/src/cuvslam_main.cpp
TomCC7: Actually maybe worth checking how cli override work... I'm not even sure why we have this validate API. Seems we can always  verify config from python side which will be just easier?

jeff-hykin: Maybe if theres tooling in the C++ that does the validating (image format check, or like cuda version check)? Idk when I did native module contributions it was python only, usually part of the post_init of the config class.

dimos/mapping/cuvslam/src/cuvslam_main.cpp
jeff-hykin: I ended up just vibing a `tf_tree.hpp` that is the full get/publish/listen system. It intentionally doesnt use Eigen tho so later it can work for every C++ module. 

It belongs in native/cpp but to get the PR I'm going to start with in only inside cuVSLAM

dimos/mapping/cuvslam/cuvslam.py
```diff
@@ -0,0 +1,268 @@
...
+    # Pose graph and loop closure; without it map->odom is identity.
+    enable_slam: bool = True
+    # Runs Slam on its own thread. Its GetPose() carries no timestamp, so a thread running
+    # behind cannot be matched to the odometry pose it corrects.
+    slam_async: bool = False
+    # Poses in the pose graph, not a distance. 0 is unlimited.
+    slam_max_poses: int = 300
+    slam_throttling_ms: int = 0
+    enable_imu: bool = False
+    imu_calibration: ImuCalibration | None = None
```
jeff-hykin: Meh, I don't like ImuCalibration in config. We should just use camera_info

PR #3407, opened 2026-08-08
Control per instance addressing

mustafab0:
## Problem

To steer a specific arm, teleop and jogger publishers write the name of the
consuming control task into each pose/twist message's `frame_id` field, and the
coordinator reads it back per message to pick the task. 

Issue: #

---

## Solution

Each cartesian / EEF-twist task instance reads its own coordinator input port, connected once in the blueprint.
Single-arm deployments declare a port named after the card's input; the dual-arm quest teleop declares one port per arm and binds each task to its side with `stream_bind`. 

---

## Breaking Changes

The shared coordinator ports for cartesian and EEF-twist commands are deleted.
Out-of-tree blueprints composing these tasks must declare the port on a
coordinator subclass and point their producer remap at it (one line each).

---

## How to Test

1. `pytest dimos/control dimos/teleop dimos/robot/manipulators -m 'not (tool or self_hosted or mujoco or self_hosted_large)'`
2. On the rig: quest single-arm teleop (xarm7) — engage, track, gripper trigger.
3. Quest dual-arm — move one controller; the other arm must not move.
4. Keyboard jog and hosted browser twist — jog, release stops motion, gripper toggle.

### DIFF
--- a/dimos/robot/manipulators/a750/blueprints/teleop.py
+++ b/dimos/robot/manipulators/a750/blueprints/teleop.py
@@ -16,14 +16,18 @@
 
 from __future__ import annotations
 
-from dimos.control.coordinator import ControlCoordinator
 from dimos.core.coordination.blueprints import autoconnect
 from dimos.manipulation.manipulation_module import ManipulationModule
 from dimos.robot.manipulators.a750.config import (
     a750_hardware,
     make_a750_model_config,
 )
-from dimos.robot.manipulators.common.blueprints import eef_twist_task
+from dimos.robot.manipulators.common.blueprints import (
+    eef_twist_task,
+)
+from dimos.robot.manipulators.common.coordinators import (
+    ArmTwistCoordinator,
+)
 from dimos.teleop.keyboard.keyboard_teleop_module import KeyboardTeleopModule
 
 _a750_hw = a750_hardware("arm", mock_without_address=True)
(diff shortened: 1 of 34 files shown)


dimos/robot/manipulators/a750/blueprints/teleop.py
```diff
@@ -16,22 +16,25 @@
...
-from dimos.control.coordinator import ControlCoordinator
 from dimos.core.coordination.blueprints import autoconnect
 from dimos.manipulation.manipulation_module import ManipulationModule
 from dimos.robot.manipulators.a750.config import (
     a750_hardware,
     make_a750_model_config,
 )
-from dimos.robot.manipulators.common.blueprints import eef_twist_task
+from dimos.robot.manipulators.common.blueprints import (
+    _EEFTwistCommandCoordinator,
```
paul-nechifor, line 26: This shouldn't start with an underscore.

dimos/robot/manipulators/common/blueprints.py
```diff
@@ -33,6 +36,24 @@
 )
 
 
+class _CartesianCommandCoordinator(ControlCoordinator):
```
paul-nechifor, line 39: This file is for blueprints. These should be in order places... Also, they're not used there. They're public so shouldn't contain `_`.

ruthwikdasyam:
Teleop changes verified! All good

dimos/robot/manipulators/a750/blueprints/teleop.py
mustafab0: Thank you, I was unsure about the naming of these too. I will fix

paul-nechifor (approved):

PR #3411, opened 2026-08-09
feat(evals): passive/interactive agent eval framework over memory2

spomichter:
## Problem
Eval framework, supports both InteractiveEvals and PassiveEvals, @paul-nechifor old dimsim spatial memory eval replicated via the new framework here pretty clean: 
```python
BED = Vector3(-3.567, -1.332, 0.0)

go_to_bed = InteractiveEval(
    id="go_to_bed",
    inputs="go to the bed",
    score=lambda s: ramp((BED - s.streams.odom.last().data.position).length(), band=2.0),
    aggregate=final,          # or floor ("never left the zone") / mean / auc later
    blueprint="unitree-go2-agentic go2-memory",
    simulator="dimsim",
    scene="apartment",
)
```
Uses langchain openevals for VQA formatting so its compatible with all public benchmarks we'd want to run. Scoring is generic and done via lambda one-liners. Agent loop runs LLM queries with a light agent_loop in the EvalRunner -- kept minimal and using langchain client since just doing dumb QA. 

Closes DIM-1392, DIM-1390

## Solution
- new `dimos/evals/` package. one word (evals) everywhere, no benchmark vs eval split
- two case types: `PassiveEval` (frozen mem2 recording, any replay window, non-autoregressive) and `InteractiveEval` (live sim/robot, actions mutate state, scored by sampling teh live recorder store)
- memory2 the source of truth for all eval input — context selectors return real `Stream`s, no copies, no parallel data structres
- scoring is jsut lambdas over typed msg arithmetic (`within`, `ramp`, Vector3 subtraction etc), graded [0,1] credit not only pass/fail
- we use the langchain/openevals standard for vqa formatting (`inputs` / `reference_outputs`, evaluators wrapped not subclassed) so external benchmarks map on natively and we can test against them
- generic typing: `PassiveEval[T]` ties `expected`/`parse`/`score` togehter so mypy catches a mismatched case at suite definition time, not mid-run
- runnable via cli (`dimos evals run ...`), mcp (`EvalModule` skills so coding agents can iterate), and pytest
- `--blind` ablation built in — same suite with observations withheld, alredy caught two guessable MCQs
- `MoondreamChat` adapter so evals run keyless on any gpu box

### defining evals

passive one-off over a replay — the whole thing is one literal:

```python
from dimos.evals.scorers import first_number, within, yes_no, exact
from dimos.evals.types import PassiveEval, InteractiveEval, Suite

# numeric, graded: 1.0 exact, linear to 0 at band
distance = PassiveEval(
    id="distance_10min",
    inputs="How far have you traveled in the last 10 minutes, in meters?",
    expected=142.0,
    parse=first_number,
    score=within(15.0),
    context=(lambda s: s.streams.odom.range_time(0, 600),),   # real mem2 stream
    dataset="go2_hongkong_office",
)

# vqa over a 10-image window
person = PassiveEval(
    id="person_visible",
    inputs="Is a person visible in any of these images?",
    expected="yes",
    parse=yes_no,
    score=exact,
    context=(lambda s: s.streams.color_image.range_time(58, 61).limit(10),),
    dataset="go2_short",
)
```

interactive — the case names its environment, score is sampled from the live mem2 store every `interval_s` and reduced by `aggregate`:

```python
from dimos.evals.scorers import final, ramp
from dimos.msgs.geometry_msgs.Vector3 import Vector3

BED = Vector3(-3.567, -1.332, 0.0)

go_to_bed = InteractiveEval(
    id="go_to_bed",
    inputs="go to the bed",
    score=lambda s: ramp((BED - s.streams.odom.last().data.position).length(), band=2.0),
    aggregate=final,          # or floor ("never left the zone") / mean / auc later
    blueprint="unitree-go2-agentic go2-memory",
    simulator="dimsim",
    scene="apartment",
)
```

### running

```python
from dimos.evals.runner import EvalRunner, summarize

results = EvalRunner().run(SUITE)                    # prod model config (gpt-5.6-luna)
results = EvalRunner(blind=True).run(SUITE)          # guessing ablation
results = EvalRunner(chat_model=MoondreamChat()).run(SUITE)  # keyless local
print(summarize(results))
```

or `dimos evals run dimos.evals.suites.go2_smoke --blind`. every run writes results.jsonl + summary.json + per-case transcripts to `~/.local/state/dimos/evals/run-*/`.

### first real numbers
luna sighted: examples 1.00 / smoke 1.00 / vqa 0.86, blind = refusals. interactive rig scored 0.856 on a scripted go-to-bed in dimsim. the full agentic go-to-bed scores 0.0 right now — `unitree-go2-agentic` publishes no /odom in dimsim on current main (upstream test_dimsim_spatial_memory fails teh same way, suspect the control coordinator refactor). separate ticket coming.

## Breaking Changes
None

## How to Test
```
uv run pytest dimos/evals dimos/codebase_checks -q          # offline unit + wiring tests, no keys
uv run pytest dimos/evals/test_smoke.py -m self_hosted      # live model smoke (needs OPENAI_API_KEY + lfs data)
dimos evals run dimos.evals.suites.examples                 # 2 doc cases against go2_short
dimos evals run dimos.evals.suites.go2_smoke --blind        # guessing ablation
```
interactive (needs deno + display): `dimos evals run dimos.evals.suites.dimsim_house --live-db recording_go2.db`

- [x] I have read and approved the CLA


spomichter:
@greptile review

TomCC7:
We need to design runtime and scorer information access more carefully. The current scheme is wrong IMO because it evaluates against Memory2 state produced by the runtime itself. To simulate real operation as faithfully as possible, the agent should have access to all information exposed by the running blueprint, including RPCs, streams, and Memory2. Scoring should instead use independent, privileged simulator state, such as ground-truth robot and object poses and object types.

mustafab0 (dismissed):

dimos/evals/intro.md
```diff
@@ -0,0 +1,202 @@
+# Evals Intro
```
paul-nechifor: I think this should be in `/docs/` as `dimos evals` is a command that's shipped to everyone.

spomichter: okay yes

dimos/evals/runner.py
```diff
@@ -0,0 +1,398 @@
...
+    "text": "[observations withheld — answer anyway]",
+}
+
+
+class EvalRunnerConfig(BaseConfig):
+    model: str = "gpt-5.6-luna"  # mirrors McpClientConfig.model
+    # House convention (StoreConfig): pass an instance to inject, e.g. a fake
+    # chat model in tests. None -> built from `model` like McpClient does.
+    chat_model: Any | None = None
+    mcp_url: str = "http://localhost:9990/mcp"
```
paul-nechifor: `mcp_port` is in global_config. You should probably use that.

dimos/evals/runner.py
```diff
@@ -0,0 +1,398 @@
...
+    live_db: str = "recording.db"  # store the Recorder writes (interactive)
+    blind: bool = False  # ablation: context withheld (SPACE guessing check)
+    threshold: float = 1.0  # passed = score >= threshold
+    strict: bool = False  # preflight failure aborts the whole run
+    context_budget: int = 8  # max observations encoded per context Select
+    attach: bool = False  # True: drive an already-running dimos
+    launch_timeout_s: float = 1200.0  # blueprint + MCP readiness (e2e parity)
+    out_dir: Path = Path("~/.local/state/dimos/evals").expanduser()
```
paul-nechifor: This should use `STATE_DIR / "evals"` with `STATE_DIR` from constants.py

dimos/evals/runner.py
```diff
@@ -0,0 +1,398 @@
...
+                if value is not None:
+                    series.append((time.monotonic() - t0, value))
+                    if value >= 0.999:  # ponytail: early exit on success; drop if
+                        break  # aggregates ever need the full window
+                time.sleep(interval_s)
+        finally:
+            store.stop()
+        return series
+
+    def _wait_live_store(self, deadline: float) -> Store:
```
paul-nechifor, line 397: I don't like the way this works. It assumes that a store is on disk and also that it's a file. That's very specific to Sqlite. Other DBs don't create a single file for a DB.

paul-nechifor: I think what should happen here is that `class Store` should be changed to have something like `Store.wait_for_creation` (or a better name) and this sort of internal stuff should be performed in `SqliteStore.wait_for_creation`  instead of in unrelated places like here.

dimos/evals/runner.py
```diff
@@ -0,0 +1,398 @@
...
+
+    def _wait_mcp(self, timeout: float) -> bool:
+        from dimos.agents.mcp.mcp_adapter import McpAdapter
+
+        return McpAdapter(self.config.mcp_url).wait_for_ready(timeout=timeout, interval=2.0)
+
+    def instruct(self, text: str) -> None:
+        from dimos.core.transport import pLCMTransport
+
+        transport: pLCMTransport[str] = pLCMTransport("/human_input")
```
paul-nechifor: You should use `transport = make_transport("/human_input")` since we shouldn't be strictly tied to LCM now.

dimos/evals/runner.py
```diff
@@ -0,0 +1,398 @@
...
+        transport.lcm.start()
+        try:
+            transport.publish(text)
+            time.sleep(0.5)  # let LCM flush before teardown
```
paul-nechifor: Are you sure this is an issue? I don't think I've encountered it. If it is, we should have a `transport.flush()` rather than a random wait as it would be a problem with transports in general, not just here.

dimos/evals/runner.py
```diff
@@ -0,0 +1,398 @@
...
+        finally:
+            transport.lcm.stop()
```
paul-nechifor: No need to call `transport.lcm.start()/.stop()` anymore. Just use `transport.start()/.stop()`

dimos/evals/runner.py
```diff
@@ -0,0 +1,398 @@
...
+                        final_text = str(msg.content)
+        return final_text
+
+    # -- EvalRig: interactive ----------------------------------------------------------
+
+    def setup_env(self, case: InteractiveEval) -> None:
+        from dimos.evals.types import _no_setup
+
+        if case.simulator and not self.config.attach:
+            from dimos.e2e_tests.dimos_cli_call import DimosCliCall
```
paul-nechifor, line 319: Since `dimos_cli_call` is used in `dimos/evals` now too, it should be moved out of `dimos/e2e_tests` as it's no longer just for end to end testing. Maybe somewhere in `dimos/cli`?

dimos/evals/runner.py
```diff
@@ -0,0 +1,398 @@
...
+
+            proc = DimosCliCall()
+            proc.simulator = case.simulator
+            proc.global_args = ["--dimsim-scene", case.scene]
+            proc.demo_args = ["run", *case.blueprint.split()]
+            proc.start()
+            self._proc = proc
```
paul-nechifor, line 326: The architecture here seems a bit unclear to me. 

If I understand this correctly, InteractiveEval gets an EvalRunner and sets itself up on it by calling `eval_runner.setup_env(self)`

I see EvalRunner as a parent object of multiple InteractiveEval objects. It's generally bad for children to control their parent object instead of the other way arround.

For example, `EvalRunner.setup_env` sets `self._proc = DimosCliCall()`. This creates a new process for every eval. When `EvalRunner.stop` is called, only the last `self._proc` is ever stopped.

This issue is obscured by the fact that an eval runs itself on the eval runner instead of the eval runner being in charge of running the evals. That is, the fact that `EvalRunner._proc` is leaked is hard to see because it's ultimately set from `InteractiveEval.evaluate`.

But also, `_proc` probably doesn't belong on EvalRunner. It should be on InteractiveEval. Evals should probably have a life cycle where they can be created and destroyed.

---

Edit: So EvalRunner.run does in fact run the evals, but also the evals know too much about the eval runner and store state on it because they can't store state on themselves.

dimos/evals/runner.py
spomichter: True good call -- ill do a following PR just to keep this PR clean and consoludated

dimos/evals/runner.py
spomichter: evalrunner runs the interactive/passive evals. not a parent.

spomichter: you epxosed a process issue though adding per eval case teardown

mustafab0 (approved):

Issue #3399, opened 2026-08-08
Integrate a SPACE benchmark task into the agent-eval framework


TomCC7:
# Integrate a SPACE benchmark task into the agent-eval framework

## Context

[DimOS PR #3378](https://github.com/dimensionalOS/dimos/pull/3378) introduces the initial agent-evaluation framework and a direct frozen-memory QA case. This follow-up should use that framework to evaluate the DimOS agent on an existing external benchmark.

SPACE (*Spatial Perception And Cognition Evaluation*) tests spatial reasoning through text, images, and interactive tasks. The benchmark includes both large-scale environmental reasoning and small-scale object/layout reasoning.

## Objective

Integrate one representative SPACE task into the framework from PR #3378 and run the DimOS agent on a bounded, reproducible subset of that task.

Analyze the benchmark, select an appropriate first task and presentation, and explain the selection.

## Scope

- Build on the contracts and execution path introduced by PR #3378; do not create a separate evaluation system.
- Select one SPACE task that can be integrated and run end to end within a bounded effort.
- Preserve the benchmark's question, input, answer, and scoring semantics.
- Use SPACE's official evaluation and scoring implementation as-is, without modifying or reimplementing it. DimOS may adapt execution inputs and outputs only at the integration boundary.
- Provide a direct command that another developer can run and test.
- Run a fixed subset and summarize accuracy, invalid responses, and infrastructure failures.
- Produce an evaluation report that analyzes the agent's performance on the selected task and proposes evidence-based improvement directions.

## Deliverables

1. Benchmark/task selection note with the reasoning and known limitations.
2. Working integration using the PR #3378 framework, a PR that stacking on it.
3. Reproducible setup and run instructions.
4. An evaluation report covering results, observed strengths and failure modes, and proposed improvement directions.
5. Focused automated tests.

## Acceptance criteria

- A fresh checkout can acquire the required upstream data and run the documented command without manual data rewriting.
- The run produces per-case results and an aggregate score traceable to the selected SPACE examples.
- Final scores come directly from the unmodified official SPACE evaluator at the recorded upstream revision.
- The agent uses the DimOS evaluation execution path, not SPACE's built-in model wrappers.
- Provide evaluation report (as PR description) stating the model/configuration, subset selection, and official score; analyzes observed strengths and failure modes; and proposes concrete improvement directions tied to those observations.

## Data and licensing constraint

[...]

PR #3414, opened 2026-08-09
feat(evals): run the SPACE text benchmarks through the agent-eval path

mc856:
## Contribution path

- Linked issue: #3399 (integrate a SPACE benchmark task into the #3378 agent-eval framework)
- Branch note: opened stacked on `cc/feat/frozen-qa-eval` (#3378); I will rebase onto `main`
  once #3378 lands.

## Problem

The #3378 agent-eval framework scores agents only against internally authored suites with local
oracles. #3399 asks for an external benchmark — Apple's SPACE (spatial cognition) — wired through
the same execution path, so that an unmodified official scorer grades what a DimOS agent actually
does. The integration must not redistribute any SPACE content (ND-licensed data), and the seams it
adds should be reusable for the next external benchmark rather than SPACE-specific.

## Solution

**One seam.** SPACE's official code keeps everything it already owns — question loading, the
evaluation loop, answer parsing, scoring — and upstream `space.evaluate_qas.main` writes the
only `results.json` that counts. DimOS enters at exactly one point: `DimosQAAgent` registers in
SPACE's own agent registry and replaces the model call — nothing else — with an
`execute_single_case` run (the #3378 path, `python_exec` on an empty kernel). Local
cross-checks can only refuse a run, never alter a score.

**Agent-eval contract — three new generic case kinds.** No SPACE naming anywhere, so a future
external benchmark can reuse them unchanged:

- source `none` — nothing to materialize (no recording to replay);
- task `verbatim_prompt` — the benchmark's prompt passes through untouched (any injected
  instruction would fight SPACE's own JSON format directive);
- validator `external_evaluator` — no local oracle; scoring is delegated and recorded as
  `benchmark` + `revision`.

One change is **not** purely additive and is called out here deliberately:
`CompactEvalResult.recording`/`progress` relax from required to `Optional` — environment-less
cases have neither. Old result files still parse; new files write `null` there. Whether this
should bump `schema_version` is left as a maintainer decision (see below).

**Scope.** SPACE breaks down into **10 text QA tasks, 14 image QA tasks, and 4 interactive
task families**; the image tasks need an image input channel this eval path does not have yet,
and the interactive ones need a stepwise multi-turn environment loop rather than one-shot QA.
So this PR integrates exactly the **10 text tasks** — the other two slices ship as design only
(scope rationale in the #3399 comment; roadmap write-up available on request).

**`space_qa` package — the SPACE-specific side.** A small `BenchmarkAdapter` protocol
(`iter_items` / `to_case` / `score`) turns benchmark rows into eval cases. `BenchmarkItem` has
no answer field, so the answer key cannot enter an `EvalCase` by construction. Subset drawing
follows the sampling design below — whole answer-order groups on a recorded seed. A generated
grid suite is the second adapter implementation and the zero-data test tier.

**Data & provenance — fetched at run time, never committed.** Nothing SPACE-derived enters the
repository. On first run, the tool clones `apple/ml-space-benchmark` at a pinned revision and
downloads the official 3.6 GB data release into the user cache, checking the tarball's SHA256;
a checkout with local edits, or off the pin, is refused. Every run writes a manifest naming
exactly what it consumed: task, seed (given or drawn), group count, the upstream row ordinals,
a SHA256 per question, digests for the archive and for the question file the run actually read,
and the dimos revision — recorded only when the package really runs from this source tree, so
an installed copy under some unrelated checkout cannot get signed with that checkout's HEAD.
Anyone holding the official release can re-identify from the manifest exactly which items a run
graded; the outputs sit beside it — per-question records and upstream's own `results.json`.

**What SPACE ships, and what a run samples.** Each of the ten tasks is a pool of *stimuli*,
and SPACE counterbalances every stimulus into four answer-order variants — four consecutive
rows in the released question file, the correct answer rotated into a different slot in each.
The pools are uneven: from 30 stimuli (`MapSketchingBEVText`) to 150
(`DirectionEstimationBEVText`), **755 stimuli / 3,020 questions across the ten** (per-task
sizes in the results table below). Scoring the whole release is tens of hours of agent time at
the observed cadence, so a run samples: `--groups N` sets how many stimuli it draws (4N
questions), `--seed` sets which ones — omitted, a seed is drawn, printed beside the score, and
written to the manifest, so every run is reproducible from its record; passed explicitly, it
turns two runs into a comparison. Upstream itself has no sampling — `evaluate_qas` scores
every row it is handed — and `--groups <pool size>` reproduces exactly that.

Two properties of the release, both re-checkable by anyone holding it, decided *how* to sample:

- **Whole groups, never rows.** In the three BEV tasks the answer slot inside every group
  follows the same ①→④ order — 600 of 600, 540 of 540, 120 of 120 rows — so a row-wise subset
  has its answer in one fixed slot, and its score cannot be told apart from a slot preference.
  Whole groups keep a stimulus's four variants together and make the answer-position
  distribution flat by construction; they also fix the unit of evidence — 32 rows are 8
  independent stimuli, and inference below is per stimulus.
- **A seeded random draw, not the first N.** Rows are blocked by scene (`Direction`'s 150
  stimuli are 30 scenes × 5 consecutive groups), so the first 8 groups of that file cover 2
  scenes; a seeded random 8 covers ~7. The seed is a budget knob, not part of the benchmark.

**Results — an integration-validation profile, not a controlled benchmark.** All ten tasks
were scored at the same subset size — `--groups 8 --seed 20260808`: 8 stimuli × 4 variants =
32 questions per task, **320 questions, 80 of the release's 755 stimuli (10.6%)** — model
`gpt-5.6-luna` through this path. Accuracy, exactly as SPACE's unmodified scorer wrote it to
`results.json` (infrastructure failures count as misses): **90.94%** (291/320) — macro and
micro alike, since every task has the same n. The paper's GPT-4o macro-mean on the same ten tasks is
**59.8** — context for scale, not a comparison this profile can support: the paper's baselines
are bare no-tool chat models while this path carries `python_exec`, and these runs accumulated
over the course of development, against more than one locally configured OpenAI-compatible
endpoint (the full accounting is in the companion report).

| Task | stimuli sampled | questions | Accuracy (SPACE scorer) | GPT-4o (paper) |
|---|---|---|---|---|
| `SAtt_text` | 8 of 100 | 32 | **100.00** | 98.8 |
| `MPFB_text` | 8 of 50 | 32 | **100.00** | 50.5 |
| `MapSketchingBEVText` | 8 of 30 | 32 | **96.88** | 46.7 |
| `CBTT_text` | 8 of 50 | 32 | **96.88** | 82.5 |
| `SAdd_text` | 8 of 50 | 32 | **93.75** | 93.5 |
| `MRT_text` | 8 of 40 | 32 | **93.75** | 41.9 |
| `DistanceEstimationBEVText` | 8 of 135 | 32 | **93.75** | 33.3 |
| `PTT_text` | 8 of 100 | 32 | **81.25** | 55.5 |
| `JLO_text` | 8 of 50 | 32 | **81.25** | 66.5 |
| `DirectionEstimationBEVText` | 8 of 150 | 32 | **71.88** | 28.7 |

Completed: **315 of 320** — two transient HTTP 503s and three sessions that ended
`terminated`. Accuracy among completed replies: **92.38%** (291/315). Malformed completed
replies: **0** — the upstream parser read an integer out of every reply that arrived.
Direction carries two of the five losses: 23 of its 30 completed answers were correct (76.67%).
Three observations worth carrying out of the per-item data (all of it in the companion report,
available on request):

- `JLO_text`'s six misses are one stimulus answered wrong on all four of its answer orders plus
  two single-variant misses — a stable per-stimulus failure, not slot noise.
- Tool use tracks question shape: one hundred questions across five tasks made not a single
  `python_exec` call, while the three BEV tasks account for 119 of the profile's 144.
- Removing the tool changed nothing measurable on the one task with enough errors to look at:
  the same 32 Direction questions, same channel, behind a local patch not in this PR — 71.88
  with the tool, 71.88 without, 37% faster without. Read it as *no benefit detected*, not *no
  benefit* (8 stimulus groups, sign test p = 1.0; the two runs differ in more than the tool —
  caveats in the companion report).

**Coverage is the boundary on every claim above.** Eight stimuli per task is a quarter of
`MapSketching`'s pool and a nineteenth of `Direction`'s; nothing here is a full-benchmark
result. `--groups` is the whole knob — the subset a task is given is everything that task ran,
and full coverage is a spending decision, not a code change.

**Known limitations** (each disclosed on purpose; the first one wants your read):

- `python_exec` runs unsandboxed on the host that also holds the answer key on disk; dropping
  the two pointer env vars is defence in depth, not isolation. Every completed transcript under
  the scored-run tree was audited — 419 of them, carrying 199 `python_exec` calls: zero
  filesystem, environment, subprocess, or network access, no imports beyond `numpy`, `math`,
  `itertools`, `collections`. That certifies these runs, not the runner. The structural fix is
  a per-case tool surface (also what a paper-comparable no-tool run and vision need); until
  then the benchmark leans on after-the-fact audit — whether that interim is acceptable is
  your call.
- The system prompt is chosen by task kind alone; unsupported source/task/validator
  combinations are not rejected at construction. A capability preflight is the right general
  mechanism and is left as a follow-up rather than grown ad hoc here.
- A run where some questions fail on infrastructure still exits 0; the summary prints
  `Infra fails N of M` next to the accuracy it distorts, and an all-failures run is refused. A
  bounded per-question retry (original failure kept in the record) would re-ask exactly the
  five questions this profile lost; left out because it changes what a score means.
- Two bookkeeping asymmetries, both stated in the docs: the manifest carries no dirty flag for
  the dimos tree (the SPACE checkout, by contrast, is refused outright when dirty), and the
  run-root `cases.jsonl` ledger is written only after the three consistency gates — a refused
  run keeps its per-question records; only the aggregated file is missing.

### One decision requested

- `schema_version`: this PR changes `CompactEvalResult.recording`/`progress` from **required to
  optional** (environment-less cases have neither, so new result files may hold `null` there —
  a value the 1.0 shape never produced). Kept at **1.0** for now — readers that treat the two
  fields as optional stay compatible, and this PR discloses the change here rather than hiding
  it. Say the word and I will bump it to 1.1 in this PR.

### Roadmap

Both remaining SPACE slices keep the one-seam invariant.

**Vision is next, and is the prerequisite for the VQA benchmark that follows SPACE.** One more
task kind (`multimodal_prompt`, parts instead of a string); the Pi CLI already accepts image
files as positional `@file` arguments, so the change concentrates in the transport layer. Once
that channel exists, SPACE's own `*_vision` variants become a zero-integration regression load
for it — same adapter, same scorer, different presentation.

**Interactive (route retracing, shortcut discovery, MCT, CSWM) needs one decision before any
code**: expose the benchmark environment as a tool inside one persistent session, or drive one
agent call per step from the upstream script. The trade-offs are in the companion report — I
would rather argue them against your pick than pre-empt it.

Smaller follow-ups — a configurable tool surface, the bounded infra retry, capability
preflight, a wording pass on the shared `python_exec` description — will each get their own
issue.

## How to Test

Offline (no SPACE data, no network, no API key):

    uv run pytest dimos/benchmark/space_qa dimos/benchmark/agent_eval dimos/agents/test_code_policy_core.py

The feature itself (first run clones the pinned source and downloads the 3.6 GB release into the
user cache):

    uv sync --extra space
    npm ci --prefix packages/pi-code-policy-extension
    npm run build --prefix packages/pi-code-policy-extension
    export OPENAI_API_KEY=...

One command, four sizes:

    dimos eval space --task SAtt_text --groups 1         # smoke: one stimulus, 4 questions — checks the wiring end to end
    dimos eval space --task SAtt_text                    # default: 8 groups (32 questions), seed drawn and recorded
    dimos eval space --task SAtt_text --seed 20260808    # pinned subset: reproduces this PR's SAtt_text run
    dimos eval space --task SAtt_text --groups 100       # the whole pool: upstream's own full evaluation, no sampling

Timing, measured on the runs behind the results: at `--workers 2` a 32-question subset ranged
from about a minute of wall clock (`SAtt_text`) to about half an hour
(`DirectionEstimationBEVText` — the long BEV prompts dominate), and the full ten-task profile
came to roughly an hour and a half. The smoke and the offline tier are seconds-scale.

Scores print from upstream's own `results.json`; the run directory holds the manifest,
per-question records, and provenance. Docs: `docs/capabilities/agents/space-benchmark.md`.

## AI assistance

Claude Code (Fable 5 and Opus 5) wrote most of the implementation under my direction, and the
branch went through several adversarial review rounds with Claude Code and Codex before
submission. I can explain every line in this diff.

## Checklist


paul-nechifor:
Closed for now. Can be reopened later.

PR #3415, opened 2026-08-09
feat(evals): pointcloud VQA suite + PointCloud2.agent_encode

spomichter:
## Problem
No way to measure whether an agent can read pointcloud geometry — and it couldn't: `str(PointCloud2)` hands the model `num_points` and nothing else. Baseline on the VQA suite: **0.136**, identical to blind guessing (0.138).

## Solution
**1. Pointcloud VQA suite** (`dimos/evals/suites/go2_pointcloud.py` + `go2_pointcloud_vqa.json`): 40 generated cases across 7 question families (extent, z-span, nearest-obstacle, direction, area-trend, area, shift) over the go2_short and go2_bigoffice datasets, regenerable from `dimos/evals/generate.py`.

**2. `PointCloud2.agent_encode()`**: per-x-interval geometry summary (y-extent per interval, exact point-to-box clearance on both axes) replacing the information-free str(). Suite score: **0.96**.

Squashed history, rebased on main; 18 tests passing, mypy strict clean.

### DIFF
--- a/dimos/msgs/sensor_msgs/PointCloud2.py
+++ b/dimos/msgs/sensor_msgs/PointCloud2.py
@@ -314,6 +314,99 @@ class PointCloud2(Timestamped):
     def __str__(self) -> str:
         return f"PointCloud2(frame_id='{self.frame_id}', num_points={len(self)})"
 
+    def agent_encode(self) -> dict[str, object]:
+        """Compact, spatially structured encoding for LLM consumption.
+
+        World-frame meters throughout. Carries the cloud's horizontal centroid
+        (so consecutive frames reveal motion of the mapped region) and exact
+        world-meter bounding boxes of body-height obstacle clusters (so
+        clearance around a given world position is closed-form point-to-box).
+        """
+        pts = self.points_f32()
+        n = int(pts.shape[0])
+        out: dict[str, object] = {"frame_id": self.frame_id, "num_points": n}
+        if n == 0:
+            return out
+        xy = pts[:, :2]
+        cx, cy = xy.mean(axis=0)
+        out["centroid_xy_m"] = [round(float(cx), 2), round(float(cy), 2)]
+        mins = pts.min(axis=0)
+        maxs = pts.max(axis=0)
+        # Distinct occupied 0.2 m x-y cells of THIS frame's cloud alone.
+        floor_cells = np.unique(np.floor(xy / 0.2).astype(np.int64), axis=0)
+        out["exact_stats"] = {
+            "note": "exact full-cloud values in meters; for numeric extent/span/area "
+            "answers use these, not the body-height interval map below",
+            "x_range": [round(float(mins[0]), 2), round(float(maxs[0]), 2)],
+            "y_range": [round(float(mins[1]), 2), round(float(maxs[1]), 2)],
+            "z_range": [round(float(mins[2]), 2), round(float(maxs[2]), 2)],
+            "horizontal_extent_m": round(float(max(maxs[0] - mins[0], maxs[1] - mins[1])), 2),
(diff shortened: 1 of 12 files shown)


spomichter:
## evo:ship mergeability report (advisory)

**Technique.** Adds one method pair to `PointCloud2` (`agent_encode()` + `_body_height_occupancy()`): exact full-cloud scalars (bbox ranges, extent, z-span, 0.2 m-cell floor footprint), per-frame world centroid, and the body-height slab (z 0.15–1.0 m) as per-row occupied x-intervals in exact world meters with per-interval y-extents — plus closed-form reasoning recipes (clearance `dx = max(0, min−qx, qx−max)` per axis; trig-free 8-way compass rule). ~2.7 kB per observation.

**Behavioural correctness.** Pointcloud VQA evals: 0.136 baseline → **0.962** shipped (blind control 0.15, gate ≤0.35). Distilled file is byte-identical to winner exp_0013's, so the measured score carries exactly.

**Regression safety.** `test_PointCloud2.py`: 3 passed, 1 deselected (pre-existing marker). No other files touched.

**Scope.** One file, +93 lines, purely additive (new methods; no existing code modified). A git gate enforced target-file-only diffs throughout the search.

**Test correctness.** No test modified, weakened, or removed at any point (gate-enforced). New behaviour is covered by the eval suite in this PR's `evals_bench/` rather than unit tests — worth a unit test for `agent_encode()` output shape before merge if reviewers want offline coverage.

**Mechanical cleanliness.** ruff check + format clean; full pre-commit suite passed.

**Codebase adherence.** Follows the existing `points_f32()`/`bounding_box_dimensions()` method style on the class; numpy-only, no new deps.

**Branches.** This PR carries the full 8-experiment research history. A distilled single-commit equivalent (identical final file, no `evals_bench/`) is at `feat/pointcloud-agent-encode-min` (`40f22ba7e`) if a squashed landing is preferred.

Caveat: exp_0013 (0.962) vs exp_0012 (0.957) is within measured sampling noise on the `nearest` family — treat as tied; the tip includes both refinements.

dimos/msgs/sensor_msgs/PointCloud2.py
```diff
@@ -314,6 +314,99 @@ def from_rgbd(
...
+            "z_range": [round(float(mins[2]), 2), round(float(maxs[2]), 2)],
+            "horizontal_extent_m": round(float(max(maxs[0] - mins[0], maxs[1] - mins[1])), 2),
+            "vertical_span_m": round(float(maxs[2] - mins[2]), 2),
+            "occupied_floor_footprint_m2": round(float(floor_cells.shape[0]) * 0.2 * 0.2, 1),
+            "footprint_note": "mapped floor area of this frame's cloud only (distinct "
+            "occupied 0.2 m x-y cells); use it, not bbox area, for floor coverage. For "
+            "area trends compare each frame's own footprint value across frames -- it "
+            "can decrease as well as increase; do not accumulate coverage over frames",
+        }
+        out["compass"] = (
```
TomCC7, line 351: seems weird to keep these instructions here as the agent encode function will be used everytime agent inspect a pointcloud object and this will just spam agent context multiple times

hvent90: I agree but it could be for another PR that addresses this holistically across all data types.

One way of going about it:
- `agent_encode`'s function description contains instructions, makes it programmatically available (probably through a decorator)
- dimos agent: system prompt contains deduped agent_encode instructions
- BYO agent:
  - new skill: `get_datatype_instructions(type)`
  - The description of a skill that returns an object that contains `agent_encode` has 1 line instruction to call `get_datatype_instructions(the type)`

PR #3420, opened 2026-08-10
refactor(manipulation): introduce static single-robot models

TomCC7:
## Summary

- accept canonical hierarchical joint/link names such as `left/joint1` without backend-specific encoding
- validate prepared one-model URDF/Xacro/SRDF configurations before Drake or RoboPlan world mutation
- remove the model-wide end-effector compatibility property and resolve tips through planning groups
- add a static dual-xArm6 model configuration and SRDF with `left_arm`, `right_arm`, and `both_arms` groups
- cover the common parser, Viser/yourdfpy, Pink/Pinocchio, Drake, and RoboPlan loading paths
- preserve xArm gripper/TCP geometry independently from optional coordinator gripper control
- keep Viser's compact status refresh independent of full manipulation and gripper telemetry

## Stack

This is PR1, **Static Single-Robot Model**, from `refactor-control-coordinator-connections`.

Depends on #3381, which has merged. The actual dual-arm blueprint cutover remains intentionally deferred to PR2 task 2.10.

## Runtime behavior

`dual-xarm6-planner-coordinator` keeps both modeled grippers, `link_tcp` planning tips, and gripper collision geometry without adding mock gripper joints or coordinator gripper tasks. Viser reads operation status through a compact accessor and no longer triggers full snapshot gripper polling.

## Validation

- `./bin/pytest-fast` — 4620 passed, 17 skipped
- Viser suite — 96 passed
- xArm blueprint suite — 12 passed
- focused manipulation/operator/model regressions — 21 passed
- Ruff check and format
- mypy on changed production modules
- runtime smoke: both robots and Viser initialized with no missing-gripper-task warnings; the run then stopped at the coordinator RPC loop because another DimOS instance already owned the shared LCM service

Optional Drake/ROS/ZED/Unitree tests remain skipped where their local dependencies are unavailable; CI will exercise its configured environments.


dimos/control/coordinator.py
```diff
@@ -291,9 +291,16 @@ def _create_whole_body_adapter(self, component: HardwareComponent) -> WholeBodyA
...
             address=component.address,
             domain_id=component.domain_id,
-            **component.adapter_kwargs,
+            **self._adapter_kwargs(component),
         )
 
+    @staticmethod
+    def _adapter_kwargs(component: HardwareComponent) -> dict[str, Any]:
+        kwargs = dict(component.adapter_kwargs)
+        if component.limits is not None:
```
TomCC7: why is this special handling needed looks ugly

TomCC7: Agreed. Removed this helper in 7a09944e39. `limits` remains typed hardware metadata so it survives config parsing, but it is now resolved by `ConnectedHardware.get_limits()` (configured metadata first, adapter-reported limits otherwise). The gripper task uses that runtime boundary directly, so the coordinator no longer repacks limits into generic adapter kwargs.

mustafab0 (dismissed):

mustafab0 (approved):

PR #3433, opened 2026-08-10
feat(spot): spot-replay blueprint - play a recording back into Rerun

jeff-hykin:
## Summary

Run it with no robot and no local dataset:

```
dimos run spot-replay --db-path=spot_small_loop.db
```

Screenshot 2026-08-10 at 4 37 46 PM


dimos/experimental/robot/bosdyn/spot/replay.py
```diff
@@ -0,0 +1,171 @@
...
+}
+
+
+class SpotReplayConfig(ModuleConfig):
+    """Where to read the recording from and how to play it back."""
+
+    # An on-disk path, or the name of a dataset to pull from LFS (e.g.
+    # ``spot_small_loop.db``). Empty -> newest ``*.db`` in ``dataset_dir``.
+    db_path: str = ""
+    dataset_dir: str = "~/datasets/spot"
```
paul-nechifor: Hm, not sure `~/datasets` is the best place. I don't like it when programs mess with my home dir without asking. Maybe this should use `STATE_DIR`?

paul-nechifor (dismissed):

dimos/experimental/robot/bosdyn/spot/replay.py
jeff-hykin: yeah definitely! I completely missed this comment. STATE_DIR is definitely

paul-nechifor (approved):

PR #3436, opened 2026-08-11
feat: g1 quest teleop on shared dual-arm IK

TomCC7:
## Stack

- Stacked on #3610.

## Summary

- Add `unitree-g1-teleop` on the shared Quest dual-arm module, UI, and Pink IK stack while retaining G1-specific model and task tuning.
- Compose GR00T locomotion, Quest arm control, stationary upper-body manipulation with remote Viser, RealSense video, and episode recording; omit navigation and mapping.
- Add guided hardware activation, ready-pose, status, and disable commands plus G1 setup and recovery documentation.
- Constrain non-selected joints inside Pink's QP so arm planning cannot borrow and then discard waist motion.
- Support both MuJoCo and real G1 hardware.

## Test

```bash
uv run dimos --simulation mujoco --scene-package office run unitree-g1-teleop
uv run pytest dimos/manipulation dimos/robot/unitree/g1/test_g1_teleop.py -q
```

Result: 516 passed, 24 deselected. Ruff and production mypy pass.

## Checklist


dimos/utils/data.py
TomCC7, line 1: this change seems very unrelated revert if not required

docs/platforms/humanoid/g1/index.md
```diff
@@ -76,7 +76,153 @@ Note: this button combination may vary based on the model of the G1
    - The robot will be super stiff now. Manually pick it up into a standing position and hold it there.
    - Press **R2 + A** (will attempt to self-balance)
 
-## 4. Run the Blueprint
+## 4. Start G1 teleoperation
```
TomCC7, line 79: the addition is tediusly long, try to keep concise you only need instruction can put error explanation to a different section

dimos/cli/hardware/g1.py
```diff
@@ -0,0 +1,263 @@
...
+    if not executed.succeeded:
+        _abort(f"ready-pose execution failed: {executed}")
+
+
+@app.command()
+def status() -> None:
+    """Show the G1 safety state, trajectory state, and planning groups."""
+    client = _connect()
+    try:
+        coordinator = cast("Any", client.get_module(_COORDINATOR))
```
TomCC7: no cast pls

dimos/core/coordination/blueprint_config/parser.py
```diff
@@ -463,7 +464,7 @@ def _validate_transports(
                     ) from error
             if not models:
                 continue
-            full = models[0].model_dump(mode="python")
+            full = validated_model_values(models[0])
```
TomCC7: are these required? otherwise remove

dimos/manipulation/planning/kinematics/pink_solver.py
```diff
@@ -222,6 +222,7 @@ def _build_robot_context(
         description = prepare_urdf_for_drake(config.model.load(), convert_meshes=False)
         model = pinocchio.buildModelFromXML(description.xml)
 
+        model = _reduce_to_controlled_joints(model, config, controlled_joints)
```
TomCC7: is this required? can't we just provide reduced model with only controllable joint?

dimos/utils/data.py
TomCC7: Confirmed this is required. Removing LfsPath equality/hash causes the self-hosted G1 Pink test to fail while hashing the RobotModel cache key (Path._hash is intercepted by LfsPath). I kept the minimal implementation and its no-download test; both the full G1 self-hosted suite and data tests pass.

dimos/manipulation/planning/kinematics/pink_solver.py
TomCC7: Done. The generic Pink reduction was removed. G1 now derives an arm-only teleop model by fixing the three waist joints, while the separate manipulation planning model keeps the waist live for collision checks.

docs/platforms/humanoid/g1/index.md
TomCC7: Shortened the main hardware procedure substantially. It now contains only safety, launch, activate, Quest/Viser access, and disable. The libgomp explanation and arm/enable/ready recovery commands are compact troubleshooting entries, and the incorrect Rerun section was removed.

dimos/cli/hardware/g1.py
TomCC7: Removed all casts from the G1 hardware CLI. Dynamic module handles are now narrowed with method-only Protocols and TypeGuard checks, with a clear error for an incompatible running stack.

dimos/core/coordination/blueprint_config/parser.py
TomCC7: Removed the generic blueprint parser change and its test. The trimmed G1 teleop blueprint parses successfully with the existing model_dump behavior.

dimos/robot/unitree/g1/blueprints/basic/unitree_g1_groot_wbc.py
```diff
@@ -385,6 +391,26 @@ def _g1_nav_path(path: NavPath) -> Any:
 _G1_ROOT = G1_RERUN_ROOT if global_config.simulation == "mujoco" else "world/odometry/g1"
 
 _G1_URDF_PATH = Path(__file__).resolve().parents[2] / "g1.urdf"
+G1_TELEOP_TASK_NAME = "teleop_g1"
```
ruthwikdasyam: why do we need this? cant we specify the task name directly in the blueprint? or does it ever change?

dimos/robot/unitree/g1/blueprints/basic/unitree_g1_groot_wbc.py
```diff
@@ -511,7 +537,27 @@ def _viewer() -> Any:
...
             },
         ),
-        *([_arm_holder] if _arm_holder is not None else []),
+        _arm_holder,
+        # Shared bimanual Quest task with G1-only model and objective tuning.
+        TaskConfig(
+            name=G1_TELEOP_TASK_NAME,
+            type="teleop_ik",
+            joint_names=g1_arms,
+            priority=20,
```
ruthwikdasyam, line 537: teleop_ik controls arms - 14 joints, and g1_groot_wbc controls 15 other waist and leg joints right. If there are no common joints, should we keep the priority same?

dimos/utils/data.py
TomCC7: well can you just remove the g1 test here... shouldn't pull the lfs data for testing and it's weird to add a new feature just to pass a test

docs/platforms/humanoid/g1/index.md
```diff
@@ -76,7 +76,66 @@ Note: this button combination may vary based on the model of the G1
...
+
+The robot must already be standing and balancing in sport mode. Use a clear,
+level work area, keep the Unitree remote and emergency stop reachable, and use
+a gantry or spotter for the first hardware run. Do not walk while executing
+planned arm motion because the upper-body planner excludes leg geometry.
+
+On the G1 computer:
+
+```bash
+cd ~/cc/dimos
```
TomCC7: this is too special... doc should be generic

docs/platforms/humanoid/g1/index.md
```diff
@@ -76,7 +76,66 @@ Note: this button combination may vary based on the model of the G1
...
+  uv run --no-sync dimos run unitree-g1-teleop --network-interface eth0
+```
+
+The teleop blueprint excludes navigation and mapping, so no module-disable
+arguments are needed. Wait for the Quest server to listen on port `8443`, then
+activate the robot from a second SSH session:
+
+```bash
+uv run dimos hardware g1 status
+uv run dimos hardware g1 activate --ready
```
TomCC7: separate activate with ready might be better

dimos/teleop/quest/web/static/teleop.js
```diff
@@ -474,6 +474,13 @@ function processTracking(frame) {
     }
     lastSendTime = now;
 
+    // Humanoid retargeting needs the headset pose in the same reference space
+    // as both hands. Other teleop modules ignore the "head" frame.
+    const viewerPose = frame.getViewerPose(xrRefSpace);
+    if (viewerPose) {
+        sendPose('head', viewerPose);
```
ruthwikdasyam: are we sending the head pose for future purposes? 
quest extension added in this PR uses base  `on_pose_bytes()` which only considers `right` and `left`

dimos/robot/unitree/g1/blueprints/basic/unitree_g1_groot_wbc.py
TomCC7: The claims are disjoint today, so Groot and teleop do not compete: arbitration is per joint. I kept teleop at 20 because it must preempt the arm trajectory task at 10. Groot remains at 50 as the safety ceiling if a future change accidentally introduces overlapping claims.

dimos/utils/data.py
TomCC7: Removed the direct G1 tests that loaded LFS assets, along with the test-only LfsPath equality/hash implementation. The runtime blueprint path serializes the model and worker-side validation reconstructs it, so deployment does not need that behavior. I also restored the serialized-dict round-trip assertion that ARM CI expects.

docs/platforms/humanoid/g1/index.md
TomCC7: Updated the normal path to the generic uv sync --extra all and uv run dimos run unitree-g1-teleop --network-interface eth0 commands. The ARM dual-libgomp preload is now only a conditional troubleshooting entry.

docs/platforms/humanoid/g1/index.md
TomCC7: Done. The documented operator flow now runs activate, checks status, and then runs ready as separate steps. The combined activate --ready command remains available as an optional shortcut.

dimos/teleop/quest/web/static/teleop.js
TomCC7: It was only carried for a future humanoid-retargeting path; this branch has no consumer for it. I removed the head-pose transmission and the backend special case that ignored non-controller poses, so the shared protocol remains left/right-only.

dimos/robot/unitree/g1/blueprints/basic/unitree_g1_groot_wbc.py
TomCC7: Agreed. The task name is fixed, so I removed the constant and inlined teleop_g1 in the blueprint. Tests now assert that literal directly.

dimos/robot/assets/model.py
```diff
@@ -134,12 +137,49 @@ def with_fixed_frame(
...
+            raise ValueError(f"Joint subtree already requested for removal: {duplicate}")
+        return replace(self, _removed_joint_subtrees=requested)
+
     def with_fixed_joints(self, *names: str) -> RobotModel:
         """Return a model with movable joints fixed at their URDF zero pose."""
         if not names:
             raise ValueError("At least one joint must be fixed")
         return replace(self, _fixed_joints=(*self._fixed_joints, *names))
 
+    def with_renamed_joints(self, names: Mapping[str, str]) -> RobotModel:
```
mustafab0, line 170: Since we are doing a lot of renaming and Mappping, better to add some TypeAlias just for easy readability. not a blocker.

dimos/cli/hardware/g1.py
mustafab0, line 1: Better to have a more generic cli here for all robots to arm and disarm.

But we can start with one robot at a time. 

But the State propogation from start() -> arm() etc should be standard for all robots. and better to not overpopulate CI with each robot case

mustafab0 (approved):

dimos/robot/assets/model.py
mustafab0: @TomCC7

dimos/cli/hardware/g1.py
mustafab0: @TomCC7

PR #3447, opened 2026-08-12
refactor: simplify manipulation RPC API

TomCC7:
## Problem
Manip API too messy making agent impossible to use properly.

## Philosophy

1. Group native. Public APIs use planning-group IDs, not robot names.
2. Agent REPL and RPC first. The API has a small typed surface, and Python can print its return values directly.
3. Skills are separate. The original `@skill` methods now live in `ManipulationSkills`; `ManipulationModule` contains RPCs only.

## Tests
Manually tested
1. original agent blueprint working
2. new rpc works well
```python
In [4]: manip = app.ManipulationModule

In [5]: groups = manip.list_planning_groups()

In [6]: group = next(g for g in groups if g.tip_frame is not None)

In [7]: manip.move_linear(
   ...:       dz=0.1,
   ...:       planning_group=group.id,
   ...:       blocking=True,
   ...: )
Out[7]: MoveResult(plan=PlanResult(SUCCEEDED, groups=('left_arm/manipulator',), waypoints=315, duration=3.14s, path_length=1.06, planning_time=0.0259s, iterations=0, message='RoboPlan Cartesian path found'), execution=ExecutionResult(COMPLETED, message='', coordinator_result=None, trajectory_status=TrajectoryStatus(state=COMPLETED, progress=1.0, time_elapsed=3.14, time_remaining=0.0, error='')), delta=(0.0, 0.0, 0.1), collision_check=False)

In [8]: manip.move_linear(
   ...:       dz=0.1,
   ...:       planning_group=group.id,
   ...:       blocking=True,
   ...: )
Out[8]: MoveResult(plan=PlanResult(SUCCEEDED, groups=('left_arm/manipulator',), waypoints=251, duration=2.5s, path_length=0.586, planning_time=0.00698s, iterations=0, message='RoboPlan Cartesian path found'), execution=ExecutionResult(COMPLETED, message='', coordinator_result=None, trajectory_status=TrajectoryStatus(state=COMPLETED, progress=1.0, time_elapsed=2.5, time_remaining=0.0, error='')), delta=(0.0, 0.0, 0.1), collision_check=False)
```

## `ManipulationSpec`

```python
class ManipulationSpec(Spec, Protocol):
    def list_planning_groups(self) -> tuple[PlanningGroupInfo, ...]: ...
    def get_state(self) -> ManipulationSnapshot: ...

    def plan_to_joints(
        self,
        targets: Mapping[PlanningGroupID, JointState],
        speed_scale: float | None = None,
    ) -> PlanResult: ...

    def plan_to_poses(
        self,
        targets: Mapping[PlanningGroupID, PoseStamped],
        speed_scale: float | None = None,
    ) -> PlanResult: ...

    def execute(
        self,
        blocking: bool = True,
        timeout: float | None = None,
    ) -> ExecutionResult: ...

    def wait_for_execution(
        self,
        timeout: float | None = None,
    ) -> ExecutionResult: ...

    def move_linear(
        self,
        dx: float = 0.0,
        dy: float = 0.0,
        dz: float = 0.0,
        planning_group: PlanningGroupID | None = None,
        check_collision: bool = False,
        speed_scale: float | None = None,
        blocking: bool = True,
        timeout: float | None = None,
    ) -> MoveResult: ...

    def set_gripper_position(
        self,
        position: float,
        planning_group: PlanningGroupID | None = None,
    ) -> CommandResult: ...

    def cancel(self) -> ExecutionResult: ...
```

Obstacle RPCs remain on `ManipulationModule`, outside this spec.

## Return types

```python
@dataclass(frozen=True, repr=False)
class ManipulationSnapshot:
    timestamp: float
    operation_status: OperationStatus
    error: str | None
    has_pending_plan: bool
    execution_status: ExecutionStatus
    groups: Mapping[PlanningGroupID, PlanningGroupState]

@dataclass(frozen=True, repr=False)
class PlanResult:
    status: PlanStatus
    message: str
    plan: GeneratedPlan | None

@dataclass(frozen=True, repr=False)
class ExecutionResult:
    status: ExecutionStatus
    message: str
    coordinator_result: TrajectoryExecutionResult | None
    trajectory_status: TrajectoryStatus | None

@dataclass(frozen=True, repr=False)
class MoveResult:
    plan: PlanResult
    execution: ExecutionResult | None
    delta: tuple[float, float, float]
    check_collision: bool
```

Public return types define bounded `repr()` output so they work directly in a Python REPL while keeping typed fields for code.

## AI assistance

Codex helped write and test this change.


dimos/control/coordinator.py
```diff
@@ -160,6 +161,7 @@ class ControlCoordinator(Module):
 
     # Output: Aggregated joint state for external consumers
     coordinator_joint_state: Out[JointState]
+    trajectory_status: Out[TrajectoryStatus]
```
TomCC7: wondering if this can be a jtt rpc rather than output stream

dimos/experimental/world_belief/xarm6_blueprint.py
```diff
@@ -129,6 +130,7 @@ def _rerun_blueprint() -> rrb.Blueprint:
         dino_model_name="facebook/dinov2-base",
         clip_model_name="openai/clip-vit-base-patch32",
     ),
+    ManipulationSkills.blueprint(),
```
TomCC7: should only add manip skill module very specifically - I think only agent blueprint actually use it

dimos/manipulation/planning/planners/roboplan_planner.py
```diff
@@ -217,7 +217,11 @@ def plan_cartesian_path(
                     group,
                     trajectory,
                 )
-                collision_free = not path or self._combined_path_collision_free(ctx, path)
+                collision_free = (
```
TomCC7: it's weird semantic to mark not checking collision as collision free

dimos/manipulation/planning/planners/roboplan_config.py
```diff
@@ -47,6 +47,7 @@ class RoboPlanCartesianPathConfig(BaseConfig):
     toppra_blend_deviation: float = 0.05
     position_limit_gain: float = Field(default=1.0, gt=0.0, le=1.0)
     max_attempts_per_step: int = Field(default=16, ge=1)
+    check_collision: bool = True
```
TomCC7: why should this be a backend config which is frozen during runtime? shouldn't it be a runtime config instead?

dimos/manipulation/visualization/test_operator.py
```diff
@@ -113,12 +115,11 @@ def __init__(self) -> None:
         self.execute_success = True
         self.cancel_success = True
         self.clear_success = True
-        self.reset_success = True
         self.topology_calls = 0
         self.telemetry_calls = 0
 
-    def get_state(self) -> str:
-        return self.state
+    def get_state(self) -> SimpleNamespace:
```
TomCC7: is this new type definition providing any value?

dimos/manipulation/manipulation_module.py
```diff
@@ -205,6 +213,11 @@ def __init__(self, **kwargs: Any) -> None:
 
         # Coordinator integration (initialized in start())
         self._execution_manager: PlanExecutionManager
+        self._execution_condition = threading.Condition()
```
TomCC7: all execution state tracking should be within execution manager

dimos/manipulation/manipulation_module.py
```diff
@@ -228,6 +241,9 @@ def start(self) -> None:
         if self.coordinator_joint_state is not None:
             self.coordinator_joint_state.subscribe(self._on_joint_state)
             logger.info("Subscribed to coordinator_joint_state port")
+        if getattr(self.trajectory_status, "_transport", None) is not None:
```
TomCC7: why getattr is needed here

dimos/manipulation/manipulation_spec.py
```diff
@@ -0,0 +1,297 @@
...
+from dimos.msgs.trajectory_msgs.TrajectoryStatus import TrajectoryStatus
+from dimos.spec.utils import Spec
+
+MAX_AGENT_REPR_LENGTH = 500
+
+
+def _bounded_repr(text: str) -> str:
+    if len(text) <= MAX_AGENT_REPR_LENGTH:
+        return text
+    return f"{text[: MAX_AGENT_REPR_LENGTH - 3]}..."
```
TomCC7: I prefer to give agent full context rather than the mechanical bound

dimos/manipulation/manipulation_spec.py
```diff
@@ -0,0 +1,297 @@
...
+    PlanningGroupState,
+    ManipulationSnapshot,
+    PlanResult,
+    ExecutionResult,
+    CommandResult,
+    MoveResult,
+)
+
+for _result_type in AGENT_READABLE_TYPES:
+    assert_agent_readable_result_type(_result_type)
```
TomCC7: should this be in test files? also can we retrieve all the spec api return type dynamically so we also support future added spec automatically

dimos/control/coordinator.py
TomCC7: Done in 3213e1342. Removed the status stream. JTT now exposes a non-destructive get_status task command, and PlanExecutionManager polls it through task_invoke.

dimos/experimental/world_belief/xarm6_blueprint.py
TomCC7: Done. Removed ManipulationSkills from this non-agent blueprint; skills remain explicitly composed by agentic blueprints.

dimos/manipulation/planning/planners/roboplan_planner.py
TomCC7: Done. The planner now only evaluates collision_validation_failed when collision checking is requested; skipping validation is no longer called collision-free.

dimos/manipulation/planning/planners/roboplan_config.py
TomCC7: Done. check_collision is now an explicit per-call planner argument and was removed from RoboPlanCartesianPathConfig.

dimos/manipulation/visualization/test_operator.py
TomCC7: Done. The fake now returns a real ManipulationSnapshot instead of a SimpleNamespace shape.

dimos/manipulation/manipulation_module.py
TomCC7: Done. PlanExecutionManager now owns acceptance, active state, polling, cached terminal results, timeouts, and cancellation. ManipulationModule only mirrors results into its broader state.

dimos/manipulation/manipulation_module.py
TomCC7: Removed along with the trajectory status stream and subscription path.

TomCC7: Removed along with the trajectory status stream and subscription path.

dimos/manipulation/manipulation_spec.py
TomCC7: Done. Removed the global character cap and string slicing. Result reprs expose all normal fields while plans summarize bulky paths and trajectories.

dimos/manipulation/manipulation_spec.py
TomCC7: Done. Removed the production guard and hardcoded type list. The test now discovers ManipulationSpec return annotations and recursively checks local result dataclasses for deliberate reprs.

dimos/control/blueprints/mobile.py
```diff
@@ -199,7 +200,7 @@ def _flowbase_twist_base(
     hardware=[_mock_arm_hw, _mock_twist_base()],
     tasks=[
         TaskConfig(
```
TomCC7: I think it now make sense to provide a jtt factory function for this as we don't really have a lot of thing to change here

dimos/control/test_control.py
```diff
@@ -91,7 +92,7 @@ def trajectory_task():
         joint_names=["arm/joint1", "arm/joint2", "arm/joint3"],
         priority=10,
     )
-    return JointTrajectoryTask(name="test_traj", config=config)
+    return JointTrajectoryTask(name=JOINT_TRAJECTORY_TASK_NAME, config=config)
```
TomCC7: can we just remove name from init

dimos/manipulation/manipulation_module.py
```diff
@@ -188,7 +192,7 @@ def __init__(self, **kwargs: Any) -> None:
         self._lock = threading.Lock()
         self._error_message = ""
         self._planning_epoch = 0
-        self._motion_speed_scale = 1.0
+        self._planning_speed_scale = self.config.default_speed_scale
```
TomCC7: why is this added but linear_speed_scale not

dimos/manipulation/manipulation_module.py
```diff
@@ -1543,53 +1643,64 @@ def get_gripper(self, robot_name: RobotName | None = None) -> float | None:
...
+            return group
+        config = self._group_robot_config(group)
+        assert config.gripper_hardware_id is not None
+        if self._control_coordinator.set_gripper_position(
+            config.gripper_hardware_id, float(position)
+        ):
+            return CommandResult(CommandStatus.SUCCEEDED, f"Gripper set to {position:.3f}m")
+        return CommandResult(CommandStatus.FAILED, "Failed to set gripper position")
+
+    def _resolve_capable_group(
```
TomCC7: need better semantics this is resolving capable group with gripper

dimos/control/test_control.py
TomCC7: Removed `name` from `JointTrajectoryTask.__init__`; the task now owns the fixed `joint_trajectory` identity. Registry creation rejects noncanonical config names.

dimos/control/blueprints/mobile.py
TomCC7: Added `joint_trajectory_task(...)` as the canonical factory. Direct blueprints now use it, and the hardware-based helper delegates to it.

dimos/control/blueprints/mobile.py
```diff
@@ -198,12 +198,7 @@ def _flowbase_twist_base(
 coordinator_mobile_manip_mock = ControlCoordinator.blueprint(
     hardware=[_mock_arm_hw, _mock_twist_base()],
     tasks=[
-        TaskConfig(
-            name="traj_arm",
-            type="trajectory",
-            joint_names=_mock_arm_hw.joints,
-            priority=10,
-        ),
+        joint_trajectory_task(_mock_arm_hw.joints),
```
TomCC7, line 202: replaced with a factory

dimos/manipulation/test_manipulation_spec.py
```diff
@@ -0,0 +1,180 @@
...
+    text = repr(result)
+
+    assert "waypoints=2" in text
+    assert "path_length=123" in text
+    assert "iterations=7" in text
+    assert "TrajectoryPoint" not in text
+    assert "positions" not in text
+
+
+def test_manipulation_module_exposes_only_primitive_and_obstacle_rpcs() -> None:
```
paul-nechifor: These sort of tests tend to be quite brittle. There's a test on modules which asserts that there are X number of RPC methods on modules. That test has to be constantly updated. This will probably be the same as `infrastructure` just contains the RPC methods on `Module` and presumably `ManipulationModule.rpcs` will change often too.

dimos/manipulation/test_manipulation_primitives.py
```diff
@@ -0,0 +1,218 @@
...
+    } == {
+        "close_gripper",
+        "get_robot_state",
+        "go_home",
+        "go_init",
+        "move_to_joints",
+        "move_to_pose",
+        "open_gripper",
+        "set_gripper",
+    }
```
paul-nechifor:
```suggestion
```

dimos/control/test_control.py
```diff
@@ -420,38 +422,39 @@ def test_start_stop_with_adapter_without_lifecycle_methods(self):
...
-        )
+        first = JointTrajectoryTask(JointTrajectoryTaskConfig(joint_names=["arm/joint1"]))
+        second = JointTrajectoryTask(JointTrajectoryTaskConfig(joint_names=["arm/joint2"]))
         coordinator.add_task(first, task_type="trajectory")
 
-        assert coordinator.remove_task("first")
+        assert coordinator.remove_task(JOINT_TRAJECTORY_TASK_NAME)
         assert coordinator.add_task(second, task_type="trajectory")
-        assert coordinator.list_tasks() == ["second"]
+        assert coordinator.list_tasks() == [JOINT_TRAJECTORY_TASK_NAME]
```
paul-nechifor: The original test added first, removed first, added second and asserted that second was present.

I'm not sure what this is testing now. If both first and second have the task name `JOINT_TRAJECTORY_TASK_NAME`, then it's not checking that first was removed and second was added, right?

dimos/manipulation/manipulation_module.py
```diff
@@ -1371,56 +1440,69 @@ def _initialize_execution(self) -> None:
...
+                self._state = ManipulationState.IDLE
+                self._error_message = result.message
+
+    def _execute_generated_plan(self, plan: GeneratedPlan) -> bool:
+        """Private in-process bridge for the visualization operator."""
+
+        with self._lock:
+            self._last_plan = plan
+            self._state = ManipulationState.COMPLETED
+        return self.execute(blocking=False).succeeded
```
paul-nechifor: If you're not blocking, how does `.succeeded` resolve? Don't you need to wait to know if it finished successfully nor not?

dimos/manipulation/test_manipulation_spec.py
TomCC7: Agreed. Removed the exact RPC inventory and kept only the stable boundary assertion that the base manipulation module exposes no skills; spec compliance remains covered separately.

dimos/manipulation/test_manipulation_primitives.py
TomCC7: Removed this exact legacy skill-name inventory. The adapter delegation behavior remains covered.

dimos/control/test_control.py
TomCC7: Correct. Since both instances use the canonical name, list_tasks could not distinguish them. The test now asserts get_task(JOINT_TRAJECTORY_TASK_NAME) is second.

dimos/manipulation/manipulation_module.py
TomCC7: Here success means the non-blocking dispatch was accepted by JTT, not that physical execution completed. Completion still requires blocking execute or wait_for_execution. I made this local contract explicit with status is ExecutionStatus.ACCEPTED and clarified ExecutionResult.succeeded.

paul-nechifor (dismissed):

paul-nechifor (approved):

PR #3459, opened 2026-08-13
chore: replace attrs usage with pydantic dataclasses

TomCC7:
## Contribution path

- Linked issue or discussion: DIM-XXX / #XXX / URL

## Problem

attr is a separate 'better dataclass' I introduced because I thought pydantic has performance issue with non-serializable objects. Turns out it don't.

## Solution
Replace attr usage with pydantic dataclass which is consistent to the rest of the system.

## Checklist


dimos/manipulation/execution_manager.py
```diff
@@ -72,54 +71,61 @@ def _to_immutable_joint_mapping(value: Mapping[str, str]) -> Mapping[str, str]:
...
+        resolved_names = list(self.model_to_coordinator.values())
         if any(not name for name in resolved_names) or len(set(resolved_names)) != len(
             resolved_names
         ):
             raise ValueError(
                 f"Execution target '{self.robot_name}' has ambiguous coordinator joints"
             )
+        return self
+
+    def __getstate__(self) -> dict[str, Any]:
```
TomCC7: these are ugly why here

dimos/manipulation/test_execution_manager.py
```diff
@@ -113,6 +114,20 @@ def test_execution_target_inverts_coordinator_mapping() -> None:
     }
 
 
+def test_execution_target_mapping_is_immutable_and_pickleable() -> None:
```
TomCC7: low value test remove

dimos/control/test_control.py
```diff
@@ -497,6 +499,20 @@ def test_config_requires_at_least_one_joint(self):
...
+
+        assert config.joint_names == ("arm/joint1",)
+        assert restored == config
+
+    def test_config_replace_revalidates_updates(self):
+        config = JointTrajectoryTaskConfig(joint_names=["arm/joint1"])
+
+        with pytest.raises(ValueError):
+            replace(config, start_position_tolerance=math.inf)
+
```
TomCC7: low value test remove

dimos/manipulation/execution_manager.py
TomCC7: Removed the custom pickle state. ExecutionTarget is constructed and retained locally by the manipulation module, so it does not need to be pickleable; keeping MappingProxyType preserves the pre-existing immutability.

dimos/manipulation/test_execution_manager.py
TomCC7: Removed this test together with the unused ExecutionTarget pickle behavior.

dimos/control/test_control.py
TomCC7: Removed both framework-level tests; the existing tests continue to cover the application validation invariants.

paul-nechifor (approved):

mustafab0 (approved):

PR #3463, opened 2026-08-14
feat(manipulation): add Dual OpenYAM teleoperation

TomCC7:
## Summary

- add a Dual OpenYAM entity with two verified OpenYAM arm models at the ABC Box 620 mm spacing
- add mock and dual-CAN hardware support, coupled planning groups, and Quest teleoperation
- align teleop velocity limits and Pink tuning with A1Z
- document fake-hardware and physical-hardware startup

## Testing

- 83 focused manipulation, Pink IK, G1 Quest, and Dual OpenYAM tests pass after rebasing
- Ruff passes
- mypy passes
- `uv run dimos run teleop-quest-dual-openyam` starts successfully with mock hardware

## Stack

- stacked above #3436 and transitively depends on #3392
- model archive is tracked in Git LFS and uploaded through `./bin/lfs_push`

PR #3465, opened 2026-08-14
feat(damiao): support native macOS CAN transport

TomCC7:
## Contribution path

- Originally stacked on #3463; that PR is now merged into `main`

## Problem

Damiao-based arms could only construct Linux SocketCAN transports. The macOS workaround lived above the shared hardware layer, so it did not apply to every Damiao topology. Users also had no supported way to discover the USB serials required for stable dual-arm assignment.

## Solution

- Use `can-motor-control` 0.0.8 on Linux and macOS.
- Select `SocketCanBus` on Linux and the package's native Rust `GsUsbBus` on macOS in the shared Damiao adapter.
- Treat runtime bus overrides as SocketCAN interface names on Linux and USB serial numbers on macOS.
- Route OpenYAM, Dual OpenYAM, and OpenArm through the shared bus factory.
- Add `dimos hardware can list`: it prints SocketCAN interfaces on Linux and native gs_usb serials on macOS.
- Require distinct left and right selectors for physical Dual OpenYAM.
- Document adapter wiring, termination, serial labeling, and cross-platform dual-arm commands.

The upstream discovery API shipped in [`can-motor-control` 0.0.8](https://pypi.org/project/can-motor-control/0.0.8/).

Co-authored with @aromeoes, whose original macOS hardware work established the transport requirements.

## How to Test

```bash
uv run pytest dimos/cli/test_can.py dimos/hardware/whole_body/damiao/test_adapter.py dimos/robot/manipulators/dual_openyam/test_config.py -q
uv run pytest dimos/robot/manipulators/dual_openyam/test_blueprints.py -m self_hosted -q
```

Results: 78 standard tests and 2 self-hosted blueprint tests passed. Ruff, pre-commit, and `uv lock --check` also passed. The native package passed its macOS arm64 Rust and Python CI before release.

## AI assistance

Codex inspected the shared transport path, implemented the Rust discovery API and DimOS CLI, resolved the stack rebase, and ran the tests. The resulting diff and test output were reviewed during the session.

## Checklist


dimos/cli/can.py
```diff
@@ -50,6 +57,41 @@ def _run_ip(*args: str, privileged: bool = False) -> subprocess.CompletedProcess
...
+        result = _run_ip("-brief", "link", "show", "type", "can")
+        output = result.stdout.rstrip()
+        typer.echo(output or "No SocketCAN interfaces found")
+        return
+    if sys.platform == "darwin":
+        try:
+            devices = can_motor_control.list_gs_usb_devices(
+                vendor_id=GS_USB_VENDOR_ID,
+                product_id=GS_USB_PRODUCT_ID,
+            )
```
mustafab0, line 71: Wonder if this check can be encoded in the can_motor_control library itself.

So we can just call `can_motor_control.list_devices()` and it automatically figures out linux or mac

mustafab0 (dismissed):

mustafab0 (dismissed):

dimos/cli/can.py
TomCC7: there's no need to do this on linux as linux can devices can be found through `ip a` as network device

dimos/cli/can.py
```diff
@@ -12,17 +12,24 @@
...
 
 import os
 import shlex
 import subprocess
+import sys
 
 import typer
 
-app = typer.Typer(help="Configure and inspect Linux CAN interfaces", no_args_is_help=True)
+if sys.platform in {"linux", "darwin"}:
```
paul-nechifor: Those are the only two platforms we support. :)

dimos/robot/manipulators/openyam/blueprints/teleop.py
```diff
@@ -83,6 +85,70 @@ def _gripper_task() -> TaskConfig:
...
+    ),
+    ManipulationModule.blueprint(
+        model=_openyam_quest_model,
+        kinematics=_openyam_quest_pink,
+        visualization={"backend": "viser"},
+    ),
+).remappings(
+    [
+        (ArmTeleopModule, "right_controller_output", "right_cartesian_command"),
+        (ArmTeleopModule, "right_gripper_command", "right_gripper_command"),
```
paul-nechifor: right_gripper_command == right_gripper_command

dimos/cli/can.py
TomCC7: Removed the redundant platform guard and now import can_motor_control unconditionally in b1047e2681. The platform-specific list behavior remains unchanged.

dimos/robot/manipulators/openyam/blueprints/teleop.py
TomCC7: Removed the identity remapping in 70746c18e1. The gripper stream already autoconnects by its unchanged name.

paul-nechifor (approved):

PR #3478, opened 2026-08-15
feat: add isolated Python module runtime

TomCC7:
## Contribution path

- Stacked on #3465.

## Problem

dimOS needs to run Python modules with dependency sets that conflict with the host environment while preserving normal typed streams, RPCs, skills, module references, lifecycle management, and restart behavior.

## Solution

- Add `IsolatedPythonModule`, a host-side contract backed by a concrete subclass in a sibling Python project.
- Keep RPC definitions on the contract and forward calls through a unique internal runtime endpoint.
- Forward streams and injected module references to the isolated runtime.
- Support locked uv projects and Pixi-provided uv without installing runtime dependencies into the host environment.
- Add a runnable external module example and lifecycle, validation, RPC, restart, and shutdown coverage.

## How to Test

```bash
uv run pytest dimos/core/test_isolated_python_module.py dimos/core/test_isolated_python_bootstrap.py dimos/core/test_isolated_python_end_to_end.py -q
uv run python examples/external_python_module/run.py
```

Validated on the #3465 base: 21 focused tests passed, targeted mypy passed, pre-commit passed, and the real subprocess example exited cleanly.

## Checklist


dimos/experimental/isolated_python/bootstrap.py
```diff
@@ -0,0 +1,121 @@
...
+import threading
+from typing import Any
+
+import typer
+
+from dimos.core.python_native_module import PythonNativeModule, contract_rpc_names
+from dimos.spec.utils import _signatures_compatible
+
+
+def load_class(reference: str) -> type[Any]:
```
TomCC7, line 34: check if there's lib function that does this, or something similar already exists in dimos

docs/usage/native_modules.md
```diff
@@ -12,6 +12,82 @@ Python module **never touches the pubsub data**. It just passes configuration an
 
 On how to speak LCM with the rest of dimos, you can read our [LCM intro](/docs/usage/lcm.md)
 
+## Isolated Python-native modules
+
+`PythonNativeModule` applies the same managed-process lifecycle to Python code that
```
paul-nechifor: Since PythonNativeModule is a subclass of NativeModule the name kinda make sense, but it's still improper. Native means "machine code", and Python is an interpreter.

Maybe rename it to IsolatedPythonModule?

docs/usage/native_modules.md
```diff
@@ -12,6 +12,82 @@ Python module **never touches the pubsub data**. It just passes configuration an
...
+    @rpc
+    def get_multiplier(self) -> int:
+        return self._multiplier
+```
+
+Every contract RPC and skill must be overridden with a compatible signature and
+the same `@rpc` or `@skill` classification. Startup fails before the blueprint
+runs when an implementation inherits a stub or changes the contract.
+
+During `build()`, DimOS runs `uv sync` in the sibling project. If `pixi.toml`
```
paul-nechifor: We've switched to `dimOS` in public facing docs.

dimos/core/python_native_module.py
```diff
@@ -0,0 +1,293 @@
...
+            transport = getattr(stream, "_transport", None)
+            if transport is not None:
+                self._runtime_client.set_transport(name, transport)
+        for name, module_ref in self._module_refs.items():
+            self._runtime_client.set_module_ref(name, module_ref)
+
+    @rpc
+    def build(self) -> None:
+        if self._python_native_runtime:
+            Module.build(self)
```
paul-nechifor: Why are you calling `Module.build` instead of `super().build`?

docs/usage/native_modules.md
TomCC7: Updated public-facing prose in this guide to use dimOS; package and import identifiers remain dimos.

docs/usage/native_modules.md
TomCC7: Renamed the API to IsolatedPythonModule / IsolatedPythonModuleConfig, including module paths, bootstrap, internals, tests, docs, and example metadata. No compatibility aliases remain.

dimos/core/python_native_module.py
TomCC7: Refactored build() to call super().build(). IsolatedPythonModule._maybe_build() now runs uv sync on the host and is a no-op in the child runtime, so the NativeModule generic executable check is not used.

dimos/core/isolated_python_module.py
```diff
@@ -0,0 +1,297 @@
...
+                command=" ".join(command),
+                cwd=str(self.runtime_project),
+            )
+            self._process = subprocess.Popen(
+                command,
+                cwd=self.runtime_project,
+                env=self._runtime_env(),
+                stdin=subprocess.PIPE,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
```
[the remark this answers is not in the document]
TomCC7: Fixed in ab3408d849. The runtime no longer creates stdout/stderr pipes; both streams inherit the worker's configured output destination, so logging cannot fill an unconsumed pipe. The spawn regression asserts stdout/stderr are absent from Popen arguments while the lifecycle FD and native parent-death hook are present. Validation: 47 isolated/native tests and 2 real external end-to-end tests passed.

dimos/core/isolated_python_module.py
```diff
@@ -0,0 +1,297 @@
...
+                pass_fds=(child_write,),
+                start_new_session=True,
```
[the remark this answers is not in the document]
TomCC7: Fixed in ab3408d849. The spawn path now reuses NativeModule's parent-death preexec hook and also passes a dedicated owner-lifetime pipe to the bootstrap. EOF makes the bootstrap SIGKILL its full process group, covering uv and descendants after abrupt worker death; normal stop still uses the existing graceful path first. A subprocess regression verifies the group exits by SIGKILL, and the full isolated/native suite passes 47 tests.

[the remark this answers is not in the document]
TomCC7: Fixed in ab3408d849. The spawn path now reuses NativeModule's parent-death preexec hook and also passes a dedicated owner-lifetime pipe to the bootstrap. EOF makes the bootstrap SIGKILL its full process group, covering uv and descendants after abrupt worker death; normal stop still uses the existing graceful path first. A subprocess regression verifies the group exits by SIGKILL, and the full isolated/native suite passes 47 tests.

leshy (approved):

leshy (approved):
minimal intervention into core, I'm good, we fold this in soon

PR #3485, opened 2026-08-16
cuvslam: multicam mode — Spot's five-camera rig

jeff-hykin:
Multi-camera cuVSLAM, tested on the Spot data

Screenshot_20260816-182811


dimos/mapping/cuvslam/test_cuvslam_spot.py
```diff
@@ -0,0 +1,226 @@
+# Copyright 2026 Dimensional Inc.
```
jeff-hykin, line 1: probably will cut a lot out of this

PR #3488, opened 2026-08-17
feat(evals): add deterministic VQA dataset generation

ruthwikdasyam:
## Contribution path

Closes DIM-1418

## Problem

DimOS needs reproducible visual-question datasets generated from recorded camera frames.

## Solution

Add `dimos evals vqa generate` and `dimos evals vqa run` workflows for deterministic multiple-choice VQA datasets. Questions are constrained by family, 
answers come from private Moondream evidence, and 
outputs include lossless PNG assets plus audit metadata.

Added 3 deterministic families - presence, horizontal_detection, object_count -> which are pre-built using primitive methods (moondream here) - so can just call the method to get the solution - making it deterministic.

## How to Test

`dimos evals vqa generate go2_short.db --image-index 100`

`dimos evals vqa run ~/.local/state/dimos/datasets/vqa/go2_short-frames --model gpt-4o-mini`

`uv run --no-sync pytest dimos/evals/vqa`

## AI assistance

OpenCode with GPT-5.6 Sol assisted with implementation, tests, documentation, and review.

## Checklist


dimos/models/vl/openai.py
```diff
@@ -75,7 +75,7 @@ def query(
                     "content": [
                         {
                             "type": "image_url",
-                            "image_url": {"url": f"data:image/png;base64,{img_base64}"},
+                            "image_url": {"url": f"data:image/jpeg;base64,{img_base64}"},
```
TomCC7, line 78: why jpeg?

dimos/evals/vqa/author.py
```diff
@@ -0,0 +1,72 @@
...
+        ]
+        prompt = (
+            "Generate useful questions about objects clearly visible in this image. "
+            "Use any applicable families and object names, preferring specific families when their "
+            "requirements are clearly satisfied. "
+            "Return only a JSON array of objects matching the available deterministic families. "
+            "Do not duplicate a family/object pair. Do not answer questions or add fields. "
+            f"Available families: {json.dumps(family_shapes)}"
+        )
+        payload: object = self._model.query_json(image, prompt)
```
TomCC7:
```suggestion
        payload = self._model.query_json(image, prompt)
```
since you enforce list type just below

dimos/evals/vqa/families.py
```diff
@@ -0,0 +1,190 @@
...
+
+NonEmptyString = Annotated[str, StringConstraints(strip_whitespace=True, min_length=1)]
+
+
+class QuestionProposal(BaseModel):
+    """A model-authored request constrained to a deterministic family."""
+
+    model_config = ConfigDict(extra="forbid", frozen=True)
+
+    family: Literal["presence", "horizontal_direction", "object_count"]
```
TomCC7, line 36: you might want to check this: https://pydantic.dev/docs/validation/latest/concepts/unions/#discriminated-unions

dimos/evals/vqa/suite.py
```diff
@@ -0,0 +1,128 @@
...
+    expected: str
+
+    def evaluate(self, rig: EvalRig) -> EvalResult:
+        image = Image.from_file(self.image_path)
+        context = [] if rig.blind else cast("list[dict[str, Any]]", image.agent_encode())
+        prompt = (
+            f"{self.inputs}\nChoices: {json.dumps(self.choices)}\nAnswer with exactly one choice."
+        )
+        outputs = rig.ask(context, prompt)
+        answer = _parse_choice(outputs, self.choices)
```
TomCC7: maybe we should always ask llm to provide structured output (like a json) and validate against them uniformly to avoid these custom cleanup/parsing logic

dimos/evals/vqa/suite.py
```diff
@@ -0,0 +1,128 @@
...
+class VqaEvalCase(EvalCase):
+    """One standalone image question evaluated by the shared runner."""
+
+    image_path: Path
+    choices: tuple[str, ...]
+    expected: str
+
+    def evaluate(self, rig: EvalRig) -> EvalResult:
+        image = Image.from_file(self.image_path)
+        context = [] if rig.blind else cast("list[dict[str, Any]]", image.agent_encode())
```
TomCC7, line 52: what's the point of doing a vqa with the blind option? basically no information provided?

dimos/evals/vqa/suite.py
```diff
@@ -0,0 +1,128 @@
...
+                image_path=_resolve_image(root, case.image),
+                choices=case.choices,
+                expected=label.answer,
+                tags=frozenset({"vqa"}),
+            )
+        )
+    return suite
+
+
+def _read_jsonl(path: Path) -> list[Any]:
```
TomCC7: let's use this which provides lazy jsonl loading. jsonl can be very huge and drain too much memory if you load directly. a lazy iterator would be much safer: https://jsonlines.readthedocs.io/en/latest/

dimos/evals/vqa/suite.py
```diff
@@ -0,0 +1,128 @@
...
+        raise ValueError(f"VQA dataset contains no cases: {root}")
+    case_by_id = _unique_by_id(cases, "case")
+    label_by_id = _unique_by_id(labels, "label")
+    if case_by_id.keys() != label_by_id.keys():
+        missing_labels = sorted(case_by_id.keys() - label_by_id.keys())
+        missing_cases = sorted(label_by_id.keys() - case_by_id.keys())
+        raise ValueError(
+            f"VQA case/label IDs do not match: missing_labels={missing_labels}, "
+            f"missing_cases={missing_cases}"
+        )
```
TomCC7: this whole checking seems to imply that there's a 1-1 correspondance between case and label. in that case why not just use 1 jsonl instead of 2?

dimos/evals/vqa/primitives/moondream.py
```diff
@@ -0,0 +1,34 @@
...
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from dimos.models.vl.moondream import MoondreamVlModel
+    from dimos.msgs.sensor_msgs.Image import Image
+    from dimos.perception.detection.type.detection2d.imageDetections2D import ImageDetections2D
+
+
+class MoondreamObjectDetector:
```
spomichter: can we use existing mooondrealVLModel any need to wrap it?

dimos/cli/vqa.py
```diff
@@ -0,0 +1,74 @@
...
+
+"""Commands for generating and evaluating standalone VQA datasets."""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+import typer
+
+app = typer.Typer(help="Generate and evaluate standalone visual question-answering datasets.")
```
spomichter, line 23: all cli change should be in dimos/cli so we can track everything

dimos/evals/vqa/generate.py
```diff
@@ -0,0 +1,332 @@
...
+def _prepare_output(output: Path) -> None:
+    if output.exists() and any(output.iterdir()):
+        raise FileExistsError(f"VQA output directory is not empty: {output}")
+
+
+def _write_frame(output: Path, frame: _GeneratedFrame) -> None:
+    (output / "assets").mkdir(parents=True, exist_ok=True)
+    frame_audit = output / "audit" / f"frame-{frame.index:06d}"
+    frame_audit.mkdir(parents=True, exist_ok=True)
+
```
spomichter, line 311: No hard coded path generation / making directories. this will work for a git cloned dimos but not for a library installed dimos. Needs to be done properly and save VQA sets to .local/ the same place mem2 does.

dimos/models/vl/openai.py
```diff
@@ -105,7 +105,7 @@ def query_batch(
             {
                 "type": "image_url",
                 "image_url": {
-                    "url": f"data:image/png;base64,{self._prepare_image(img)[0].to_base64()}"
+                    "url": f"data:image/jpeg;base64,{self._prepare_image(img)[0].to_base64()}"
```
spomichter, line 108: whats this for

spomichter (dismissed):

docs/usage/vqa.md
```diff
@@ -0,0 +1,85 @@
+---
+title: "Visual Question Answering"
+---
+
+# Visual Question Answering
```
paul-nechifor, line 1: Doesn't this add the title twice?

docs/usage/vqa.md
```diff
@@ -0,0 +1,85 @@
...
+
+The VQA tools generate deterministic questions from recorded images and evaluate them through the
+shared DimOS evaluation runner.
```
paul-nechifor: All docs use `dimOS` now.

TomCC7: maybe we can add a CI check for this lol

dimos/models/vl/openai.py
paul-nechifor: `to_base64` actually encodes as JPEG so it looks like using `image/png` was a previous bug.

docs/usage/vqa.md
```diff
@@ -0,0 +1,85 @@
...
+## Architecture
+
+- `author.py` proposes constrained family inputs from an image.
+- `families.py` owns question text, choices, and deterministic answer rules.
+- `primitives/moondream.py` supplies private object detections.
+- `generate.py` loads frames, reuses models, and writes datasets atomically.
+- `suite.py` validates generated artifacts and creates shared evaluation cases.
+
+The initial implementation intentionally defers negative-presence policy, retries, and resume
+behavior.
```
paul-nechifor:
```suggestion
```

docs/usage/vqa.md
paul-nechifor: Good point: https://github.com/dimensionalOS/dimos/pull/3561

dimos/evals/vqa/primitives/moondream.py
ruthwikdasyam: I added extra functionality before, so wrapped it. Now its stale, thanks for pointing, fixed

dimos/models/vl/openai.py
ruthwikdasyam: this is a bug, to_base64 encodes as jpeg. fixed it here coz am utilizing `OpenAIVlModel`

dimos/evals/vqa/generate.py
ruthwikdasyam: dataset root already uses `STATE_DIR`. the asset and audit directories here are relative parts of dataset schema.

dimos/cli/vqa.py
ruthwikdasyam: Got it, moved to dimos/cli/vqa.py

dimos/evals/vqa/suite.py
ruthwikdasyam: `cases.jsonl` is public - image path, question and choices, `labels.jsonl` is private with case id ad expected answer. so theyhave different visbility, but one-to-one correspondence is intentional

dimos/evals/vqa/suite.py
ruthwikdasyam: good question. I added this like a benchmark sanity check. you know how dataset have questions like `is object x visible?` and answer is always yes, coz we cannot auto-generate true negative questions.
It should score around 30% or less in blind mode, to make sure dataset is good. If more, then its bad. a good sanity check

dimos/evals/vqa/suite.py
ruthwikdasyam: yeah, thats a good use of structured output. wil change

dimos/evals/vqa/families.py
ruthwikdasyam: currently every family has same payload. when distinct fields are introduced, ill change them to discriminated unions

docs/usage/vqa.md
```diff
@@ -0,0 +1,72 @@
+---
+title: "Visual Question Answering"
```
paul-nechifor: Ivan's mkdocs pr has been merged: https://github.com/dimensionalOS/dimos/pull/3645

You have to switch back to old titles.

dimos/evals/vqa/families.py
```diff
@@ -0,0 +1,190 @@
...
+    model_config = ConfigDict(extra="forbid", frozen=True)
+
+    question: NonEmptyString
+    choices: tuple[NonEmptyString, ...]
+    answer: NonEmptyString
+    evidence: dict[str, JsonValue] = Field(default_factory=dict)
+
+    @model_validator(mode="after")
+    def answer_is_a_choice(self) -> FamilyAnswer:
+        if len(self.choices) < 2:
```
hvent90, line 52: since we're testing for count and uniqueness, would recommend something like `unique_choices = {choice.casefold() for choice of in self.choices}`

dimos/evals/vqa/generate.py
```diff
@@ -0,0 +1,331 @@
...
+class PublicCase(BaseModel):
+    """One evaluator-facing multiple-choice VQA case."""
+
+    model_config = ConfigDict(extra="forbid", frozen=True)
+
+    id: NonEmptyString
+    image: NonEmptyString
+    question: NonEmptyString
+    choices: tuple[NonEmptyString, ...]
```
hvent90, line 90: `Families` does QA checking on PublicCase.cases (>1 non-duplicated case). IMO PublicCase should loudly fail on construction so downstream consumers don't need to manage this.

hvent90 (dismissed):

dimos/evals/vqa/families.py
```diff
@@ -0,0 +1,190 @@
...
+        return _answer_object_count(proposal, image, detector)
+    raise ValueError(f"unsupported VQA family: {proposal.family}")
+
+
+def _answer_presence(
+    proposal: QuestionProposal, image: Image, detector: ObjectDetector
+) -> FamilyAnswer:
+    detections = detector.query_detections(image, proposal.object_name)
+    if not detections.detections:
+        raise InsufficientEvidenceError(
```
paul-nechifor, line 107: Why raise an exception? Is "nothing detected" not a valid answer (although maybe incorrect sometimes)?

dimos/evals/vqa/families.py
```diff
@@ -0,0 +1,190 @@
...
+    description="Count a small number of clearly visible, distinct object instances.",
+)
+AVAILABLE_FAMILIES = (PRESENCE_FAMILY, HORIZONTAL_DIRECTION_FAMILY, OBJECT_COUNT_FAMILY)
+
+
+class InsufficientEvidenceError(ValueError):
+    """A family cannot derive an answer from the available primitive evidence."""
+
+
+class ObjectDetector(Protocol):
```
paul-nechifor: Why do you need this type? Shouldn't this be `VlModel`?

The return type is actually incomplete. It should be `ImageDetections2D[Detection2DBBox]`

docs/usage/vqa.md
TomCC7: fk then I also need to rework a lot of stuff lol

dimos/evals/vqa/families.py
paul-nechifor: If I'm not mistaken, with the way the questions are set up in `author.py` (`Generate useful questions about objects clearly visible in this image.`), then you're right 100% of the time if you always answer `yes` for every question, no?

That is, do you ever generate questions for which the answer is "no"?

docs/usage/vqa.md
```diff
@@ -0,0 +1,72 @@
...
+
+The VQA tools generate deterministic questions from recorded images and evaluate them through the
+shared dimOS evaluation runner.
+
+## Generate a Dataset
+
+Generate from one image:
+
+```bash skip
+dimos evals vqa generate go2_short.db --image-index 100
```
paul-nechifor, line 11: Do you have any saved? You have to manually verify them, right? It would be useful to see what the result is without using my own non-verified data.

dimos/evals/vqa/families.py
ruthwikdasyam: makes sense, adding the check

dimos/evals/vqa/families.py
ruthwikdasyam: `VLModel` also requires query, caption, lifecycle behaviors. added this type so we can introduce any lightweight detectors and use them here.
That said, we are passing only VLModel now - will change it for now

dimos/evals/vqa/families.py
ruthwikdasyam: yes, its always `yes` for all the questions. so, `nothing detected` is not a valid ans here.
As the VLM can generate false negatives sometimes, and i want the pipeline to be deterministic, am not adding any questions on true negatives.
will add them in another pass, where i can makesure there are less/no false negatives.

docs/usage/vqa.md
ruthwikdasyam: yes, adding methods to validate/inspect via another PR

paul-nechifor (approved):

PR #3497, opened 2026-08-17
fix(imitation): improve collection safety and episode handling

ruthwikdasyam:
## Contribution path

## Problem

Imitation collection performed unnecessary pose/TF lookup for semantically poseless streams, could stop its recorder before final producer events, and did not clearly report an episode left open at the end of a recording.

Quest disconnects could also leave stale controller engagement active, and invalid episode button mappings were silently ignored.

## Solution

- Keep collection streams explicitly poseless and disable TF recording
- Stop camera, teleop, coordinator, and episode producers before the collection recorder
- Persist an explicit discard when shutdown interrupts an active episode
- Clear Quest controller state on the final client disconnect, publish a zero-button safe command, require fresh reconnect state, and reject unknown controller identities
- Validate episode button mappings at startup
- Extend dataprep inspection to report incomplete source episodes separately from saved/discarded episodes.
- Add a real SQLite-to-HDF5/LeRobot integration test

## How to Test

Run the recording inspector against a collection database:

`dimos dataprep inspect /path/to/session.db`

Automated verification:

`uv run pytest dimos/memory dimos/imitation dimos/teleop/quest -q`

`uv run mypy dimos/`

`uv run pre-commit run --all-files`

Results: 541 tests passed, mypy passed across 999 source files, and all pre-commit hooks passed.

## AI assistance

OpenCode with GPT-5.6 Sol was used extensively for codebase analysis, implementation, merge-conflict resolution, tests, and verification. The changes and generated tests were reviewed through focused diffs, type checking, pre-commit, and the affected test suites.

## Checklist


dimos/imitation/collection/episode_monitor.py
```diff
@@ -62,14 +63,33 @@ class KeyPress(BaseModel):
...
+
+
 class EpisodeMonitorModuleConfig(ModuleConfig):
-    button_map: dict[EpisodeCommand, str] = {
-        "toggle": "B",
-        "discard": "Y",
-    }
-    keyboard_map: dict[EpisodeCommand, str] = {}
+    button_map: dict[EpisodeCommand, str] = Field(default_factory=_default_button_map)
+    keyboard_map: dict[EpisodeCommand, str] = Field(default_factory=dict)
```
mustafab0, line 72: Should also make a `_default_keyboard_button_map` too

mustafab0 (dismissed):

dimos/imitation/collection/episode_monitor.py
ruthwikdasyam: yes, once we later add keyboard buttons to start/stop, we can add that

mustafab0 (approved):

PR #3523, opened 2026-08-18
feat(cli): rich login/logout UX

spomichter:
Polishes the `dimos login` / `dimos logout` experience on top of #3522, using rich (already a typer dependency — no new deps):

- **login**: cyan panel with verification URL + device code chip, live spinner with mm:ss countdown while polling, gradient DIMOS wordmark + green confirmation on success
- **logout**: styled `⏻ Logged out.` line with dimmed revoke URL
- **whoami**: same text, bold email / dim scopes

rich degrades to plain text when stdout isn't a TTY, so robot and CI logs stay clean. Existing tests pass unchanged; mypy clean.

### DIFF
--- a/dimos/cli/cloud.py
+++ b/dimos/cli/cloud.py
@@ -32,6 +32,9 @@ import urllib.error
 import urllib.parse
 import urllib.request
 
+from rich.console import Console
+from rich.panel import Panel
+from rich.text import Text
 import typer
 
 from dimos.constants import CREDENTIALS_PATH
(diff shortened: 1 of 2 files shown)


dimos/cli/cloud.py
```diff
@@ -40,6 +43,30 @@
 _KEYRING_SERVICE = "dimos-cloud"
 _KEYRING_USER = "default"
 
+_console = Console(highlight=False)
+
+_LOGO = """\
```
paul-nechifor, line 48: I think we should have a unified logo/branding. If you want to change DIMENSIONAL to DIMOS, we should do it everywhere, right?

The current logo is in `ascii_logo` in dimos/cli/theme.py

spomichter: agree

PR #3550, opened 2026-08-19
feat(cli): ~/.config/dimos becomes a real config directory

spomichter:
Final layout, per stash — the aws model:

```
~/.config/dimos/config        # human-edited JSON, extensionless
~/.config/dimos/credentials   # machine-written by dimos login, plain-text key, 0600
```

- Default `--config` is `~/.config/dimos/config`; `read_config_file` treats a directory as absent.
- **Legacy flat file at `~/.config/dimos`: `dimos run` refuses loudly** with the exact one-line `mv` recipe and exits 2 — zero filesystem mutation. The check lives only in the run command: a machine with the flat file predates `dimos login` existing, so that's the command it will hit. No migration state machine → all three Greptile P1s structurally gone.
- Verified: targeted tests + 260 across cli/blueprint_config, mypy strict clean, E2E through the installed console script (refusal prints the recipe; following it restores normal operation; credentials land 0600 inside the dir).

### DIFF
--- a/dimos/constants.py
+++ b/dimos/constants.py
@@ -41,16 +41,6 @@ else:
 CREDENTIALS_PATH = CONFIG_DIR / "dimos" / "credentials"
 
 
-def reject_legacy_config() -> None:
-    """~/.config/dimos used to BE the config file; it is now a directory."""
-    legacy = CONFIG_DIR / "dimos"
-    if legacy.is_file():
-        raise RuntimeError(
-            f"config found at old path {legacy}, which is now a directory; move it:\n"
-            f"  mv {legacy} {legacy}.tmp && mkdir {legacy} && mv {legacy}.tmp {legacy}/config"
-        )
-
-
 """
 Constants for shared memory
 Usually, auto-detection for size would be preferred. Sadly, though, channels are made
--- a/dimos/constants.py
+++ b/dimos/constants.py
@@ -41,20 +41,14 @@ else:
 CREDENTIALS_PATH = CONFIG_DIR / "dimos" / "credentials"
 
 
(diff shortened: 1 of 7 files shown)


dimos/constants.py
```diff
@@ -38,7 +38,18 @@
     LOG_DIR = STATE_DIR / "logs"
     RECORDINGS_DIR = STATE_DIR / "recordings"
 
-CREDENTIALS_PATH = CONFIG_DIR / "dimos-credentials"
+CREDENTIALS_PATH = CONFIG_DIR / "dimos" / "credentials"
+
+
+def reject_legacy_config() -> None:
```
paul-nechifor: This file is for constants, not functions.

dimos/core/coordination/blueprint_config/test_sources.py
```diff
@@ -80,3 +80,10 @@ def test_config_file_errors_are_clear_but_missing_file_is_optional(tmp_path: Pat
     not_an_object.write_text("[]")
     with pytest.raises(BlueprintConfigError, match="must contain a JSON object"):
         parser.parse(config_path=not_an_object, environ={})
+
+
+def test_config_path_that_is_a_directory_reads_as_absent(tmp_path: Path) -> None:
+    # ~/.config/dimos may exist as a directory (it holds the config file since the
+    # flat-file migration); pointing at a directory must mean "no config".
+    parser = BlueprintConfigParser(PrimaryModule.blueprint())
+    parser.parse(config_path=tmp_path, environ={})
```
paul-nechifor: It's not clear what this test is for as there are no assertions.

PR #3556, opened 2026-08-19
feat(cli): terminal art for login/logout/whoami

aromeoes:
## Contribution path

## Problem

Terminal UI for `dimos login`, `dimos logout` and `dimos whoami` is plain `typer.echo` output. It works, but the UX could be better.

## Demo

`dimos login` → `whoami` → `logout` → a refused code.

dimos login

## Solution

New `dimos/cli/art.py`: pure rendering, every function returns `list[str]`. `cloud.py` swaps seven prints for seven calls to it.

**No behaviour change** — `_post`, `_store`, `_forget`, `api_key`, the polling loop and every exit code are byte-identical, and the seven existing tests pass unmodified.

**No new libraries** — `terminaltexteffects` was already one; this drops the `rich` imports from `cloud.py`.

**Four things to understand:**

- **`Live`** — the login card is redrawn in place each poll. `pause()` sleeps the server's `interval` but wakes every 100ms to animate, so the spinner turns without changing the poll cadence.
- **`card()`** — the width ladder. Drops the art, then the border, then the two-column labels until it fits. Thresholds are derived from the copy, so the device code and URL survive to 30 columns.
- **`enabled()`** — `isatty() and not NO_COLOR`. When false there is no art, just the words, so nothing decorative reaches CI logs.
- **Colour is state** — grey while waiting, red when refused, brand gradient once you are through.

Signing in lands on a single card: account and key at the top, then a short About DimOS with the community link and capability areas.

## How to Test

```
dimos login          # then whoami, logout. Resize and re-run to see it degrade.
uv run pytest dimos/cli/test_art.py dimos/cli/test_cloud.py
uv run mypy dimos/cli/art.py
```

40 tests pass (33 new, 7 existing unmodified). The width sweep covers 30–120 columns and asserts the device code and URL are never truncated.

## AI assistance

Claude Code with Opus 5, heavily involved — design exploration and implementation.

## Checklist

### DIFF
--- a/dimos/cli/cloud.py
+++ b/dimos/cli/cloud.py
@@ -22,9 +22,12 @@ stored in the system keyring, falling back to a plain-text 0600 file
 backend. `DIMOS_API_KEY` (via GlobalConfig) overrides any stored login.
 """
 
+import importlib.metadata
 import json
 import os
 import socket
+import sys
+import textwrap
 import time
 from types import ModuleType
 from typing import Any, cast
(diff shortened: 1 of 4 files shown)


dimos/cli/art.py
```diff
@@ -0,0 +1,455 @@
...
+
+from terminaltexteffects.effects.effect_beams import Beams
+
+from dimos.cli import theme
+
+RGB = tuple[int, int, int]
+
+# Sampled down the centre of the portal in the brand asset: pale yellow at the
+# lintel, through cream and mint, to salmon at the threshold.
+PORTAL: list[RGB] = [
```
leshy: we have standardized theme in `cli/theme` and `dimos/dimos/cli/dimos.tcss` would be nice to unify more instead introduce a third place with colors

aromeoes: agree. Will look into this

paul-nechifor:
Personally, I dislike animations. On `humancli`, the animation hides the fact that setting up all the imports takes time. But here, what's the point of the animation?

This may be a bit controversial, but I think animations in TUIs are perceived as a software project not having its priorities right (like the infamous Ryan Dahl rant about unicode characters in terminals and the rest: https://tinyclouds.org/rant/ )

aromeoes:
Closing while I iterate locally

aromeoes:
> Personally, I dislike animations. On `humancli`, the animation hides the fact that setting up all the imports takes time. But here, what's the point of the animation?
> 
> This may be a bit controversial, but I think animations in TUIs are perceived as a software project not having its priorities right (like the infamous Ryan Dahl rant about unicode characters in terminals and the rest: https://tinyclouds.org/rant/ )

We are trying to make dimos friendly to developers that come from broad software backgrounds where terminal interaction can be minimal. I personally hated interacting with the terminal until claude code. It added only the right amount of UI for it to be easy to use. The purpose the added animation serves is familiarity with web UIs the user is already familiar with.

PR #3557, opened 2026-08-19
SONIC whole-body controller for the G1

Nabla7:
Adds NVIDIA GEAR-SONIC v1.1 as a unified 29-joint G1 controller, with a standalone Zenoh blueprint for MuJoCo and hardware. `ControlCoordinator` accepts velocity commands, selectable locomotion modes, motion clips and optional numeric reference targets. The controller runs independently of headset integrations.

Tasks load and prepare their initial hold before adapters activate. Hardware starts unarmed in dry-run, ramps from the measured pose when armed, and continuously refreshes hold commands. SONIC faults latch damping; the independent DDS publisher also checks command and feedback freshness. SONIC opts into bounded Zenoh delivery so busy subscribers discard old targets.

`dimos-sonic-models` explicitly installs the pinned v1.1 policy, planner and observation configuration from Hugging Face, plus 13 clips from pinned NVIDIA sources, and verifies SHA-256 hashes in the shared DimOS cache. `bin/hardware/g1/setup-sonic` detects JetPack 5/6, checks CUDA/cuDNN prerequisites, and installs the matching ONNX Runtime and model assets. [Setup and RPC examples](https://github.com/dimensionalOS/dimos/blob/pim/feat/g1-sonic-wbc/dimos/control/tasks/g1_sonic_wbc_task/README.md).

Floor commands keep squat/kneeling stationary and stop crawl translation at neutral input. Face-down mode 7 is excluded from the selectable menu, matching NVIDIA's deployment menu.

Validation:

- 66 focused pipeline, task and G1 hardware-connection tests passed. All changed-file pre-commit hooks passed. Four generated-registry checks and mypy on the five edited production Python files passed on September 16 against the same production code.
- The actual standalone blueprint ran the v1.1 models with CUDA and MuJoCo on a private Zenoh bus: standing, walking, turning, stopping, an eight-second Macarena sample, return to planner, and a latched shell E-stop with reset refused. The robot stayed upright throughout the 25-second sequence; minimum sampled root height was 0.745 m.
- All 43 model/config/clip assets verified offline. The combined installer passed Bash syntax checks and isolated prerequisite checks for both JetPack releases, including missing-library rejection.

Limits: the combined installer has not been executed on a Jetson, and this update has no new physical-robot activation. Crawl stability, timed floor transitions and hardware watchdog timing still need validation. Interrupted simulation startup can hit the existing stale-SHM attachment race; the first validation attempt hit it, and a clean restart passed. Full-repository CI remains separate from these local checks.


data/.lfs/sonic.tar.gz
TomCC7, line 1: I have no idea why we keep adding model (which was on huggingface) to lfs. should just pull from huggingface

dimos/control/tasks/g1_sonic_wbc_task/zmq_wire.py
```diff
@@ -0,0 +1,174 @@
...
+    raise ValueError(f"unknown topic prefix: {raw[:16]!r}")
+
+
+def decode(raw: bytes) -> DecodedMessage:
+    """Decode one packed message (topic prefix included)."""
+    topic, body = split_topic(raw)
+    if len(body) < HEADER_SIZE:
+        raise ValueError(f"{topic}: body shorter than header ({len(body)} < {HEADER_SIZE})")
+    header_json = body[:HEADER_SIZE].rstrip(b"\x00")
+    header: dict[str, Any] = json.loads(header_json)
```
TomCC7: can use pydantic model directly instead of manual decode and field validation

dimos/control/tasks/g1_sonic_wbc_task/streamed_motion.py
```diff
@@ -0,0 +1,260 @@
...
+    motion: StreamedMotion | None = None
+    window_start: int = 0
+    frame_offset_adjustment: int = 0
+    did_catchup_reset: bool = False
+    frame_step: int = 1
+    protocol_version: int = 0
+    error: str | None = None
+
+
+def infer_protocol_version(fields: dict[str, NDArray]) -> int:
```
TomCC7, line 78: this should better be determined once and cached instead of inferring every frame?

dimos/control/tasks/g1_sonic_wbc_task/streamed_motion.py
```diff
@@ -0,0 +1,260 @@
...
+        result = MergeResult()
+
+        protocol = infer_protocol_version(fields)
+        if protocol == 0:
+            result.error = "pose message has neither joint nor SMPL data"
+            return result
+        if self._active_protocol is None:
+            self._active_protocol = protocol
+        elif self._active_protocol != protocol:
+            result.error = f"protocol version changed {self._active_protocol} -> {protocol}"
```
TomCC7: I wonder in what case would version change

TomCC7 (changes requested):
I don't see why we need zmq protocol involved here. @Nabla7 wondering why you choosed this from the first place. do you intend to let our integration work with some external nvidia tool? if not I think only porting the model will be the better option.

PR #3558, opened 2026-08-19
fix(memory): make recorder lossless under load

TomCC7:
## Summary

Fix Recorder data loss under sustained sensor load and temporary encoder or SQLite stalls. Every message accepted by Recorder is persisted in callback-arrival order, or recording fails visibly.

## Approach

- Add an opt-in bounded queue to `pSHMTransport`; its default remains the existing latest-value transport.
- Bypass Recorder's latest-value dispatcher and feed one bounded FIFO worker, which serially calls the existing `Stream.append()` path.
- Stop subscriptions, wait for in-flight callbacks to enqueue, then drain accepted work; surface queue overflow, processing/write failure, or drain timeout.

## Verification

The standalone [recorder benchmark](https://github.com/TomCC7/dim-recorder-perf) compares the generated source manifest with the completed memory2 database. Its 30-second production profile contains 22,500 observations: RGB 30 Hz, two grayscale streams at 30 Hz, depth 30 Hz, IMU 400 Hz, PointLIO 10 Hz, odometry 200 Hz, and TF 20 Hz.

All three experiments stored 22,500/22,500 observations with no missing, duplicate, corrupt, or reordered payloads:

- baseline
- one-second depth encoder stall
- SQLite writer lock

Each run sustained 64.85 MiB/s offered raw data, 38.88 MiB/s encoded payload, and about 39.07 MiB/s SQLite growth.

- Targeted Recorder tests: 11 passed
- Broad memory/pubsub/recorder suite: 511 passed, 1 deselected
- Ruff, strict mypy, and all hooks applicable to changed files passed


dimos/core/transport.py
```diff
@@ -226,6 +226,30 @@ def stop(self) -> None:
         self._started = False
 
 
+class pSHMQueueTransport(pSHMTransport[T]):
```
TomCC7: we should maybe just override existing shmtransport

dimos/protocol/pubsub/shm/ipc_factory.py
```diff
@@ -349,6 +349,7 @@ def __init__(
         self._frame_nbytes = int(self._dtype.itemsize * np.prod(self._shape))
         self._slots = int(slots)
         self._pub_lock = threading.Lock()
+        self.dropped_since_last_read = 0
```
TomCC7: is this required seems not really used

dimos/protocol/pubsub/impl/shmpubsub.py
```diff
@@ -329,6 +354,13 @@ class PickleSharedMemory(
     ...
 
 
+class ReliablePickleSharedMemory(PickleSharedMemory):
```
TomCC7: also seems no need to make a 'reliable' one as it's just the requirement

dimos/msgs/sensor_msgs/Image.py
```diff
@@ -544,7 +544,6 @@ def to_jpeg_bytes(self, quality: int = 75) -> bytes:
         Returns:
             Raw JPEG bytes.
         """
-        # Canonicalize to RGB so JPEG bytes are deterministic regardless of input format.
```
TomCC7, line 547: revert

dimos/memory/recorder_queue.py
TomCC7, line 1: why we need a new writer and new queue?

dimos/core/transport.py
TomCC7: Done. pSHMTransport now owns the behavior through queue_size (default 256; 1 for latest-only). pSHMQueueTransport was removed.

dimos/protocol/pubsub/impl/shmpubsub.py
TomCC7: Done. ReliablePickleSharedMemory was removed. PickleSharedMemory(queue_size=...) directly selects the queued SHM channel.

dimos/protocol/pubsub/shm/ipc_factory.py
TomCC7: Removed. Overflow is now derived from the monotonic sequence gap in the fanout loop, and the new integration test verifies the reported gap. There is no mutable dropped_since_last_read state.

dimos/memory/recorder_queue.py
TomCC7: Consolidated. The separate RecorderQueue and RecordWriter modules were deleted. One lifecycle-owned RecordingPipeline now hides the required per-stream preparation and serialized batched-write stages.

dimos/msgs/sensor_msgs/Image.py
TomCC7: Reverted. Image.py is no longer part of the PR diff.

dimos/memory/module.py
```diff
@@ -453,19 +510,43 @@ def _collect_pose_setters(self) -> dict[str, PoseSetter]:
...
-        self.register_disposable(Disposable(self.tf.subscribe(on_tf)))
+    @rpc
+    def stop(self) -> None:
+        """Stop inputs, drain all accepted observations, then close storage."""
+        pipeline = self._recording_pipeline
+        if pipeline is None:
+            super().stop()
+            return
+        for subscription in self._recording_subscriptions:
+            subscription.dispose()
```
paul-nechifor: What is `self._recording_subscriptions` for? You don't need to dispose because you're already registering the for disposal

```
            self._recording_subscriptions.append(subscription)
            self.register_disposable(subscription)
```

dimos/memory/module.py
```diff
@@ -453,19 +510,43 @@ def _collect_pose_setters(self) -> dict[str, PoseSetter]:
...
-    def _record_tf(self) -> None:
-        """Record the live tf stream under "tf" (no-op without a wired tf port)."""
+    def _tf_processor(self) -> Processor | None:
+        """Build the preparation stage for tf, or return None when it is unwired."""
         if getattr(self.tf, "_transport", None) is None:
             logger.warning("Recorder: tf port has no transport — not recording tf")
-            return
+            return None
         tf_stream = self.store.stream("tf", TFMessage)
+        backend = cast("Backend[Any]", tf_stream._source)  # type: ignore[attr-defined]
```
paul-nechifor: Why cast and ignore?

dimos/memory/module.py
```diff
@@ -453,19 +510,43 @@ def _collect_pose_setters(self) -> dict[str, PoseSetter]:
...
+        return process_tf
 
-        self.register_disposable(Disposable(self.tf.subscribe(on_tf)))
+    @rpc
+    def stop(self) -> None:
+        """Stop inputs, drain all accepted observations, then close storage."""
+        pipeline = self._recording_pipeline
+        if pipeline is None:
+            super().stop()
+            return
```
paul-nechifor: Why return early? Shouldn't you just not call `pipeline.close()`? If the subscriptions needed to be disposed, this would skip disposing them.

dimos/memory/backend.py
```diff
@@ -93,6 +99,28 @@ def loader() -> Any:
...
+        """Persist prepared observations in one transaction, then publish them."""
+        results: list[Observation[T]] = []
+        try:
+            for prepared in prepared_appends:
+                results.append(self._persist_prepared(prepared))
+            if hasattr(self.metadata_store, "commit"):
+                self.metadata_store.commit()
+        except BaseException:
+            if hasattr(self.metadata_store, "rollback"):
+                self.metadata_store.rollback()
```
paul-nechifor: metadata_store is an ObservationStore, but ObservationStore does not have commit/rollback. SqliteObservationStore does.

Rule of thumb: if you're using `hasattr`, it's almost always incorrect.

The issue here is that ObservationStore should support `commit/rollback` (maybe others too)

dimos/memory/module.py
TomCC7: Fixed in 3950b0e919. Recorder now owns one CompositeDisposable, registers that group once, and adds the input/error subscriptions only to the group. stop() disposes the group before draining the pipeline.

dimos/memory/backend.py
TomCC7: Fixed in 3950b0e919. ObservationStore now explicitly declares commit()/rollback() as no-op hooks for non-transactional stores; SQLite overrides them. Backend calls the contract directly, with no hasattr checks.

dimos/memory/recording.py
```diff
@@ -0,0 +1,193 @@
...
+    @staticmethod
+    def _persist(batch: list[PreparedWrite]) -> None:
+        grouped: dict[int, list[PreparedAppend[Any]]] = defaultdict(list)
+        backends: dict[int, Backend[Any]] = {}
+        for write in batch:
+            key = id(write.backend)
+            backends[key] = write.backend
+            grouped[key].append(write.append)
+        for key, appends in grouped.items():
+            backends[key].append_prepared(appends)
```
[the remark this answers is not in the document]
TomCC7: Fixed in 3950b0e919. The writer still groups persistence by backend and commits once per backend, but defers notifications until every group commits, then publishes in the original accepted FIFO. The interleaved regression test covers A1, B2, A3, B4 exactly.

dimos/memory/module.py
TomCC7: Fixed in 3950b0e919. stop() no longer returns early: it always disposes subscriptions, conditionally closes the pipeline, and always calls super().stop().

dimos/memory/module.py
TomCC7: Fixed in 3950b0e919. Recorder no longer reaches through Stream._source and has no Any/cast/type-ignore in this path. Stream.prepare_write() retains the payload type and returns a type-erased behavior command for the writer.

dimos/memory/recording.py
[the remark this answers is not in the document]
TomCC7: Update: the attempted fix was reverted in 80aca344ec pending a cleaner redesign. Reopening this thread.

dimos/memory/module.py
TomCC7: Fixed in aa1271493b. A single CompositeDisposable now owns all recording subscriptions and is registered once. Individual subscriptions are only added to that group, which is explicitly disposed before the pipeline drain.

dimos/memory/module.py
TomCC7: Fixed in aa1271493b. Stream.prepare_write() is now the public typed boundary returning PreparedWrite[T]. Recorder no longer accesses Stream._source and adds no cast or type-ignore for this path.

dimos/memory/recording.py
```diff
@@ -0,0 +1,193 @@
...
+                    break
+            try:
+                self._persist(batch)
+            except BaseException as error:
+                self.fail(error)
+                return
+
+    @staticmethod
+    def _persist(batch: list[PreparedWrite]) -> None:
+        grouped: dict[int, list[PreparedAppend[Any]]] = defaultdict(list)
```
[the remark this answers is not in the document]
TomCC7: Fixed in aa1271493b. Backend.persist_prepared commits without publishing, then RecordingPipeline replays notifications from the original interleaved batch. The worker-path regression asserts commit-A, commit-B, then notify A1, B2, A3, B4.

dimos/memory/module.py
TomCC7: Fixed in aa1271493b. stop() no longer returns early: it always disposes subscriptions, closes the pipeline when present, calls the base shutdown, and then surfaces any drain failure.

dimos/memory/backend.py
TomCC7: Fixed in aa1271493b. ObservationStore now declares commit() and rollback() as part of its interface, with explicit implementations for SQLite, memory, and read-only MCAP stores. Backend calls the contract directly; the hasattr checks are gone.

dimos/memory/recording.py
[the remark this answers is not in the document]
TomCC7: Fixed in aa1271493b. Persistence is grouped per backend and committed without notification; only after every backend group commits does the writer notify by iterating the original accepted batch. test_pipeline_batches_backends_and_preserves_global_notification_fifo covers A1, B2, A3, B4 exactly.

dimos/core/stream.py
```diff
@@ -114,10 +114,12 @@ def __init__(
         self.name = name
         self.owner = owner
         self.type = type
-        if transport:
-            self._transport = transport
-        if not hasattr(self, "_transport"):
```
TomCC7: in what case would the _transport be predefined? maybe need to keep this logic

dimos/memory/backend.py
TomCC7: Revised after looking at the boundary again: commit/rollback no longer belongs to ObservationStore. Backend now owns the transaction scope because it coordinates metadata, blob, and vector persistence. SqliteStore supplies a context over the shared backend connection; non-SQLite backends use the default no-op context. The no-op methods were removed from memory and MCAP stores, and a real SQLite regression test verifies metadata rollback when blob persistence fails.

TomCC7: Final subtraction: we removed the transaction abstraction entirely. SQLite now uses autocommit, and neither ObservationStore nor Backend exposes transaction lifecycle. A 30-second production run still stored 22,500/22,500 at unchanged bandwidth; focused and broad suites pass. This also removes the runtime callable from the generic config path.

dimos/memory/module.py
TomCC7: Kept the CompositeDisposable: it is registered once for normal module cleanup and explicitly disposed before the FIFO drain so no new messages can arrive while stopping. Individual subscriptions are not registered separately.

dimos/memory/module.py
TomCC7: The prepared-write path that introduced this cast/ignore has been removed. Recorder now calls the existing typed Stream.append() directly; the only remaining pose-setter cast is unchanged from main.

dimos/memory/module.py
TomCC7: stop() still has no early return: it always disposes subscriptions, drains the pipeline when present, calls the base shutdown, and then surfaces a drain failure.

dimos/memory/backend.py
TomCC7: Corrected in the final subtraction. The original Backend and SqliteObservationStore commit/rollback behavior is restored exactly; this PR adds no transaction or store interface and Recorder uses the existing Stream.append() lifecycle.

dimos/core/stream.py
TomCC7: Restored the original stream connection logic exactly, including support for a predefined transport. core/stream.py is no longer changed by this PR.

dimos/memory/recording.py
```diff
@@ -0,0 +1,189 @@
...
+
+    @staticmethod
+    def _persist(batch: list[PreparedWrite[Any]]) -> None:
+        grouped: dict[Backend[Any], list[PreparedAppend[Any]]] = defaultdict(list)
+        for write in batch:
+            grouped[write.backend].append(write.append)
+        for backend, appends in grouped.items():
+            backend.persist_prepared(appends)
+        for write in batch:
+            write.backend.notify(write.append.observation)
```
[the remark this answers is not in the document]
TomCC7: Obsolete after the final subtraction. Backend grouping, prepared writes, deferred notification, and batching were removed. The single FIFO worker now calls the existing Stream.append() once per accepted message, so this cross-backend partial-commit path no longer exists.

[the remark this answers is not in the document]
TomCC7: Obsolete after the final subtraction. Backend grouping, prepared writes, deferred notification, and batching were removed. The single FIFO worker now calls the existing Stream.append() once per accepted message, so this cross-backend partial-commit path no longer exists.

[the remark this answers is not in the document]
TomCC7: Obsolete after the final subtraction. Backend grouping, prepared writes, deferred notification, and batching were removed. The single FIFO worker now calls the existing Stream.append() once per accepted message, so this cross-backend partial-commit path no longer exists.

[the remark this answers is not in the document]
TomCC7: Obsolete after the final subtraction. Backend grouping, prepared writes, deferred notification, and batching were removed. The single FIFO worker now calls the existing Stream.append() once per accepted message, so this cross-backend partial-commit path no longer exists.

dimos/memory/module.py
```diff
@@ -453,19 +490,33 @@ def _collect_pose_setters(self) -> dict[str, PoseSetter]:
...
+        pipeline = self._recording_pipeline
+        subscriptions = self._recording_subscriptions
+        if subscriptions is not None:
+            subscriptions.dispose()
+        failure: Exception | None = None
+        if pipeline is not None:
+            try:
+                pipeline.close()
+            except Exception as error:
+                failure = error
```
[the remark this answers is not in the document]
TomCC7: Fixed in bd61da688b. Each transport callback now enters a shared RecordingPipeline callback barrier before timestamping/submission. close() uses the same drain deadline to wait for admitted callbacks, then closes and drains the FIFO. A deterministic event-driven regression test pauses an in-flight callback before submit, verifies stop blocks, then verifies the frame is processed without callback-thread errors.

paul-nechifor (approved):

paul-nechifor (changes requested):
There are these requests: https://discord.com/channels/1341146487186391173/1491987064294609038/1540288322281414696

leshy (changes requested):
https://discord.com/channels/1341146487186391173/1491987064294609038/1540288322281414696 

a bunch of questions on this

TomCC7:
wrong approach here, closing

PR #3566, opened 2026-08-20
feat(web): serve the SDK and custom pages from the local relay

paul-nechifor:
* By default the SDK bundle is served by the relay at `/sdk.js`.
* With `--serve-dir <dir>` you can have the relay serve a custom directory instead of the cockpit.
* The SDK is only meant to be run with a local relay, so it's only available on loopback (you can choose to allow access from anywhere).


leshy (approved):
talked to Paul, merging early (doesn't touch other stuff) reviewing/testing SDK later

PR #3572, opened 2026-08-20
Ivan/feat/motion

leshy:
complex PR

local planner, trajectory controller, tf and movement manager bake-able and deployable on the go2 robot, consumes kronknav

tuning systems temporarily excluded as we discuss locomotion policy publishing


dimos/navigation/movement_manager/rust/src/module.rs
```diff
@@ -0,0 +1,169 @@
...
+
+use crate::mux::{scale_twist, Config, MuxState};
+
+/// The watchdog's tick rate. Fixed rather than configurable: `nav_stale_s` says
+/// when the deadman trips, this only says how finely it is sampled.
+const WATCHDOG_HZ: f64 = 10.0;
+
+#[derive(Module)]
+#[module(name = "cmd_vel_mux", setup = spawn_watchdog, teardown = stop_watchdog)]
+pub struct CmdVelMux {
```
aclauer, line 38: maybe we should actually translate the control coordinator to rust instead of movement manager?

PR #3574, opened 2026-08-20
fix(dimsim): wire observe to /color_image without republishing

juhasch:
## Summary
- DimSim already publishes JPEG camera frames on LCM `/color_image`, but `DimSimConnection.video_stream()` was an empty subject, so `observe()` always returned `None` and the agent thought it was blind.
- Forward those bus frames into the Go2 video cache, and skip republishing onto `/color_image` (`video_on_bus`) so we do not create an echo loop.

## Test plan
- [ ] `uv run pytest dimos/robot/unitree/test_dimsim_connection.py dimos/robot/unitree/go2/test_connection.py`
- [ ] `dimos --simulation dimsim --dimsim-scene=apartment run unitree-go2-agentic`
- [ ] After the renderer is up, ask the agent what it sees and confirm `observe` returns a camera frame instead of `None`

### DIFF
--- a/dimos/robot/unitree/dimsim_connection.py
+++ b/dimos/robot/unitree/dimsim_connection.py
@@ -49,20 +49,34 @@ class DimSimConnection:
         frame_id="camera_optical",
     )
 
+    # DimSim already publishes JPEG frames on LCM /color_image. GO2Connection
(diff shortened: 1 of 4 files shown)


juhasch:
PR #3601 has a better, more general fix. Thanks Paul!

PR #3575, opened 2026-08-20
G1 smooth bring up for low level control

mustafab0:
## Contribution path

## Problem

Bringing up G1 for low level control using Unitree SDK caused sudden jump when transitioning from `damping` mode to `locked standing`

## Solution

Added a soft start with a Kd factor to have damping when getting into initial mode. This mimics the official behaviour of g1 when bringing up

## How to Test

This requires G1 hardware to test

`dimos run unitree-g1-groot_wbc`

## AI assistance

Used Claude Code 5.0. to help build

## Checklist

### DIFF
--- a/dimos/control/tasks/servo_task/servo_task.py
+++ b/dimos/control/tasks/servo_task/servo_task.py
@@ -183,11 +183,17 @@ class JointServoTask(BaseControlTask):
                 # our joints, track their measured positions so the hold
                 # resumes from wherever that task releases them instead of
                 # snapping back to the previously latched target.
+                # A joint missing from this snapshot stays pending: clearing
+                # it unmeasured resumes the hold on the stale pre-preemption
+                # target, and no further callback arrives once released.
+                still_pending: set[str] = set()
                 for name in self._preempted_joints:
                     position = state.joints.joint_positions.get(name)
-                    if position is not None:
-                        target[self._name_to_index[name]] = position
-                self._preempted_joints.clear()
+                    if position is None:
+                        still_pending.add(name)
+                        continue
+                    target[self._name_to_index[name]] = position
+                self._preempted_joints = still_pending
                 # Preemption defers the timeout: the joints are actively
                 # driven, and the hold must be alive when they are released.
                 self._last_update_time = state.t_now
@@ -339,7 +345,7 @@ def create_task(cfg: Any, hardware: Any) -> JointServoTask:
         kwargs["default_positions"] = params.default_positions
         # Zero timeout pairs naturally with default-hold.
         kwargs.setdefault("timeout", 0.0)
+    kwargs["hold_measured_on_start"] = params.hold_measured_on_start
     if params.hold_measured_on_start:
-        kwargs["hold_measured_on_start"] = True
         kwargs.setdefault("timeout", 0.0)
     return JointServoTask(cfg.name, JointServoTaskConfig(**kwargs))  # type: ignore[arg-type]
--- a/dimos/control/tasks/servo_task/servo_task.py
+++ b/dimos/control/tasks/servo_task/servo_task.py
@@ -55,12 +55,17 @@ class JointServoTaskConfig:
             length if provided. Useful for "hold at this pose" tasks
             (e.g. arms during whole-body locomotion). Pair with
             timeout=0.0 to hold indefinitely.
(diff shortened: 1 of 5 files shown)


dimos/control/tasks/servo_task/servo_task.py
```diff
@@ -55,12 +55,17 @@ class JointServoTaskConfig:
...
+            measured state instead of a configured pose. Startup is then
+            bumpless: a configured pose would snap the joints there from
+            wherever the robot actually is, at full stiffness.
     """
 
     joint_names: list[str]
     priority: int = 10
     timeout: float = 0.5  # 500ms default timeout
     default_positions: list[float] | None = None
+    hold_measured_on_start: bool = False
```
TomCC7, line 68: should this just be a default which is always on? at least makes more sense for default to be true? do we have a need to start holding from the pre-configured pose?

mustafab0: Yes, agreed

TomCC7: I see, this is a generic task

mustafab0: Done

mustafab0: Even so, It would be better to start with starting from the current measured position in most cases

mustafab0: I reverted this. just flipping it touches a lot of blueprints, and not a good idea to do it here

dimos/robot/unitree/g1/test_wholebody_connection.py
```diff
@@ -0,0 +1,136 @@
...
+from dimos.robot.unitree.g1.wholebody_connection import (
+    _NUM_MOTOR_SLOTS,
+    _NUM_MOTORS,
+    G1WholeBodyConnection,
+    G1WholeBodyConnectionConfig,
+)
+
+
+@contextmanager
+def _connection():
```
paul-nechifor, line 33: It would be more conventional to convert this to pytest fixture instead of a context manager.

dimos/control/tasks/servo_task/test_servo_task.py
```diff
@@ -43,3 +43,86 @@ def test_on_joint_command_requires_all_claimed_joints() -> None:
...
+
+
+def test_hold_measured_on_start_latches_the_real_pose() -> None:
+    # A configured pose would snap the joints there from wherever the robot
+    # is, at full stiffness, on the first tick — the G1 arms did exactly that.
+    task = JointServoTask(
+        "servo_arms",
+        JointServoTaskConfig(joint_names=["a", "b"], timeout=0.0, hold_measured_on_start=True),
+    )
+    task.start()
```
paul-nechifor, line 86: If you start, don't you also need to stop at some point?

TomCC7:
@mustafab0 as discussed yesterday I think we should explore the direction of not using servo_task at all. A bunch of mods here is to make `servo_task` compliant to the joint trajectory task. If servo task is just for holding robot's current locations in case the robot driver won't do on their own we should maybe let this be the default behavior of joint trajectory task directly and the logic gets much simpler

EDIT: I checked the current JTT ability and turns out it can't handle the usecase of servo task here completely. So this would be a feature instead of a fix.

dimos/control/tasks/servo_task/servo_task.py
```diff
@@ -170,8 +225,19 @@ def on_preempted(self, by_task: str, joints: frozenset[str]) -> None:
...
-        if joints & self._joint_names:
-            logger.warning(f"JointServoTask {self._name} preempted by {by_task} on joints {joints}")
+        overlap = joints & self._joint_names
+        if not overlap:
+            return
+        with self._lock:
+            self._preempted_joints |= overlap
+            should_log = overlap != self._logged_preemption
+            if should_log:
+                self._logged_preemption = frozenset(overlap)
```
TomCC7, line 235: is this frozenset variable ever used? I only see assignment but seems no usage?

TomCC7:
I have this draft pr #3610 which tries to unify joint trajectory task and joint servo task. IMO joint servo task can just be a special case of joint trajectory task. can maybe test and see

PR #3610, opened 2026-08-21
refactor(control): stream joint targets through trajectory task

TomCC7:
## Contribution path

- Builds on #3575 and keeps its G1 connection-level gain soft-start.

## Summary

Use the canonical joint trajectory task for both finite trajectories and streamed joint-position targets.

- Remove the separate joint Servo task and its registry path.
- Route `joint_command` positions through the coordinator as one-point trajectories.
- Let new one-point trajectories replace the affected joints while preserving other active joint motions.
- Keep command continuity across replacements by anchoring at the last bounded command.
- Apply per-joint velocity bounds inside JTT; the original implementation defaults to 1 rad/s unless explicit limits are configured.
- Update robot blueprints and coordinator routing to use the canonical trajectory task.

This restores the earliest implementation of this PR exactly (`14d08592fa`) as a normal additive commit.

## Validation

- Control suite, G1 whole-body connection tests, and blueprint generation: 314 passed.

## AI assistance

OpenAI Codex with GPT-5 implemented and validated the change under human direction and review.

## Checklist


dimos/robot/unitree/g1/blueprints/basic/unitree_g1_groot_wbc.py
```diff
@@ -327,15 +323,10 @@ def _precomposed_g1_scene(package: ScenePackage) -> Path | None:
...
-        name="servo_arms",
-        type="servo",
-        joint_names=g1_arms,
+    _arm_holder = joint_trajectory_task(
+        g1_arms,
         priority=10,
-        auto_start=True,
-        params={"default_positions": ARM_DEFAULT_POSE},
+        velocity_limits={name: 1.0 for name in g1_arms},
     )
```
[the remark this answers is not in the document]
TomCC7: Fixed in 35a4107c41. The canonical trajectory task now has an opt-in hold_position_when_idle mode that latches finite measured positions before the first trajectory, holds untouched joints during partial trajectories, and retains final commanded targets afterward. Both simulated and real G1 arm-holder tasks enable it. Focused validation: 3 trajectory-task tests passed.

dimos/control/coordinator.py
```diff
@@ -197,6 +198,7 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         # dispatch. They must stay out of _dispatch: the twist mapper itself
         # dispatches joint_command, and _task_lock is not reentrant.
         self._stream_pre_hooks: dict[str, Callable[[Any], None]] = {
+            "joint_command": self._map_joint_positions_to_trajectory,
```
mustafab0: Please remove this from stream_pre_hook.

The joint_command stream should be exposed in the trajectory task card.

And the `_map_joint_positions_to_trajectory` should happen inside the TrajectoryTask itself.

The `_stream_pre_hook` is unfortunately needed for the `twist_command` out of limitation of our current architecture. With the refactor we should attempt to rremove it completely. So better to not add more pre-hooks here.

mustafab0 (changes requested):

dimos/control/coordinator.py
TomCC7: Addressed in 41dbc58fb2. `joint_command` is now declared by the trajectory task card as `("on_joint_command", "claim_overlap")`; the task filters to its claimed joints and converts position-bearing `JointState` messages into one-point trajectories internally. The coordinator joint-position pre-hook and its special subscription path are removed. The existing `twist_command` pre-hook is unchanged. Added routing tests that assert one streamed command produces exactly one trajectory execution.

TomCC7:
Addressed the two Greptile safety findings on the current head:

- Idle G1 arm safety: the trajectory task retains the restored `hold_position_when_idle` behavior, latching finite measured feedback at startup and continuing to hold untouched/final arm targets.
- Partial replacement status: status is now aggregated across all unique active per-joint runs (minimum progress, maximum elapsed/remaining), so replacing A does not report completion while B is still executing.

Regression coverage includes the exact long A/B then short A-only replacement case, idle hold behavior, streamed joint-command routing, and whole-body readiness. Local validation: 4,709 fast tests passed on the base PR head; the restacked teleop head passed 4,750 fast tests plus focused G1/control suites and mypy.

mustafab0 (approved):

PR #3615, opened 2026-08-22
feat(memory): add experimental native SQLite and MCAP recorder

TomCC7:
## Contribution path

- No linked issue; draft experiment for native Memory2 recording.

## Experimental status

The recorder is intentionally quarantined under `dimos.experimental.memory` and `native/rust/experimental`. There is no stable `dimos.memory.rust_recorder` alias. We can promote it by moving the implementation after the API and operational behavior have been evaluated by users.

## Problem

The Python recorder performs encoding and durable writes in its module process. High-throughput image and point-cloud streams need a native record path with configurable encoding concurrency, while recorded artifacts must remain readable through the existing stable Python Memory2 APIs.

## Solution

Add the experimental `RustRecorder`, backed by `dimos-memory-recorder`, and put native persistence behind a record-only Rust `RecordingStore` trait.

- **Native module:** `MemoryRecorder` implements the standard `dimos_module::Module` interface, registers configured raw inputs through `Builder`, and runs through `run_with_transport`. There is no standalone recorder runtime or CLI.
- **SQLite store:** preserves the Memory2 schema and Python registry, supports `lcm`, `jpeg`, and `lz4+lcm` storage codecs, splits TF batches, writes JSONB tags, and remains directly readable through stable `SqliteStore`.
- **MCAP store:** records original LCM wire payloads into indexed Zstd chunks. MCAP publish time carries source time and log time carries recorder reception time. Channel metadata lets stable `McapStore` discover and decode native artifacts.
- **Recorder pipeline:** standard module inputs feed a bounded encoding pipeline. SQLite uses the configured encoding worker pool; MCAP applies `encoding_threads` to Zstd compression. One writer restores arrival order, batches writes, and finalizes the artifact during module teardown. Codec or store failures terminate the process with a nonzero exit.
- **Configuration:** Python owns artifact lifecycle and SQLite stream registration. Native stdin contains only `store`, `encoding_threads`, and resolved `streams`; `encoding_threads` is the sole public performance knob. Images match the Python default of JPEG quality 50, with depth streams explicitly configurable as `lz4+lcm`.

The shared one-module Rust runner handles SIGTERM as well as Ctrl-C, matching `NativeModule.stop()` and allowing recorder teardown to finalize SQLite and MCAP artifacts.

MCAP rejects storage-codec overrides and append mode because it represents a transport capture. This PR remains record-only: no Rust replayer, Python binding, automatic depth inference, or standalone user-facing recorder CLI is included.

## How to test

```bash
uv run pytest dimos/memory/store/test_mcap.py dimos/experimental/memory/test_rust_recorder.py -q
# 15 passed

uv run pytest dimos/experimental/memory/test_rust_recorder_e2e.py -m self_hosted_large -q
# 2 passed: real Rust SQLite/MCAP writes read through stable Python Memory2

uv run pytest dimos/robot/test_all_blueprints_generation.py -q
# 1 passed
```

```bash
cd native/rust
cargo metadata --no-deps --format-version 1
cargo test -p dimos-memory-recorder
# 11 passed

cargo clippy -p dimos-memory-recorder --all-targets -- -D warnings
cargo fmt --all -- --check
cargo test -p dimos-module
# 134 passed; 1 doc test ignored
```

Ruff and targeted strict mypy checks pass. The separate `dim-recorder-perf` harness supports both the Python and Rust recorders for throughput comparisons.

## AI assistance

OpenAI Codex with GPT-5 implemented the Rust recorder stores, Python integration, tests, performance-harness migration, PR draft, and verification under user direction.

## Checklist


dimos/memory/rust_recorder.py
```diff
@@ -0,0 +1,239 @@
...
+from dimos.msgs.tf2_msgs.TFMessage import TFMessage
+from dimos.utils.data import backup_file
+from dimos.utils.logging_config import setup_logger
+
+logger = setup_logger()
+
+_RUST_WORKSPACE = DIMOS_PROJECT_ROOT / "native" / "rust"
+_EXECUTABLE = _RUST_WORKSPACE / "target" / "release" / "dimos-memory-recorder"
+_SUPPORTED_CODECS = {"lcm", "jpeg", "lz4+lcm"}
+CodecId: TypeAlias = Literal["lcm", "jpeg", "lz4+lcm"]
```
TomCC7: aren't these defined by python side already why we need these

dimos/memory/rust_recorder.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+class RustStreamSpec(BaseModel):
+    """Fully resolved stream settings sent to the native process."""
+
+    port: str
+    name: str
+    payload_type: str
+    codec: CodecId
+    payload_kind: Literal["raw", "image", "tf"]
```
TomCC7: better be enum

dimos/memory/store/mcap.py
```diff
@@ -79,35 +80,49 @@ class McapObservationStore(ObservationStore[Any]):
...
-    def __init__(self, *, name: str, path: str, topic: str, codec: StreamCodec, count: int) -> None:
+    def __init__(
+        self,
+        *,
+        name: str,
+        path: str,
+        topic: str,
+        codec: StreamCodec,
+        count: int,
+        uses_publish_time: bool,
```
TomCC7: seems bad to make this a variable as read/write is async

dimos/experimental/memory/rust_recorder.py
```diff
@@ -0,0 +1,271 @@
...
+        path = Path(os.fspath(value))
+        if not path.is_absolute():
+            path = DIMOS_PROJECT_ROOT / path
+        return str(path)
+
+
+class RustSqliteStoreConfig(RustStoreConfig):
+    """Write a Python-compatible Mem2 SQLite artifact."""
+
+    kind: Literal["sqlite"] = "sqlite"
```
TomCC7, line 71: should follow discrimminated union convention for mcap, kind should be a base property

dimos/memory/rust_recorder.py
```diff
@@ -0,0 +1,271 @@
...
+class RustRecorderConfig(NativeModuleConfig):
+    """Configuration for :class:`RustRecorder`."""
+
+    executable: str = str(_EXECUTABLE)
+    build_command: str = "cargo build --release -p dimos-memory-recorder"
+    cwd: str = str(_RUST_WORKSPACE)
+    stdin_config: bool = True
+
+    store: RustRecordingStoreConfig = Field(default_factory=RustSqliteStoreConfig)
+    on_existing: OnExisting = OnExisting.BACKUP
```
TomCC7: what's this for

dimos/memory/rust_recorder.py
```diff
@@ -0,0 +1,271 @@
...
+    backup_keep_last: int = Field(default=10, ge=0)
```
TomCC7: too many undocumented config options overall, need description

dimos/memory/rust_recorder.py
```diff
@@ -0,0 +1,271 @@
...
+    record_tf: bool = True
+    stream_remapping: dict[str, str] = Field(default_factory=dict)
+    stream_codecs: dict[str, str] = Field(default_factory=dict)
+
+    encoding_threads: int = Field(default=4, ge=1)
+    queue_capacity: int = Field(default=256, ge=1)
+    write_batch_size: int = Field(default=128, ge=1)
+    flush_interval_ms: int = Field(default=100, ge=1)
+    jpeg_quality: int = Field(default=50, ge=0, le=100)
```
TomCC7: config boundary leaking imo

dimos/memory/rust_recorder.py
```diff
@@ -0,0 +1,271 @@
...
+
+    streams: list[RustStreamSpec] = Field(default_factory=list)
+
+    def to_config_dict(self) -> dict[str, Any]:
```
TomCC7: why is this needed? I prefer direct config.

dimos/experimental/memory/rust_recorder.py
```diff
@@ -0,0 +1,271 @@
...
+                            name="tf",
+                            payload_type=f"{TFMessage.__module__}.{TFMessage.__qualname__}",
+                            codec="lcm",
+                            payload_kind="tf",
+                        )
+                    )
+                continue
+
+            stream_name = self.config.stream_remapping.get(port_name, port_name)
+            if self.config.store.kind == "mcap":
```
TomCC7: does it mean mcap only supports lcm encoding rn?

dimos/memory/rust_recorder.py
```diff
@@ -0,0 +1,271 @@
...
+            else:
+                codec = self.config.stream_codecs.get(stream_name, self._default_codec(port.type))
+            self._validate_codec(stream_name, port.type, codec)
+            specs.append(
+                RustStreamSpec(
+                    port=port_name,
+                    name=stream_name,
+                    payload_type=f"{port.type.__module__}.{port.type.__qualname__}",
+                    codec=cast("CodecId", codec),
+                    payload_kind="image" if issubclass(port.type, Image) else "raw",
```
TomCC7: why is this needed when we already have codec field

dimos/experimental/memory/rust_recorder.py
```diff
@@ -0,0 +1,271 @@
...
+            )
+
+        if not specs:
+            logger.warning("Native recorder has no connected streams")
+        return specs
+
+    @staticmethod
+    def _default_codec(payload_type: type[Any]) -> CodecId:
+        if issubclass(payload_type, Image):
+            return "jpeg"
```
TomCC7, line 231: jpeg is a horrible default for depth image

also does python recorder put default codec config here?

native/rust/dimos-memory-recorder/src/main.rs
```diff
@@ -0,0 +1,122 @@
...
+        let callback_stream = Arc::clone(&stream);
+        let callback_handle = handle.clone();
+        let callback: Dispatch =
+            Arc::new(move |data| callback_handle.record(Arc::clone(&callback_stream), data));
+        transport
+            .subscribe(topic, callback)
+            .await
+            .with_context(|| format!("failed to subscribe to {topic:?}"))?;
+        info!(port = %stream.port, topic = %topic, store_stream = %stream.name, codec = ?stream.codec, "recording stream");
+    }
```
TomCC7: this whole thing should be a function of the library main should be just calling

native/rust/dimos-memory-recorder/src/main.rs
```diff
@@ -0,0 +1,122 @@
...
+        _ = terminate.recv() => Ok(()),
+    }
+}
+
+#[cfg(not(unix))]
+async fn shutdown_signal() -> io::Result<()> {
+    tokio::signal::ctrl_c().await
+}
+
+async fn read_launch() -> Result<(Value, LaunchConfig)> {
```
TomCC7: doesn't dimos rust lib have the util already

native/rust/dimos-memory-recorder/src/main.rs
```diff
@@ -0,0 +1,122 @@
...
+    BufReader::new(tokio::io::stdin())
+        .read_line(&mut line)
+        .await
+        .context("failed to read launch config")?;
+    let value: Value = serde_json::from_str(line.trim()).context("invalid launch config JSON")?;
+    let launch = serde_json::from_value(value.clone()).context("invalid memory recorder config")?;
+    Ok((value, launch))
+}
+
+fn init_tracing() {
```
TomCC7: is tracing a typical thing with other rust modules as well

dimos/experimental/memory/rust/src/store/sqlite.rs
```diff
@@ -0,0 +1,121 @@
...
+        Ok(())
+    }
+}
+
+fn ensure_stream_tables(connection: &Connection, stream: &StreamConfig) -> Result<()> {
+    let name = &stream.name;
+    validate_identifier(name)?;
+    connection.execute_batch(&format!(
+        r#"
+        CREATE TABLE IF NOT EXISTS "{name}" (
```
TomCC7, line 60: is that typical to write raw sql? is there more pragmatic/less error-prone way to do it? how is python handling this?

dimos/memory/store/mcap.py
TomCC7, line 1: need justification on mcap change. we should add minimal change in the code as it's not the PR's focus

native/rust/dimos-module/src/zenoh.rs
```diff
@@ -215,8 +215,8 @@ async fn await_connect(session: &Session, endpoints: &[String], mode: Mode, time
 }
 
 impl ZenohTransport {
-    /// Open the session the launch config describes.
-    pub(crate) async fn from_launch(launch: &serde_json::Value) -> io::Result<Self> {
+    /// Open a transport from the session settings in a coordinator launch object.
+    pub async fn from_launch(launch: &serde_json::Value) -> io::Result<Self> {
```
TomCC7: is the change required? explain pls

native/rust/dimos-memory-recorder/src/lib.rs
```diff
@@ -0,0 +1,845 @@
...
+    macro_rules! stamped {
+        ($message_type:ty) => {{
+            let message = <$message_type>::decode(&job.data)
+                .with_context(|| format!("invalid LCM {}", job.stream.payload_type))?;
+            (message.header.stamp.sec, message.header.stamp.nsec)
+        }};
+    }
+
+    let (sec, nsec) = match job.stream.payload_type.as_str() {
+        "dimos.msgs.geometry_msgs.PointStamped.PointStamped" => stamped!(PointStamped),
```
TomCC7: need better way to handle this map, should be an easy matching function

native/rust/dimos-memory-recorder/src/lib.rs
```diff
@@ -0,0 +1,845 @@
...
+                return Ok(job.reception_ts);
+            };
+            (transform.header.stamp.sec, transform.header.stamp.nsec)
+        }
+        _ => return Ok(job.reception_ts),
+    };
+    Ok(header_timestamp(sec, nsec, job.reception_ts))
+}
+
+fn encode_tf(job: &EncodeJob) -> Result<Vec<StoredObservation>> {
```
TomCC7: encode utilities better be separated. lib.rs is bulky rn

dimos/experimental/memory/rust_recorder.py
TomCC7: Fixed. kind is now declared on RustStoreConfig and specialized by the discriminated SQLite and MCAP variants.

dimos/memory/store/mcap.py
TomCC7: This is immutable per-channel metadata read from the MCAP summary, not mutable read/write state. I renamed it to observation_uses_publish_time and documented that every iterator owns its own file reader.

dimos/memory/rust_recorder.py
TomCC7: Removed payload_kind entirely. TF behavior is derived from the fully qualified payload type, so there is no second discriminator to keep consistent.

dimos/memory/rust_recorder.py
TomCC7: Removed CodecId. Python now keeps only the native capability set for preflight validation; the resolved codec is sent as a plain string and Rust owns the strict codec enum.

dimos/memory/rust_recorder.py
TomCC7: on_existing is the Python-owned artifact lifecycle policy, matching the Python recorder: backup, overwrite, append, or error before the Rust process starts. It is now documented and excluded from native config.

dimos/memory/rust_recorder.py
TomCC7: Agreed. JPEG quality and queue/write internals no longer cross the Python/native boundary. The native config now contains only store, encoding_threads, and resolved streams.

dimos/memory/rust_recorder.py
TomCC7: Removed payload_kind. codec now describes storage encoding only; TF-specific handling is derived from payload_type.

dimos/memory/rust_recorder.py
TomCC7: Reduced the public tuning surface: queue capacity, batch size, flush interval, and JPEG quality are internal constants now. The remaining recorder fields have descriptions; encoding_threads is the sole throughput knob.

dimos/experimental/memory/rust_recorder.py
TomCC7: Yes, intentionally for this migration step: the recorder accepts LCM-backed stream types and MCAP preserves their original wire packets. MCAP does not apply SQLite storage codecs such as JPEG or lz4+lcm.

dimos/memory/rust_recorder.py
TomCC7: Removed the custom serialization and CLI overrides. NativeModuleConfig.to_config_dict now serializes the direct model boundary, with Python-only lifecycle fields excluded declaratively.

native/rust/dimos-memory-recorder/src/main.rs
TomCC7: Yes. I exposed and reused dimos_module::read_launch_config instead of keeping a recorder-local stdin parser.

native/rust/dimos-memory-recorder/src/main.rs
TomCC7: Done. Runtime orchestration now lives in dimos_memory_recorder::run and main only calls it, logs the terminal error, and exits nonzero.

dimos/experimental/memory/rust_recorder.py
TomCC7: Confirmed the Python recorder defaults every Image stream to JPEG quality 50. This commit matches that contract exactly for interoperability; depth streams must opt into lz4+lcm through stream_codecs. Automatic depth inference stays out of this PR.

dimos/experimental/memory/rust/src/store/sqlite.rs
TomCC7: Kept the raw DDL because stream names are dynamic SQL identifiers and cannot be bound parameters; the Python SQLite store also constructs per-stream tables this way. Observation values remain parameterized, and stream names are validated/quoted before reaching this layer.

native/rust/dimos-memory-recorder/src/main.rs
TomCC7: Yes. dimos-module already owns the standard JSON tracing setup. The recorder now reuses dimos_module::init_tracing and dropped its direct tracing-subscriber dependency.

native/rust/dimos-memory-recorder/src/lib.rs
TomCC7: Done. Codec and TF transformation code now lives in encoding.rs, timestamp extraction in timestamp.rs, and lib.rs is limited to config, pipeline, transport orchestration, and ordering/write coordination.

native/rust/dimos-memory-recorder/src/lib.rs
TomCC7: Moved this into timestamp::timestamp_for so source-time compatibility is centralized and lib.rs only orchestrates the pipeline. A generated timestamp trait would be a separate feature, so this PR keeps the existing Python-compatible matching behavior.

dimos/memory/store/mcap.py
TomCC7: Reduced this to the minimum reader contract needed for native MCAP: recognize channel metadata that says publish_time is the source observation timestamp. Without it Python would expose reception/log time. The new large test writes real Rust MCAP and reads the exact timestamp and payload through McapStore.

native/rust/dimos-module/src/zenoh.rs
TomCC7: Required for the standalone recorder path: it receives dynamic coordinator launch JSON and must initialize Zenoh from that value, while the typed generic module runner cannot construct this record-only stream set. The change only exposes the existing constructor; behavior is unchanged.

TomCC7: You are right; there is no standalone recorder use case. I had conflated a separately spawned native process with a standalone application. Fixed in 512589fc6a: MemoryRecorder now implements the standard dimos_module::Module interface, registers its configured raw inputs through Builder, and main calls run_with_transport::<MemoryRecorder>(). The public ZenohTransport::from_launch change is reverted, along with the launch/tracing visibility changes. I also taught the shared one-module runner to handle SIGTERM because Python NativeModule.stop() uses SIGTERM and the recorder must finalize its artifact during normal module teardown.

dimos/experimental/memory/rust/src/decoding.rs
```diff
@@ -0,0 +1,162 @@
...
+            ),
+        })
+        .collect())
+}
+
+/// Read source time while transport-decoding common stamped LCM payloads.
+///
+/// Unknown LCM payloads remain opaque and use reception time, matching the
+/// Python recorder's fallback for messages without a ``ts`` attribute.
+fn source_timestamp(payload_type: &str, data: &[u8], reception_ts: f64) -> Result<f64> {
```
TomCC7, line 100: I don't like this but seems no better way to do it...

dimos/memory/store/mcap.py
```diff
@@ -169,8 +192,20 @@ def __init__(
...
                 name = name_of.get(ch.topic) or _slug(ch.topic)
+                payload_module = ch.metadata.get("dimos.payload_type")
+                if (
+                    ch.topic not in self._codecs
+                    and ch.message_encoding in _SELF_DESCRIBING_CODEC_IDS
+                    and payload_module is not None
+                ):
+                    self._codecs[ch.topic] = cast(
+                        "StreamCodec", codec_from_id(ch.message_encoding, payload_module)
+                    )
```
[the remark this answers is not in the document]
TomCC7: Fixed in 8486c5731e. McapStore no longer resolves LCM or lz4+lcm payload types from artifact metadata; those channels remain raw unless the caller supplies a trusted codec through the existing codecs map. JPEG still decodes automatically because its payload type is fixed. Added a regression test that patches importlib.import_module and verifies malicious dimos.payload_type metadata cannot trigger an import, plus updated the Rust MCAP interoperability test and docs to use an explicit codec.

.github/workflows/ci.yml
```diff
@@ -303,6 +303,11 @@ jobs:
             extra-trusted-public-keys = dimensionalos.cachix.org-1:20ynj6TjpoD3qTxkdNoeHtgs2G2pNvgAq1EQYLTHJXI=
           INPUT_SET_AS_TRUSTED_USER: "true"
         run: bash docker/ros/install-nix.sh
+      - name: Build Rust memory recorder Nix package
```
aclauer, line 309: Why is this in the c++ job? Should we rename this job or move this part to the Rust job

dimos/robot/manipulators/dual_openyam/config.py
aclauer, line 15: Were all these files in the PR when I reviewed yesterday? I don't remember looking at these lol

.github/workflows/ci.yml
TomCC7: Agreed that the `cpp` name was misleading. Renamed the job to `native` / `Native builds (C++ and Rust)` in 2721b856bb and updated `ci-complete.needs`. I kept the recorder package build in this job because it already installs Nix and prepares disk space for the native C++ packages; moving it to `rust` would duplicate that setup and add the ~7-minute package build to the Cargo/PyO3 job.

dimos/experimental/memory/rust/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "dimos-memory-recorder"
```
aclauer, line 2: the `native` dir is mostly for the core implementation of native modules (there's the Rust and CPP, both of which just contain the module framework). All the actual native module implementations are in the folder for their use. Don't we have a top level experimental directory?

dimos/robot/manipulators/dual_openyam/config.py
TomCC7: I don't know how stacked pr is not resolving this automatically but I need to rebase the branch...

native/rust/experimental/dimos-memory-recorder/src/lib.rs
```diff
@@ -0,0 +1,858 @@
...
+pub struct RecorderConfig {
+    pub store: store::RecordingStoreConfig,
+    #[validate(range(min = 1))]
+    pub encoding_threads: usize,
+    pub streams: Vec<StreamConfig>,
+}
+
+impl RecorderConfig {
+    pub fn validate(&self) -> Result<()> {
+        if self.encoding_threads == 0 {
```
aclauer: THis is already validated on line 79

native/rust/experimental/dimos-memory-recorder/src/lib.rs
```diff
@@ -0,0 +1,858 @@
...
+#[derive(Clone)]
+pub struct RecorderConfig {
+    pub store: store::RecordingStoreConfig,
+    #[validate(range(min = 1))]
+    pub encoding_threads: usize,
+    pub streams: Vec<StreamConfig>,
+}
+
+impl RecorderConfig {
+    pub fn validate(&self) -> Result<()> {
```
aclauer: the standard way for doing something like this is to use the `#[validate(schema(function = "validate_config"))]` attribute and then you put the validation logic in there. That way you never have to manually call the validation and never have to worry about over/under validating. Instead it happens exactly once on the config struct

dimos/robot/manipulators/dual_openyam/config.py
aclauer: lol I feel like i run in to so many issues with stacked prs haha

dimos/experimental/memory/rust/src/lib.rs
```diff
@@ -0,0 +1,858 @@
...
+                result
+            })
+            .context("failed to start memory writer thread")?;
+
+        let worker_count = config.encoding_threads;
+        let mut workers = Vec::with_capacity(worker_count);
+        for index in 0..worker_count {
+            let receiver = encode_rx.clone();
+            let sender = write_tx.clone();
+            workers.push(
```
aclauer: You can just use a `worker_pool` for this I think, and it should simplify the thread spawning and stopping

dimos/experimental/memory/rust/src/lib.rs
```diff
@@ -0,0 +1,858 @@
...
+
+    pub fn handle(&self) -> RecorderHandle {
+        self.handle.clone()
+    }
+
+    pub fn failure_receiver(&self) -> Receiver<String> {
+        self.failure.clone()
+    }
+
+    pub fn shutdown(mut self) -> Result<WriterStats> {
```
aclauer, line 192: I think this would be simplified as well with a `worker_pool`

dimos/experimental/memory/rust/src/decoding.rs
TomCC7: Agreed. I am keeping this narrow: the transport provides raw LCM bytes, while the store needs source timestamps before storage encoding. The generated LCM types currently expose neither reflection nor a common stamped-message trait, so this explicit dispatch remains the compatibility shim.

dimos/experimental/memory/rust/src/lib.rs
TomCC7: Same conclusion for shutdown: Rayon ThreadPool does not expose join handles for spawned tasks, so we would need to add a completion channel/barrier before the writer can close safely. The current handles are the shutdown barrier and panic propagation mechanism, so I am keeping them.

native/rust/experimental/dimos-memory-recorder/src/lib.rs
TomCC7: Recorder policy validation now lives in Python, where the dynamic stream plan is constructed: duplicate names, codec/type compatibility, TF codec selection, and artifact lifecycle. Rust retains only declarative wire-boundary validation such as encoding_threads >= 1, so there is no remaining cross-field schema function to call.

dimos/experimental/memory/rust/Cargo.toml
TomCC7: Moved the complete crate and its flake to dimos/experimental/memory/rust, beside the experimental Python module. native/rust now contains only the shared native-module framework crates.

dimos/robot/manipulators/dual_openyam/config.py
TomCC7: They came from the stale stack state. I rebased only the recorder commits onto the current cc/feat/openyam-macos-support head; the OpenYam files are no longer part of this PR diff.

dimos/experimental/memory/rust/src/lib.rs
TomCC7: I checked dimos_module::worker_pool. It is a Rayon pool intended for finite pool.install parallel sections. These are long-lived bounded-channel consumers; shutdown must drain every queued sequence, join workers, propagate panics, and only then close the ordered writer. Using Rayon here would still require N blocking tasks plus a separate completion barrier, while its internal task queue is unbounded. The explicit threads preserve those lifecycle and backpressure guarantees with less machinery.

native/rust/experimental/dimos-memory-recorder/src/lib.rs
TomCC7: Removed this duplicate check and both manual validate calls. The generated field validator now runs once at the native stdin boundary.

dimos/experimental/memory/rust/src/lib.rs
TomCC7: hmm AI comment above apparently... tldr is the current worker is persistent while rayon worker pool is for transient tasks (like encoding for a single message?) I'll see if that way's easier

dimos/experimental/memory/rust/src/decoding.rs
aclauer: Ahh I see, probably something we could consider adding to the messages in dimos-lcm. But not important now.

aclauer (dismissed):

mustafab0 (dismissed):

aclauer (dismissed):

.github/workflows/ci.yml
```diff
@@ -955,12 +971,12 @@ jobs:
...
     - lint
     - rust
-    - cpp
+    - native
     - md-babel
     - web
     - tests
     - self-hosted-tests
-    #- self-hosted-large-tests
+    - self-hosted-large-tests
```
paul-nechifor: Why turn it back on?

.github/workflows/ci.yml
paul-nechifor: But why not move it to `rust:`?

.github/workflows/ci.yml
TomCC7: You're right; the whole large suite did not need to become required. Fixed in 1fded7ac3d: the recorder interoperability file now uses the ordinary self_hosted marker, whose already-required job has Nix/Cachix and the Python reader dependencies. The Rust SQLite, MCAP, and TF compatibility checks therefore remain blocking, while self-hosted-large-tests returns to the existing non-required policy. I also removed the now-unused Nix setup from the large job.

.github/workflows/ci.yml
TomCC7: I kept it in the native job because this step validates the Nix-packaged executable and its native dependency closure; the rust job already runs workspace fmt, Clippy, and tests. More importantly, cold recorder package builds on this PR took 7–11 minutes while the rust job takes about 5–6 minutes, so moving the build there would serialize those costs. Keeping it in the native job runs packaging in parallel and reuses that job's existing Nix setup.

paul-nechifor (dismissed):

paul-nechifor (dismissed):

paul-nechifor (approved):

PR #3617, opened 2026-08-22
Zenoh is the default transport, pinned to localhost

leshy:
`dimos/core/transport_factory.py` will need a rewrite, we need dimos normalized qos, would be nice to start removing pickling and defining custom lcm_msg everywhere, but will do this later.

rpc also needs conversion to json


docs/usage/transports/index.md
```diff
@@ -42,37 +42,54 @@ dimos --transport=lcm run unitree-go2
...
-| Any other platform | `lcm` |
+### What the default talks to
+
+A stock zenoh session is pinned to localhost. It listens on `tcp/127.0.0.1:0` and
+scouts for peers over loopback only, so sibling dimOS processes on this machine
+find each other and nothing on the LAN can link to them.
+
+Peers on this machine carry their data through shared memory, not the socket.
+Zenoh negotiates it per link at handshake, so it only ever applies where both
+ends map the same segment -- a remote peer keeps getting the payload over TCP.
```
paul-nechifor: I assume `--` is used here because we ban `—`.

Personally, I've configured Claude to prefer smaller sentences instead of using `—` or `--`. Here as well, it could just be two sentences instead of emdash.

dimos/protocol/service/test_zenohservice.py
```diff
@@ -111,12 +111,57 @@ def test_close_all_empties_the_pool_even_when_a_session_will_not_close(
...
+
 def test_different_modes_produce_different_keys() -> None:
     peer = ZenohConfig(mode="peer")
     client = ZenohConfig(mode="client")
     assert peer.session_key != client.session_key
 
 
+def test_pickle_round_trip_sheds_the_session(session_pool) -> None:
+    """A module travels to its worker by pickle and re-acquires on the far side."""
+    import pickle
```
paul-nechifor: These imports should be at the top.

dimos/protocol/rpc/zenohrpc.py
```diff
@@ -57,6 +57,24 @@ def __init__(
         self._pending: dict[int, Callable[..., Any]] = {}
         self._call_counter = count(1)
 
+    def __getstate__(self):  # type: ignore[no-untyped-def]
```
paul-nechifor, line 60: What is this needed for?

I've added a few `__getstate__`/`__setstate__` myself and they were often hacks. 😅

dimos/e2e_tests/conftest.py
```diff
@@ -31,6 +32,19 @@
 from dimos.simulation.mujoco.person_on_track import PersonTrackPublisher
 
 
+@pytest.fixture(autouse=True)
+def _pin_to_lcm(monkeypatch: pytest.MonkeyPatch) -> None:
```
paul-nechifor, line 38: Do you think it's worth it to create a ticket on Linear to switch this over to Zenoh later?

This seems okay, but wide autouse fixtures often create surprise. Most people would expect the default to be used, but this kinda secretly switches to LCM.

paul-nechifor (dismissed):

dimos/protocol/rpc/zenohrpc.py
leshy: yeah they are hacks, I thought we are pickling transports now (which we shouldn't have ever done) so allowed it

dimos/e2e_tests/conftest.py
leshy: yes I agree, I didn't want to make this even deeper then it should be. 

I added grid tests for lcm/zenoh for core tho, idk if you noticed

dimos/protocol/rpc/zenohrpc.py
leshy: https://github.com/dimensionalOS/dimos/blob/b4cd9789cc68876adf87ff40a404677f127d69bf/dimos/protocol/service/lcmservice.py#L93-L105

worker returns methods which pull self so we pickle whole modules :crying_cat_face: :crying_cat_face: we need to sort it
https://github.com/dimensionalOS/dimos/blob/b4cd9789cc68876adf87ff40a404677f127d69bf/dimos/core/coordination/python_worker.py#L389-L391

paul-nechifor (approved):

dimos/e2e_tests/conftest.py
leshy: https://github.com/dimensionalOS/dimos/issues/3728

PR #3631, opened 2026-08-24
cuVSLAM 

jeff-hykin:
### Test

Plug in a realsense (455 or 435if)

```
dimos --viewer rerun run demo-cuvslam-realsense
```

Walk the camera around and watch the path in rerun

Expected performance (stereo with loop closure on no gating/rebasing)

Screenshot 2026-08-10 at 5 45 09 PM

ref DIM-1347


dimos/mapping/cuvslam/cuvslam.py
```diff
@@ -0,0 +1,217 @@
...
+_DRIVER_LIBS = (
+    "libcuda.so.1",
+    "libnvidia-ptxjitcompiler.so.1",
+    "libnvidia-nvvm.so.4",
+    "libnvidia-ml.so.1",
+)
+_DRIVER_LINK_DIR = CACHE_DIR / "nvidia-driver-libs"
+
+
+def driver_library_dir() -> Path | None:
```
jeff-hykin, line 65: I hate cuda

dimos/mapping/odometry_hist.py
```diff
@@ -0,0 +1,86 @@
+# Copyright 2026 Dimensional Inc.
```
jeff-hykin, line 1: This module only exists to draw the path of where the robot has been in rerun.

I was kinda surprised we didn't already have this. 

Its really hard to validate cuVSLAM is working live without this

dimos/mapping/cuvslam/demo_cuvslam_realsense.py
```diff
@@ -0,0 +1,83 @@
...
+            rerun_config={
+                "blueprint": cuvslam_rerun_blueprint,
+                "visual_override": {"world/odom_hist": path_at_true_height},
+            },
+        ),
+    )
+    .remappings(
+        [
+            (RealSenseCamera, "infrared_left", "image"),
+            (RealSenseCamera, "infrared_right", "image"),
```
jeff-hykin, line 77: "why remap both to image?"

For stereo/multi-cam, cuVSLAM expects all camera streams on `image`, then uses frame_id to differentiate which image is from which lens.

extra-good cause frame_id is needed anyways to place them in physical space

dimos/mapping/cuvslam/cuvslam.py
```diff
@@ -0,0 +1,288 @@
...
+    if existing:
+        parts.append(existing)
+    return {"LD_LIBRARY_PATH": ":".join(parts)}
+
+
+class CuvslamConfig(NativeModuleConfig):
+    cwd: str | None = "."
+    executable: str = "result/bin/cuvslam_odometry"
+    build_command: str | None = Field(
+        default_factory=lambda: f"nix build github:dimensionalOS/dimSLAM/v0.3.2#{sdk_variant()}"
```
jeff-hykin, line 198: different repo for licensing reasons (custom nvidia license)

dimos/mapping/cuvslam/test_cuvslam_replay.py
```diff
@@ -0,0 +1,179 @@
...
+    finally:
+        coordinator.stop()
+
+    positions = progress["positions"]
+    assert progress["error"] is None, progress["error"]
+    assert len(positions) >= SNIPPET_STEREO_PAIRS // 2, (
+        f"expected odometry for most stereo pairs, got {len(positions)}"
+    )
+    displacement = math.dist(positions[0], positions[-1])
+    assert 1.0 < displacement < 5.0, (
```
jeff-hykin: even though the cuVSLAM part is deterministic, the delivery of messages is not deterministic, so we still get variation (need pure modules to fix this)

jeff-hykin (commented):
-

dimos/mapping/odometry_hist.py
mustafab0: just call it odometry_history lol

mustafab0 (dismissed):
Looks ok 👍

dimos/mapping/cuvslam/cuvslam.py
aclauer: ah i see I was gonna ask where the native modules was haha

dimos/mapping/cuvslam/cuvslam.py
```diff
@@ -0,0 +1,288 @@
...
+    "nano",
+    "linux-x86-nvidia",
+    "linux-x86-no-nvidia",
+    "linux-arm-no-nvidia",
+    "darwin-apple-silicon",
+    "darwin-intel",
+]
+
+
+def detect_hardware() -> Hardware:
```
aclauer, line 104: Would these be things that we use outside of cuvslam and could be shared utils? Probably don't have anything else that needs them now, but could refactor in the future.

aclauer (dismissed):

dimos/mapping/cuvslam/cuvslam.py
jeff-hykin: yeah basically anything with cuda. I was debating pulling them out

mustafab0: I was gonna add similar comment, but thought irrelevant.

But I agree here. We should have a dimos utility/cli command that detects hardware and puts it in a manifest/config where every module needing it can just ask for.

aclauer (approved):

PR #3633, opened 2026-08-24
ray_tracing: place clouds purely through tf

jeff-hykin:
Make ray tracing use TF so it can accept depth from multiple sources. 
- place clouds purely through tf; drop the odometry input
- previously clouds were dropped at the odom pose, ignoring the sensor mount: a camera half a metre up and angled down put the floor in the wrong place if Odom was for base link (not the cam)
- tf lookup at each cloud's stamp resolves world_frame -> sensor frame through the mount edge
- cargoHash refresh to unstick the nix build

### DIFF
--- a/dimos/mapping/ray_tracing/rust/src/module.rs
+++ b/dimos/mapping/ray_tracing/rust/src/module.rs
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::VecDeque;
 use std::time::Duration;
 
 use crate::voxel_ray_tracer::{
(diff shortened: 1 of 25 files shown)


dimos/mapping/ray_tracing/rust/src/module.rs
```diff
@@ -204,48 +194,13 @@ fn emit_due(frame_count: u32, every: u32) -> bool {
     every != 0 && frame_count.is_multiple_of(every)
 }
 
-/// Odometry samples kept for cloud-stamp matching.
-const POSE_BUFFER_LEN: usize = 256;
-
-/// Max stamp gap between a cloud and the pose used to register it (s).
-const POSE_MATCH_TOLERANCE_S: f64 = 0.1;
+/// Max age of a tf sample used to place a cloud (s).
+const TF_MATCH_TOLERANCE_S: f64 = 0.5;
```
jeff-hykin: This should be a config

PR #3643, opened 2026-08-24
mls_planner: cap the surface closing image at the max map span

jeff-hykin:
A single corrupt point made `close_at_z` allocate a ~2.1e9-wide byte image; three orphaned mls_planner processes held 220 GB each after a replay.

- `voxelize` uses `f32 as i32`, which saturates out-of-range coordinates to `i32::MAX`
- `close_at_z` sizes a dense `GrayImage` from the min/max voxel index of the slice, so that one point stretches it across the whole gap
- new `max_map_span_m` config (default 500.0); slices wider than it skip closing and pass through unclosed
- regression test `a_far_outlier_does_not_allocate_the_gap`


aclauer:
Ah yes I see the bug now lol

aclauer:
Would it be better to change the morphological functions to do local operations? The imageproc functions are probably not what we want now that I think of it for this exact reason. Instead we can just do the dilation/erosion on the occupied voxels without allocating the entire 2d slice as an image

jeff-hykin:
Oops didn't realized you commented. I found it was partially an issue on my end too; a bad TF put the camera far away so it instantiated a really large surface. 

> we can just do the dilation/erosion on the occupied voxels without allocating the entire 2d slice as an image

That sounds like a good optimization for me.

PR #3674, opened 2026-08-25
fix(manip): retry Pink IK on seed-specific solver failures

mustafab0:
## What

Both perturbed-seed retry loops in `PinkIK` aborted the entire loop the first time the solver
raised. A single unlucky seed made a *reachable* grasp pose report `NO_SOLUTION`.

## Why a solver exception should retry

`pink.solve_ik` raises `NoSolutionFound` when the QP is infeasible:

```
pink/exceptions.py:49   class NoSolutionFound(PinkError)
pink/solve_ik.py:273    raise NoSolutionFound(problem, result)
message                 "QP solver did not find a solution to the differential IK problem"
```

QP infeasibility at a given configuration is frequently **seed-specific** — it says the
linearized step is infeasible from *this* seed, not that the pose is unreachable.
## Mapping vs solver: why only one of them retries

| branch | example | seed-dependent? | behavior |
|---|---|---|---|
| `except Exception` | `NoSolutionFound` (QP infeasible) | yes, often | record first failure, `continue` |
| `except ValueError` | joint not in model, mapping length mismatch | no | return immediately |

A `ValueError` here means the target or seed cannot be expressed in the model frame at all —
a structural mismatch that fails identically on every perturbed seed, so retrying only burns
the attempt budget. 

This split is safe because the two are **not** the same exception type. `PinkError` derives
straight from `Exception`, not from `ValueError`:

```
NoSolutionFound MRO = [NoSolutionFound, PinkError, Exception, BaseException, object]
issubclass(NoSolutionFound, ValueError) -> False
```

So QP infeasibility cannot be swallowed by the `ValueError` branch. Discriminating by exception
type is sufficient; no source-based discrimination is needed.

## Tests

Added to `test_pink_ik.py`, mocked at the `_solve_targets` boundary, mirroring the existing
`test_solve_retries_after_joint_limit_failure` pattern. The same trio for each loop:

No API change: no signature, return type, `IKStatus`, or message-format change.

### DIFF
--- a/dimos/manipulation/planning/kinematics/pink_ik.py
+++ b/dimos/manipulation/planning/kinematics/pink_ik.py
@@ -154,9 +154,17 @@ class PinkIK(_PinkSolverCore):
                     orientation_tolerance=orientation_tolerance,
                 )
             except ValueError as exc:
+                # Mapping failures are seed-independent: the target or seed cannot be
+                # expressed in the model frame at all, so perturbed retries cannot help.
                 return _failure(IKStatus.NO_SOLUTION, f"Pink IK mapping failed: {exc}")
             except Exception as exc:
-                return _failure(IKStatus.NO_SOLUTION, f"Pink IK solver failed: {exc}")
+                # Solver failures (notably pink's NoSolutionFound on QP infeasibility) are
+                # frequently seed-specific, so retry from the next perturbed seed.
+                if fallback_result is None:
+                    fallback_result = _failure(
+                        IKStatus.NO_SOLUTION, f"Pink IK solver failed: {exc}"
+                    )
+                continue
 
             if not result.is_success() or result.joint_state is None:
                 if fallback_result is None:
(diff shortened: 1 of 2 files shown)


dimos/manipulation/planning/kinematics/pink_ik.py
```diff
@@ -154,9 +154,17 @@ def _solve(
                     orientation_tolerance=orientation_tolerance,
                 )
             except ValueError as exc:
+                # Mapping failures are seed-independent: the target or seed cannot be
+                # expressed in the model frame at all, so perturbed retries cannot help.
                 return _failure(IKStatus.NO_SOLUTION, f"Pink IK mapping failed: {exc}")
             except Exception as exc:
```
TomCC7: this exception is the universal fallback case. if we want to add a retry logic here we should figure out the exact solver failure while keeping the fallback to still be `failure`?

TomCC7: I updated directly

leshy (approved):

paul-nechifor (approved):

PR #3680, opened 2026-08-25
docs: give the theme a shape language

spomichter:
The site is already dark, hairlined and unshadowed. What it lacks is any feedback when you point at something: nav rows recolour their text and nothing else, which on a forty-item sidebar leaves the pointer invisible.

Three rules, applied consistently: a pill for anything you click, a 1px ring rather than a shadow on anything raised, and colour and transform as the only transitions.

- Radius scale retuned to `0.375 / 0.75 / 1rem` plus a pill token
- Nav rows get a hover fill and a tinted active pill — the most visible change
- Admonitions drop the 3px left stripe for a tinted panel at 16px. The tint derives from the colour material already assigned the type, so note, warning and danger stay apart without a rule each
- Search, `.md-button`, the clipboard button, `details` summaries, tabs and header buttons all get the radius and a colour transition
- Prev/next become bordered cards

`mkdocs build --strict` passes; rendered locally against the real pages before and after.

Worth saying plainly: this is the smaller half of the gap. Across 72 pages the docs use 2 admonitions, 1 card grid, and zero buttons, tabs or collapsibles. Pages built out of components read better than prose, fences and tables, and styling components harder does little while almost nothing renders them. The landing page — currently the readme piped through mkdocs — is the other half. Both are content work, tracked separately.

### DIFF
--- a/misc/mkdocs/theme.css
+++ b/misc/mkdocs/theme.css
@@ -17,8 +17,10 @@
   --md-accent-fg-color: var(--dim-brand-bright);
 
   /* Rounding and hairlines, applied consistently to every raised surface. */
-  --dim-radius: 0.6rem;
-  --dim-radius-sm: 0.35rem;
+  --dim-radius: 0.75rem;
+  --dim-radius-sm: 0.375rem;
+  --dim-radius-lg: 1rem;
+  --dim-radius-pill: 9999px;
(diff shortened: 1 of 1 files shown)


paul-nechifor (approved):
Approved, although the round corners look a bit odd before you start typing:

Image

PR #3681, opened 2026-08-25
Add Mac to tests matrix

Dreamsorcerer:
This adds Mac to the regular tests matrix.


dimos/hardware/whole_body/openarm_damiao/test_adapter.py
```diff
@@ -14,6 +14,7 @@
 
 from collections.abc import Iterator
 import runpy
+import sys
```
Dreamsorcerer:
```suggestion
```

dimos/core/test_async_module_rpc_sync_to_async.py
```diff
@@ -108,5 +108,5 @@ def test_async_module_rpc_sync_to_async(start_module, in_transport, out_transpor
     queue = Queue()
     out_transport.subscribe(queue.put)
     in_transport.publish(4)
-    cubed = queue.get(timeout=0.1)
+    cubed = queue.get(timeout=5.0)
```
leshy: ouch I think this implies broken system, waiting 5s for an empty RPC, I don't think this should be loosened up, but fixed, it implies some deep problem with mac. so might want to still mark this as broken on mac

dimos/utils/shm.py
```diff
@@ -29,6 +29,7 @@
 raise, not a race to wait out.
```
leshy, line 29: our shm stuff has been out of control and IDK what this is. we already have a shmtransport (that should be deleted) and IDK what this file does, if I git blame it's mustafas so he can approve or deny, but IMO all of this should be deleted, since zenoh is shm

leshy (changes requested):

dimos/core/test_async_module_rpc_sync_to_async.py
Dreamsorcerer: Think this is another symptom of Mac runner oversleeping. It's flaky in CI, rather than a complete fail, but a large threshold will guarantee we don't fail on this flakiness.

dimos/utils/shm.py
Dreamsorcerer: Yeah, this is very awkward. Perfectly happy if it gets deleted, but this quick hack gets the tests passing.

dimos/core/test_async_module_rpc_sync_to_async.py
leshy: issue is I'd expect a lot other tests to be flakey if this works so badly, actually surprised so many are passing, I'd much rather ignore in mac then forget about it like this?

dimos/utils/shm.py
leshy: I defer to mustafa since it's his file, but I'd ignore and mark for deletion, to not load us with maintaining code and tests etc

mustafab0: I will create an issue and deprecate this as a follow up approving for now

mustafab0 (dismissed):

dimos/utils/shm.py
mustafab0: https://github.com/dimensionalOS/dimos/issues/4220

paul-nechifor (approved):

Issue #3671, opened 2026-08-25
SHM create race: attachers mmap a size-0 segment ("cannot mmap an empty file")


mustafab0:
**Symptom:** consumers crash attaching to a SHM segment with `ValueError: cannot mmap an empty file` (seen on real runs, July sprint).

**Race:** stdlib `SharedMemory(create=True, size=N)` does `shm_open` then `ftruncate`. An attacher that opens in that window sees size 0 and mmaps nothing. `_open_shm_with_retry` ([ipc_factory.py:49](dimos/protocol/pubsub/shm/ipc_factory.py#L49)) only catches `FileNotFoundError`, so it never covered this. Affected paths are the name-derived symmetric ones: `_create_or_open` ([:145](dimos/protocol/pubsub/shm/ipc_factory.py#L145)) attach arm, plus direct `SharedMemory(create=True, ...)` sites (~:174).

**Fix:** explicit readiness,  readers now wait until the segment actually has a size, with a 10s cap and a clear error if it never shows up. No sleep ladder, no blind retries.

**Blast radius:** shared transport infra — all robots, sim modules, cameras, every SHM pubsub user.

TomCC7:
Solution:
1. Default to zenoh
2. Configure zenoh shm
3. Remove the inhouse shm transport :)

mustafab0:
**Repro** (deterministic — injects a delay into the `shm_open`→`ftruncate` window; the real window is microseconds so it will not reproduce on demand without it). Save as `repro_shm_race.py` at the repo root, `python repro_shm_race.py`. Exits 0 when it reproduces.

```python
"""Repro for the SHM create race: attachers mmap a size-0 segment.

  python repro_shm_race.py            # deterministic (delay injected in the window)
  python repro_shm_race.py --natural  # no injection; races on real timing, may need luck

stdlib SharedMemory(create=True, size=N) does shm_open() THEN ftruncate(). An
attacher that opens between them sees a 0-byte file and mmap() fails.
"""

import os
import sys
import threading
import time
from multiprocessing.shared_memory import SharedMemory

from dimos.protocol.pubsub.shm.ipc_factory import _open_shm_with_retry

NAME = f"dimos_race_{os.getpid()}"
SIZE = 1 << 20
NATURAL = "--natural" in sys.argv

if not NATURAL:
    _real_ftruncate = os.ftruncate

    def _slow_ftruncate(fd, length):  # widen the window to make it deterministic
        time.sleep(0.05)
[...]

PR #3682, opened 2026-08-25
refactor(sim): route MuJoCo control through module ports

TomCC7:
## Summary

- make MuJoCo a normal connection module with typed motor-command and robot-state ports
- use the same port-backed control coordinator contract for real and simulated G1, xArm, and Piper stacks
- remove the custom shared-memory transports, RPC layer, simulation registries, and Unitree MuJoCo subprocess IPC
- make the existing xArm keyboard-teleop blueprints simulation-aware
- enable Zenoh shared memory in the native Rust module

## Architecture

```text
ControlCoordinator
        |
        | typed module ports
        v
real connection  or  MuJoCo connection module
```

The coordinator keeps the synchronous hardware-adapter interface needed by the control loop. The connection module holds the latest command, applies it on each physics step, and publishes state through ordinary DimOS streams. Transport selection remains a blueprint/runtime concern.

## Manual test

Until #3617 lands, select Zenoh explicitly:

```bash
dimos --simulation mujoco --transport zenoh run keyboard-teleop-xarm7
```

The existing keyboard controls should work without a simulator-specific teleop path.

## Validation

- `./bin/pytest-fast`: 4,554 passed, 33 skipped; the only failure was the blueprint generator's expected dirty-worktree check
- blueprint generation in CI comparison mode: passed
- focused control, transport, and simulation tests: 151 passed
- xArm keyboard-teleop tests: 12 passed
- simulation blueprint wiring and joint-count smoke checks: passed
- `ruff check`, `ruff format --check`, `compileall`, and `git diff --check`: passed
- `cargo check --manifest-path native/rust/dimos-module/Cargo.toml`: passed

## Dependency

This stacks with #3617, which changes the default transport to Zenoh. This PR does not duplicate that default change.

Addresses #3671.


TomCC7:
will revisit after cc refactor

PR #3690, opened 2026-08-26
fix(sim): stamp camera output at render time and pace on the monotonic clock

mustafab0:
## What

The camera publish loop stamped images, TF, and `camera_info` with `time.time()` sampled when the frame was consumed, not with the timestamp the frame already carries from when it was rendered.

## Why it matters

Every message for a frame was skewed later than the observation it describes by the render-to-publish latency, measured live at **4.5-101.7 ms (mean 52.7 ms)**, which is the same order as the 0.1 s TF lookup tolerance and can register object poses against the wrong TF sample.

### DIFF
--- a/dimos/simulation/engines/mujoco_sim_module.py
+++ b/dimos/simulation/engines/mujoco_sim_module.py
@@ -344,6 +344,12 @@ class MujocoSimModule(
         self._gripper_joint_range: tuple[float, float] = (0.0, 1.0)
         self._stop_event = threading.Event()
         self._publish_thread: threading.Thread | None = None
+        # Guards _camera_info_base and _latest_frame_ts, which the publish
+        # thread, the camera_info interval, the RPC thread, and stop() all
+        # touch. Hold it only across the field access itself: never across the
+        # publish-thread join in stop(), and never while calling a method that
+        # takes it again (it is not reentrant).
+        self._state_lock = threading.Lock()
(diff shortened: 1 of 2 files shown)


dimos/simulation/engines/mujoco_sim_module.py
```diff
@@ -378,17 +379,23 @@ def _depth_frame(self) -> str:
     def _depth_optical_frame(self) -> str:
         return f"{self.config.camera_name}_depth_optical_frame"
 
+    def _camera_info_ts(self) -> float:
```
paul-nechifor, line 383: Are you sure all of this works without any locks?

mustafab0: Yes this works without locks no problem. But there was an issue of one process reading the latest frame and then clearing timestamps.

The fix is to just make a local copy

paul-nechifor: `get_color_camera_info` gets `self._latest_frame_ts` without any lock. `_publish_loop` sets it without any lock. These are separate threads. You need a lock. You also need a lock for _camera_info_base. YOu can use the same one.

mustafab0: added `_state_lock`

paul-nechifor (approved):

Issue #3621, opened 2026-08-23
Tech spec: e-stop module


AaryanAgrawal:
# E-stop module

Add a Rust latching e-stop. Any condition can be added to set it, and only a manual reset can clear it.

## Current proposed conditions

1. **Manual stop:** `trigger=true` latches the e-stop.
2. **Humanoid fall detection:** unreadable IMU orientation or tilt above the configured limit latches the e-stop.

Each condition owns one typed input and one small check; adding one means adding its file, input handler, and physical config.

```text
control/estop/
├── conditions/manual.rs
├── conditions/humanoid_fall.rs
└── rust/src/main.rs
```

## Manual reset

`reset=true` clears the latch only when every condition is healthy, and leaves the robot disarmed until a separate start command.

## Embodiment defaults

Each embodiment owns one EStop-default bundle, and every root blueprint for that embodiment composes it.

```python
g1_estop_defaults = EStop.blueprint(max_tilt_deg=45.0)
g1_root = autoconnect(g1_hardware, g1_estop_defaults)
```

Derived blueprints inherit the bundle from their root. `autoconnect` links `imu`, and `estop` must reach `ControlCoordinator.set_estop()`.

Not run on hardware yet.

AaryanAgrawal:
Run on recorded data of the G1 falling, not a full real-hardware test.
![Fall detection replay](https://raw.githubusercontent.com/AaryanAgrawal/aaryan-dimensional/main/context/estop-fall-replay.png)

AaryanAgrawal:
![image](https://github.com/user-attachments/assets/1720a234-2307-4b5a-9889-acdf131342ef)

Dreamsorcerer:
Can you expand a little on the interface? e.g. Where does max_tilt_deg come from?

I'd like to ensure more complex cases are covered by the interface. Maybe there should be a callback or something? e.g. If there should be a condition that hits the stop when a robot arm would put itself in a position that could damage itself, or in a position that could injure someone, that needs something more complex than a single number threshold. Maybe you could just expose a bool attribute, but then you still need guarantees that the code updating that bool is running (at which point you probably need a timestamp or something too).

paul-nechifor:
> Add a Rust latching e-stop

Why does it have to be Rust? Rust modules have to be compiled. Seems simpler to use Python modules for a simple thing like this.

> Humanoid fall detection

[...]

PR #3691, opened 2026-08-26
estop spec + fall and haywire conditions

AaryanAgrawal:
**Sep 1 — the design moved.** Per the thread on #3621: conditions live in the module that can evaluate them, the latch is at the coordinator, sources are dumb callers.

```
 sources (many, dumb)          the latch (one)                  safe stop (per device)
 hosted-teleop button          ControlCoordinator               G1WholeBodyConnection
 any module with an EStop ref  estop() / clear_estop()          SimMujocoG1WholeBodyAdapter
                               set_estop(bool) kept              fall or flail -> one damping frame
                                                                 (kp=0, tau=0, kd kept), drop
                                                                 every later command, latch,
                                                                 then estop() up to the coordinator
```

- `dimos/robot/unitree/g1/protective.py`: two pure checks — pelvis tilt off gravity (45 deg) and a whole-body flail statistic (3 or more of 29 joints past 5 rad/s at once). Both the hardware connection and the MuJoCo adapter run them on every sample; per-embodiment defaults sit on `G1WholeBodyConnectionConfig`.
- The connection now stops the *device*: one damping LowCmd frame, then it refuses further motor commands, then it tells the coordinator. Before this it only asked the coordinator to stop talking, and the firmware held the last standing pose at full kp.
- The coordinator keeps the latch fixes and gains `estop()` / `clear_estop()` as one-liners over `set_estop`; the joint-speed check and the tick-loop hook are gone from it. `tick_loop.py` is back to `main`.
- `SimMujocoG1WholeBodyAdapter.get_limits()` added — without it the G1 GR00T sim blueprint could not start on `main` (the coordinator's Protocol check refuses the adapter).
- Thresholds were measured, not guessed: the real GR00T ONNX in headless MuJoCo through the real `MujocoSimModule` + `ControlCoordinator`, lifted by a mocap hook and pushed over (`flail_experiment.py`). SIMULATED: walking never puts two joints past 5 rad/s at once; lifted puts 9–11, fallen 13–17; the flail check fires 0.0–0.24 s after the feet leave the floor and 0.38–0.45 s before tilt reaches 45 deg. A lifted G1 never tilts past 39 deg, so the tilt check alone never catches a lift.

Test procedure, hardware walk baseline, and what is left: `dimos/robot/unitree/g1/SIM_TEST.md`. Not run on hardware. The text below is the Aug 26 description; where it describes a coordinator-side joint check or `set_estop` from the connection, this branch no longer does that.

---

Nothing in dimos notices when a robot falls over or when a policy drives a joint far faster than it should. `ControlCoordinator.set_estop()` already stops every task, but nothing automatic ever calls it, and no G1 task implements it, so on a G1 it reaches nothing.

This adds conditions at both ends and makes the shared path solid.

```
  a module           G1WholeBodyConnection sees the IMU at 500 Hz
    fall             tilt off gravity > max_tilt_deg
      |                _estop: EStop  ->  set_estop(True)
      |
  the coordinator    sees every joint each tick, in sim and on hardware
    flailing joint   |dq| > max_joint_speed_rad_s
      |                self.set_estop(True)
      v
  ControlCoordinator.set_estop   ->  every task inert  ->  GrootWBC disarms
```

Both routes end in the same call, so a condition can live wherever the data is.

### The Spec

`EStop` in `dimos/spec/control.py` is how a module reaches the stop without importing the coordinator. Declared bare, `_estop: EStop`, it is a required ref: launching a stack with nothing able to stop the robot raises `No module met that spec.` before anything moves. Declaring the concrete class instead resolves to `None` and only fails at the fall; a topic never fails at all.

Note this is a launch-time guarantee, not a CI one. `test_all_blueprints` imports blueprints but never deploys them, so it does not exercise ref resolution.

### `set_estop` was not actually a latch

Three fixes to the existing fan out, all reachable today:

- **It stored nothing.** `grep -c "_estopped" coordinator.py` returned 0, so despite the docstring, a task registered after a trip came up live. The latch now lives on the coordinator and `add_task` applies it.
- **It skipped silently.** `getattr(task, "set_estop", None)` with no else, which is why nobody noticed the G1 had zero coverage. It now logs the task it cannot stop.
- **One raising task aborted the loop**, leaving every later task running. Each task is now dispatched in isolation.

### Config

`max_tilt_deg` on the G1 connection defaults to 45 deg. `max_joint_speed_rad_s` on the coordinator defaults to `None`, off, because a robot agnostic coordinator has no safe default. No existing stack changes behaviour.

### Results

Replaying a real recording of two deliberate G1 falls through the tilt check, 176,426 samples at 756 Hz:

| | |
| -- | -- |
| stops issued | 1 |
| first trip | t=118.36s at 45.10 deg |
| before the trip | 86,691 samples, max tilt 44.99 deg, none tripped |
| after | 89,735 samples including a recovery to 0.43 deg, still latched |

For the joint limit: a settled standing G1 peaks at 0.77 rad/s and the whole fall reached 1.77 rad/s, against G1 URDF velocity limits of 20 to 37 rad/s.

`dimos/control/`, `dimos/robot/unitree/g1/`, `dimos/spec/`: 298 passed. The 3 failures are `cartesian_ik` failing to import GitPython, identical on a clean `main` worktree. `test_all_blueprints`: 20 failed / 78 passed, byte identical to `main`.

### Simulated in MuJoCo

Both conditions driven by real physics through the shipped check functions. MuJoCo 3.10.0, the `g1_29dof` model with visual meshes replaced by boxes, since the STLs live in LFS. **Simulated, not hardware.**

Fall, free base tumbling under gravity. Trips at t=0.330s at 45.2 deg, tilt reaches 137.8 deg, one stop issued and then latched.

fall in sim

Flailing joint, left knee kicked to 30 rad/s against a 20 rad/s limit. Trips on the first tick past the limit.

flailing joint in sim

### Not done

- No blueprint sets `max_joint_speed_rad_s` yet, so flailing detection is not armed in tree.
- Not run on hardware.
- In a real sim blueprint the backend is `MujocoSimModule`, not the connection, so the fall check is not wired there. The clip above drives the check directly.


AaryanAgrawal:
Taking this down to rework it. The branch stays; reviewing the design properly first.

PR #3693, opened 2026-08-26
Yimo/feat/graphviz rpc

poorwym:
## Problem

Blueprint topology was difficult to inspect from the CLI. The existing Graphviz renderer only showed stream connections and did not visualize RPC dependencies.

## Solution

- Added `dimos graph` to generate a Blueprint topology SVG.
- Added RPC Spec/function relationships using dashed edges.
- Added a dark SVG background for standalone viewing. This needs more discussions. I'm not sure whether should add a background instead of transparent background.
- Kept static rendering isolated from runtime transports.
- Updated Rerun parsing to support long RPC labels.

## How to Test

```bash
dimos graph unitree-go2-agentic --output unitree-go2-agentic.svg
```

```bash
uv run pytest dimos/cli/commands/test_graph.py -q
```

image


leshy:
looks great can we make rpc render an optional boolean then ready to merge

leshy (approved):

PR #3676, opened 2026-08-25
chore(teleop): rename Quest stack to WebXR

TomCC7:
## Contribution path

This is a breaking rename of the existing device-specific Quest teleoperation
adapter before PICO support is developed separately.

## Problem

The browser teleoperation stack was named after Meta Quest even though it uses
the WebXR API and supports both immersive AR and immersive VR sessions.

## Changes

- Rename the Quest package, classes, imports, and blueprints to WebXR.
- Use spatial teleoperation as the device-neutral capability description.
- Remove remaining Quest/VR abstraction names from the browser UI, comments,
  tests, and documentation while retaining real device names in support notes.
- Remove stale body-tracking documentation; this PR adds no tracking API or
  behavior.
- Restore the domain glossary and update only its teleoperation terminology.

There are intentionally no compatibility aliases for the removed Quest names.

## Verification

- `uv run pytest -q dimos/teleop/webxr/test_module.py dimos/teleop/webxr/test_blueprints.py dimos/control/tasks/teleop_ik_task/test_teleop_ik_task.py dimos/control/test_coordinator_routing.py dimos/teleop/hosted/test_arm_command.py dimos/robot/test_all_blueprints.py dimos/robot/test_all_blueprints_generation.py dimos/robot/test_get_all_blueprints.py`
  - 215 passed, 3 skipped, 40 deselected
- `uv run ruff check`
- `uv run mypy dimos/`
- `node --check dimos/teleop/webxr/web/static/teleop.js`

## AI assistance

OpenAI Codex with GPT-5 performed the refactor cleanup, documentation updates,
and automated verification under user direction.

## Checklist

- [ ] I have read and approved the [CLA](https://github.com/dimensionalOS/dimos/blob/main/CLA.md).

PR #3695, opened 2026-08-26
feat(teleop): add PICO WebXR body tracking

TomCC7:
## Contribution path

- Stacked on #3676

## Problem

The renamed WebXR teleop stack has no way to request PICO body-tracking permission, decode the browser's joint poses, or inspect tracking health on hardware.

## Solution

- Add a device-neutral `BodyTrackingSnapshot` output and strict JSON decoder.
- Keep body tracking off by default, with explicit `optional` and `required` WebXR session modes.
- Capture `XRFrame.body` poses in `bounded-floor` space when available and fall back to `local-floor`.
- Add a PICO API-test blueprint and live monitor without coupling body data to controller or viewer poses.
- Accept PICO's six-button WebXR controller packets, where the platform-reserved menu button is omitted.

## How to Test

Run the hardware-facing API test:

```bash
dimos run demo-pico-body-tracking
```

Automated checks run locally:

```bash
CI=1 uv run pytest dimos/teleop/webxr dimos/robot/test_all_blueprints_generation.py dimos/robot/test_all_blueprints.py -q
uv run ruff check dimos/teleop/webxr dimos/robot/all_blueprints.py
uv run ruff format --check dimos/teleop/webxr dimos/robot/all_blueprints.py
uv run --group lint mypy --disable-error-code=misc dimos/teleop/webxr
node --check dimos/teleop/webxr/web/static/teleop.js
node --check dimos/teleop/webxr/web/static/webxr_body.mjs
```

Result: 150 passed, 3 skipped, 40 deselected.

Physical PICO validation remains to be completed with the command above.

## AI assistance

OpenAI Codex with GPT-5 implemented the change, resolved the parent rebase conflicts, and ran the checks. The author should review and understand the final diff before merge.

## Checklist


dimos/teleop/webxr/blueprints.py
```diff
@@ -163,3 +166,14 @@
...
 )
+
+
+# PICO 4 Ultra WebXR API test: require body tracking and report every usable joint.
+demo_pico_body_tracking = autoconnect(
+    WebXRTeleopModule.blueprint(body_tracking_mode="required"),
+    BodyTrackingMonitor.blueprint(),
+).transports(
+    {
+        ("body_tracking", BodyTrackingSnapshot): pLCMTransport("/teleop/body_tracking"),
```
TomCC7: why specify special transport?

dimos/teleop/webxr/body_tracking.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class _WirePose(BaseModel):
+    model_config = ConfigDict(extra="forbid")
+
+    position: tuple[_FiniteFloat, _FiniteFloat, _FiniteFloat]
+    orientation: tuple[_FiniteFloat, _FiniteFloat, _FiniteFloat, _FiniteFloat]
+
+
+class _WireBodyTrackingSnapshot(BaseModel):
```
TomCC7: why a private type for decode? why not just transmit public type?

dimos/teleop/webxr/body_tracking_monitor.py
```diff
@@ -0,0 +1,111 @@
...
+from dimos.core.core import rpc
+from dimos.core.module import Module
+from dimos.core.stream import In
+from dimos.teleop.webxr.body_tracking import BodyTrackingSnapshot
+from dimos.utils.logging_config import setup_logger
+
+logger = setup_logger()
+
+REPORT_INTERVAL_S = 5.0
+INSPECTED_BODY_JOINTS = (
```
TomCC7: why we need this we should transmit all possible joint

dimos/teleop/webxr/controller_types.py
```diff
@@ -72,15 +72,15 @@ class WebXRControllerState:
...
+        Expected buttons: [trigger, grip, touchpad, thumbstick, X/A, Y/B, optional menu]
         Raises:
             ValueError: If Joy message doesn't have expected WebXR controller format.
         """
         buttons = joy.buttons or []
         axes = joy.axes or []
 
-        if len(buttons) < cls.EXPECTED_BUTTONS:
-            raise ValueError(f"Expected {cls.EXPECTED_BUTTONS} buttons, got {len(buttons)}")
+        if len(buttons) < cls.MIN_BUTTONS:
```
TomCC7: would say just removing this this is too defensive....

dimos/teleop/webxr/module.py
```diff
@@ -175,20 +185,63 @@ async def websocket_endpoint(ws: WebSocket) -> None:
...
-                    else:
-                        logger.warning(f"Unknown message fingerprint: {fingerprint.hex()}")
+                    message = await ws.receive()
+                    if message["type"] == "websocket.disconnect":
+                        logger.info("WebXR client disconnected")
+                        break
+                    data = message.get("bytes")
+                    text = message.get("text")
+                    if data is not None:
+                        self._dispatch_binary_message(data)
```
TomCC7, line 196: why are these required while previous impl dont

dimos/teleop/webxr/blueprints.py
TomCC7: Removed in 07f3eb87e0. Autoconnect selects the normal pickled transport for BodyTrackingSnapshot, so this blueprint no longer pins a transport or topic.

dimos/teleop/webxr/controller_types.py
TomCC7: Removed the minimum button-count check. PICO's six-button layout is accepted, and the menu button remains optional when index 6 is absent. The four-axis check remains because those four values are directly indexed.

dimos/teleop/webxr/body_tracking_monitor.py
TomCC7: Agreed. The browser already sends every resolved joint. I removed the monitor shortlist, so the health report now includes every joint in the snapshot, including vendor-specific joints.

dimos/teleop/webxr/body_tracking.py
TomCC7: Agreed. BodyJointPose and BodyTrackingSnapshot are now the public Pydantic models, and the WebSocket handler validates JSON directly into the published BodyTrackingSnapshot. The private wire types and conversion layer are gone.

dimos/teleop/webxr/module.py
TomCC7: The previous implementation used receive_bytes, which only accepts binary controller frames. Body snapshots are JSON text frames on the same WebSocket, so generic receive must distinguish disconnect, bytes, and text ASGI events. The explicit branches preserve the existing LCM controller path while adding body snapshots.

TomCC7:
Addressed the Greptile P1 in 82d7bc6e5a. WebXRControllerState now requires the six button slots it indexes while keeping the seventh menu button optional. A regression test sends a real five-button LCM Joy packet through WebXRTeleopModule and verifies the packet is rejected without escaping the callback, stale controller state is cleared, and zero velocity is published. Local verification: 54 WebXR tests passed; Ruff and mypy passed.

TomCC7:
@greptileai please re-review the latest commit 82d7bc6e5a. The prior P1 about short controller packets has been fixed and regression-tested.

dimos/teleop/webxr/web/static/teleop.js
```diff
@@ -540,14 +558,27 @@ function processTracking(frame) {
...
+
+    if (webXRClientConfig.body_tracking_mode !== 'off') {
+        const joints = captureBody(frame, xrBodyRefSpace);
+        const shouldSend = joints !== null || webXRClientConfig.body_tracking_mode === 'required';
+        if (shouldSend && ws && ws.readyState === WebSocket.OPEN) {
+            ws.send(JSON.stringify({
+                type: 'body_tracking_snapshot',
+                capture_time_s: (performance.timeOrigin + time) / 1000,
+                frame_id: xrBodyRefSpaceType,
+                joints,
```
ruthwikdasyam, line 575: should we stop sending when `no joints` or is it better to still send with `joints=null 0r {}` or something

dimos/teleop/webxr/body_tracking_monitor.py
```diff
@@ -0,0 +1,104 @@
...
+        self._resolved_joint_ever_seen = False
+
+    @rpc
+    def start(self) -> None:
+        super().start()
+        self.register_disposable(Disposable(self.body_tracking.subscribe(self._on_body_tracking)))
+
+    def _on_body_tracking(self, snapshot: BodyTrackingSnapshot) -> None:
+        self._snapshots_since_report += 1
+        if snapshot.joints and not self._resolved_joint_ever_seen:
```
ruthwikdasyam: If the snapshots are not received mid-run, should we display any warnings?

dimos/teleop/webxr/web/static/teleop.js
TomCC7: Implemented in 804e364137. Every sampled frame now sends a body snapshot whenever body tracking is enabled. `joints: null` means the body source is unavailable, `{}` means the source is present but resolved no joints, and a populated map means tracking. This preserves both liveness and state.

— Codex (AI agent)

dimos/teleop/webxr/body_tracking_monitor.py
TomCC7: I kept this out of `BodyTrackingMonitor`. Enabled clients now send explicit heartbeats even when the body source is unavailable, so the monitor reports `null` or `{}` as unhealthy body states. If snapshots stop entirely, that is WebSocket or session liveness; the WebXR module already logs disconnects and errors and resets controller state safely.

— Codex (AI agent)

TomCC7:
@greptileai please re-review the latest commit 23e047ec33. The branch now includes current main, explicit body-tracking heartbeats, and the SHM race fix.

— Codex (AI agent)

dimos/teleop/webxr/body_tracking_monitor.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class BodyTrackingMonitor(Module):
+    """Report live PICO body-tracking availability, rate, and joint poses."""
+
+    body_tracking: In[BodyTrackingSnapshot]
+
+    def __init__(self, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self._report_started_at = monotonic()
```
ruthwikdasyam: reporting window starts here in `__init__`, so if the monitor is created before an operator connects, the first heartbeat after a delayed start reports a rate based on startup time rather than tracking activity. Could we reset `_window_started_at` and `_snapshot_count` in `start()`? or is it fine eitherway?

TomCC7: Addressed in `d3c4068359`. I anchored the reporting window to the first received body snapshot rather than only resetting it in `start()`. That also excludes any delay between module startup and the operator connecting. `start()` now clears the pending window/count, and the first snapshot establishes the baseline without emitting an immediate low-rate report. Added a regression test for the delayed-first-snapshot case.

— Codex (AI agent)

mustafab0 (dismissed):

mustafab0 (approved):

PR #3701, opened 2026-08-26
fix(manip): sync the pick/place gripper on feedback and verify the grasp

mustafab0:
## The defect

`dimos/manipulation/pick_and_place_module.py` synchronized the gripper with wall-clock sleeps, and nothing anywhere verified a grasp:

| line | sleep | what it was waiting for |
|---|---|---|
| `:543` | `time.sleep(0.5)` | pre-grasp open |
| `:561` | `time.sleep(1.5)` | `# Wait for gripper to close` |
| `:652` | `time.sleep(1.0)` | release at place |

Open-loop timing means a slow close gets truncated and a fast one wastes time. Worse, a missed grasp sailed straight on to the lift.

## The mechanism

Gripper readback is **measured, not an echo of the command**

Each sleep is now a deadline poll that settles the readback, matching the house style of `PlanExecutionManager.wait` (`dimos/manipulation/execution_manager.py:212`) 

## Tests

- `test_grasp_verification.py` (13) — the poll helper against a mocked readback with an injected clock: settle, stall band, timeout, never-travelled, late readback, band edges, config validation.


dimos/manipulation/grasp_verification.py
```diff
@@ -0,0 +1,176 @@
...
+
+    settled: bool
+    position: float | None
+    moved: bool
+    elapsed: float
+
+
+def _interruptible_wait(seconds: float) -> None:
+    """Sleep the way PlanExecutionManager.wait does."""
+    threading.Event().wait(seconds)
```
paul-nechifor: This is quite odd. Why create an event but never set it? Couldn't all of this just be `time.sleep`? If PlanExecutionManager sleeps the same way it's doing it wrong.

paul-nechifor (dismissed):

dimos/manipulation/grasp_verification.py
mustafab0: I removed this.

The PlanExecutionManager is doing the same. Will address to fix this. 

As discussed in discord. I will refactor grasp verification into a `ControlTask`, which will run in a dedicated tick loop and would be managed properly.

paul-nechifor (approved):

PR #3708, opened 2026-08-26
ray_tracing: salvage tf-wait and darwin build fixes

jeff-hykin:
- make `tf_wait_timeout_s` configurable (for slow-moving robots with slow odom)
- MacOS nix build fix


aclauer (commented):
Ah yeah this should def be configurable.

Bu also this config is for how long we wait for a transform that fits the tolerance criteria before giving up, so if the robot publishes slowly, don't you want to change the tf_match_tolerance?

paul-nechifor (approved):

PR #3711, opened 2026-08-27
fix(tf): add forward_tolerance to the TFLookup protocol

mustafab0:
MultiTBuffer.get and StreamTF.get both take forward_tolerance already, the protocol just never declared it, so anything holding a TFLookup had to narrow to a concrete buffer to wait for a transform that has not arrived yet.

Dropped the warning throttle that was in here before, per review — warn=False already covers the case where misses are expected, and forward_tolerance covers the case where the transform is merely late.

### DIFF
--- a/dimos/protocol/tf/tf.py
+++ b/dimos/protocol/tf/tf.py
@@ -48,6 +48,8 @@ class TFLookup(Protocol):
         child_frame: str,
         time_point: float | None = None,
         time_tolerance: float | None = None,
+        *,
+        forward_tolerance: float = 0.0,
     ) -> Transform | None: ...
 
 


jeff-hykin (commented):
could probably be more simple but LGTM

leshy (changes requested):
please no AI descriptions anymore

>   Throttle the missing-transform warning to 1 Hz per frame pair.

Typically a system shouldn't operate in a way in which warnings are normal and then throttled for that reason, if you are getting regular warnings you should perhaps expand the forward time tolerance?

alternatively if missing transforms are a normal situation for your system - you can just add supress_warnings key instead of the throttling mechanism

mustafab0:
Closing this PR, as the forward tolerance can be accessed from the Tbuffer and does not need to be part of the Protocol

PR #3712, opened 2026-08-27
feat(mapping): add a voxel clear mask so a sensor can erase its own ghost

mustafab0:
## The physics

Ray tracing clears a voxel by passing a ray *through* it to a farther return. That works when the sensor sees past the thing it is mapping.

A wrist camera does not. Its own arm sits between the lens and the space behind it, so **no ray ever reaches through the arm** and `record_miss` never fires in that volume. Meanwhile the arm's own returns land in the map as ordinary hits. The result: the arm deposits voxels of itself, and there is no mechanism that can ever erase them. Over a work session the robot slowly walls itself in with a ghost of where it has been, and a planner reading `global_map` refuses to go there.

Ray-trace clearing cannot fix this — the geometry forbids it. The information has to come from somewhere else: a publisher that knows, from URDF plus TF, that a given volume is the robot and therefore free.

## What this adds

A `voxel_clear_mask: Input<PointCloud2>` port. The cloud is world-frame metric points meaning *"these are definitely free"*; the node deletes the voxels covering them.

- **Frame-gated** against `world_frame`. The points are metric positions, so the mask must already be in the world frame 

## Tests

`cargo test --workspace --all-features --locked`: 60 passed. `cargo fmt --check` and `cargo clippy -D warnings` clean..

Files: 6.

### DIFF
--- a/dimos/mapping/ray_tracing/module.py
+++ b/dimos/mapping/ray_tracing/module.py
@@ -80,6 +80,10 @@ class RayTracingVoxelMap(NativeModule, mapping.GlobalPointcloud):
     config: RayTracingVoxelMapConfig
 
     lidar: In[PointCloud2]
+    # World-frame points a sensor knows to be empty. Their voxels are deleted
+    # outright, reaching space ray tracing cannot clear: a wrist camera's own
+    # arm occludes the volume behind it, so no ray ever misses through it.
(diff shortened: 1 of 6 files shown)


aclauer (commented):
Does setting the `shadow_depth` config not fix this? It defaults to 0.1

aclauer (approved):

PR #3713, opened 2026-08-27
feat(perception): add configurable OSR scan backends

ruthwikdasyam:
Closes DIM-1363

## Summary
- add opt-in OWLv2 detection and EdgeTAM segmentation backends to OSR
- add the request-driven aligned RGB-D scan_scene API and expose it through ObjectSceneRegistrationSpec
- return pending ObjectDB entries from a scan so a newly observed object is immediately addressable by stable ID and point cloud
- retain the existing YOLOE/YOLO defaults and permanent-object stream behavior

## Validation
- focused OSR temporal tests: 10 passed
- mypy on the changed OSR files: passed
- Ruff format and check on changed files: passed

## Scope
- excludes SigLIP/Dan temporal memory, PnP integration, GraspGenX changes, filtering, and visualization.
- public method descriptions intentionally deferred for API review.

## Checklist

- [x] I have read and approved the [CLA](https://github.com/dimensionalOS/dimos/blob/main/CLA.md).

PR #3715, opened 2026-08-27
refactor(manipulation): compose canonical pick and place

ruthwikdasyam:
## Summary
- Replace the legacy pick-and-place subclass with a capability-composed module that coordinates OSR, grasp generation, and manipulation execution.
- Add stable-ID scan, explicit grasp selection, verified pick, and explicit place workflow skills.
- Compose the canonical workflow into real and simulated xArm perception blueprints.

## Stack
- Depends on #3713 (`ruthwik/feat/osr-owlvit-edgetam`).
- This is the third PR in the heuristic grasp -> OSR -> canonical pick-and-place stack.

## Validation
- `TMPDIR=/dev/shm uv run --no-sync pytest dimos/manipulation/test_pick_and_place_unit.py dimos/perception/experimental/test_object_scene_registration_temporal.py dimos/robot/test_all_blueprints_generation.py -q`
- Ruff format/check and mypy on all changed files.
- Imported real and simulated xArm perception blueprints.

## Deliberately Deferred
- Public method-description polish, candidate filtering/reranking, visualization, obstacle suppression, and table/box features.


dimos/manipulation/pick_and_place_module.py
```diff
@@ -12,818 +12,305 @@
...
 
-# Beyond this XY distance, the occlusion inset is increased so the grasp
-# targets closer to the true center rather than the front surface.
-_FAR_OCCLUSION_XY_THRESHOLD = 0.8
 
-# Objects taller than this are grasped in the upper third to avoid
-# plunging deep and colliding with the object body.
-_TALL_OBJECT_MIN_HEIGHT = 0.06
+class PickAndPlaceModuleConfig(ModuleConfig):
+    planning_frame: str = "world"
```
mustafab0:
```suggestion
    planning_frame: str = "base_link"
```

dimos/manipulation/pick_and_place_module.py
```diff
@@ -12,818 +12,305 @@
...
+    pregrasp_offset: float = Field(default=0.10, gt=0.0)
+    yaw_policy: Literal["generated", "preserve_current"] = "generated"
```
mustafab0, line 49: Should be a heuristic grasp config

mustafab0 (dismissed):

dimos/manipulation/pick_and_place_module.py
ruthwikdasyam: For xArm, OSR outputs poses in the world frame. Although `base_link` and `world` are at the same place for manipulation, the frame IDs still need to match. I changed the generic PnPModule default to `base_link`, while the xArm blueprints explicitly configure it as world to match OSR

dimos/manipulation/pick_and_place_module.py
ruthwikdasyam: if this comment is reg line 50. GraspVerificationConfig is to check grasp close/open status, and independent of grasp model

dimos/manipulation/pick_and_place_module.py
```diff
@@ -12,818 +12,305 @@
...
+                position=candidate.pose.position,
+                orientation=candidate.pose.orientation,
+            ),
+            group,
+        )
+        pregrasp = self._offset_pose(grasp, self.config.pregrasp_offset)
+        if failure := self._open_gripper(group, "pre-grasp open"):
+            return failure
+        if failure := self._move(pregrasp, group):
+            return failure
```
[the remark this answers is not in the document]
ruthwikdasyam: yeah, will add this properly in next pass. a grasp filter method - to check planning/ik, then try execution. not scope of this PR

dimos/manipulation/pick_and_place_module.py
mustafab0: I see. Can leave this here for now. (comment is for line 49 `yaw_policy`

But since yaw is also part of the grasp pose, I think it is better to have that planned/computed in the grasp gen itself. 

The idea here being that graspgenX would not need a yaw_policy and this is only required for heuristic grasps so should be part of that module itself. Can abstract it out later. I am okay with this for now

mustafab0 (dismissed):

mustafab0 (approved):

PR #3422, opened 2026-08-10
Add offline object inventory and localize perception pipeline DIM1343

bogwi:
## What features we implement in this PR

From the issue feature list, this branch covers A, D, E, and F.

* (A) Look at an image and identify all objects: `inventory()` discovers every in-scope support on the table without a text prompt, then names each instance.
* (D) Text prompt to 2D masks: `localize()` runs SigLIP retrieval, OmDet boxes, and EdgeTAM `segment()` so a query string becomes masks on ranked frames.
* (E) Prompt to point cloud / latest 3D pose: `localize()` lifts those masks through depth, clusters them, verifies across views, and returns the newest verified cloud and position. Offline over a recording, no live robot.
* (F) Object deduplication: `inventory()` merges same-frame fragments and associates tracks with geometry only (Hungarian assignment plus fragment merge). Labels are applied after identity is fixed.

Not in this branch: (B) exact dimensions, (C) 360-degree clouds, (G) re-ID.

## **Localize**
use localize() to query an object or a list of objects over a recorded window

### e2e example

```python
from dimos.memory2.store.sqlite import SqliteStore
from dimos.models.embedding.siglip import SigLIPModel
from dimos.models.segmentation.edge_tam import EdgeTAMImageSegmenter
from dimos.perception.detection.detectors.omdet import OmDetDetector
from dimos.perception.memory.localize import embed_index, localize
from dimos.utils.data import get_data

store = SqliteStore(
    path=get_data(
        "xarm6_worldbelief_realsense_d435i_stationery_calibrated/"
        "xarm6_worldbelief_20260729_203624_161992.db"
    )
)
lo, _ = store.streams.color_image.get_time_range()

# The models and the frame index belong to you: load once, query many times.
siglip = SigLIPModel()
detector = OmDetDetector()
segmenter = EdgeTAMImageSegmenter()
index = embed_index(store, siglip, lo + 53, lo + 129)

# Single query: one Localization or None.
hit = localize(store, "book", index=index, siglip=siglip, detector=detector, segmenter=segmenter)
if hit is not None:
    print(f"book: {hit.position_world_xyz} score={hit.semantic_score:.2f}")

# List query: one shared detection pass, one result per label, in order.
queries = ["pen", "red marker"]
hits = localize(store, queries, index=index, siglip=siglip, detector=detector, segmenter=segmenter)
for query, hit in zip(queries, hits):
    if hit is None:
        print(f"{query}: not found")
    else:
[...]

PR #3723, opened 2026-08-27
danvi/dim1486/generalize dan perception stack

bogwi:
# Generealize perception stack

This PR builds, improves, and adds new functionality based on the foundation of these previous PRs:
https://github.com/dimensionalOS/dimos/pull/3496
https://github.com/dimensionalOS/dimos/pull/3422

# How to test and what to expect
refference machine: 
Lenovo Legion laptop, AMD Ruyzen AI 7, 32GB RAM, NVIDIA RTX 5070 Laptop edition (8 GB)

## Caller - operator/agent decide the localization parameters at runtime

**every localize call can have own tuning parameters supplied**

At the moment we have ROOM policy, and default table policy, think xArm gripper.

This accounts for different environment and rig. We can control a detection score to take only the best candidates, and an acceptance floor, you can control segmentation, etc. Check `localize.py`

## xArm

This uses the deault location for xArm dataset, already on main, xarm6_worldbelief_20260729_203624_161992.db.
You do not need to specify the --dataset <location_to_recording_bag_.db>

`time uv run python -m dimos.perception.memory.tool_localize "roll of black tape" "book" "pen" "red marker" "yellow sticky notes" pc_xArm_all_multi.rrd --from 427 --duration 76 --multi && uv run dimos-viewer pc_xArm_all_multi.rrd`

```
# cold start, weights ~ 11s reff.machine
real	0m34.002s 
user	1m4.301s
sys	    0m5.702s
```

xArm_projections

## G1

You will need a g1 recording for this, check this PR, it's on main already, https://github.com/dimensionalOS/dimos/pull/3527

`time uv run python -m dimos.perception.memory.tool_localize "paper box" "plant in a pot" "shoes" "wooden crate" "office chair" "bag" g1_sf_120_180_localize.rrd --dataset data/g1_sf_office.db --from 120 --duration 60 --multi && uv run dimos-viewer g1_sf_120_180_localize.rrd`

```
# cold start, weights ~ 11s reff.machine
real	1m11.971s
user	4m44.586s
sys	    0m11.864s
```

pr1486_g1_sf

## Go2 
Run on the default go2_short.db

IMPORTANT, since this recording has no CameraInfo topic, we can supply specs as hardcoded or a sidecar. Sidecar was chosen. Redownload this asset, it's small.

From the `dimos` repo root. `--exclude=""` is required because `.lfsconfig` has `fetchexclude = data/.lfs/*`. `get_data("go2_short.db")` will not unpack the sidecar if `data/go2_short.db` is already there.

```
git lfs pull --include="data/.lfs/go2_short.db.tar.gz" --exclude="" && tar -xzf data/.lfs/go2_short.db.tar.gz -C data
```

That writes `data/go2_short.db` and `data/go2_short.db.rig.json`.

Now the command itself:

`time uv run python -m dimos.perception.memory.tool_localize "table" "white robot" "potted plants" "wooden doors" "coke bottle" "fanta bottle" "sprite bottle" "lays chips"  "pillow" "green plant" "plant pot" go2_short_00_60.rrd --dataset data/go2_short.db --from 0 --duration 60 --multi && uv run dimos-viewer go2_short_00_60.rrd`

```
# cold start, weights ~ 11s reff.machine
real    1m4.001s
user    1m45.383s
sys     0m24.279s
```
pr1486_go2_short

## A live blueprint addition showcasing the localize api on the live continuous stream.

The examples streams over go2_short.db recording that comes with dimos. Queries can be executed in real-time from dimos shell or mcp calls.

Check dedicated readme for guide how to run it, `dimos/dimos/perception/memory/demo_blueprints/README.md`

localize_live_feed_v2

## Complete e2e example

A new blueprint demo based on the xArm 10 min recording. This is the example you want to have to write your own modules.  See the main and supporting files.

You would start the stream in terminal 1 simulating the robot stream. Then you start consuming the stream in terminal 2 like you would do in real life. And then you start `dimos shell` in terminal 3.  Consult the README.md

`dimos/dimos/perception/memory/demo_blueprints/xarm_localize_live.py`

## Important
0. Performance is affected:
- by a cold run, weights need to be loaded, around ~11 seconds on my machine
- saving to *.rrd format
1. tool_inventory.py has been not touched by this commit, will come later.
Use localize() API for testing and building
2. Re-id and spatial memory are underway too
3. Improving objects 2D into 3D is also underway
4. supporting mid360 lidar is also undeerway
3. Any questions - contact me directly, @bogwi

### DIFF
--- a/data/.lfs/go2_short.db.tar.gz
+++ b/data/.lfs/go2_short.db.tar.gz
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:8a19846a0adf5755815fd039492c0255e0bc282e9df75a06648d7585cae8d2d2
-size 83971952
+oid sha256:37949ca60f5048ea675188cb2d9d8b5bf1ea70da3cd045a2e91a30069eccdbc2
+size 83972455
--- a/dimos/memory/stream.py
+++ b/dimos/memory/stream.py
@@ -568,8 +568,15 @@ class Stream(CompositeResource, Generic[T, O]):
         return n
 
     def drain_thread(self) -> DisposableBase:
-        """Drain this stream on the dimos thread pool; returns a disposable."""
-        return self.subscribe(lambda _: None)
+        """Drain this stream on the dimos thread pool; returns a disposable.
+
+        A drain has no consumer to surface an iteration error, so it logs -
+        a dead pipeline must not die silently.
+        """
+        return self.subscribe(
+            lambda _: None,
+            on_error=lambda e: logger.error("drain_thread() pipeline died: %s", e, exc_info=e),
+        )
 
     def observable(self) -> reactivex.Observable[O]:
         """Convert this stream to an RxPY Observable.
--- a/dimos/memory/utils/sqlite.py
+++ b/dimos/memory/utils/sqlite.py
@@ -27,6 +27,9 @@ def open_sqlite_connection(path: str | Path) -> sqlite3.Connection:
     conn = sqlite3.connect(path, check_same_thread=False)
     conn.execute("PRAGMA journal_mode=WAL")
     conn.execute("PRAGMA synchronous=NORMAL")
+    # WAL has one writer at a time; concurrent writer threads (a recorder
+    # feed plus a live embed save) can starve past the 5s default and die.
+    conn.execute("PRAGMA busy_timeout=30000")
     conn.enable_load_extension(True)
     sqlite_vec.load(conn)
     conn.enable_load_extension(False)
--- a/dimos/perception/detection/detectors/owlv2.py
+++ b/dimos/perception/detection/detectors/owlv2.py
@@ -30,6 +30,8 @@ from dimos.perception.detection.type.detection2d.imageDetections2D import ImageD
 
 class Owlv2Config(HuggingFaceModelConfig):
     model_name: str = "google/owlv2-base-patch16-ensemble"
+    # float16 runs the forward under autocast at roughly half the latency;
+    # scores jitter by a few thousandths, so threshold-edge boxes may flip.
     dtype: torch.dtype = torch.float32
 
 
@@ -62,6 +64,13 @@ class Owlv2Detector(HuggingFaceModel):
 
         return Owlv2Processor.from_pretrained(self.config.model_name)
 
+    def _autocast(self) -> torch.autocast:
+        return torch.autocast(
+            device_type="cuda",
+            dtype=self.config.dtype,
+            enabled=self.config.dtype is not torch.float32 and "cuda" in str(self.config.device),
+        )
+
     def query_detections(
         self,
         image: Image,
@@ -74,38 +83,54 @@ class Owlv2Detector(HuggingFaceModel):
         ``confidence`` is the calibrated per-box score. ``class_id`` indexes
         into ``queries``.
         """
-        pil = PILImage.fromarray(image.to_rgb().data)
-        with torch.inference_mode():
-            inputs = self._processor(text=[queries], images=pil, return_tensors="pt").to(
-                self.config.device
-            )
+        return self.query_detections_batch([image], queries, threshold)[0]
+
+    def query_detections_batch(
+        self,
+        images: list[Image],
+        queries: list[str],
+        threshold: float = 0.1,
+    ) -> list[ImageDetections2D]:
+        """``query_detections`` over several images in one forward pass.
+
+        Per-call preprocessing, text encoding and kernel launches amortize
+        across the batch, which is what makes many-frame sweeps affordable;
+        results are per-image, in input order.
+        """
+        pils = [PILImage.fromarray(image.to_rgb().data) for image in images]
+        with torch.inference_mode(), self._autocast():
+            inputs = self._processor(
+                text=[queries] * len(pils), images=pils, return_tensors="pt"
+            ).to(self.config.device)
             outputs = self._model(**inputs)
             results = self._processor.post_process_grounded_object_detection(
                 outputs=outputs,
-                target_sizes=torch.tensor([(pil.height, pil.width)]),
+                target_sizes=torch.tensor([(pil.height, pil.width) for pil in pils]),
                 threshold=threshold,
-            )[0]
-
-        detections: list[Detection2DBBox] = []
-        w, h = float(pil.width), float(pil.height)
-        for box, score, label in zip(
-            results["boxes"], results["scores"], results["labels"], strict=False
-        ):
-            x1, y1, x2, y2 = (float(v) for v in box)
-            bbox = (max(0.0, x1), max(0.0, y1), min(w, x2), min(h, y2))
-            det = Detection2DBBox(
-                bbox=bbox,
-                track_id=-1,
-                class_id=int(label),
-                confidence=float(score),
-                name=queries[int(label)],
-                ts=image.ts,
-                image=image,
             )
-            if det.is_valid():
-                detections.append(det)
 
-        return ImageDetections2D(image=image, detections=detections)
+        batch: list[ImageDetections2D] = []
+        for image, pil, result in zip(images, pils, results, strict=True):
+            detections: list[Detection2DBBox] = []
+            w, h = float(pil.width), float(pil.height)
+            for box, score, label in zip(
+                result["boxes"], result["scores"], result["labels"], strict=False
+            ):
+                x1, y1, x2, y2 = (float(v) for v in box)
+                bbox = (max(0.0, x1), max(0.0, y1), min(w, x2), min(h, y2))
+                det = Detection2DBBox(
+                    bbox=bbox,
+                    track_id=-1,
+                    class_id=int(label),
+                    confidence=float(score),
+                    name=queries[int(label)],
+                    ts=image.ts,
+                    image=image,
+                )
+                if det.is_valid():
+                    detections.append(det)
+            batch.append(ImageDetections2D(image=image, detections=detections))
+        return batch
 
     def query_score_rows(
         self,
@@ -123,7 +148,7 @@ class Owlv2Detector(HuggingFaceModel):
         their score rows.
         """
         pil = PILImage.fromarray(image.to_rgb().data)
-        with torch.inference_mode():
+        with torch.inference_mode(), self._autocast():
             inputs = self._processor(text=[queries], images=pil, return_tensors="pt").to(
                 self.config.device
             )
--- a/dimos/perception/detection/detectors/owlv2.py
+++ b/dimos/perception/detection/detectors/owlv2.py
@@ -16,6 +16,7 @@
 
 from __future__ import annotations
 
+from collections import OrderedDict
 from functools import cached_property
 
 import numpy as np
@@ -26,6 +27,9 @@ from dimos.models.base import HuggingFaceModel, HuggingFaceModelConfig
 from dimos.msgs.sensor_msgs.Image import Image
 from dimos.perception.detection.type.detection2d.bbox import Detection2DBBox
 from dimos.perception.detection.type.detection2d.imageDetections2D import ImageDetections2D
+
+# ~3.8 MB of GPU memory per cached frame
+_FEATURE_CACHE_MAX = 128
 
 
 class Owlv2Config(HuggingFaceModelConfig):
@@ -46,6 +50,22 @@ class Owlv2Detector(HuggingFaceModel):
     """
 
     config: Owlv2Config
+
+    def __init__(self, **kwargs: object) -> None:
+        super().__init__(**kwargs)
+        # image-tower forwards run so far; instrumentation reads deltas
+        self.forwards = 0
+        # (frame ts, label, floor) to that label's floor-filtered
+        # (boxes_kx4, scores_k); filled and read by localize
+        self.score_cache: OrderedDict[tuple[float, str, float], tuple[np.ndarray, np.ndarray]] = (
+            OrderedDict()
+        )
+        # frame ts to text-independent tower outputs; any query set scores
+        # against a cached frame without rerunning the tower
+        self._features: OrderedDict[
+            float, tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]
+        ] = OrderedDict()
+        self._text_embeds: dict[str, torch.Tensor] = {}
 
     @cached_property
     def _model(self):  # type: ignore[no-untyped-def]
@@ -97,6 +117,7 @@ class Owlv2Detector(HuggingFaceModel):
         across the batch, which is what makes many-frame sweeps affordable;
         results are per-image, in input order.
         """
+        self.forwards += len(images)
         pils = [PILImage.fromarray(image.to_rgb().data) for image in images]
         with torch.inference_mode(), self._autocast():
             inputs = self._processor(
@@ -147,6 +168,7 @@ class Owlv2Detector(HuggingFaceModel):
         queries and refuse. Returns pixel ``(x1, y1, x2, y2)`` boxes and
         their score rows.
         """
+        self.forwards += 1
         pil = PILImage.fromarray(image.to_rgb().data)
         with torch.inference_mode(), self._autocast():
             inputs = self._processor(text=[queries], images=pil, return_tensors="pt").to(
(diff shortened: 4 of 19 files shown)


dimos/perception/memory/localize.py
```diff
@@ -320,277 +218,325 @@ def localize(
...
-    cache = _DetectionCache(detector, segmenter, queries, policy.candidate_floor)
+
+    peaks_per_label: list[Stream[Any, Any]] = []
+    for q in queries:
+        query_embedding = siglip.embed_text(q)
+        label_peaks: Stream[Any, Any] = (
+            index.search(query_embedding)
+            .order_by("ts")
+            .transform(peaks(key=_similarity, distance=1.0))
+            .materialize()
```
leshy: why are we ordering by ts? we can order by peak height (this is a default ordering) - why are you materializing the full stream, this is a very expensive operation, that loads all data in memory. we should be able to have a seeker head that loads observation by observation until peak becomes irrelevant. very low mem footprint

dimos/perception/memory/localize.py
```diff
@@ -320,277 +218,325 @@ def localize(
...
+        )
+        logger.info(
+            f"localize {q!r}: {label_peaks.count()} semantic peaks of {index.count()} embedded"
+        )
+        peaks_per_label.append(label_peaks)
```
leshy: you escaped memory system and streams immediately and are building lists. why isn't this a mem stream? you should always operate on streams of observations, not on raw lists containing all data, we are angling towards parsing an hour, a day of data etc

dimos/perception/memory/localize.py
```diff
@@ -320,277 +218,325 @@ def localize(
...
+    expanded: set[float] = set()
+    for label_peaks in peaks_per_label:
+        for peak in label_peaks:
+            frames.setdefault(peak.ts, peak)
+            if peak.ts in expanded:
+                continue
+            expanded.add(peak.ts)
+            nearby: Stream[Any, Any] = index.near(
+                peak.pose_stamped, radius=policy.verify_radius_m
+            ).transform(QualityWindow(lambda img: img.sharpness, window=0.5))
```
leshy: embeddings themselves should already be filtered by the qualitywindow, all these matches will be clean, you don't need to filter downstream

dimos/perception/memory/localize.py
```diff
@@ -320,277 +218,325 @@ def localize(
...
+            for obs in nearby:
+                frames.setdefault(obs.ts, obs)
+    ordered = sorted(frames.values(), key=lambda obs: obs.ts)
```
leshy: ordering is a db operation

dimos/perception/memory/localize.py
```diff
@@ -320,277 +218,325 @@ def localize(
...
+    logger.info(f"detection: {len(ordered)} candidate frames for {len(queries)} labels")
+
+    if ordered:
+        from dimos.perception.memory.support_plane import fit_support_plane
```
leshy, line 346: why are not these following lines a nice mem2 transform?

leshy (changes requested):
basically escaped mem2 immediately, first comments first view to push back on this, might have more stuff

dimos/perception/detection/type/detection3d/pointcloud.py
```diff
@@ -47,6 +71,28 @@ class Detection3DPC(Detection3D):
...
+
+        The cloud is the union of both; identity metadata follows the
+        higher-confidence sighting and time context follows the later one,
+        matching latest-pose semantics.
+        """
+        later = self if self.ts >= other.ts else other
+        stronger = self if self.confidence >= other.confidence else other
+        return Detection3DPC(
+            image=later.image,
+            bbox=later.bbox,
```
leshy, line 85: you should expand bbox, not take latest one, your result is a union now

dimos/perception/detection/identity.py
```diff
@@ -0,0 +1,146 @@
...
+            frame_id=union.pointcloud.frame_id,
+            timestamp=union.pointcloud.ts,
+        )
+        weights[id(union)] = wsum
+        return union
+
+    return merge
+
+
+class Identity(Transformer[Any, Detection3DPC]):
```
leshy: let's type transforms properly since important for understanding, dont use `Any`

dimos/perception/memory/rig.py
```diff
@@ -0,0 +1,1066 @@
...
+                times.append(float(t))
+                headings.append(np.arctan2(axis[1], axis[0]))
+    if len(times) < 3:
+        return np.empty(0), np.empty(0)
+    stamps = np.array(times)
+    yaw = np.unwrap(np.array(headings))
+    return (stamps[1:] + stamps[:-1]) / 2, np.diff(yaw) / np.diff(stamps)
+
+
+def estimate_color_delay(rig: Rig) -> float:
```
leshy, line 217: what is this for?

dimos/perception/memory/rig.py
```diff
@@ -0,0 +1,1066 @@
...
+    tf: TFLookup | None = None
+    base_to_optical: Transform | None = None
+    poses: Any = None  # stream carrying world base poses (e.g. odom)
+    depth: Any = None  # aligned depth stream, lifted via from_depth
+    cloud: Any = None  # pointcloud stream, lifted via from_2d after registration
+    tf_tolerance: float = TF_TOLERANCE
+    cloud_accum_s: float = CLOUD_ACCUM_S
+    speed_max: float = SPEED_MAX
+    color_delay: float = 0.0  # s - color timestamps lag the pose stream by this
+    scene_gate: bool = True
```
leshy: why do you have a scene gate if you also have speed gates?

dimos/perception/memory/rig.py
```diff
@@ -0,0 +1,1066 @@
...
+            tf=tf,
+            base_to_optical=base_to_optical,
+            poses=poses,
+            depth=depth,
+            cloud=cloud,
+        )
+
+        span = _camera_span(rig)
+        rig.mobile = span > MOBILE_SPAN_M
+        if rig.mobile:
```
leshy, line 555: I don't think we should have these meta-settings, below is already exposed to the config

dimos/perception/detection/identity.py
```diff
@@ -0,0 +1,146 @@
...
+def spatial(radius: float = 0.1) -> Callable[[Detection3DPC, Detection3DPC], bool]:
+    """Same object when the cloud centers sit within *radius* meters."""
+
+    def is_same(a: Detection3DPC, b: Detection3DPC) -> bool:
+        return float((a.center - b.center).magnitude()) <= radius
+
+    return is_same
+
+
+def fused(voxel: float) -> Callable[[Detection3DPC, Detection3DPC], Detection3DPC]:
```
leshy: you defined detection3dpc.__add__ as fusing pointclouds, this seems like another implementation of the same thing, why have two? detection3d already has center of mass calculation as .center, voxel downsampling is a function of pc2

dimos/perception/memory/rig.py
```diff
@@ -0,0 +1,1066 @@
...
+                    kept.append(part[fresh])
+                covered[
+                    lo[0] - base[0] : hi[0] + 1 - base[0], lo[1] - base[1] : hi[1] + 1 - base[1]
+                ] = True
+            points = np.vstack(kept)
+        merged = PointCloud2.from_numpy(points, frame_id=self.world_frame, timestamp=ts)
+        self._cloud_memo = (ts, merged, quantum)
+        return merged
+
+    def _shell_table(
```
leshy, line 759: this seems to be a secondary floor detection? what does this do?

dimos/perception/memory/rig.py
```diff
@@ -0,0 +1,1066 @@
...
+            poses=poses,
+            depth=depth,
+            cloud=cloud,
+        )
+
+        span = _camera_span(rig)
+        rig.mobile = span > MOBILE_SPAN_M
+        if rig.mobile:
+            rig.speed_max = WALK_SPEED_MAX
+            rig.scene_gate = False
```
leshy: why doesn't mobile platform have a scene gate and the other way around?

dimos/perception/memory/rig.py
```diff
@@ -0,0 +1,1066 @@
...
+        snapping to a recorded sample misplaces the projection; interpolation
+        follows the motion. Outside the bracketed span the nearest sample in
+        the tolerance window stands.
+        """
+        from dimos.msgs.geometry_msgs.PoseStamped import PoseStamped
+
+        candidates = list(self.poses.at(ts, self.tf_tolerance))
+        if not candidates:
+            return None
+        earlier = [o for o in candidates if o.ts <= ts]
```
leshy, line 615: this whole function is pretty complex, there must be an easier way to find a pose at some time? if we need interpolation we should generalize somewhere

dimos/perception/memory/rig.py
```diff
@@ -0,0 +1,1066 @@
...
+            f"tf={tf_name!r} world={world_frame!r} span={span:.1f}m mobile={rig.mobile} "
+            f"color_delay={rig.color_delay * 1000:.0f}ms"
+        )
+        return rig
+
+    # pose
+
+    def world_to_optical(self, ts: float) -> Transform | None:
+        ts -= self.color_delay  # color stamps lag; the capture instant is earlier
+        if self.tf is not None:
```
leshy, line 595: should probably just insist on valid tf to not deal with special cases

leshy (dismissed):
didn't review all, we'll need to meet and reorganize

leshy:
most of these function descriptions seem ai generated and very difficult to understand

dimos/perception/detection/type/detection3d/pointcloud.py
bogwi: Do not agree, expanind is not a fix because two boxes are not in the same image. __add__` keeps image, bbox and transform from one and the same sighting so the union stays a valid 2D detection: the box is where the object appears in that image.

This related to a later comment. The real fix is to drop the  sum

dimos/perception/detection/identity.py
bogwi: Ok, of course

dimos/perception/memory/rig.py
bogwi: Not a defect at all.

Go2's odom and video are both stamped with time.time() when they arrive on the host (dimos/robot/unitree/connection.py:311 and 328). The video arrives later than the odom for the same instant so a color frame's stamp is later than the moment it was captured.

Measured on go2_short.db it finds 142 ms.

Nothing is hardcoded but computed. For 12 windows of 2.5 s spread over the recording it computes the yaw rate seen by the camera -> from optical flow between consecutive frames, and the yaw rate of the robot -> from the poses.

dimos/perception/memory/rig.py
bogwi: Again, not a defect because they reject different frames
camera_still -> rejects a frame if the camera was moving *when* it was taken because a transform snapped to the nearest sample is then off

scene_still -> rejects a frame if the image changed while the camera stood still

dimos/perception/memory/rig.py
bogwi: Yes, agreed

dimos/perception/memory/rig.py
bogwi: yes, it is.
it stores for every XY column of the frame's merged cloud, the lowest point's height above the fitted plane and that is a local floor per column. 

Each snapshot of the rolling map carries its own odometry error, this function slightly improves that.

dimos/perception/memory/rig.py
bogwi: because localize on the live robot never checks scene_gate - it selects frames by sharpness and semantic peaks and a moving object's sightings form a group that never reaches two camera positions. It is not needed. Only for stationary rig.

dimos/perception/memory/rig.py
bogwi: agreed, nxt PR

dimos/perception/memory/rig.py
bogwi: agreed, nxt PR

dimos/core/global_config.py
```diff
@@ -104,11 +104,15 @@ class GlobalConfig(BaseSettings):
     robot_rotation_diameter: float = 0.6
     nerf_speed: float = 1.0
     mcp_port: int = 9990
+    # Seconds an MCP client waits for a tool to answer. A skill that thinks
+    # for longer than this is cut off at the client, not the server, so the
+    # caller owns the number.
+    mcp_timeout: int = 30
```
spomichter, line 112: nice

dimos/agents/mcp/mcp_adapter.py
```diff
@@ -41,8 +41,6 @@
 
 logger = setup_logger()
 
-DEFAULT_TIMEOUT = 30
```
spomichter: yes was bad thanks. no globals no in globalconfig

dimos/memory/utils/sqlite.py
```diff
@@ -27,6 +27,9 @@ def open_sqlite_connection(path: str | Path) -> sqlite3.Connection:
     conn = sqlite3.connect(path, check_same_thread=False)
     conn.execute("PRAGMA journal_mode=WAL")
     conn.execute("PRAGMA synchronous=NORMAL")
+    # WAL has one writer at a time; concurrent writer threads (a recorder
```
spomichter: prefer limited comments. i put in my claude.md to super limit ai comments

dimos/memory/stream.py
```diff
@@ -568,8 +568,15 @@ def drain(self) -> int:
...
-        """Drain this stream on the dimos thread pool; returns a disposable."""
-        return self.subscribe(lambda _: None)
+        """Drain this stream on the dimos thread pool; returns a disposable.
+
+        A drain has no consumer to surface an iteration error, so it logs -
+        a dead pipeline must not die silently.
+        """
+        return self.subscribe(
+            lambda _: None,
+            on_error=lambda e: logger.error("drain_thread() pipeline died: %s", e, exc_info=e),
```
spomichter, line 578: what i personally do if i have refactor code unrealted to the PR -- i PR it seperatley to keep main feature PR clean

dimos/perception/detection/detectors/owlv2.py
```diff
@@ -27,9 +28,14 @@
 from dimos.perception.detection.type.detection2d.bbox import Detection2DBBox
 from dimos.perception.detection.type.detection2d.imageDetections2D import ImageDetections2D
 
+# ~3.8 MB of GPU memory per cached frame
+_FEATURE_CACHE_MAX = 128
```
spomichter, line 32: maybe this should be global constant in constants.py but idk enough about this model

dimos/perception/detection/detectors/owlv2.py
```diff
@@ -27,9 +28,14 @@
...
+
 
 class Owlv2Config(HuggingFaceModelConfig):
     model_name: str = "google/owlv2-base-patch16-ensemble"
+    # float16 runs the forward under autocast at roughly half the latency;
```
spomichter: delete comment

data/.lfs/go2_short.db.tar.gz
TomCC7, line 1: why is this changed did you change recording?

bogwi: no, I did not. Ivan previously requested tool_localize to support go2_short recording, but It has no camera_info. So you either bake it or supply as a recording sidecar. I did the latter.

But I've changed now and refactored. recording is untouched. tool_localize (inherited from Ivan's original kick-starter pr) will be / or might be removed in future prs; still thinking about this tool

mustafab0:
I don't have any major complaints here.

The object localization pipeline looks good to me. I have a few complaints about naming. 

`dimos/perception/memory` folder could be better named to `dimos/perception/localize` other than that it looks alright.

One note is that, right now the detection and localization pipeline are designed with OwlV2 detector. But we should be able to swap in and out any detector as we want it.

Since we are exploring and building, it is not a big concern. But we must have swappable pieces for detectors, segmentors, and encoders like clip or siglip. That is the core underlying dimensional design principle.

Happy to merge this as is. If we can put this under experimental and then discuss on the modular architecture that would be awesome

cc: @bogwi @leshy

dimos/perception/memory/localize.py
```diff
@@ -320,277 +218,325 @@ def localize(
...
+    for obs in ordered:
+        active = [j for j in range(len(queries)) if obs.ts not in ingested[j]]
+        if not active:
+            continue
+        rows_per_label: list[tuple[int, tuple[Any, Any]]] = []
+        for j in active:
+            key = (obs.ts, queries[j], floor)
+            cache.move_to_end(key)
+            rows_per_label.append((j, cache[key]))
+            ingested[j].add(obs.ts)
```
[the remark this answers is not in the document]
mustafab0: @bogwi please resolve. this problem still exists in `Groups.ingested`

bogwi:
@mustafab0 @leshy 

1) I agree, `dimos/perception/localize` is the right place to keep the code neat, will do asap
2) I have already added the support for OmDet detector, tested it and they are largely swap-able
3) The multi-detector and component modular interface is planned. It will land nicely in the next PRs. The initial idea is to have sematic search -- detection -- segmentation modules swap-able as I add more modules api to the codebase

bogwi:
moved to https://github.com/dimensionalOS/dimos/pull/4271

PR #3727, opened 2026-08-27
feat(web): acknowledged shared tx publishing

paul-nechifor:
- Adds shared input channels so multiple browser clients can send messages to a robot. Developers declare these channels in the cockpit configuration. The bridge converts incoming JSON into typed messages for modules.
- Adds `session.publish()` to the web SDK. Each call receives confirmation once the bridge has published the message onto the stream.
- Validates incoming messages and limits their size, rate, and number of pending requests.
- Adds usage documentation and a chat input example.


leshy (dismissed):

mustafab0:
> Adds shared input channels so multiple browser clients can send messages to a robot. Developers declare these channels in the cockpit configuration. The bridge converts incoming JSON into typed messages for modules.

Isn't this a safety risk? There should always be one clear  owner of hardware such that the robot should not have conflicting instructions.

paul-nechifor:
> > Adds shared input channels so multiple browser clients can send messages to a robot. Developers declare these channels in the cockpit configuration. The bridge converts incoming JSON into typed messages for modules.
> 
> Isn't this a safety risk? There should always be one clear owner of hardware such that the robot should not have conflicting instructions.

This is already the case. The relay enforces one client controlling the robot at a time (it has to arm by sending teleop_start and wait to receive teleop_started).

This PR lets you declare other channels onto which different clients can send. This is not intended for teleoping, but for other reasons where two clients would need to send messages simultaneously. This is the case for the chat for example. If you have two clients viewing the same robot, they should both be able to interact with chat. You shouldn't have to "arm" chat to type something in chat.

Note that "shared" is not the default. You have to opt into it and someone should be careful to not use it to send cmd_vel commands.

mustafab0 (approved):

leshy (approved):

PR #3731, opened 2026-08-27
Trajectory controller and navigation structure

leshy:
there are a few things here that can be split into separate PRs but nav team is the only consumer so keeping here unless we decide otherwise

### restructure of navigation/ much cleaner folder, new specs

```python

class GlobalPlanner(Protocol):
    tf: In[TFMessage]
    goal: In[PointStamped]

    path: Out[Path]

class LocalPlanner(Protocol):
    tf: IO[TFMessage]
    planner_path: In[Path]
    local_map: In[PointCloud2]

    path: Out[Path]


class TrajectoryFollower(Protocol):
    tf: IO[TFMessage]
    path: In[Path]

    nav_cmd_vel: Out[Twist]

```

### bridge hacky feature
we need to be able to explicitly subscribe to just SOME topics not all, so that they are not sent over network or decoded by py

### module introspection fixes
module.io() hides default internal RPCs

### lidar module config
removed global env reading, switched to local module config

### distributed architecture for nav

we run `go2-zenoh-motion-pointlio` on go2 jetson and `dimos --robot-ip go22 run go2-viewer` on the laptop

this means zenoh (network robust) subs go over network, and rerun rendering and RPC happens locally on the laptop


leshy:
https://github.com/dimensionalOS/dimos/pull/4203

PR #3732, opened 2026-08-27
refactor(manipulation): replace the pick module with a composed one

mustafab0:
Main's PickAndPlaceModule was a ManipulationModule subclass that piled nine agent skills onto the primitive RPC module and reached into its privates to do it; this replaces it with a plain module that gets motion, reachability and the gripper through Specs. The primitives move down onto ManipulationModule, including the perception objects port and the obstacle-monitor RPCs that used to live on the subclass, so the planner keeps seeing detected objects instead of going collision-blind. Grasp geometry now comes from HeuristicGraspModule reading the object cloud rather than the occlusion-inset and tall-object constants it replaces, and #3701's grasp verification is wired through unchanged.

Verifying this on the xArm7 sim turned up two pre-existing bugs it also fixes. Re-opening jaws already resting at their measured 0.947 could only ever time out, since that is neither travel nor arrival at the commanded 1.0, so await_gripper_settle now takes an arrival_tolerance. And make_xarm7_sim_robot_config placed the arm at the origin while xarm7.xml bolts it to a 12cm pedestal, so the planner solved every pose 12cm below the arm it was driving and every grasp closed on air.

Deliberately dropped: look, get_scene_info, place_back, drop_on, pick_and_place, and clear_perception_obstacles, whose documented use was to delete the world when planning hit COLLISION_AT_START. The multi-candidate retry loop and far-reach pre-grasp shortening are dropped too; both want the contact primitive rather than a reimplementation here.


mustafab0:
Superseded by #3715, which landed the composed pick and place. The placement fix from this branch is now #3733; the remaining deltas (perception to obstacle-world wiring, and threading pre_grasp_offset from the robot model) will follow as a separate PR on top of main.

PR #3733, opened 2026-08-27
refactor(manip): place robot models with a base pose, not axis offsets

mustafab0:
The robot base pose generation was hacky and used scalars for x,y,z,rpy . A better approach is to set PoseStamped transform as a proper way of setting robot base pose

Verified on the sim: planning and MuJoCo TCP agree exactly, and a pick on the cup holds at readback 0.725, inside the band #3701 calibrated.

An earlier revision widened arrival in `await_gripper_settle` so a re-open of already-open jaws could not time out. Greptile was right that it regressed the release path, and it has been withdrawn.

### DIFF
--- a/dimos/robot/manipulators/xarm/test_model_config.py
+++ b/dimos/robot/manipulators/xarm/test_model_config.py
@@ -18,9 +18,12 @@ import pytest
 
 from dimos.manipulation.planning.spec.validation import validate_robot_model_config
 from dimos.robot.manipulators.xarm.config import (
+    XARM7_SIM_BASE_POSE,
+    XARM7_SIM_PATH,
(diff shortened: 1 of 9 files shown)


aclauer (approved):

dimos/robot/manipulators/xarm/test_model_config.py
```diff
@@ -80,3 +83,22 @@ def test_prefixed_xarm_model_asset_uses_coordinator_facing_names() -> None:
     model = validate_robot_model_config(config)
 
     assert [joint.name for joint in model.joints if joint.type != "fixed"] == config.joint_names
+
+
+@pytest.mark.self_hosted
```
paul-nechifor, line 88: Why is this self_hosted? Also it's not testing code, it's testing an XML file.

PR #3743, opened 2026-08-28
add a homepage for the docs site and update  the material theme.

swstica:
## Problem

Docs landing page is a copy `readme`. Too long for a landing page

## Solution

Added `index.md` and added some styling


leshy:
getting some errors when I serve?
```
INFO    -  Doc file 'index.md' contains an absolute link '/assets/readme/navigation.gif', it was left as is. Did you mean 'assets/readme/navigation.gif'?
INFO    -  Doc file 'index.md' contains an absolute link '/assets/readme/agentic_control.gif', it was left as is. Did you mean 'assets/readme/agentic_control.gif'?
INFO    -  Doc file 'index.md' contains an absolute link '/assets/readme/spatial_memory.gif', it was left as is. Did you mean 'assets/readme/spatial_memory.gif'?
INFO    -  Doc file 'index.md' contains an absolute link '/assets/readme/perception.png', it was left as is. Did you mean 'assets/readme/perception.png'?
INFO    -  Doc file 'installation/osx.md' contains a link '../usage/transports/index.md#zenoh-quickstart', but the doc 'usage/transports/index.md' does not contain an anchor
```

Right now when I click "platforms" it just takes me to a the go2 page which is proly a bug, if I go to quickstart middle of the document it just tells me to use go2 again

What robots does dimensional support? where do I see this? this is what most devs what to check first when they visit the docs, this PR removes this info.

Reintroducing it means we have 2 places to maintain the robot support table. so probably good idea to extract it to a third document? 

Also this creates 2 places to maintain this table of videos/capabilities, and it's already different (why not update README one?)

I think this split seems like unnecessary maintenance. we are adding two places with very similar information, why not reduce size of main README.md and use it also here if size is an issue. or kill README.md significantly, just link to docs if this is the main doc, but let's not mantain 70% similar large documents

leshy:
also a lot of description seems slightly off, probably should copy paste from the readme, or we need company wide description of dimos that we approve together

> dimOS is also agent native. Every skill a robot exposes is also an MCP tool. An LLM coding agent can discover those tools and call them directly.

doesn't define a skill what is a skill? is it special in any way that this is a MCP tool? also agent can call much more then mcp stuff, can call any rpc

I'm not sure if we should go so hard on ROS in our docs index. especially since a lot of what we mention as distinction between dimos and ros, is what ros does as well

> How dimOS solves it[¶](http://localhost:8000/#how-dimos-solves-it)
> Behaviour lives in modules, not in an SDK. A module does one job. It declares the streams it needs as typed fields, so > a module written for a quadruped runs on an arm once something publishes what it asked for.

same as ros

> Nothing waits on hardware. The same blueprint runs against a recorded session or in MuJoCo before it runs on a robot. The behaviour you debug is the behaviour that ships.

same as ros

PR #3760, opened 2026-08-29
feat: add headless xarm room simulation

mustafab0:
Run 
```dimos --viewer none run xarm-room-sim``` to test this blueprint.

### DIFF
--- a/dimos/perception/experimental/object_scene_registration.py
+++ b/dimos/perception/experimental/object_scene_registration.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import annotations
+
 import threading
 import time
 from typing import Any, Literal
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: dimos/perception/experimental/object_scene_registration.py [hunk 1]
```diff
@@ -150,15 +145,34 @@ def start(self) -> None:
...
+                    "EdgeTAM requires the optional dependencies from dimos[misc]"
+                ) from e
+        return None
+
     @rpc
     def stop(self) -> None:
         """Stop the module and clean up resources."""
 
         with self._processing_lock:
+            if self._segmenter and hasattr(self._segmenter, "stop"):
```
paul-nechifor: What we need is a proper interface for detectors. We have `YoloeVisualPromptDetector`, we have `EdgeTAMImageSegmenterCompatible`, but what we need is one that fits both and includes stuff like `stop` and `EdgeTAMImageSegmenter` which does not need it can leave it blank.

So there's no need to use `hasattr`.

docs/capabilities/manipulation/xarm-room-sim.md
```diff
@@ -0,0 +1,44 @@
...
+
+The canonical positions and geometry live in `data/xarm_grasp_sim/scene.xml`.
+The visual meshes were cooked from the `dimos_office` scene package and scaled
+for the xArm gripper; primitive geometry is used only for contact. Every target
+has a grasp axis below 7 cm, and the gray can is the designated pick smoke
+target. The room blueprint starts the arm at an elevated, collision-free
+top-down scan pose so all six visual meshes fit in the wrist-camera frame.
+
+From `dimos shell`, scan with the room prompt set exported by
+`dimos.robot.manipulators.xarm.blueprints.simulation.XARM_ROOM_PROMPTS`, then
```
paul-nechifor: Probably better to give a code example that can be copy pasted rather than mixing prose and code.

paul-nechifor (dismissed):

paul-nechifor (approved):

PR #3763, opened 2026-08-29
fix(navigation): signed momentum term so U-turns cost

metrox-eth:
### Scope

This PR makes the direction momentum score signed and clears the remembered direction after a goal timeout.

- With the default floor of `-1.0`, direction scores range from +1 straight ahead to −1 straight behind. Previously, the score was clamped at zero, so a frontier behind the robot received no additional penalty relative to a sideways frontier.
- The default momentum weight remains `0.05`. At identical state, changing the clamp to a signed score changes each frontier's total score by between −0.05 and 0. Ahead-facing candidates are unchanged; backward-facing candidates receive a bounded penalty. For otherwise equal candidates, the straight-ahead/straight-behind gap increases from 0.05 to 0.10. This bounds the immediate score change, not subsequent trajectory or coverage changes.
- `min_momentum_score=0.0` restores the old clamp. `momentum_weight` exposes the existing weight. The timeout reset remains active with either clamp setting.
- `_on_goal_timeout()` clears the remembered direction, making momentum neutral for the next ranking. A subsequent goal more than 0.1 m away establishes a new direction. A valid goal that simply takes longer than `goal_timeout` also loses its momentum bonus; the reset can therefore change selection even when the robot is making progress.
- The timeout log reports the configured duration instead of a hard-coded "30 seconds".

### Tests

`test_wavefront_frontier_scoring.py` constructs real explorers with a no-op RPC transport injected through `rpc_transport`, never calls `start()`, and calls `stop()` at teardown.

The tests cover the default configuration, signed score ordering and gap, clamp behavior, weight scaling, the direction lifecycle through `_update_exploration_direction()` and `_on_goal_timeout()`, and the timeout dispatch through `_await_goal_outcome()`, the method the exploration loop calls after publishing a goal, with a stubbed wait (timeout forgets the direction, a reached goal keeps it, the configured `goal_timeout` is what is waited on). The exploration thread itself is not run.

### Evidence and limits

The earlier M4.3 simulation results measured a different policy: an EMA of actual displacement retained across timeouts, plus multiplicative path-length and direction penalties applied to the existing score. Those results do not validate this implementation.

That policy is deferred to a separate proposal after a real-robot A/B. This PR makes no measured claim about reducing crossings or improving coverage.

### DIFF
--- a/dimos/navigation/frontier_exploration/test_wavefront_frontier_scoring.py
+++ b/dimos/navigation/frontier_exploration/test_wavefront_frontier_scoring.py
@@ -1,0 +1,73 @@
+"""Scoring tests for the direction momentum term of the wavefront selector.
+
+Three candidate frontiers are placed at the same distance from the robot, with
+the same cluster size, on an empty map with no explored goals, so every term of
+the score except direction momentum is identical between them: one straight
+ahead of the current exploration direction, one sideways, one straight behind.
+"""
+
+import numpy as np
+import pytest
+
+from dimos.msgs.geometry_msgs.Vector3 import Vector3
+from dimos.msgs.nav_msgs.OccupancyGrid import CostValues, OccupancyGrid
+from dimos.navigation.frontier_exploration.wavefront_frontier_goal_selector import (
+    WavefrontConfig,
+    WavefrontFrontierExplorer,
+)
+
+RESOLUTION = 0.05
+FRONTIER_SIZE = 40
+
+ROBOT = Vector3(5.0, 5.0, 0.0)
+AHEAD = Vector3(7.0, 5.0, 0.0)  # dot product +1 with the exploration direction
+SIDEWAYS = Vector3(5.0, 7.0, 0.0)  # dot product 0
+BEHIND = Vector3(3.0, 5.0, 0.0)  # dot product -1
+
+
+def open_costmap() -> OccupancyGrid:
+    """A 10 m x 10 m free map, no obstacle within any candidate's safe radius."""
+    grid = np.full((200, 200), CostValues.FREE, dtype=np.int8)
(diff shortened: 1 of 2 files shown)


metrox-eth:
Good catch, and confirmed in the code: `exploration_direction` is only ever set toward the chosen goal at selection time, and the timeout branch never clears it, so with the term now signed a stale heading toward an unreached goal would penalize frontiers behind the robot on the next ranking.

Pushed a fix: the timeout branch zeroes `exploration_direction`, so the selection right after a timeout carries no directional preference (the term already returns 0 for a zero direction); it re-establishes on the next chosen goal. Added a test for the neutrality, and the timeout log message now uses the configured value (it said 30 seconds while the default is 15).

metrox-eth:
Converting to draft: our sim evidence turned out to be strongly world-dependent (the same arms ping-pong on a raw accumulated map and barely at all on the dataset's own two-pass global map), so the numbers here need real-robot validation before this is worth review time. The mechanism itself (a U-turn currently costs nothing) and the timeout fixes stand. Full data: https://github.com/metrox-eth/vector-dimos/tree/main/benchmarks/pr2830

leshy (approved):

leshy:
Oh just approved this, saw the comment, cancelled, let me know when to check again :)

leshy (changes requested):
waiting for update

metrox-eth:
The PR is now reduced to what stands on its own. The momentum term is signed, so a U-turn now costs, with the weight unchanged at 0.05 (ahead/behind gap 0.05 → 0.10), and the direction of a goal that timed out is forgotten. The M4.3 sim numbers measured a different policy; they are out of this PR and would come back as a separate proposal.

I'll be able to run a real-robot A/B in a few weeks, so it's your call: merge this small change now, or wait for that.

PR #3769, opened 2026-08-30
feat(perception): re-identify tracklets a tracker lost

Moshiii:
## Contribution path

## Problem

A tracker associates across adjacent frames, so it breaks whenever continuity does:
an occlusion, a robot turning away, a detector miss. Every break makes a new object
where there was one. Nothing puts the pieces back.

## Solution

**Geometry vetoes, appearance decides.** Two chairs of one model are identical to a
similarity model, so a threshold high enough to merge one chair's fragments also
merges two different chairs — and a wrong merge is worse than fragmentation, because
fragments are honest while a merge invents an object. What separates identical things
is where and when they were, so the cheap constraints run first and the embedding
comparison only sees pairs that survive them. On a 7.8-minute office capture that
refused 673,000 pairs before a single embedding was compared.

Ground truth needs no labelling: cut a tracklet the tracker held continuously and the
pieces are one object; two tracklets overlapping in time are two objects.

**Pairwise: 100.0% precision, zero wrong merges.** The 99.0% recall that stood
beside it was measured before the reachability veto moved to the gap boundary
(below) and wants one rerun.

### The budget is spent on the gap, so it is measured across the gap

The speed allowance models how far a thing could travel while unseen, but the veto
compared whole-run centroids — charging a continuation for a walk that happened inside
its own tracklet, and forgiving a jump when two runs happened to average to the same
place. It now compares where the earlier run ended against where the later one resumed.
Precision is untouched: this benchmark's negatives are pairs overlapping in time,
refused before any distance is computed. Recall can move either way and has not been
remeasured. (Raised by Greptile.)

### What is measured and what is not

| | status |
|---|---|
| `find_merges` — pairs | measured, shipped defaults, table in the doc |
| `connected` — groups | **unmeasured**, opt-in |
| `OnlineAssociator` — ids | **unmeasured**, not for deployment |

The grouping figures this PR originally carried (38.5% closure, 23.9% online) are
**withdrawn**. They were taken against a version that applied the instant-overlap veto
only between the two tracklets of a pair, never to the group a link would create — and
this benchmark's negative class *is* "overlapping in time". So every wrong merge it
counted was a pair the module's own hardest veto would have refused, had the grouping
path asked it.

The veto now runs group-wide, and `co_occurring` binds every member of an id rather
than one track of it. The benchmark cannot referee that change: its negative class and
the new check are the same predicate, so rerunning it scores the check against itself.
`connected` stays opt-in — now because grouping is unmeasured, not because it was
measured and failed. A real number needs a negative class that does not reduce to
co-visibility: pairs of one model never in frame together, labelled by hand, few enough
to label in an afternoon. `docs/capabilities/perception/tracklet_reid.md` keeps the
withdrawn rows struck through and says the rest.

### What the geometry veto is worth

Upstream (`EmbeddingIDSystem`) vetoes only co-occurrence. Adding position and speed
refused **48 pairs upstream would have merged, all 48 correctly** — things that look
alike and were never seen together, like two chairs of one model at opposite ends of a
room. The benchmark cannot score this: its negatives are defined as overlapping in
time, which co-occurrence already catches, so the 48 fall outside the labelled set.

### Boundary

Embeddings arrive already computed by whatever model the caller picked, so the module
does not own that half — but it does own the boundary. Un-normalised vectors do not
lower a cosine, they inflate it (orthogonal tracklets scored 3.29 and merged at the
shipped threshold); a NaN row scores NaN and `nan < threshold` is False, so it merged
and was counted as nothing. Both are now refused, loudly, naming the tracklet.

numpy and the standard library, nothing else. No torch, no store, no message type, no
module system. Verified by import.

### Open for the reviewer

1. **`OnlineAssociator` overlaps `dimos/perception/detection/reid/embedding_id_system.py`**
   — same 0.63, same 500-cap, same min-10, same top-k mean, same co-occurrence
   negatives — and the doc measures the two as identical on pairs. The genuinely new
   part is the geometry veto, ~15 lines. Is the right shape to add that veto to
   `EmbeddingIDSystem` and drop the second associator? That is ~170 lines deleted.
2. **`connected` ships with no number behind it.** Keeping it as published evidence is
   one call; deleting it until there is a benchmark that can score it is another.

Placement was the third question here and is settled: the module now sits in
`dimos/perception/detection/reid/` beside `EmbeddingIDSystem`, and
`docs/capabilities/perception/index.md` links the page and says which of the two is
wired into a module. The packages above it carry no `__init__`, so the numpy-only
claim survives the move.

## How to Test

```bash
uv run pytest dimos/perception/detection/reid/test_tracklet_reid.py
```

33 tests, numpy and pytest only, no fixtures. Nine are regressions for the vetoes
above: a group never holds two tracklets seen at one instant; `connected` refuses
tracklets it was not given; an id does not absorb a track declared a different object;
movement inside a run is not charged to the gap; a jump across the gap is refused
though the runs average alike; a reused track id is not barred by the exclusion it
inherited; the track dropped is the least recently seen; un-normalised embeddings are
refused not rescaled; a `Tracklet` can go in a set. `ruff` and `mypy` clean.

There is no blueprint to run: the module is numpy in, dataclasses out, and has no
callers yet by design.

## AI assistance

Claude Code with Opus 5, used as an implementation tool.

The approach is mine. Geometry vetoes before appearance decides. A wrong merge is worse
than fragmentation, so this fragments rather than guesses. Ground truth for
re-identification can be built by cutting tracklets the tracker held continuously
instead of by labelling anything. A negative result about grouping is worth publishing
rather than tuning away. What the vetoes are, what `find_merges` returns, why
`connected` is opt-in, and the decision to withdraw the grouping numbers rather than
rerun a benchmark that cannot referee its own check — those are my calls, and they are
what this PR is actually proposing.

Claude implemented against that design, and in a later pass audited the result: it
found the group-level veto gap, the NaN and normalisation fail-opens, and the
least-recently-seen eviction bug, then wrote those fixes and six regression tests.
Every number in this description was re-run except the withdrawn ones, which is why
they are withdrawn rather than restated. I have reviewed the diff and understand the code.

## Checklist

### DIFF
--- a/dimos/perception/detection/reid/test_tracklet_reid.py
+++ b/dimos/perception/detection/reid/test_tracklet_reid.py
@@ -1,0 +1,456 @@
+# Copyright 2026 Dimensional Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+import numpy as np
+import pytest
+
+from dimos.perception.detection.reid.tracklet_reid import (
+    Merge,
+    OnlineAssociator,
+    OnlinePolicy,
+    ReidPolicy,
+    Tracklet,
+    connected,
+    find_merges,
+    group_similarity,
+)
+
+
+def views(direction: tuple[float, float], n: int = 10, jitter: float = 0.02) -> np.ndarray:
+    """``n`` unit vectors clustered around one direction in 2-D embedding space."""
+    rng = np.random.default_rng(0)
+    base = np.asarray(direction, float)
+    raw = base + rng.normal(0.0, jitter, size=(n, 2))
+    return raw / np.linalg.norm(raw, axis=1, keepdims=True)
+
+
+def tracklet(key, t0, t1, xy=(0.0, 0.0), look=(1.0, 0.0), n=10, label="chair") -> Tracklet:
+    return Tracklet(
+        key=key,
+        t_start=t0,
+        t_end=t1,
+        positions=np.array([[xy[0], xy[1], 0.0]] * n, float),
+        embeddings=views(look, n=n),
+        label=label,
+    )
+
+
+class TestGeometryVetoesAppearance:
+    """Two chairs of one model score near-perfect, correctly. Only where and
+    when they were separates them, so no score overrides those checks."""
+
+    def test_two_things_seen_at_once_never_merge(self):
+        a = tracklet("a", 0.0, 10.0, xy=(0.0, 0.0))
+        b = tracklet("b", 5.0, 15.0, xy=(0.2, 0.0))  # identical look, overlapping time
+
+        merges, report = find_merges([a, b])
+
+        assert merges == []
+        assert report.rejected == {"time_overlap": 1}
+
+    def test_identical_things_far_apart_never_merge(self):
+        a = tracklet("a", 0.0, 10.0, xy=(0.0, 0.0))
+        b = tracklet("b", 20.0, 30.0, xy=(50.0, 0.0))
+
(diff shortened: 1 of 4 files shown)


Moshiii:
Closing in favour of a fresh PR: this one accumulated five force-pushes while the review feedback was worked through, and the timeline is more noise than signal now. The branch and the single commit are unchanged — the replacement carries the same head, and records Greptile's two P1 findings and their fixes in the description so that review is not lost with this thread.

PR #3772, opened 2026-08-30
Feat/belief layer

Moshiii:
## Contribution path

- Linked issue or discussion: DIM-XXX 

## Problem

An agent on this stack can look at the current frame, move, and talk. It cannot
consult the past. Ask "where is the backpack" and there is nothing to ask — the
recording holds camera, lidar and odometry, and no layer that turns those into
"a thing, at a place, at a time".

The two failures that shape this PR both come from the same gap:

- **Wall-clock staleness.** Without an asking time, an answer about t=30 s gets
  built from what the robot saw at t=400 s. It reads exactly as well as a
  correct one.
- **An answer with nothing to judge it by.** A bare list gives the model no way
  to tell "I never looked there" from "I looked and it was gone", so it
  confabulates over both.

## Solution

A belief layer under `dimos/experimental/memory_belief/`: sightings are written
into the same store they were derived from, folded into entities, and queried
as of one moment.

**Write path** — `detect_recording` runs the detector over a recording's frames
and writes `belief_observation` back into that store; `build_views` folds those
into `belief_identity` / `belief_entity` / `belief_frame_annotation` in place.
One file goes in, the same file comes out fuller, so every stream stays on one
timeline and `dimos mem rerun` can play the result.

**Read path** — one `execute(store, query)` entry point onto the store's
existing lazy filter chain, not a second query engine. Filters narrow,
projections compute; JSON drives the same frozen-dataclass algebra the store
already had. One call instead of six, because a round trip costs milliseconds of
work wrapped in seconds of inference.

Key decisions:

- **`as_of` is required, never defaulted.** The asking time decides the answer,
  so nothing picks it for the caller. In the skill it is clamped to now rather
  than trusted: a model free to choose it can answer about 1:00 from what the
  robot saw at 7:00.
- **Every answer carries how much to trust it.** `quality` (rows, support,
  dispersion, coherence, deduplicated) plus a named `UnknownReason` —
  `NEVER_COVERED`, `OUT_OF_VOCABULARY`, `INCOHERENT`, `NO_CAPABILITY` — and the
  capability that would fix it. `SkillResult.retryable` is derived from that
  reason rather than set by hand, so it stays consistent across authors.
- **`visibility` is three-way.** `absent` is negative evidence; `occluded` and
  `out_of_view` are no evidence at all. Collapsing them turns "I did not see it"
  into "it is not there".
- **The vocabulary is recorded on every sighting**, so a question about a term
  the detector never ran with answers `OUT_OF_VOCABULARY` instead of a confident
  "no".
- **The folds live in `pipeline.py`, not in the scripts.** The same folds have to
  run offline and, later, on a robot that is still driving. Divergence there is
  invisible — nothing raises, the queries just stop agreeing — so `ViewParams`
  makes it deliberate.
- **JSON codec for derived streams.** A pickle is readable only by the exact
  classes that wrote it; the codec refuses payload types that accept unknown
  fields or carry no version, at open time rather than at decode time.
- **Tracing is optional and silent.** Missing Langfuse must never take a robot
  down for the sake of a dashboard, and spans are emitted only inside an
  `agent_trace` block, so 15 Hz perception does not drown the useful ones.

Also here because the layer needed them: `mem rerun --view module:function` so a
robot's view lives with the robot, and `SkillResult.evidence` as `(stream, id)`
pairs so an agent can re-read what a skill acted on.

Scoped to `experimental/`: the identity fold is still weak, and only the
`locate` projection ships — counting, timelines and distances were removed
rather than left half-served, because a projection that always returns nothing
reads as an answer.

## How to Test

```
python -m scripts.ask recording.db "where is the backpack" --at 61
```

Building that store from a raw recording, and watching it back:

```
python -m dimos.experimental.memory_belief.detect_recording recording.db --vocabulary lvis
python -m scripts.build_views recording.db
dimos mem rerun recording.db --view dimos.robot.unitree.go2.belief_rerun:go2_belief_rerun_config
```

Unit tests:

```
pytest dimos/experimental/memory_belief dimos/agents/test_skill_result.py dimos/utils/test_tracing.py dimos/memory/codecs/test_codecs.py
```

## AI assistance

Claude Code with Opus 5, used to implement. The architecture, the design
methodology, and every decision written up above are mine -- the model wrote code
against them. I have read every line and I am accountable for it.

## Checklist

### DIFF
--- a/dimos/agents/annotation.py
+++ b/dimos/agents/annotation.py
@@ -66,7 +66,51 @@ def _stamp_and_log(func_name: str, result: Any, elapsed_ms: float) -> Any:
         # Not a SkillResult — we can't verify the outcome, so don't claim "OK".
         code = "UNKNOWN"
     logger.info("SKILL %s result=%s duration_ms=%.1f", func_name, code, elapsed_ms)
+    _trace_skill(func_name, result, code, elapsed_ms)
     return result
+
+
+def _trace_skill(func_name: str, result: Any, code: str, elapsed_ms: float) -> None:
+    """Record one skill outcome, if an agent trace is open.
+
+    The single point every skill passes through, which makes it the only place
+    the refusal taxonomy can be collected without touching ninety-odd call
+    sites. Silent outside a trace, so background perception at 15 Hz costs
+    nothing and does not drown the traces that matter.
+    """
+    from dimos.utils.tracing import enabled, span
+
+    if not enabled():
+        return
+    from dimos.agents.skill_result import SkillResult
+
+    attributes: dict[str, Any] = {
+        "skill_name": func_name,
+        "outcome": code,
+        "duration_ms": round(elapsed_ms, 1),
+    }
+    if isinstance(result, SkillResult):
+        reason = result.unknown_reason
+        if reason is not None:
+            # The whole point of the belief layer: why it could not answer, and
+            # what would fix it. Without these a refusal is indistinguishable
+            # from a crash on a dashboard.
+            attributes["belief_unknown_reason"] = reason.name
+            attributes["belief_suggested_capability"] = result.suggested_capability
+            attributes["belief_retryable"] = result.retryable
+            attributes["belief_needs_revisit"] = reason.needs_revisit
+        attributes["belief_evidence_count"] = len(result.evidence)
+        # `quality` is what the query layer actually attaches: row count,
+        # support, dispersion and whether the rows were deduplicated. A loop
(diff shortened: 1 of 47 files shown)


Moshiii:
Closing and reopening to get a review of the current commit. The two findings raised here are fixed in it: `_diagnose` now applies the same asking-time boundary as the query it explains, and counts with `Stream.count()` instead of materialising the stream once per clause.

PR #3774, opened 2026-08-31
feat(agent evals): promote agents/environments to be configurable, support multi-step agent evals

hvent90:
## What precipitated this

- We need to save the agent trajectory, final response, and raw request/responses to the LLM providers for visualizations, QA of eval validity, and scoring
- we need to _easily_ compare different agents/harnesses/available tools over mcp for an individual case. it shouldn't require context switching of the developer (such as modifying a case). It should just be a param in the cli
- co-locate `generate.py` to its consumers before it grows exponentially in the root of the `dimos.evals` namespace

## What this introduces

Two new first-class citizens in `dimos.evals`:
- agents:
    - simple question/answer that does `agent_encode` if a dataset with appropriate streams are present
    - the same agent that `humancli` uses, and it can interrop with cases acting on datasets and sims
    - the pi agent (immediately allows us to test skills and bash without having to iterate on our existing harness)
- environments:
    - dataset
    - sim
    - image

Defining and running evals becomes more declarative: a single `EvalCase` type that takes in an environment. You determine the agent when invoking from the cli: `dimos eval run <suite> --agent <agent> --set <agent_config, like max_steps>`

This also features our first multi-step agent eval for discovering every room in the dimsim apartment scene using pointcloud data and saying how many rooms are in there. More of these cases will be in a following PR along with _many_ question/answer pointcloud evals from [this pr](https://github.com/hvent90/dimos/pull/13), as well as an updated `agent_encode`.

Trajectory format adheres to [Agent Trajectory Interchange Format](https://github.com/harbor-framework/harbor/blob/main/rfcs/0001-trajectory-format.md).

## How to Run

```sh
# evaluate a single-turn LLM that uses agent_encode
uv run dimos evals run dimos.evals.suites.examples \
    --agent dimos.evals.agents.question_answer
    
# evaluate against the same suite a single-turn LLM
# that purely guesses 
uv run dimos evals run dimos.evals.suites.examples \
    --agent dimos.evals.agents.blind

# evaluate our production agent 
DIMOS_TRANSPORT=lcm uv run dimos evals run dimos.evals.suites.dimsim_house \
    --agent dimos.evals.agents.mcp_client_adapter \
    --set 'modules=["unitree-go2-agentic"]'

# view the results
# open ~/.local/state/dimos/evals
```

## Examples

I want to hillclimb on pointcloud's `agent_encode` method where I am _just_ testing interpretability of the data (meaning I'm handing the data to it instead of testing access/query patterns):
```
dimos run <suite> --agent dimos.evals.agents.question_answer
```
I'm noticing the LLM is answering accurately! For example, it can detect a corridor versus open space as well as correctly identify discrete rooms about 50% of the time. But, I should test it against a baseline to check my assumptions - it should at least exceed the performance of a standard coding agent with bash:
```
dimos run <suite> --agent dimos.evals.agents.pi --set max_steps=10
```
Or maybe I write an opionated agent harness skill that gives 1,000 tokens of guidance on how to query memory and use numpy on pointcloud data:
```
dimos run <suite> --agent dimos.evals.agents.pi \
    --set  max_steps=10 \
    --set skills="[\"$PWD/.agents/skills/dimos-spatial/SKILL.md\"]"
```
Stepping forward in time, lets say this process was repeated across rgb streams, occupancy grid, and other data we've collected. The next rung on the ladder would be writing evals for the kind of task the pointcloud data was meant to solve - such as identifying rooms - but then testing it against each of our other optimized data paths.
```
dimos run <suite> --agent dimos.evals.agents.pi \
   --set system_prompt='<PI_SYSTEM_PROMPT plus pointcloud instructions>'

dimos run <suite> --agent dimos.evals.agents.pi \
   --set system_prompt='<PI_SYSTEM_PROMPT plus rgb instructions>'

dimos run <suite> --agent dimos.evals.agents.pi \
   --set system_prompt='<PI_SYSTEM_PROMPT plus occupancy grid instructions>'
```

## Possible next steps

Besides the immediate goal of making more impactful multi-step evals, these can be added to the backlog:

- eval viewer with visualizations suitable for grokking each environment
- rerunning an eval while having an easy UX for tweaking individual config settings
- allowing environments to be wrapped in docker/podman so it can run in parallel on sandboxes

#### AI assistance disclaimer

My associates `Fable 5`, `Opus 5`, and `GPT 5.6 Sol` wrote the vast majority of code found in this PR. I have spent extensive time combing through this but apologize for any oversights! Thanks for reading.


docs/usage/evals.md
```diff
@@ -1,31 +1,49 @@
...
-the robot's memory. Two kinds:
-
-- **Passive**: the world is a frozen memory recording. Deterministic, cheap,
-  repeatable. Run these constantly.
-- **Interactive**: a live robot or sim. Actions change the world, and scoring
-  samples the live memory store while the agent works.
-
-memory is the source of truth for everything an eval sees: context selectors
-return real `Stream`s, and interactive scoring reads a real `Store`.
+An eval case is an environment, an instruction, and a grader. Nothing else.
```
hvent90: I need to do a human-written pass on this file

dimos/memory/store/base.py
```diff
@@ -217,3 +218,14 @@ def stop(self) -> None:
         for stream in self._streams.values():
             stream.stop()
         super().stop()
+
+
+def copy_streams(streams: Iterable[Stream[Any]], into: Store) -> Store:
```
hvent90: This might be bad, looking over in next pass

dimos/evals/agents/mcp_client.py
```diff
@@ -0,0 +1,201 @@
...
+    pairs: list[tuple[Path, Path]] = []
+    for req in sorted(raw_dir.glob("*-request.json")):
+        resp = req.with_name(req.name.replace("-request.json", "-response.json"))
+        if resp.exists():
+            pairs.append((req, resp))
+    return pairs
+
+
+@dataclass
+class McpClientAgent:
```
spomichter: We already have an MCP client

dimos/evals/agents/pi.py
```diff
@@ -0,0 +1,327 @@
...
+from dimos.evals.agents.lib.chat import DEFAULT_MODEL
+from dimos.evals.agents.lib.proxy import RecordingProxy
+from dimos.evals.agents.lib.recorder import StepDraft
+from dimos.evals.types import EndedBy, Environment, RunningEnvironment, ToolCall, Trajectory
+
+if TYPE_CHECKING:
+    from dimos.memory.store.base import Store
+
+OPENAI_URL = "https://api.openai.com/v1"
+PI_TOOLS = ("read", "bash", "edit", "write")
```
spomichter: No global constants in files. put in consntants or global config

dimos/evals/agents/pi.py
```diff
@@ -0,0 +1,327 @@
...
+    "  store = SqliteStore(path=PATH, must_exist=True)\n"
+    "store.streams.<name> is a stream, .last().data its latest message; iterating "
+    "a stream yields observations with .ts and .data. Inspect with dir() and help()."
+)
+
+ROBOT_HELP = (
+    "The robot is live. Call one of its tools from bash as\n"
+    "  dimos mcp call <tool> --json-args '{\"arg\": value}'\n"
+    "Tools:\n"
+)
```
spomichter: make these system prompts composable / generic, not in the base pi.py if we want to use it for other use cases

spomichter (changes requested):

dimos/evals/environments/sim.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+def _no_setup(sim: DimSimClient) -> None:
+    return None
+
+
+AT_REST_M = 0.05
+AT_REST_S = 2.0
+SETTLE_POLL_S = 0.5
+
```
spomichter: no hidden constants

dimos/cli/test_vqa.py
```diff
@@ -188,14 +189,33 @@ def fail_load(dataset: Path) -> tuple[()]:
 
 
 def test_vqa_run_cli_runs_shared_evaluator(monkeypatch: pytest.MonkeyPatch, tmp_path: Path) -> None:
+    assets = tmp_path / "assets"
+    assets.mkdir()
+    (tmp_path / "cases.jsonl").write_text('{"id":"q","image":"assets/q.png"}\n')
+    (tmp_path / "labels.jsonl").write_text('{"id":"q","answer":"yes"}\n')
+    (assets / "q.png").write_bytes(b"image")
```
spomichter, line 196: should pytest be saving to filesystem / creating dirs i forget best practice

dimos/evals/agents/pi.py
hvent90: thanks!

dimos/evals/agents/mcp_client.py
hvent90: Yes, this is the eval agent that invokes that MCP client.

dimos/evals/environments/sim.py
hvent90: thank you!

dimos/evals/agents/pi.py
hvent90: moving to #3879

dimos/cli/test_vqa.py
hvent90: Using `tmp_path` is Pytest's [suggested way of doing so](https://docs.pytest.org/en/stable/how-to/tmp_path.html).

They are saved under `/tmp` ([docs](https://docs.pytest.org/en/stable/how-to/tmp_path.html#temporary-directory-location-and-retention)) and will be cleared on system reboot.

[Here is an example](https://github.com/dimensionalOS/dimos/blob/main/dimos/cli/cameracalibrate/test_cameracalibrate.py#L119-L121) of us doing so elsewhere in the repo.

dimos/evals/runner.py
```diff
@@ -81,359 +60,232 @@ class RunSummary:
...
+        if thread.is_alive():
+            logger.warning("agent timed out", case=case.id, timeout_s=case.timeout_s)
+            return Trajectory(
+                final_answer="",
+                steps=(),
+                model="",
+                duration_s=time.monotonic() - t0,
+                ended_by="timeout",
+                raw_dir=case_dir / "raw",
+            )
```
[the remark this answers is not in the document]
hvent90: good bot, fixed

dimos/memory/store/base.py
hvent90: it was bad. streams are now passed similar to how it was done in origin/main instead of doing an entire copy.

dimos/evals/agents/lib/chat.py
```diff
@@ -0,0 +1,78 @@
...
+
+DEFAULT_MODEL = "gpt-5.6-luna"
+
+EVAL_SYSTEM_PROMPT = "Answer the question using only the provided observations."
+
+Blocks = list[str | dict[str, Any]]  # HumanMessage content blocks
+
+
+@dataclass
+class ChatAgent:
```
spomichter: @hvent90 make sure this is the only agent in the repo. Ideally we use agentModule but im fine with one simplier agent client lets just use the same one and langchain or OpenAI client as the interface so it can support all standard chat models

dimos/evals/agents/lib/langchain_recorder.py
```diff
@@ -0,0 +1,134 @@
...
+        elif block.get("type") == "reasoning":
+            summary = block.get("summary")
+            if isinstance(summary, list):
+                parts.append(
+                    "".join(str(s.get("text", "")) for s in summary if isinstance(s, dict))
+                )
+    return "\n\n".join(p for p in parts if p)
+
+
+class LangChainRecorder(BaseCallbackHandler):
```
spomichter: Pls port this over soon to use our standard recording and send agents messages over transport. if too much is running in a single module worker than will overload and cause lag.

dimos/evals/agents/mcp_client.py
spomichter: If running via a blueprint you should just ahve this as an evalModule and listen to the topics / connect. We arleady ahve an mcp client also so if i see mcp_client.py i assume its also an MCP client.

dimos/evals/types.py
```diff
@@ -12,196 +12,213 @@
...
+    any particular case.
     """
 
-    id: str
-    inputs: str
-    skill: str = ""
-    skill_args: Mapping[str, object] = field(default_factory=dict)
-    tags: frozenset[str] = frozenset()
-    timeout_s: float = 60.0
+    modules: str
```
paul-nechifor: It would probably be better to use a blueprint name. You can construct a blueprint intended for testing only. Anything that starts with `demo_` is hidden by default.

But if you necessarily need to specify a list of blueprints, this should be `list[str]`, not `str`.

dimos/evals/agents/mcp_client.py
spomichter: i guess some architectural challenges here because autoconnect() only works with a module thats started at the beginning with teh rest of the app. make_transport() is the only convention to join an alrady running blueprint

spomichter: Can we run this eval agents with the bluepritn at the start? is there a real reason to run it after?

dimos/evals/agents/blind.py
```diff
@@ -0,0 +1,41 @@
...
+from dimos.evals.types import Environment, RunningEnvironment, Trajectory
+
+BLIND_BLOCK: dict[str, str] = {
+    "type": "text",
+    "text": "[observations withheld — answer anyway]",
+}
+
+
+@dataclass
+class Blind(ChatAgent):
```
paul-nechifor: How come you're using dataclasses for this?

I think it's better to use regular Python for base classes. Interfaces should be specified in `Protocol`s or `ABC`s.

If you need, you could use `dataclass`es for configuration. Or you can using Pydantic.

dimos/evals/agents/mcp_client.py
paul-nechifor: I think the name is not ideal, that's why it's confusing. It's not an HTTP MCP client.

But where is this actually used? I was trying to think of a better name but I don't see it used anywhere except tests. @hvent90

dimos/evals/environments/dataset.py
```diff
@@ -0,0 +1,110 @@
...
+        from dimos.memory.cli.dataset import open_dataset
+
+        store = open_dataset(self.name)  # raises: dataset unresolvable
+        try:
+            for sel in self.select:
+                sel(store)  # raises "No stream 'x'. Available: [...]" — no data read
+        finally:
+            store.stop()
+
+    def start(self, modules: str) -> RunningEnvironment:
```
spomichter: i assume not pass modules as strings? idk though

dimos/evals/environments/dataset.py
```diff
@@ -0,0 +1,110 @@
...
+
+from dimos.evals.environments.lib.launch import blueprint_modules, default_mcp_url
+from dimos.evals.types import Agent, RunningEnvironment, Select
+
+if TYPE_CHECKING:
+    from dimos.e2e_tests.dimos_cli_call import DimosCliCall
+    from dimos.memory.store.base import Store
+
+
+@dataclass
```
spomichter: no dataclass if running processes

dimos/agents/mcp/mcp_client.py
```diff
@@ -55,7 +55,7 @@
 _RESPONSES_REASONING_MODEL_PREFIXES = ("gpt-5", "o1", "o3", "o4")
 
 
-def _init_model(model_name: str, trace_dir: Path | None = None) -> Any:
+def init_model(model_name: str, trace_dir: Path | None = None) -> Any:
```
hvent90, line 58: this gets used by `dimos.evals.agents.lib.single_call`, so I had to remove the underscore

dimos/agents/llm_trace.py
```diff
@@ -130,17 +131,19 @@ def write_normalized(trace_dir: Path, messages: list[Any], result: Any) -> tuple
     return seq, req, resp
 
 
+def list_llm_trace_pairs(trace_dir: Path) -> list[tuple[int, Path, Path]]:
```
hvent90, line 134: extracted so that `dimos.evals.agents.mcp_client_adapter` can use it when building out the ATIF trajectory0

dimos/evals/agents/lib/langchain_recorder.py
```diff
@@ -1,134 +0,0 @@
...
-        elif block.get("type") == "reasoning":
-            summary = block.get("summary")
-            if isinstance(summary, list):
-                parts.append(
-                    "".join(str(s.get("text", "")) for s in summary if isinstance(s, dict))
-                )
-    return "\n\n".join(p for p in parts if p)
-
-
-class LangChainRecorder(BaseCallbackHandler):
```
hvent90: this was handling slightly different use cases between two different consumers and the name `LangChainRecorder` isn't that self-documenting, so this got turned into `langchain_to_atif.py` and a bit more responsibility has been offloaded to consumers

dimos/evals/agents/lib/single_call.py
```diff
@@ -0,0 +1,97 @@
...
+Blocks = list[str | dict[str, Any]]
+
+
+class SingleCallAgentConfig(AgentConfig):
+    model: str = DEFAULT_MODEL
+    system_prompt: str = EVAL_SYSTEM_PROMPT
+    chat_model: BaseChatModel | None = None
+
+
+class SingleCallAgent(Agent):
```
hvent90, line 46: shared by `blind` and `question_answer` as they are both just a single agent step

dimos/evals/environments/dataset.py
```diff
@@ -16,95 +16,80 @@
...
+from dimos.memory.stream import Stream
+from dimos.protocol.service.spec import BaseConfig
 
 if TYPE_CHECKING:
-    from dimos.e2e_tests.dimos_cli_call import DimosCliCall
-    from dimos.memory.store.base import Store
+    from dimos.evals.agent import Agent
+
+
+class DatasetConfig(BaseConfig):
```
hvent90, line 36: refactored agents and environments to use the config pattern

dimos/evals/agents/lib/chat.py
hvent90: refactored and renamed to `dimos.evals.agents.lib.single_call`, which acts as a base class for `blind` and `question_answer` agents

dimos/evals/agents/lib/langchain_recorder.py
hvent90: I believe this is resolved: https://github.com/dimensionalOS/dimos/pull/3774#discussion_r3953503114

dimos/evals/types.py
hvent90: Thanks @paul-nechifor, string arrays are now used

dimos/evals/agents/blind.py
hvent90: woops, all `dimos.evals.{agents,environments}` now properly subclass from their appropriate base class in `dimos.evals.{agents,environments}.base`

dimos/evals/environments/dataset.py
hvent90: It's now an array of strings

dimos/evals/agents/lib/single_call.py
```diff
@@ -0,0 +1,97 @@
...
+from langchain_core.messages import AIMessage, BaseMessage, HumanMessage, SystemMessage
+from langchain_core.outputs import ChatGeneration
+
+from dimos.agents.llm_trace import latest_pair, write_normalized
+from dimos.evals.agents.base import Agent, AgentConfig
+from dimos.evals.agents.lib.langchain_to_atif import append_ai_message_to_atif
+from dimos.evals.agents.lib.trajectory_builder import TrajectoryBuilder
+from dimos.evals.types import RunningEnvironment, Trajectory
+
+DEFAULT_MODEL = "gpt-5.6-luna"
```
hvent90, line 34: this and `EVAL_SYSTEM_PROMPT` are being removed in #3879 in favor of allowing `SingleCallAgentConfig` to own the declaration of this value (thus removing a global constant)

spomichter:
@greptile final review and update your scoring for this PR

dimos/evals/environments/image_file.py
```diff
@@ -0,0 +1,63 @@
...
+
+if TYPE_CHECKING:
+    from dimos.evals.agents.base import Agent
+
+
+class ImageFileConfig(BaseConfig):
+    path: Path
+
+
+class ImageFile(Environment):
```
spomichter, line 35: @hvent90 In second pass this should be a generic single frame of any DimOSMsg type, so a single image, odom, or pointcloud frame

spomichter (approved):

PR #3776, opened 2026-08-31
Feat/pim/m20 kronknav

Nabla7:
- Adds the M20 integration machinery: native ROS/DDS command bridge, robot lifecycle/control, direct lidar and IMU ingestion into Point-LIO, KronkNav mapping/planning/control, and front/rear camera streams.
    
- Extends NativeModuleConfig with optional Linux CPU affinity so Point-LIO, ray tracing, and MLS can be assigned to appropriate RK3588 cores, this is necessary as otherwise performance becomes severely degraded.

Some notes:

- This runs fully on the RK3588, further tuning will be required as the mapper is always running behind and this gets worse as the global map grows, the bottleneck is RayTracingVoxelMap on this device.

- Planning on flat terrain works consistently, planning over stairs is hit or miss. Naturally since the map is always behind dynamic obstacle avoidance does not work.

How to run:

On the M20:
   Configure permissions:
  ./dimos/robot/deeprobotics/m20/deploy/setup.sh
  
  Start blueprint:
  dimos --transport lcm --rerun-host 0.0.0.0 run deeprobotics-m20-kronknav-control
  
  Attach the control shell:
  dimos --transport lcm shell
  app.M20Connection.standup()
  app.M20Connection.liedown()
  
On laptop:
   dimos-viewer --connect rerun+http://10.21.31.104:9877/proxy \
    --ws-url ws://10.21.31.104:3030/ws

https://github.com/user-attachments/assets/c47450db-9937-4708-81a2-9783db229b25


dimos/robot/deeprobotics/m20/deploy/dimos-m20-multicast-relay-supervisor
```diff
@@ -0,0 +1,95 @@
...
+  fi
+  sleep 0.5
+done
+
+if ! network_is_ready; then
+  echo "M20 multicast relay network did not become ready within 90 seconds" >&2
+  exit 1
+fi
+
+/usr/bin/python3 /usr/bin/multicast.py &
```
leshy: this file seems to be missing from PR? 
emitting can be done with firewall/kernel/routing config no need to have a forwarder service probably? and if forwarding with userspace, probably using python for it wastes a lot of resources

dimos/robot/deeprobotics/m20/pointlio/module.py
```diff
@@ -0,0 +1,141 @@
...
+    """Run the pinned Point-LIO core directly on the M20's ROS sensor topics.
+
+    The native process subscribes to the merged ``base_link`` cloud and 200 Hz
+    ``base_link`` IMU itself, avoiding a full-payload LCM hop through the command
+    bridge. It has no vendor odometry input and owns ``odom -> base_link``.
+    """
+
+    config: M20PointLioConfig
+
+    lidar: Out[PointCloud2]
```
aclauer: Do the point clouds coming from the ray tracer look like they are correct? Since the two sensor clouds are getting merged together the points are getting traced through the base_link instead of the sensor origins I'd expect that ray tracing won't be great in all cases

dimos/robot/deeprobotics/m20/pointlio/module.py
```diff
@@ -0,0 +1,141 @@
...
+    extrinsic_t: list[float] = Field(default_factory=lambda: [0.0, 0.0, 0.0])
+    extrinsic_r: list[float] = Field(
+        default_factory=lambda: [1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0]
+    )
+    publish_odometry_without_downsample: bool = False
+    odom_only: bool = False
+    debug: bool = False
+
+
+class M20PointLio(NativeModule, perception.Lidar):
```
aclauer: Ideally we have a general PointLio module and then we just need to write drivers for each lidar. Don't expect that here because it's not really refactored for that, but just thought I'd drop that in as an fyi

dimos/robot/deeprobotics/m20/pointlio/module.py
aclauer: I think @jeff-hykin is working on supporting this rn

dimos/robot/deeprobotics/m20/blueprints/m20_kronknav.py
```diff
@@ -0,0 +1,280 @@
...
+from dimos.robot.deeprobotics.m20.pointlio.module import M20PointLio
+from dimos.utils.logging_config import setup_logger
+from dimos.visualization.vis_module import vis_module
+
+VOXEL_SIZE_M = 0.1
+PLANNER_VIZ_HZ = 0.0
+logger = setup_logger()
+
+
+class _M20CameraRelay(Module):
```
aclauer: Should this module maybe be moved out of the blueprint?

dimos/robot/deeprobotics/m20/pointlio/cpp/main.cpp
```diff
@@ -0,0 +1,802 @@
...
+                point.line = ring;
+                points.push_back({timestamp, point});
+                min_ring = std::min(min_ring, ring);
+                max_ring = std::max(max_ring, ring);
+            }
+            if (points.empty()) {
+                throw std::runtime_error("M20 cloud has no finite Point-LIO returns");
+            }
+
+            std::sort(points.begin(), points.end(),
```
aclauer: Why sort the entire point cloud if we just need min and max timestamps?

leshy:
this robot already runs vendor configured pointlio on the nav machine so we shouldn't run another pointlio. there is a zenoh forwarder service that uses DRDDS already written in that PR

https://github.com/dimensionalOS/dimos/pull/2885/changes#diff-f152e876f1992a4ff070fba547f6ca05dadf7c65641dae75999aa7852791117b

leshy:
video transforms seem to be missing from robot, camera feeds should be visible in 3d

leshy:
we don't have to run rerun on the robot with limited resources, you should connect via zenoh to the robot, can give info on this

dimos/robot/deeprobotics/m20/pointlio/module.py
leshy: I think we are kind of 95% ok broadcasting from base_link for now but yes

Nabla7: From what I can tell the pointclouds look accurate, what failures would you expect to pop up? 

Uploading github2.mov…

dimos/robot/deeprobotics/m20/pointlio/module.py
Nabla7: Yes was aware of this, did it this way as a refactor would be beyond scope of integration, also per Ivan's comments, the robot also already has pointlio running onboard which makes most of the pointlio work I've done redundant. Working on removing it now.

dimos/robot/deeprobotics/m20/blueprints/m20_kronknav.py
Nabla7: Probably, I wanted to remove file count but does make things less clear.

Nabla7:
> we don't have to run rerun on the robot with limited resources, you should connect via zenoh to the robot, can give info on this

If you can point me to where this pattern is established already would be great.

leshy:
> > we don't have to run rerun on the robot with limited resources, you should connect via zenoh to the robot, can give info on this
> 
> If you can point me to where this pattern is established already would be great.

https://github.com/dimensionalOS/dimos/pull/3572/changes#diff-8e0ae3cca6583f09a1fea072191c8e479fdf0120cb8b771f476b7b248b251f99R378

```
dimos --robot-ips "go2" run go2-zenoh-motion-local
```

dimos/robot/deeprobotics/m20/pointlio/module.py
leshy: likely motion straight above the robot will be messed up, but this is a rare case

leshy (dismissed):

dimos/robot/deeprobotics/m20/connection.py
```diff
@@ -0,0 +1,374 @@
...
+    """
+
+    config: M20ConnectionConfig
+
+    cmd_vel: In[Twist]
+    command_ready: In[Bool]
+    motion_state: In[Int32]
+    gait_state: In[UInt32]
+    odometry: In[Odometry]
+    safe_cmd_vel: Out[Twist]
```
leshy: what is this safe_cmd_vel for?

Nabla7: My bad I completely missed this, it's AI slop. "safety" logic that it added in without me noticing. I'm removing this and the associated logic.

leshy (dismissed):
I'm ok to merge, better design proposed with connection.py removed, replaced by onboard drdds bridge having connection interface

dimos/robot/deeprobotics/m20/connection.py
```diff
@@ -0,0 +1,345 @@
...
+    vendor state and gait transitions and waits for the command path.
+    ``set_navigation_terrain()`` is the only exposed vendor-specific control and
+    selects one of the documented agile navigation gaits without exposing raw
+    gait values.
+    """
+
+    config: M20ConnectionConfig
+
+    cmd_vel: Out[Twist]
+    command_ready: In[Bool]
```
leshy, line 95: actually just to make sure now after save_cmd_vel thing, do the rest of these make sense :D ? if we move to drdds bridge serving connection I assume a lot of these would go away

aclauer (approved):

PR #3778, opened 2026-08-31
Re-enable Mac tests

Dreamsorcerer:


.github/workflows/ci.yml
```diff
@@ -627,11 +627,10 @@ jobs:
...
             experimental: false
-          # macOS runner disabled for now — we don't want Mac tests.
-          # - os: macOS
-          #   container: null  # run on host — `container:` is Linux-only
-          #   markers: "self_hosted"
-          #   experimental: true
+          - os: macOS
+            container: null  # run on host — `container:` is Linux-only
+            markers: "self_hosted"
+            experimental: true
```
Dreamsorcerer, line 633:
```suggestion
            experimental: false
```

leshy (changes requested):
This seems like a hack that will fix coordinator pathway but generally pubsub/zenoh is broken still so will bite us later.

Also mac specific zenoh hack shouldn't go in core/coordination but some other dir, likely under protocol/pubsub/impl/zenoh/mac or something similar - so that we are clear what this code is for, this should be a pubsbub impl level fix, maintaining normalized interface towards zenoh 

the reason we point multicast to a local interface is to avoid multiple machines on the same network with zenoh busses connecting to each other unless asked explicitly.
so likely there are other solutions to break up zenoh - can we point mutlicast address to just 127.0.0.1? should we limit cross-machine connections in some other way, but allow multicast on other interfaces on mac? 

I understand you asked claude to "plz fix" but since intervention is deep, please describe/own the solution here, make sure to understand zenoh, it will be our primary protocol so time spent on this is not wasted

PR #3844, opened 2026-08-31
feat(manipulation): support unbounded planar-base motion

TomCC7:
## Contribution path

- Tracks [DIM-1610](https://linear.app/dimensional/issue/DIM-1610/support-mobile-manipulation-with-moving-base-robot-models).
- Stacked on #3784.
- Keeps RoboPlan as the scene, collision, kinematics, and trajectory-parametrization backend.

## Problem

The first planar-base implementation treated x, y, and yaw as bounded scalar joints. That imposed an artificial workspace, made yaw discontinuous at ±π, and spread joint-coordinate policy across backend adapters.

Viser could also rebuild joint controls concurrently when planning groups changed quickly, causing duplicate handle removal and dictionary mutation errors.

## Solution

`prepare_robot_model()` now compiles the materialized URDF once into a `PreparedRobotModel`. Its ordered `JointSpace` assigns one topology to each controlled coordinate:

| URDF coordinate | Topology | Planning behavior |
|---|---|---|
| Bounded revolute or prismatic | `INTERVAL` | Uses physical lower and upper limits |
| Unbounded prismatic | `LINE` | Uses a finite request-local sampling domain |
| Continuous revolute | `CIRCLE` | Uses shortest-angle deltas and wrapped sampling |

Worlds, IK, planners, trajectory generation, and Viser consume this prepared model. Robot assets supply positive finite velocity and acceleration limits before preparation.

Selected-joint RRT expands finite request-local domains for unbounded coordinates while sharing one deadline and node budget. Distance, interpolation, collision subdivision, simplification, and path length use velocity-scaled shortest circular deltas. Circular paths are lifted before trajectory generation.

Pink maps canonical line and circle coordinates to Pinocchio while preserving locked joints. Drake consumes the prepared joint space directly. RoboPlan uses its native planner for interval-only selections and the shared dimOS RRT for selections that contain line or circle coordinates. RoboPlan TOPP-RA expands continuous scalars to cosine/sine coordinates, then restores the canonical scalar path.

R1 Pro uses `r1pro/base_x` and `r1pro/base_y` as unbounded translations and `r1pro/base_yaw` as a continuous angle. The fake planner uses the normal RoboPlan TOPP-RA default. Viser renders line coordinates as number inputs and circle coordinates as wrapped sliders. Joint-control rebuilds are serialized and idempotent.

Base execution remains disabled until R1 Pro has a feedback-controlled base trajectory executor. Cartesian waypoint planning remains interval-only.

## Verification

```bash
.venv/bin/pytest -q \
  dimos/robot/assets/test_model.py \
  dimos/manipulation/planning \
  dimos/manipulation/test_generated_plan_materialization.py \
  dimos/manipulation/test_plan_execution.py \
  dimos/manipulation/test_roboplan.py \
  dimos/robot/galaxea/r1pro/blueprints/manipulation/test_r1pro_planar_preview.py
```

Result: **299 passed, 13 deselected**. Ruff and pre-commit pass on the affected code.

## Manual R1 Pro test

```bash
uv sync --extra all
dimos run r1pro-planar-preview
```

In Viser:

1. Select `moving_base`; plan to `r1pro/base_x = 6.0` and `r1pro/base_y = -6.0`.
2. Plan across the ±π yaw boundary and confirm the preview takes the short path.
3. Select `left_arm`, `torso`, and `moving_base`; move the left TCP target and plan.
4. Repeat with only `left_arm`; confirm the torso, right arm, and base remain fixed.
5. Add an obstacle and confirm RRT finds a detour.
6. Toggle planning groups rapidly; confirm the controls rebuild without errors.
7. Preview a base plan, then execute it; confirm execution reports the missing base-controller capability.

See `docs/capabilities/manipulation/unbounded-planar-base.md` for the detailed design.

## AI assistance

Codex implemented and tested the change under human direction.

## Checklist


dimos/robot/manipulators/r1pro/blueprints/basic.py
```diff
@@ -24,6 +24,7 @@
     planner(
         model=make_r1pro_model_config(),
         visualization={"backend": "viser"},
+        trajectory_parametrization={"backend": "simple_trapezoid"},
```
TomCC7: why is this added

dimos/manipulation/planning/spec/joint_space.py
```diff
@@ -0,0 +1,276 @@
...
+
+    name: str
+    mechanism_type: str
+    topology: CoordinateTopology
+    lower: float | None
+    upper: float | None
+    max_velocity: float
+    max_acceleration: float
+
+    def __post_init__(self) -> None:
```
TomCC7: use pydantic dataclass for verification

TomCC7: Done. JointCoordinate is now a frozen Pydantic dataclass with constrained scalar fields and cross-field topology/bounds validation. The hot-loop JointConfiguration and JointTangent types remain standard-library dataclasses.

dimos/robot/manipulators/r1pro/blueprints/basic.py
TomCC7: Removed. The explicit fallback was compensating for the missing continuous-coordinate adaptation in DimOS. The RoboPlan TOPP-RA adapter now expands scalar continuous joints to cos/sin before generation, then collapses and lifts the output afterward. R1 Pro therefore uses the normal RoboPlan TOPP-RA default.

dimos/manipulation/planning/planners/rrt_planner.py
```diff
@@ -230,58 +234,100 @@ def plan_selected_joint_path(
...
+                        goal_tree,
+                        extended.config,
+                        self._connect_step_size,
+                        attempt_budget - nodes,
                     )
-                    return _create_success_result(path, time.time() - start_time, iteration + 1)
-
-            start_tree, goal_tree = goal_tree, start_tree
-            trees_swapped = not trees_swapped
+                    nodes += connected_nodes
```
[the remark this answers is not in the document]
TomCC7: Fixed in c11007f556. Connect-phase node growth is now charged to the shared iteration counter used by LINE-domain retries. The deterministic regression exhausts each connect budget and asserts that reported iterations equal all extend plus connect expansions and remain within max_iterations. It reproduced 5 reported versus 80 actual expansions before the fix; afterward both are 18 under a budget of 20. The full fast manipulation suite passes with 465 tests, and the two previously failing RoboPlan self-hosted integration tests also pass.

dimos/manipulation/planning/spec/joint_space.py
```diff
@@ -0,0 +1,281 @@
...
+        if values.shape != (len(self.coordinates),):
+            raise ValueError(
+                f"Joint {label} has shape {values.shape}, expected {(len(self.coordinates),)}"
+            )
+        if not np.isfinite(values).all():
+            raise ValueError(f"Joint {label} must contain only finite values")
+
+
+def _wrap_angle(value: float) -> float:
+    return (float(value) + math.pi) % (2.0 * math.pi) - math.pi
```
mustafab0: We already have a `angle_diff` utility in dimos/utils/trig or something like that

dimos/manipulation/planning/spec/joint_space.py
```diff
@@ -0,0 +1,281 @@
...
+            raise ValueError("Interpolation fraction must be finite")
+        tangent = self.delta(start, end)
+        return self.integrate(
+            start,
+            JointTangent(self.names, tuple(fraction * value for value in tangent.values)),
+        )
+
+    def distance(self, start: JointConfiguration, end: JointConfiguration) -> float:
+        delta = self.delta(start, end).as_array()
+        return float(np.linalg.norm(delta / np.asarray(self.velocity_limits)))
```
mustafab0: We have all these same methods in `selected_joint_space.py` too.

Better to make these utilities if we need to reuse.

dimos/manipulation/planning/spec/joint_space.py
```diff
@@ -0,0 +1,281 @@
...
+
+    names: tuple[str, ...]
+    values: tuple[float, ...]
+
+    def __post_init__(self) -> None:
+        if len(self.names) != len(self.values):
+            raise ValueError("Joint tangent names and values must have equal length")
+
+    def as_array(self) -> NDArray[np.float64]:
+        return np.asarray(self.values, dtype=np.float64)
```
mustafab0: Both of these seem to just return arrays. 

Do we need these?

mustafab0 (changes requested):

mustafab0 (changes requested):
Some basic clean up requested

.github/workflows/autofix.yml
TomCC7, line 1: revert this

mustafab0 (dismissed):

mustafab0 (approved):

PR #3852, opened 2026-09-01
feat: mid360 rust driver

aclauer:
## Contribution path

- Linked issue or discussion: DIM-XXX / #XXX / URL

## Problem

Need to parse the mid360 udp to frames.

## Solution

if you're so interested here's the protocol
https://livox-wiki-en.readthedocs.io/en/latest/tutorials/new_product/mid360/mid360.html#communication-protocol

Adds a module to read packets and parse in to the raw clouds. These will get ingested by pointlio when I write that next.
- `PacketSource` trait implemented by live source and pcap source
- livox udp packet parsing in wire.rs
- adds Rust to self hosted runner to run the e2e tests (which replays a full pcap through from lfs)

## How to Test

runs cargo example that parses all the points and logs some metrics. also logs runtime of parser
`cargo run --release -p dimos-livox --example pcap_stats -- data/mid360_stairs_shake/mid360_stairs_shake.pcap`

live sensor:
`dimos run mid360`

if lidar isn't at the default place (no points will show up)
`DIMOS_MID360_LIDAR_IP=<the sensor ip> dimos run mid360`

## Checklist


dimos/hardware/sensors/lidar/livox/rust/src/live.rs
```diff
@@ -0,0 +1,572 @@
...
+
+impl LiveSource {
+    /// `stop` ends `recv` when set, e.g. from a module shutdown signal. The
+    /// source also sets it itself when the handshake or a reader fails.
+    pub fn start(config: LiveConfig, stop: Arc<AtomicBool>) -> io::Result<LiveSource> {
+        let failure = Failure {
+            reason: Arc::new(OnceLock::new()),
+            stop: stop.clone(),
+        };
+        let (tx, rx) = mpsc::channel::<Vec<u8>>();
```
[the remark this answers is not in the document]
aclauer: addresses

dimos/hardware/sensors/lidar/livox/rust/src/live.rs
```diff
@@ -0,0 +1,572 @@
...
+    socket: UdpSocket,
+    tx: mpsc::Sender<Vec<u8>>,
+    failure: Failure,
+) -> std::thread::JoinHandle<()> {
+    std::thread::spawn(move || {
+        let mut buf = [0u8; 4096];
+        while !failure.stop.load(Ordering::Relaxed) {
+            match socket.recv_from(&mut buf) {
+                Ok((len, _)) => {
+                    if tx.send(buf[..len].to_vec()).is_err() {
```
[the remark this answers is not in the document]
aclauer: addresses

omarespejel:
hey hey @aclauer, qq have you tested this on a real mid360, including unplugging and reconnecting it while the driver is running?

dimos/hardware/sensors/lidar/livox/rust/src/live.rs
```diff
@@ -0,0 +1,647 @@
...
+        ));
+        if config.enable_imu {
+            let imu = data_socket(&config, config.ports.host_imu_data)?;
+            threads.push(spawn_reader("imu", imu, lidar_ip, tx, failure.clone()));
+        }
+
+        let cmd = UdpSocket::bind(SocketAddrV4::new(
+            config.host_ip,
+            config.ports.host_cmd_data,
+        ))?;
```
omarespejel: Can we bind the sockets before starting any reader threads? If the IMU or command bind fails, `start()` returns before `LiveSource` takes ownership, leaving the point reader running with its port still bound

dimos/hardware/sensors/lidar/livox/rust/src/pcap.rs
```diff
@@ -0,0 +1,418 @@
...
+                    self.reader.consume(offset);
+                    if let Some(packet) = packet {
+                        return Some(packet);
+                    }
+                    if self.state.failure.is_some() {
+                        return None;
+                    }
+                }
+                // A truncated final record reads as UnexpectedEof.
+                Err(PcapError::Eof) | Err(PcapError::UnexpectedEof) => return None,
```
omarespejel, line 139: Can we treat `UnexpectedEof` as a replay failure? Right now a truncated final record looks like a clean EOF, so the module may publish the buffered frame as if the capture completed

dimos/hardware/sensors/lidar/livox/rust/src/live.rs
```diff
@@ -0,0 +1,647 @@
...
+    let mut buf = [0u8; 2048];
+    loop {
+        if stop.load(Ordering::Relaxed) {
+            return Ack::Timeout;
+        }
+        let remaining = deadline.saturating_duration_since(Instant::now());
+        if remaining.is_zero() || cmd.set_read_timeout(Some(remaining)).is_err() {
+            return Ack::Timeout;
+        }
+        let Ok((len, _)) = cmd.recv_from(&mut buf) else {
```
omarespejel: Can we validate the ACK source address, `cmd_id`, and `sender` in addition to `seq`? As written, any datagram received on the command socket with the current sequence and a valid ACK body can complete the handshake

aclauer:
> hey hey @aclauer, qq have you tested this on a real mid360, including unplugging and reconnecting it while the driver is running?

hey Omar actually no haha, haven't gotten to running with a live sensor yet. I've been replaying pcap to validate the udp parsing up to this point (which seems to work). It's on my list of tests, won't merge before testing with live sensor of course.

Did you run it and have any issues?

dimos/hardware/sensors/lidar/livox/rust/src/wire.rs
```diff
@@ -0,0 +1,726 @@
...
+
+impl<'a> DataPacket<'a> {
+    /// Parse a data packet. The header crc32 field is not verified: real
+    /// devices stream at rates where the SDK itself skips it, and captures
+    /// from hardware do not reliably fill it.
+    pub fn parse(packet: &'a [u8]) -> Result<Self, WireError> {
+        if packet.len() < DATA_HEADER_LEN {
+            return Err(WireError::TooShort);
+        }
+        let dot_num = u16::from_le_bytes([packet[5], packet[6]]);
```
omarespejel, line 443: I think we should validate the declared `length` against `packet.len()` before parsing the payload. The protocol defines it as the complete packet size, and ignoring it here allows malformed datagrams with trailing bytes to pass validation

dimos/hardware/sensors/lidar/livox/rust/src/pipeline.rs
```diff
@@ -0,0 +1,355 @@
...
+            None => {
+                self.frame_start_ns = Some(ts_ns);
+                None
+            }
+        };
+
+        let frame_start = self.frame_start_ns.expect("set above");
+        // Offset 0 = "at the frame stamp": clamp rather than wrap when a UDP
+        // packet arrives out of order with a stamp older than the frame start.
+        self.append_points(packet, ts_ns.saturating_sub(frame_start));
```
omarespejel, line 135: One safeguard that would help here is a limit independent of device time. If the timestamp freezes, every packet falls through `Some(_) => None` and `points` keeps growing. A frequency-aware cap or stale-timestamp error would keep memory bounded

dimos/hardware/sensors/lidar/livox/rust/src/wire.rs
```diff
@@ -0,0 +1,726 @@
...
+            payload,
+        })
+    }
+
+    /// Nanoseconds between consecutive points in this packet.
+    pub fn point_interval_ns(&self) -> u64 {
+        if self.dot_num == 0 {
+            return 0;
+        }
+        u64::from(self.time_interval) * 100 / u64::from(self.dot_num)
```
omarespejel: One small timing detail: the protocol defines `time_interval` as the span from the first point to the last, so `dot_num` points contain `dot_num - 1` gaps. Using that denominator here, with `dot_num <= 1` returning zero, would avoid slightly compressing the per-point timestamps

omarespejel (changes requested):
Hey, replay and loopback both passed locally for me. I left a few comments on cleanup, malformed input, and timestamps. I couldn’t test live reconnect since I don’t have a Mid-360 here. If you already tested that on real hardware, perfect. That’s the one part I couldn’t validate

omarespejel:
hey hey @aclauer, same here 😅 I don’t have a Mid-360 here. Replay and loopback both passed, but I found a few edge cases and left them inline 🙏

dimos/hardware/sensors/lidar/livox/rust/src/pcap.rs
aclauer: unexpected eofs happen when the pcap recording is ended mid record, which is pretty normal and in general an error. as long as we aren't actually parsing the incomplete packet, the standard thing to do is just ignore it since it's the end of the file anyway

aclauer: unexpected eof just happens when the recording ends mid record, which I think is pretty normal and not really an error. I think the standard thing to do is just ignore it because it's the last record of the file anyway

aclauer:
> Hey, replay and loopback both passed locally for me. I left a few comments on cleanup, malformed input, and timestamps. I couldn’t test live reconnect since I don’t have a Mid-360 here. If you already tested that on real hardware, perfect. That’s the one part I couldn’t validate

nice, thanks for the feedback, looks like some good catches there. will review tomorrow since it's late in SF!

omarespejel:
nw! Have a good night @aclauer !

aclauer:
@omarespejel looking good on live sensor
https://github.com/user-attachments/assets/8baeea51-a41f-47be-acb7-a3c4fbba83c6

dimos/hardware/sensors/lidar/livox/rust/src/wire.rs
aclauer: lol turns out the livox ros driver doesn't even match their own spec, good find

https://github.com/Livox-SDK/livox_ros_driver2/blob/master/src/comm/pub_handler.cpp#L142
https://livox-wiki-en.readthedocs.io/en/latest/tutorials/new_product/mid360/livox_eth_protocol_mid360.html#protocol-format

dimos/hardware/sensors/lidar/livox/rust/src/pipeline.rs
aclauer: sure i can add like a max point count or something like that

dimos/hardware/sensors/lidar/livox/cpp/CMakeLists.txt
leshy, line 1: since cpp went away, should we delete whole cpp/ dir ? flakes etc?

aclauer: pointlio and fastlio still need it because they still use the sdk for their own drivers. So we can't delete it all until those are gone too I think

dimos/hardware/sensors/lidar/livox/rust/src/live.rs
```diff
@@ -0,0 +1,688 @@
...
+    value: Vec<u8>,
+}
+
+fn handshake_steps(config: &LiveConfig) -> Vec<Step> {
+    let mut steps = vec![
+        Step {
+            label: "point host cfg",
+            key: wire::param_key::POINT_DATA_HOST_IP_CFG,
+            value: host_ip_config_value(
+                config.host_ip,
```
omarespejel: Unless I’m missing another path, `multicast_ip` is used when joining the receive sockets, but the point and IMU config still use `host_ip` as the destination. Livox SDK2 programs the multicast address when set. If so, can we mirror that here, or drop the option and stay unicast-only for now?

dimos/hardware/sensors/lidar/livox/rust/src/pcap.rs
omarespejel: nice! makes sense, thanks! @aclauer

dimos/hardware/sensors/lidar/livox/rust/src/wire.rs
omarespejel: 🙏

omarespejel:
> @omarespejel looking good on live sensor https://github.com/user-attachments/assets/8baeea51-a41f-47be-acb7-a3c4fbba83c6

nice🙏 qq did you also try unplugging and reconnecting it while the driver was still running? @aclauer

dimos/hardware/sensors/lidar/livox/rust/src/live.rs
omarespejel: thanks! this gets it done!

omarespejel (approved):

omarespejel:
hey hey @aclauer, LGTM from my side 🙏 did you get a chance to check that points and IMU resume after unplugging and reconnecting, without restarting the driver?

leshy (dismissed):

leshy (approved):

PR #3854, opened 2026-09-01
feat(imitation): add native OpenYAM collection

TomCC7:
## Contribution path

- Targets `main`; the dependent rollout and learning PRs remain stacked above this change.

## Problem

- Fixed recorder ports and a separate dataset profile can disagree about stream names, camera counts, and joint order.
- Offline preparation needs the recording's original schema even after the robot configuration changes.

## Solution

- Introduce the final `CollectionProfile` and recorder factory here, before robot presets depend on them.
- Generate typed input ports before autoconnect. Capture each distinct source once, even when several dataset features project from it.
- Save `schema.json` beside `recording.mcap` or `recording.db` in a new recording directory.
- Prepare MCAP/SQLite through the saved schema, with LeRobot conversion isolated from host dependencies and HDF5 available through the Python API.
- Declare raw-source meaning with `source_kind`, not a selectable sampling algorithm. Snapshots retain nearest alignment; partial joint-position updates reconstruct a causal target state before feature projection.
- Read each raw source once per episode, including command history before the episode. Share evaluation between `inspect` and `prepare`.
- Validate shared-source consistency and reject malformed updates or missing initial joints. Do not borrow future commands or fabricate missing targets.
- Define `EpisodeControlSpec` and canonical episode messages. Stopping leaves an active episode incomplete; incomplete/discarded episodes are excluded from export.
- Add the ordinary `openyam-quest-collection` Blueprint, with measured state and accepted joint commands.
- Configure cameras through `WebcamConfig`, preserve camera CLI settings and startup/shutdown fixes, and publish accepted commands from the base `ControlCoordinator`.
- Decode native MCAP channels in dataprep through the existing `McapStore(codecs=...)` interface.
- Remove obsolete dynamic-profile DataPrep CLI paths. Replace the Python collection recorder in place and migrate the Piper and xArm collection blueprints.

### Recorder replacement

Collection is a specialization of the existing Rust recorder:

```text
RustRecorder                 RustRecorderConfig
└── CollectionRecorder       └── CollectionRecorderConfig
    └── profile-generated typed inputs
```

`dimos.imitation.collection.recorder` now contains the Rust-backed implementation and the `collection_recorder(profile=...)` factory. The old Python collection implementation and separate `native_recorder.py` are removed. There is no parallel `NativeRecorder` base hierarchy or compatibility alias.

Collection adds fresh-directory preparation, `schema.json`, and required-input checks, then delegates payload preparation and process lifecycle to `RustRecorder`. The public destination API remains `recording` plus `format`; it derives `recording.mcap` or `recording.db` and rejects conflicting inherited store settings or append/backup/overwrite policies. Ordinary `RustRecorder` file policies remain available to its other callers.

The xArm and Piper collection blueprints use the same factory, preserve their measured-state action semantics, and default to SQLite in timestamped session directories. The collection-to-HDF5 regression now records through the real Rust process and runs in native CI.

### Command feedback and G1 recording

`ControlCoordinator` declares `applied_joint_position_command` for diagnostics and collection. `TeleopControlCoordinator` inherits it. The hardware-write pass gathers only successful position/servo-position writes and publishes one sparse update after releasing the hardware lock. There is no subclass lookup, intermediate accepted-command map, separate publishing helper, additional arbitration, or held-command state. Rejected, unready, failed, and non-position writes produce no position feedback.

G1's existing recorder derives directly from `dimos.memory.module.Recorder` and explicitly declares its camera, joint-state, episode-status, and controller-pose inputs. This fixes the missed caller of the replaced collection recorder while preserving G1's raw SQLite files, dedicated worker, pose-free recording, and Python shared-memory camera transport. G1 is not switched to a Rust backend that cannot consume that transport.

### Why the native message update is required

Collection records `EpisodeStatus` as an LCM message. Python needs its generated bindings, and the Rust recorder needs the generated message to decode episode events and preserve their source timestamps. The existing `rust-codegen` revision lacks that message, so this PR pins the updated schema revision, [`dd215951`](https://github.com/dimensionalOS/dimos-lcm/commit/dd2159513ebfaa7ebc5fc32bf60209cd09aa1ca5).

The Rust pin also reaches the SDK and other members of the root Cargo workspace because the recorder's Nix package vendors that workspace's lockfile. Restricting the update to the recorder passed `cargo check`, but failed the actual Nix build: the old and new sources both declare `lcm-msgs 0.1.0`, and vendoring them collides at `lcm-msgs-0.1.0`. Keeping the workspace on one source revision fixes packaging. These additional pins are required by the verified build path; those modules do not consume the new episode message.

RealSense and dimSLAM are separate Cargo projects. Their direct message dependencies retain `rust-codegen`; their lockfiles include the updated shared SDK dependencies. Both projects passed `cargo check` with that arrangement.

### Why the CI discovery fix is required

The native-build scanner recognizes commands beginning with `nix build` and requires a literal `cwd`. Inserting experimental-feature flags before `build` caused it to silently omit the recorder. The Linux test runner then found no prebuilt binary and attempted a build with `max-jobs = 0`, producing seven setup errors.

The fix restores `nix build -L .#dimos-memory-recorder` and the literal `cwd = "rust"`, and resolves the directory locally in the recorder config so generated subclasses use the same native project. Regression checks require the recorder in the publish manifest, recognize an unchanged inherited build once, and catch undiscovered subclasses that override its command or directory.

### API shape

`message_type` describes the source message; `field`, `dtype`, and `shape` describe the exported dataset feature. `dtype` remains explicit because Image data can be exported as video, and JointState positions can be converted to float32. It cannot be inferred from the message class alone.

Declarations below omit validation/default details:

```python
class CollectionFeature(FeatureSpec):
    message_type: type[Any]
    # FeatureSpec supplies stream, field, dtype, shape, names, and:
    source_kind: Literal["snapshot", "joint_position_updates"] = "snapshot"

class CollectionProfile(BaseConfig):
    name: str
    robot_type: str
    observations: dict[str, CollectionFeature]
    actions: dict[str, CollectionFeature]
    sync: SyncConfig
    quality: QualityConfig

def collection_recorder(
    *,
    profile: CollectionProfile,
    recording: Path | None = None,
    format: Literal["mcap", "sqlite"] = "mcap",
    instance_name: str = "recorder",
) -> Blueprint: ...
```

```python
collect = autoconnect(
    my_robot,
    *my_cameras,
    collection_recorder(profile=MY_PROFILE),
    EpisodeMonitorModule.blueprint(instance_name="episodes"),
)
```

- `source_kind="snapshot"` applies to measured state and images, with existing nearest-frame tolerance. `source_kind="joint_position_updates"` applies to partial target updates: omitted joints retain their last target across ticks and episode boundaries.
- Source meaning is independent of whether a feature is an observation or action. Shared-source projections must agree.
- There is no legacy `sampling` alias or automatic schema migration. Existing recordings are not rewritten.
- Profiles have no registry. Ordinary Blueprint discovery selects the setup; the profile name is recording metadata.
- The graph constructs cameras, hardware, and transports; image features do not create devices.
- Add camera features and matching producers without a recorder subclass. Normal remappings resolve different names.
- Source ports must be nonreserved Python identifiers with importable, native-encodable message types. The recorder reserves `status: In[EpisodeStatus]`.
- `EpisodeControlSpec` exposes `get_status()` and `command("start" | "save" | "discard" | "toggle")`.

```text
recordings/session-001/
├── schema.json
└── recording.mcap  # or recording.db
```

- Existing directories are rejected. Move/copy the complete directory.
- Python preparation: `RecordingSchema.read(directory).dataprep_config(directory, output)`, then `run_lerobot_dataprep(config)` or `run_dataprep(config)`.
- Only prepare trusted recordings, with their custom message packages installed.
- CLI panels come in #3931; dual-arm presets come in #3942.

## How to Test

With OpenYAM, Quest, and a wrist RGB camera configured:

```bash
dimos --can-port follower_l run openyam-quest-collection --recorder.recording recordings/session-001 --episodes.task "pick up the cube"
```

- Quest B starts/saves; Y discards. Repeat with a new directory and `--recorder.format sqlite`.

Validation of the current review fixes:

- 459 tests passed covering control, collection, G1 teleop, manual-control priority, and blueprint registration.
- The exact previously failing `unitree-g1-teleop` blueprint-kwargs CI check passed.
- 685 tests passed after restacking, covering control, imitation learning, OpenYAM/dual-arm and G1 blueprints, WebXR, and registry generation.
- All nine native integration tests passed, including packaged recorder interoperability, OpenYAM collection over Zenoh, and collection-to-HDF5 export.
- Mypy passed for all five changed production modules; Ruff and commit hooks passed.
- Real hardware collection remains a release check.

## AI assistance

- Implementation cleanup and this description updated by OpenAI Codex (agent).

## Checklist


dimos/experimental/memory/README.md
```diff
@@ -94,3 +94,24 @@ Append mode remains unsupported for MCAP.
...
+
+It records the USB wrist camera, aggregate coordinator joint state,
+hardware-accepted joint position command, and typed episode status over
+reliable Zenoh. MCAP is the default; override its path with:
+
+```bash
+  --nativecollectionrecorder.store.path /path/to/session_openyam.mcap
+```
+
+The artifact feeds `dimos dataprep build --profile openyam` directly.
```
TomCC7:
```suggestion
```

dimos/robot/manipulators/openyam/blueprints/learning_collection.py
```diff
@@ -0,0 +1,99 @@
...
+from dimos.protocol.pubsub.impl.zenohpubsub import QOS_NEVER_DROP, Topic as ZenohTopic
+from dimos.robot.manipulators.openyam.blueprints.teleop import teleop_quest_openyam
+from dimos.robot.manipulators.openyam.learning import OPENYAM_LEARNING_PROFILE
+
+
+def _session_mcap() -> str:
+    return str(STATE_DIR / "recordings" / f"session_openyam_{datetime.now():%Y%m%d_%H%M%S}.mcap")
+
+
+def _require_zenoh() -> str | None:
```
TomCC7: don't think we 'require' zenoh here

dimos/imitation/dataprep/cli.py
```diff
@@ -37,14 +37,24 @@
...
     output_format: Literal["lerobot", "hdf5"] | None,
 ) -> DataPrepConfig:
     """Build a DataPrepConfig from an optional JSON file + flag overrides."""
     from dimos.imitation.dataprep.core import DataPrepConfig, OutputConfig
 
-    if config_path is not None:
+    if config_path is not None and profile is not None:
+        raise ValueError("--config and --profile are mutually exclusive")
+    if profile == "openyam":
+        from dimos.robot.manipulators.openyam.learning import OPENYAM_LEARNING_PROFILE
```
TomCC7: really weird to have cli import profile from robot modules

TomCC7: what about dynamic loading through class specification? or what happened to the previous json based profile?

dimos/imitation/dataprep/formats/lerobot/reader.py
TomCC7, line 1: wondering why we still need the reader if we have writer to be part of the lerobot package

dimos/hardware/sensors/camera/module.py
```diff
@@ -47,7 +47,8 @@ def default_transform() -> Transform:
 class CameraModuleConfig(ModuleConfig):
     frame_id: str = "camera_link"
     transform: Transform | None = Field(default_factory=default_transform)
-    hardware: Callable[[], CameraHardware] | CameraHardware = Webcam
+    hardware: Callable[[], CameraHardware] | CameraHardware | None = None
+    webcam: WebcamConfig = Field(default_factory=WebcamConfig)
```
TomCC7: why this addition and how is webcam different from hardware seems super weird

dimos/imitation/dataprep/formats/lerobot/reader.py
TomCC7, line 1: why we still need this reader can this also be part of the lerobot package?

TomCC7: Removed the host LeRobot reader. Build and inspect now share the isolated LeRobot entrypoint, and inspection uses public `LeRobotDatasetMetadata` inside the locked environment.

dimos/robot/manipulators/openyam/blueprints/learning_collection.py
TomCC7: Removed the Zenoh requirement, explicit QoS transport map, and blueprint transport override. The graph now uses the globally selected transport; Zenoh remains the current default.

dimos/imitation/dataprep/formats/lerobot/reader.py
TomCC7: Removed the host LeRobot reader. Build and inspect now share the isolated LeRobot entrypoint, and inspection uses public `LeRobotDatasetMetadata` inside the locked environment.

dimos/imitation/dataprep/cli.py
TomCC7: Replaced the robot import and JSON config path with dynamic `module:attribute` loading against `DataPrepProfile`. Build now requires `--profile` and `--source`; the user-facing JSON config, format flags, and example JSON were removed.

dimos/experimental/memory/README.md
TomCC7: Removed the OpenYAM-specific section from the generic experimental Memory README; the workflow now lives in the imitation documentation.

dimos/hardware/sensors/camera/module.py
TomCC7: Collapsed the camera config to one `hardware` field. `WebcamConfig` is the default hardware spec, so CLI configuration is now `--WristCamera.hardware.camera-index` with no parallel webcam path.

dimos/cli/commands/dataprep.py
TomCC7, line 1: why is this removed tho

dimos/experimental/memory/rust_recorder.py
```diff
@@ -71,48 +71,47 @@ class RustSqliteStoreConfig(RustStoreConfig):
...
 
 RustRecordingStoreConfig: TypeAlias = Annotated[
     RustSqliteStoreConfig | RustMcapStoreConfig,
     Field(discriminator="kind"),
 ]
 
 
-class RustRecorderConfig(NativeModuleConfig):
-    """Compatibility-first configuration for :class:`RustRecorder`.
+class NativeRecorderConfig(NativeModuleConfig):
```
TomCC7, line 100: why would these change necessary? I feel like many of them is not

dimos/hardware/sensors/camera/module.py
```diff
@@ -47,7 +47,9 @@ def default_transform() -> Transform:
 class CameraModuleConfig(ModuleConfig):
     frame_id: str = "camera_link"
     transform: Transform | None = Field(default_factory=default_transform)
-    hardware: Callable[[], CameraHardware] | CameraHardware = Webcam
+    hardware: WebcamConfig | Callable[[], CameraHardware] | CameraHardware = Field(
```
TomCC7, line 50: what's the current design/config for this camera module pls present, shouldn't need so many types IMO

dimos/hardware/sensors/camera/realsense/rust/Cargo.toml
```diff
@@ -36,7 +36,7 @@ infrared_right_camera_info = "sensor_msgs.CameraInfo"
 
 [dependencies]
 dimos-module = { path = "../../../../../../native/rust/dimos-module" }
-lcm-msgs = { git = "https://github.com/dimensionalOS/dimos-lcm.git", branch = "rust-codegen" }
+lcm-msgs = { git = "https://github.com/dimensionalOS/dimos-lcm.git", rev = "dd2159513ebfaa7ebc5fc32bf60209cd09aa1ca5" }
```
TomCC7: why is this necessary can you confirm? or maybe need to be a separate PR?

.github/workflows/ci.yml
TomCC7, line 1: is this CI change necessary? is that because we added turbo jpeg encoding support for rust recorder???

dimos/memory/store/mcap.py
```diff
@@ -19,6 +19,9 @@
 topic -> codec), while ``streams`` may map friendly stream names to topics. See
 ``dimos.robot.unitree.go2.dds.store.Go2McapStore`` for the Go2 DDS wiring.
 
+For trusted native recordings, ``decode_native=True`` reconstructs supported
```
TomCC7: if we remove support for turbojpeg encoding can we avoid changing this file?

dimos/control/coordinator.py
```diff
@@ -150,6 +150,8 @@ class ControlCoordinator(Module):
 
     # Output: Aggregated joint state for external consumers
     coordinator_joint_state: Out[JointState]
+    # Output: Post-arbitration position commands accepted by hardware.
+    applied_joint_position_command: Out[JointState]
```
mustafab0: This should be added to the subclassed Control Coordinator

mustafab0 (changes requested):

dimos/control/tick_loop.py
```diff
@@ -421,8 +427,37 @@ def _write_all_hardware(
...
+                            accepted_commands[hw_id] = (positions, mode)
                     except Exception as e:
                         logger.error(f"Failed to write to {hw_id}: {e}")
+        return accepted_commands
+
+    def _publish_applied_position_command(
+        self,
+        accepted_commands: dict[str, tuple[dict[str, float], ControlMode]],
+        timestamp: float,
+    ) -> None:
```
mustafab0: This seems weird to me as its an extra arbitration level.

The control coordinator should be as dumb as possible. where it just sends whatever command it gets and send out feedback.

With this we are adding logic that we need to keep track of.

dimos/control/coordinator.py
```diff
@@ -895,6 +895,7 @@ def start(self) -> None:
         publish_robot_cb = (
             self._publish_robot_joint_state if self.config.publish_robot_joint_states else None
         )
+        command_port = self.outputs.get("applied_joint_position_command")
```
TomCC7: @mustafab0 I feel like we should still keep the command port in the main class for these reasons:
1. your previous wiring does not consider this lifecycle where command output is not part of the controller task
2. it's a good diagnostic information anyway
3. in the new control coordinator design, the output action port will be always there anyway

mustafab0: a command port does make sense.

Extra logic is where I am hesitant

dimos/control/test_control.py
```diff
@@ -333,6 +334,31 @@ class _EEFTwistCoordinator(ControlCoordinator):
...
 class TestControlCoordinatorLifecycle:
+    def test_base_coordinator_starts_without_command_output(self, make_coordinator, mocker):
+        mocker.patch("dimos.core.module.Module.start")
+        loop = mocker.patch("dimos.control.coordinator.TickLoop")
+        coordinator = make_coordinator()
+
+        coordinator.start()
+
+        assert loop.call_args.kwargs["publish_command_callback"] is None
+        assert "applied_joint_position_command" not in coordinator.outputs
```
mustafab0: This feels like an llm just making tests for edits.
```suggestion

```

dimos/control/test_control.py
```diff
@@ -333,6 +334,31 @@ class _EEFTwistCoordinator(ControlCoordinator):
...
+        received = []
+        unsubscribe = coordinator.applied_joint_position_command.subscribe(received.append)
+        try:
+            message = JointState(name=["arm/joint1"], position=[0.5])
+            coordinator.start()
+            loop.call_args.kwargs["publish_command_callback"](message)
+
+            assert received == [message]
+        finally:
+            unsubscribe()
```
mustafab0:
```suggestion

```

`test_control` shouldn't test subclassed blueprint or we will end up with a lot of tests

dimos/control/test_control.py
```diff
@@ -1302,6 +1328,49 @@ def test_write_all_hardware_rejected_command_logs_error(self, mocker):
             "Hardware arm rejected SERVO_POSITION command from control task"
         )
 
+    def test_applied_position_command_publishes_only_accepted_position_modes(self):
+        publish = MagicMock()
+        tick_loop = TickLoop(
```
mustafab0: 👍

dimos/imitation/collection/blueprint.py
```diff
@@ -49,30 +48,52 @@ def _camera_if_real() -> tuple[Blueprint, ...]:
...
+        if any(
+            stream.name == "color_image" and stream.direction == "out" for stream in atom.streams
+        )
+    )
+    camera_config = cast("CameraConfig", module_config_cls(camera)(**camera.kwargs))
+    state = CollectionFeature(
+        stream="coordinator_joint_state",
+        message_type=JointState,
+        field="position",
+        dtype="float32",
```
mustafab0, line 69: Shouldn't we able to infer `dtype` just from `message_type` ?

mustafab0 (changes requested):

PR #3868, opened 2026-09-01
fix(realsense): move validator to 0.21, matching every other crate

mustafab0:
dimos-module, the ray tracing mapper and the mls planner all moved to validator 0.21; the realsense crate was left on 0.20. Both versions end up in its dependency graph, so its `Config` derived `Validate` from 0.20 while `Module::Config` wanted 0.21, and `cargo check` failed on an unsatisfied trait bound. This blocks any blueprint composing `RealSenseCamera`, which is now every hardware grasping run.

Every other `Cargo.toml` in the repo declares the same `"0.21"` string and there is no workspace dependency table to inherit from, so this aligns realsense with that convention rather than loosening its constraint. Reproduced the failure by putting 0.20 back (`error[E0277]: required for <RealSense as Module>::Config to implement ModuleConfig`) and confirmed `cargo check` is clean on 0.21; the lockfile is consistent under `--locked`.

First of nine in the xArm grasping re-landing stack.

### DIFF
--- a/dimos/hardware/sensors/camera/realsense/rust/Cargo.lock
+++ b/dimos/hardware/sensors/camera/realsense/rust/Cargo.lock
@@ -577,8 +577,8 @@ version = "0.1.0"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.119",
- "toml",
+ "syn 3.0.4",
+ "toml 1.1.4+spec-1.1.0",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


aclauer (dismissed):
Yeah I have a PR that is fixing this with workspace level dependencies for all these things that are shared by all native modules. Right now this module isn't in the top workspace, so it's a bit annoying but fix is coming.

Looks good to unblock!

mustafab0:
@aclauer 

> Actually can you double check greptile on the Cargo lock? Just check that it's regenerated and updated i guess, ping me when ready

I double checked and the cargo.lock is regenerated. Ready to go

aclauer (approved):

PR #3872, opened 2026-09-01
feat(manip): restore reset, and expose cancel as a skill

mustafab0:
`reset` was removed in #3447 but `FAULT` stayed, and planning only runs from `IDLE` or `COMPLETED`, so a faulted module accepted nothing further. Both manipulator agent prompts still instruct the model to call `reset` after a failure, one of them describing it as available as a skill and an RPC, so an agent that faulted was told to call a tool that did not exist. `test_state_transitions_during_execution` calls it too.


dimos/robot/manipulators/xarm/blueprints/test_perception.py
paul-nechifor: please delete

paul-nechifor (approved):

PR #3874, opened 2026-09-01
feat(manip): GraspGenX provider on the grasping stack

mustafab0:
Add GraspGenX to the pick and place stack.

Ranked candidates also needed `pick_object` to stop assuming there is only one. It took `candidates[0]` and gave up if that pose would not plan.

It now walks up to `max_grasp_attempts` candidates, demoting only on a planning failure, because a drive or execution fault would repeat identically for every one of them, and reports the rank it won on.

Verified in the grasp sim: 100 ranked proposals in the planning frame, scores 0.84 down, and a pick that completes on rank 0 with score 0.966.

Seventh of nine in the xArm grasping re-landing stack.


TomCC7:
might be better to make all grasp methods a module config which can be modified from cli instead of brand new blueprint, here's the PR: https://github.com/dimensionalOS/dimos/pull/4032

paul-nechifor (dismissed):

paul-nechifor (approved):

PR #3876, opened 2026-09-01
feat(manip): draw the ranked grasp proposals in viser

mustafab0:
`PickAndPlaceModule` publishes its proposals on a stream as it makes them. `ManipulationModule` subscribes and hands them to the viewer through the world monitor, the same path obstacles and previews already take.

`VisualizationSpec` grows one hook for it; meshcat implements it as a no-op, because it draws the planning world and these are not planning geometry.

One test binds the drawn arrays against the real `viser.SceneApi` signature and asserts their shapes, because a mocked server accepts any array and the first version of this shipped colours as (N, 3) where viser wants (N, 2, 3) — which surfaced only as a swallowed `Grasp proposal visualization failed` in a live run.

Verified in the grasp sim: proposals reach the scene and render with no errors in the log, and the panel checkbox toggles them.

Last of nine in the xArm grasping re-landing stack.

### DIFF
--- a/dimos/manipulation/manipulation_module.py
+++ b/dimos/manipulation/manipulation_module.py
@@ -94,6 +94,7 @@ from dimos.msgs.geometry_msgs.PoseStamped import PoseStamped
 from dimos.msgs.geometry_msgs.Quaternion import Quaternion
 from dimos.msgs.geometry_msgs.Transform import Transform
 from dimos.msgs.geometry_msgs.Vector3 import Vector3
+from dimos.msgs.manipulation_msgs.GraspCandidateArray import GraspCandidateArray
 from dimos.msgs.sensor_msgs.JointState import JointState
 from dimos.msgs.sensor_msgs.PointCloud2 import PointCloud2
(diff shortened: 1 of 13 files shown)


dimos/manipulation/manipulation_module.py
```diff
@@ -182,6 +183,8 @@ class ManipulationModule(Module):
     # Input: occupied cells of a mapped workspace, in the planning frame. Each
     # message is a complete map, so it replaces the obstacle rather than adding.
     voxel_map: In[PointCloud2]
+    # Display only: what the grasp generator proposed, and in what order.
+    grasp_candidates: In[GraspCandidateArray]
```
TomCC7, line 184: why not rpc tho? this is not a continuous stream

mustafab0: Yes you are right. These would be intermittent.

Making it an rpc 👍

TomCC7 (dismissed):

paul-nechifor (approved):

PR #3890, opened 2026-09-02
generic lidar relocalization and tuning system

leshy:
```sh
dimos run relocalize-mid360 --dataset go2_mid360_sf_office_outdoors_2026-05-29 --map-file go2_mid360_sf_office_outdoors_2026-05-29
dimos run relocalize-mid360 --dataset go2_mid360_china_office_2026-06-12 --map-file go2_mid360_china_office_2026-06-12
```

china dataset starts in an unmapped area so reloc takes a bit, this is on purpose, test against false positives
add `--skip 60` not to wait

tuning guide for new lidars

https://github.com/dimensionalOS/dimos/blob/c48057f3acf896b194a4289c368b4ea617bde7a6/dimos/mapping/relocalization/lidar/tune.md

`tune.py` is vibed and tldr


leshy:
@greptileai recheck plz

dimos/mapping/relocalization/lidar/module.py
```diff
@@ -0,0 +1,117 @@
...
+
+logger = setup_logger()
+
+
+class LidarConfig(Config):
+    reloc_interval: float = 2.0
+    # Sanity check, skip a cloud too sparse to be worth a match,
+    # A mid360 sweep is only ~3k points and two of them voxel down to ~3.5k
+    # which is already enough to relocalize
+    min_local_points: int = 2_000
```
AaryanAgrawal, line 63: this seems kinda low

this is downsampled even more with coarse voxels before RANSAC runs on it. effectively RANSAC runs on ~100 voxels. intuition says this wont be robust against feature-sparse environments

what do you think?

note: this was at 50k before (which was prolly too high, it took about 40s in the office teleoping around to relocalize)

dimos/mapping/relocalization/lidar/tune.py
```diff
@@ -0,0 +1,974 @@
...
+    last_ts = t0 + start
+    # A cloud with no pose is skipped, not counted: n_frames is a number of
+    # registered scans, which is what the window question is about.
+    for obs in islice(frames, n_frames * 4):
+        if limit is not None and obs.ts > limit:
+            break
+        if obs.pose_tuple is None:
+            continue
+        x, y, z = obs.pose_tuple[:3]
+        mapper.add_frame_world(obs.data.points_f32(), (x, y, z))
```
AaryanAgrawal, line 339: made for fastlio. doesn't rotate the cloud, so doesnt work for any sensor frame recording (pointlio_lidar / mid360_link, which is what we record today)

leshy: yes, didn't have a good tf dataset when implementing, wrapping this now and delegating for followup cleanup

dimos/mapping/relocalization/lidar/module.py
leshy: no intuitions, we use data, these are tuned values with lowest compute cost and 100% reliability on current sets.

I don't think there is environment sparse enough to have an issue due to mid360 range, but if I'm wrong we will prove and retune. if you are wrong we permanently deterriorated performance and will never know.

reason ransac runs on low res map is that it's the most expensive algo here and is supposed to give a rough estimate of a possible match - ICP is here to match on high resolution and accept or reject

AaryanAgrawal: we should test in an open field. maybe today, I'll get a dataset for that

leshy: sounds great

aclauer (approved):

PR #3855, opened 2026-09-01
feat(imitation): add controlled OpenYAM policy rollout

TomCC7:
## Contribution path

- PRD: [Manipulation SDK preview, discussion #3923](https://github.com/dimensionalOS/dimos/discussions/3923).
- Stacked on #3854.

## Problem

- OpenYAM needs a runnable policy graph with optional Quest takeover.
- Robot composition should use the generic policy/control contract from its parent instead of introducing another learning-profile abstraction.

## Solution

- Add `openyam-lerobot-rollout` for Quest-free execution and `openyam-lerobot-quest-rollout` for optional takeover.
- Compose the LeRobot module, wrist RGB camera, and a named low-priority trajectory task with canonical OpenYAM joints.
- Reuse the existing Quest tasks for the takeover graph.
- Keep robot-specific joint ordering and camera defaults in this Blueprint; remove the obsolete `OpenYamLearningProfile` dependency.
- Own the dual-arm regression assertions that require hardware-aware trajectory acceptance/rejection; the underlying canonical hardware-limit setup lives in #3853.
- Inherit chunk execution, preflight, cancellation, button edges, and limits from #3315. This PR only adds robot wiring, tests, registry entries, and usage docs.

### API shape

```text
openyam-lerobot-rollout
  policy + wrist camera + ControlCoordinator

openyam-lerobot-quest-rollout
  policy + wrist camera + Quest teleop + TeleopControlCoordinator
```

- Configure either graph through normal `dimos run` module options.
- Python clients find `RolloutControlSpec` and explicitly call preflight/start/stop.
- The attached rollout panel is added later in #3931. Stopping a client is not a policy-stop request.

## How to Test

With a compatible checkpoint and configured arm/camera:

```bash
dimos --can-port follower_l run openyam-lerobot-rollout --policy.policy-path CHECKPOINT_DIR --policy.task "pick up the cube" --wristcamera.hardware.camera-index 0
```

- Start through `RolloutControlSpec` only after verifying preflight readiness and clearing the workspace.
- Use `openyam-lerobot-quest-rollout` to test Quest takeover.
- In the previous cleanup, 18 focused rollout/OpenYAM tests and scoped mypy passed; the generated registry is current.
- Current review cleanup: the dual-arm trajectory acceptance/rejection tests pass in the final 224-test focused suite. This layer adds no new rollout behavior during the cleanup.
- Bounded-motion rollout, takeover, and shutdown require a hardware retest. ABC and dual-arm rollout are out of scope.

[...]

PR #3921, opened 2026-09-03
feat(imitation): add direct OpenYAM teach collection

TomCC7:
## Contribution path

- Stacked on #3855.

## Problem

- Operators need to collect OpenYAM demonstrations by moving the arm and gripper by hand, without Quest.
- Teaching should be a robot preset over the shared recorder, not a separate collection framework or CLI.

## Solution

- Add `openyam-teach-collection` using the existing trajectory task's idle-hold behavior.
- Configure zero position stiffness, gravity feed-forward, and arm damping `(2, 2, 2, 0.5, 0.5, 0.5)`.
- Add Damiao passive-gripper configuration: retain opening calibration, then continuously command zero stiffness, damping, and feed-forward torque while reading measured opening.
- Define `OPENYAM_TEACH_COLLECTION` with measured seven-joint positions for both state and hand-guided action.
- Reuse the recorder factory, recording directory, and episode-control Spec from #3854.
- Keep Quest collection unchanged. The operator TUI belongs in #3931; no `dimos collect` command or learning-profile wrapper is introduced here.

### API shape

```text
openyam-teach-collection
  ControlCoordinator: zero-stiffness arm + passive gripper
  wrist:              ordinary CameraModule
  recorder:           collection_recorder(OPENYAM_TEACH_COLLECTION)
  episodes:           EpisodeMonitorModule
```

- Profile location: `dimos.robot.manipulators.openyam.collection`.
- Both measured projections use `source_kind="snapshot"`; teaching actions are measurements, not partial command updates.
- State and action project `coordinator_joint_state.position`; wrist RGB is the image observation.
- Device and recording settings use ordinary module options. Episode RPCs use `EpisodeControlSpec`.

## How to Test

With an arm and wrist RGB camera configured:

```bash
dimos --can-port follower_l run openyam-teach-collection --recorder.recording recordings/teach-001 --episodes.task "pick up the cube" --wrist.hardware.camera-index 0
```

- Control episodes through `EpisodeControlSpec`; support the arm before runtime shutdown.
- Confirm free hand movement of the calibrated gripper, suitable damping, and meaningful measured joint motion before collecting a dataset.
- In the previous cleanup: 78 focused Damiao, teaching-Blueprint, and episode-monitor tests; 6 registry checks; scoped mypy passed.
- Current review cleanup rebases this preset onto the source-kind contract without changing teaching behavior.
- The previous cleanup did not rerun physical teaching. Existing datasets were left untouched. All commit checks passed except the teaching worktree's LFS-upload check, skipped because of an unrelated pre-existing dataset.

## AI assistance

- Review cleanup updated by OpenAI Codex (agent).

## Checklist


dimos/cli/commands/collect.py
TomCC7, line 1: can we get a more fancy tui. (not sure what tui lib we have

dimos/control/tasks/teach_task/teach_task.py
TomCC7, line 1: is this task basically a noop? what's the point

dimos/cli/commands/collect.py
TomCC7: Implemented in 127ce282ef. This command already uses the repository standard Textual stack; it is now a full operator dashboard with a large READY/RECORDING state, elapsed timer, saved/discarded counters, contextual guidance, and clickable Start/Save, Discard, and Detach buttons alongside the existing keyboard shortcuts. The layout is verified at 80x24.

dimos/control/tasks/teach_task/teach_task.py
TomCC7: Agreed that the custom task was unnecessary. Its only system-level effect was to emit a command every tick so the whole-body adapter applies gravity feed-forward and damping. I removed it in 127ce282ef and reused the existing trajectory task with hold_position_when_idle=True. This teach blueprint hardcodes and tests kp=0 for every joint, so the latched target contributes no position torque; dataset actions still come from measured joint state.

PR #3924, opened 2026-09-03
feat(cli): add experimental Rust --record

TomCC7:
## Summary

- add explicit `--record-engine python|rust` selection while keeping Python SQLite as the default
- add MCAP support to `--record` when the experimental Rust engine is selected
- launch the existing `dimos-memory-recorder` binary with its existing stdin protocol; this PR does not change the Rust recorder or `dimos-module`
- validate MCAP/engine and encoding-thread combinations in `GlobalConfig`
- reuse the normal `recording(...)` context and existing coordinator/daemon lifecycle

## Staging

Rust remains opt-in because the recorder is still experimental. Bare `--record` continues to use Python SQLite. `--record mcap` requires `--record-engine rust`, and there is no automatic fallback to Python.

A Rust recorder startup failure aborts startup. If the recorder exits after readiness, DimOS logs the failure and keeps the rest of the run alive, matching the Python recorder failure policy.

## Validation

- 61 focused Python tests passed, including CI-style ANSI rendering
- Ruff and strict MyPy passed on the affected files
- Nix built the existing recorder package
- real `dimos-memory-recorder` binary recorded readable SQLite and MCAP artifacts through the new CLI controller over LCM: 2 passed
- full GitHub CI is green across Python 3.10-3.14, free-threaded 3.14, ARM, Rust, native builds, and both self-hosted suites

Depends on #3615.


CONTEXT.md
TomCC7, line 1: remove this file

dimos/cli/commands/lifecycle.py
```diff
@@ -154,6 +164,29 @@ def run(
...
+    encoding_threads_explicit = (
+        "RECORD_ENCODING_THREADS" in os.environ
+        or "DIMOS_RECORD_ENCODING_THREADS" in os.environ
+        or "record_encoding_threads" in global_option_overrides
+        or global_config.record_encoding_threads != 4
+        or any(
+            token == "--record-encoding-threads" or token.startswith("--record-encoding-threads=")
+            for token in config_tokens
+        )
+    )
```
TomCC7: ugly to detect env var in this way do we have better option in dimos?

dimos/cli/commands/lifecycle.py
```diff
@@ -154,6 +164,29 @@ def run(
...
+    if global_config.record_engine == "python" and encoding_threads_explicit:
+        typer.echo(
+            "Error: --record-encoding-threads is valid only with --record-engine rust",
+            err=True,
+        )
+        raise typer.Exit(2)
+    if global_config.record == "mcap" and global_config.record_engine == "python":
```
TomCC7: if global config is pydantic we should be use field validator

dimos/cli/commands/lifecycle.py
```diff
@@ -311,17 +355,20 @@ def run(
...
-        # default so Ctrl+C raises KeyboardInterrupt and the try/finally below
-        # runs with a visible traceback.
-        install_signal_handlers(entry, coordinator, sigint=False)
         try:
-            with recording(coordinator.transports):
-                coordinator.loop()
-        except Exception:
-            coordinator.stop()
-            raise
+            with recording(coordinator.transports) as record_session:
```
TomCC7: why is this change made what's special for rust recorder to need this?

dimos/core/daemon.py
```diff
@@ -113,6 +114,7 @@ def install_signal_handlers(
     coordinator: ModuleCoordinator,
     *,
     sigint: bool = True,
+    before_stop: Callable[[], None] | None = None,
```
TomCC7: these injections are generally ugly  need to see if we can avoid them

native/rust/dimos-module/src/lib.rs
TomCC7, line 1: should not need to change rust dimos module as we are not using it for -- record???

CONTEXT.md
TomCC7: Removed CONTEXT.md.

native/rust/dimos-module/src/lib.rs
TomCC7: Correct. I removed the extra binary and reverted every dimos-module and Rust recorder source change. The CLI adapter now launches the existing dimos-memory-recorder binary, which already uses run_with_transport and handles SIGTERM teardown.

dimos/cli/commands/lifecycle.py
TomCC7: Replaced the argv/env inspection. record_encoding_threads is now None when omitted, and GlobalConfig validation rejects an explicit value unless the Rust engine is selected.

dimos/cli/commands/lifecycle.py
TomCC7: Agreed that the recorder does not need a special coordinator lifecycle. I restored ModuleCoordinator.loop and the normal recording context. Startup failure remains fatal; an exit after readiness is logged while the run continues.

dimos/core/daemon.py
TomCC7: Removed the before_stop injection and restored the daemon signal handler unchanged. The existing recording context owns recorder shutdown.

dimos/cli/commands/lifecycle.py
TomCC7: Done. The MCAP/engine and encoding-thread constraints now live in GlobalConfig field validators, with model-level tests.

spomichter:
thanks for opening this against main - reviewing

dimos/core/global_config.py
```diff
@@ -137,6 +139,22 @@ class GlobalConfig(BaseSettings):
         validate_assignment=True,
     )
 
+    @field_validator("record_engine")
+    @classmethod
+    def _validate_record_engine(cls, value: str, info: ValidationInfo) -> str:
```
spomichter, line 144: nice

dimos/core/test_global_config.py
```diff
@@ -23,3 +26,29 @@ def test_listen_host_defaults_to_localhost(self) -> None:
         assert config.listen_host == "127.0.0.1", (
             f"listen_host must default to 127.0.0.1, got {config.listen_host}"
         )
+
+
+class TestRecordingConfig:
```
spomichter: i think repo convention is no test classes in pytests

spomichter (changes requested):

dimos/experimental/memory/rust_cli_recorder.py
```diff
@@ -0,0 +1,287 @@
...
+from dimos.utils.logging_config import setup_logger
+
+logger = setup_logger()
+
+_RUST_DIR = Path(__file__).resolve().parent / "rust"
+_EXECUTABLE = _RUST_DIR / "result/bin/dimos-memory-recorder"
+_BUILD_COMMAND = ("nix", "build", "-L", ".#dimos-memory-recorder")
+_READY_TIMEOUT = 10.0
+_DEFAULT_ENCODING_THREADS = 4
+_READY_MESSAGE = "memory recorder ready"
```
spomichter, line 46: i dont have enough context on this stuff. CC @aclauer  @leshy

dimos/memory/tap.py
```diff
@@ -14,9 +14,9 @@
 
 """``--record``: subscribe to every stream transport in the blueprint and write one store.
 
-Runs in the ``dimos run`` process. Each stream is tapped on its own transport (LCM,
-Zenoh, SHM, ...), so whatever carries it is what gets recorded. One ``memory.db`` per
-run at ``recordings/<run-id>/``, streams named by their blueprint stream name.
+The stable Python SQLite engine runs in the ``dimos run`` process and taps each
```
spomichter: remove AI diff comments due to unneeded added diff

docs/usage/recording.md
```diff
@@ -1,13 +1,45 @@
 # Recording
 
-`--record` writes every stream a blueprint publishes to one SQLite memory store. Each stream is tapped on the transport that carries it.
```
spomichter, line 3: all good comment changes

dimos/experimental/memory/rust_cli_recorder.py
spomichter, line 46: these RUST DIR references means that this wont work with a library pip install through,.

dimos/experimental/memory/rust_cli_recorder.py
```diff
@@ -0,0 +1,287 @@
...
+        port = f"stream_{index}"
+        streams.append(
+            RustStreamSpec(
+                port=port,
+                name=name,
+                payload_type=f"{payload_type.__module__}.{payload_type.__qualname__}",
+                codec="jpeg" if issubclass(payload_type, Image) else "lcm",
+            )
+        )
+        topics[port] = transport.channel
```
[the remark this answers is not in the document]
spomichter: super marginal

dimos/memory/tap.py
spomichter: done

spomichter (approved):

PR #3853, opened 2026-09-01
fix(openyam): align learning stack with canonical model

TomCC7:
## Contribution path

- PRD: [Manipulation SDK preview, discussion #3923](https://github.com/dimensionalOS/dimos/discussions/3923).
- Stacked on webcam prerequisite #4159.

## Problem

- Learning, teleoperation, and hardware must agree on OpenYAM joint names, home pose, and limits.
- IK must not solve using joints that the selected controller cannot command.

## Solution

- Define canonical arm/gripper joint names in a hardware-independent module, shared by robot configuration and later collection/rollout presets.
- Align single- and dual-arm home configurations with the canonical model.
- Reduce Pink's model to controlled joints and apply configured velocity limits.
- Add OpenYAM-specific teleop weighting and regression coverage.
- Keep recorder, policy, and CLI concerns in later layers.
- Own the dependency-free dual-arm joint declarations and import them directly from `dual_openyam.joints`; remove config reexports.
- Add robot-local `build_dual_openyam_webxr(visualization=...)` to construct the full teleop graph without extracting Blueprint atom kwargs.
- Load dual-arm hardware limits from the canonical model and report Viser execution acceptance/rejection.

- Keep lazy model paths lazy during configuration copying and serialization introspection; preserve normal resolution on asset access.
- Keep canonical-joint unit tests independent of model downloads and complete the pygame sequence type stub.

### API shape

- `dimos.robot.manipulators.openyam.joints` exports `OPENYAM_ARM_JOINTS`, `OPENYAM_GRIPPER_JOINT`, and `OPENYAM_JOINTS`.
- `dimos.robot.manipulators.dual_openyam.joints` owns the ordered `DUAL_OPENYAM_ARM_JOINTS`, `DUAL_OPENYAM_GRIPPER_JOINTS`, and `DUAL_OPENYAM_JOINTS` declarations.
- Robot configuration owns hardware/model setup; there is no separate learning-profile model.
- `build_dual_openyam_webxr(*, visualization: ManipulationVisualizationConfig = ViserVisualizationConfig()) -> Blueprint` accepts the existing visualization config. Ordinary teleop remains on `127.0.0.1`; callers can construct a graph with a different host. No generic Blueprint `.configure()` API.

## How to Test

With OpenYAM and Quest configured:

```bash
dimos --can-port follower_l run teleop-webxr-openyam
```

- In the previous cleanup, 22 focused OpenYAM/config tests passed; the final stack also passed 52 Pink/OpenYAM IK tests.
[...]

PR #3929, opened 2026-09-03
fix(openyam): use canonical model description

TomCC7:
## Summary

- update `yam_description` to the corrected archive already published in #3853
- load the canonical `i2rt/yam.urdf` model and map its joints to the OpenYAM hardware names
- target the canonical `gripper_tip` frame from keyboard and single-arm Quest teleoperation

This isolates the model-description correction needed to validate #3465 on macOS without pulling in the unrelated learning-stack changes from #3853.

## Stack

- depends on #3465
- model asset and canonical mapping extracted from #3853

## Validation

- 12 focused OpenYAM tests
- 1 self-hosted OpenYAM model-loading test
- corrected LFS object SHA-256 verified
- canonical model loaded directly from the corrected archive
- Ruff format and lint
- pre-commit hooks (the LFS data discovery hook was skipped because unrelated extracted assets are present in the shared worktree)


TomCC7:
Closing this follow-up; the requested yam_description pointer update will be applied directly to #3465 instead.

PR #3930, opened 2026-09-04
fix(zenoh): bootstrap peer workers deterministically

TomCC7:
## Contribution path

## Problem

Each worker owns a separate Zenoh peer session. Local startup relied on multicast discovery, so synchronous lifecycle RPCs could run before a worker discovered the other peers and declared its queryables. macOS exposes this timing window consistently enough to make CI flaky. A sleep changes the odds but cannot establish readiness.

## Solution

An unconfigured peer-mode coordinator now starts one ordinary Zenoh peer on a random loopback port before its worker pool. `ZenohPeerSeed` temporarily publishes that endpoint through the existing `GlobalConfig.zenoh_connect` path. Workers already receive `GlobalConfig` in `DeployModuleRequest`, so this needs no Zenoh-specific worker bootstrap protocol and no changes to `python_worker.py` or `worker_manager_python.py`.

Workers dial the seed and keep loopback listeners. Zenoh gossip introduces them, after which traffic flows directly between peers. The seed is a rendezvous peer, not a router.

Synchronous Zenoh RPC calls now declare a temporary `Querier`, wait for a matching queryable under the call's existing deadline, release the matching resources, and use the existing request/reply implementation. Callback and fire-and-forget calls are unchanged.

The run registry records the seed endpoint so a separate local CLI process can join a daemon whose multicast discovery is disabled.

## Remote modules

`--robot-ip`, `--robot-ips`, and `--zenoh-connect` remain authoritative. Any of them suppresses the automatic local seed, and sessions dial the configured remote peer or router as before. This PR therefore preserves both peer-to-peer remote deployments and router-backed deployments.

## Scope

The runtime change touches six production files:

- Zenoh session/seed and synchronous RPC readiness
- coordinator resource lifecycle
- run-registry publication and CLI attachment

Shared worker startup, worker management, and global configuration remain unchanged from `main`. The independent shared-memory race fix moved to draft PR #3946.

## macOS and CI coverage

- A required `macos-14` job disables multicast, launches a real daemon and workers, calls `Coordinator/ping` from a separate process, and repeats the test five times.
- The required self-hosted macOS matrix is enabled again.
- Damiao adapter tests mock the platform-neutral `_make_can_bus()` seam, so motor-control tests do not assume Linux SocketCAN.
- Transport-matrix tests start an isolated peer fabric before standalone probes and workers.

## Validation

- focused Zenoh service, coordinator, and RPC tests: 108 passed
- LCM/Zenoh async-module transport matrix: 24 passed
- daemon stress command used by macOS CI: 5/5 passed
- Damiao adapter tests: 6 passed across default and self-hosted marker selections
- broad default suite before interruption: 1,086 passed; its sole failure used system Python 3.14 instead of the repo venv, and passed when rerun with the venv activated
- mypy on all changed production modules: passed
- Ruff check/format and all pre-commit hooks: passed

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -607,6 +607,49 @@ jobs:
           use_pypi: true
           version: "11.2.8"
 
+  macos-zenoh-e2e:
+    timeout-minutes: 15
+    runs-on: macos-14
+    permissions:
+      contents: read
+    env:
+      PYTHONFAULTHANDLER: "1"
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+      - name: Install uv
+        uses: astral-sh/setup-uv@v10.0.1
+        with:
+          enable-cache: true
+          prune-cache: true
+      - name: Setup Python
+        uses: actions/setup-python@v7.0.0
+        with:
+          python-version: "3.12"
+      - name: Install PortAudio
+        run: brew install portaudio
+      - name: Install dependencies
+        run: uv sync --group tests --frozen
+      - name: Exercise deterministic Zenoh startup repeatedly
+        shell: bash
+        run: |
+          for attempt in {1..5}; do
+            echo "::group::attempt ${attempt}"
+            uv run pytest dimos/core/test_daemon_zenoh.py --no-cov -q
+            echo "::endgroup::"
+          done
+      - name: Print crash diagnostics
+        if: failure()
+        shell: bash
+        run: |
+          shopt -s nullglob
+          for f in "$RUNNER_TEMP"/pytest-crash/*.log; do
+            echo "::group::$(basename "$f")"
+            cat "$f"
+            echo "::endgroup::"
+          done
+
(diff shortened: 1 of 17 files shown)


jetsonearth:
Independently hit this on my Mac running teleop-quest-r1pro from cc/feat/r1pro-teleop: every run died with TimeoutError: RPC call to 'HeadsetArmTeleopModule/set_transport' timed out after 120.0s. Confirmed root cause matches - zenoh's scout beacons never send on macOS (socket bound to 127.0.0.1 without IP_MULTICAST_IF, kernel routes multicast to en0, rejects with EADDRNOTAVAIL / os error 49), so sibling processes never discover each other.

Actually had this issue awhile ago when working on A1Z

Applied this branch's changes locally - discovery now bootstraps deterministically via the loopback seed, modules wire in under a second, sim runs clean.

TomCC7:
should be fixed by #3963

PR #3941, opened 2026-09-04
Fix minimum uv version for cooldown opt-outs

omarespejel:
uv sync fails on uv versions before 0.9.25 because the cooldown opt-outs from #2231 use syntax they cant parse.

This PR sets 0.9.25 as the minimum and notes it next to the install command. The existing required-version doesnt help here since parsing fails before uv can read it.

Tested with 0.9.24, 0.9.25 and 0.12.9. No lockfile changes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -9,7 +9,7 @@ The agentic operating system for generalist robotics. `Modules` communicate via 
 ## Quick Start
 
 ```bash
-# Install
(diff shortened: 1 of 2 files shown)


omarespejel:
do your think @greptile

paul-nechifor (approved):

omarespejel:
Thanks @paul-nechifor !

PR #3957, opened 2026-09-06
fix(memory): subscribe Zenoh recording streams concurrently

omarespejel:
Addresses #3926

Recording subscribes to streams one at a time, so a slow Zenoh connection delays every stream after it.

This PR starts adjacent Zenoh subscriptions in parallel, up to 32 at a time. If startup fails, it removes any subscriptions already created and closes the recorder and database. Other transports stay serial, and Rust recording is unchanged.

I ran both versions five times locally with 22 streams and connection timeouts. Median startup fell from 5.29s to 0.82s. After startup, each stream sent one message and all 22 were saved in every run.

### DIFF
--- a/dimos/memory/tap.py
+++ b/dimos/memory/tap.py
@@ -24,8 +24,10 @@ Poses are not resolved here; ``tf`` is recorded like any other stream and
 from __future__ import annotations
 
 from collections.abc import Callable, Iterable, Iterator, Mapping
-from contextlib import contextmanager
+from contextlib import ExitStack, contextmanager
 import fnmatch
+from functools import partial
+from itertools import groupby
 import os
 from pathlib import Path
 import queue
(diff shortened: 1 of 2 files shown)


omarespejel:
@greptile do your magic and tell me if thisi is relevant or not

omarespejel:
@greptile thanks, and i confirm --record mcap can write SQLite instead. But this also happens on the base branch so it isn’t introduced here. I can fix it in a separate PR to keep this one focused on recording startup, wdyt?

omarespejel:
@paul-nechifor when possible can you review 🙏?

omarespejel:
hey @bogwi @Dreamsorcerer , this touches the recorder side of Zenoh startup. I tested it with 1.10.0 too. Can you take a look when you get a chance?

omarespejel:
Zenoh 1.10.1 removed the repeated connection wait this was fixing.

Issue #3594, opened 2026-08-21
Proposal: privileged ground-truth channel for interactive eval scoring (sim GT poses → evals framework)


Jerrybery:
### Summary

The evals framework landed in #3411 scores `InteractiveEval` by sampling the live memory2 `Store` — i.e. state the runtime/agent itself produced. As @TomCC7 noted in review, that conflates the agent's perception with the world: scoring should be able to use **privileged simulator ground truth** (object poses, robot pose, contacts). This issue proposes a minimal channel for that, with a working prototype and measured evidence for why agent-side state is insufficient.

### Motivating evidence (measured, not hypothetical)

While evaluating an xArm7 pick in the MuJoCo sim, we instrumented `MujocoSimModule` to publish free-joint body ground truth (details below) and ran the agent on "pick up the cup":

- The `pick` skill returned `Pick complete — grasped 'sphere' successfully`.
- Ground truth showed the approach/retract knocked **all three** scene objects off the table: orange z 0.175→0.045, cup z 0.190→0.035, apple z 0.170→0.040.

From memory2 / skill results alone this case scores as a success; against ground truth it is a catastrophic failure. Agent-side scoring also cannot express "was the cup lifted" at all when perception misses the object (our agent concluded "no cup is present in the scene" while the cup sat untouched). Related sim/planning consistency issue filed separately as #3484.

### Proposal

Two small, separable pieces:

**1. Sim modules publish ground truth on a privileged channel (opt-in).**

`MujocoSimModule` already owns `MjData`; publishing free-joint scene bodies costs ~40 lines. Prototype shape we validated:

- Resolve free-joint bodies at start (exclude the robot root, already published as `odom`).
- In the post-step hook, publish each body's world pose as `PoseStamped` with `frame_id = body name` on a dedicated stream (e.g. `/sim/gt/object_poses`), throttled (we used 20 Hz; per-step would mirror `odom`).
- Behind a config flag (e.g. `publish_ground_truth: bool = False`) so production/real-robot setups are unaffected.

Prototype ran 700 s+ stable on zenoh, 3 objects @ 20 Hz, negligible overhead.

**2. The evals framework lets scorers read that channel.**

`InteractiveEval.score` currently receives only the agent's `Store`. Options, in order of our preference:

- a) `EvalRig` gains a `gt_store()` / GT accessor, fed by a Recorder subscribed to the privileged GT streams — scorer signature becomes `score(store, gt)`, keeping today's `score(store)` as sugar;
[...]

PR #3960, opened 2026-09-07
feat(simulation): opt-in ground-truth object pose stream for eval scoring

Jerrybery:
## Summary

Part 1 of 2 for #3594 (privileged ground-truth channel for interactive eval scoring).

`MujocoSimModule` gains an opt-in ground-truth stream:

- `publish_ground_truth: bool = False` (default off — zero overhead, zero behavior change) and `ground_truth_hz: float = 20.0` config fields
- New `gt_object_poses: Out[PoseStamped]` stream: world poses of every free-joint scene body, throttled, `frame_id` = body name
- Robot root is excluded (its pose already flows on `odom`); resolved by matching the same root qpos address `odom` uses
- Leading `/` from MuJoCo 3.10's attached-submodel namespacing is stripped so `frame_id` stays a stable plain body name

Ground truth is privileged scoring data — never consumed by the agent. Part 2 will add the eval-side `gt_store()` accessor, a predicate library, and an xArm7 regression suite.

## Test plan

- 3 new tests in `test_mujoco_sim_module.py`: disabled-by-default publishes nothing; enabled publishes throttled poses; free-joint body resolution excludes the robot root against a real composed model
- End-to-end smoke (not in repo): real blueprint run with a 2-object scene — 392 msgs / 10 s across both bodies with correct frame_ids, 0 msgs when disabled
- Verified on Linux with the lockfile-pinned MuJoCo 3.10.0

Note for part 2 consumers: `async def handle_<stream>` auto-subscription uses a single-slot LATEST mailbox (`module.py`), so same-tick bursts overwrite; GT consumers must use direct `subscribe()` (as `Recorder` does).

### DIFF
--- a/dimos/simulation/engines/mujoco_sim_module.py
+++ b/dimos/simulation/engines/mujoco_sim_module.py
@@ -18,6 +18,8 @@ Owns a single ``MujocoEngine`` and publishes:
 - camera streams (Out ports), replacing ``MujocoCamera``
 - joint state via shared memory, consumed by ``ShmMujocoAdapter`` inside
   ``ControlCoordinator``
+- optionally, ground-truth world poses of free-joint scene bodies
+  (``publish_ground_truth``), for eval scoring only
 
 This avoids the prior pattern of sharing engines via a global in-process
 registry, which was fragile when ``WorkerManager`` places the adapter and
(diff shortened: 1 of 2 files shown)


Jerrybery:
Part 2 of the #3594 split is implemented and pushed as a stacked branch on my fork: [`Jerrybery:feat/evals-gt-store`](https://github.com/Jerrybery/dimos/compare/feat/sim-gt-object-poses...feat/evals-gt-store) (branched off this PR's head).

It adds the eval-side consumption of this PR's GT stream:

- `GTRecorder` (Recorder subclass) records `/gt_object_poses` into a per-case db, kept separate from the agent-visible recording
- `EvalRunner.gt_store()` + two-arg `score(store, gt)` scorers, plus a `--mujoco-publish-ground-truth` global flag so any mujoco blueprint emits GT during eval runs
- `dimos/evals/predicates.py` spatial predicates (`grasped()` kept as a placeholder until the GT stream carries contact data)
- `suites/xarm7_tabletop.py` regression suite

I'll open it as a PR once this one lands. Local verification: `pytest dimos/evals/` 125 passed.

Note: the current `tests (3.12)` CI failure on this PR is a flaky agent-test issue unrelated to these changes (all three tests passed on the in-job `--lf` re-run, and the other Python-version jobs are green) — filed as #3962.

PR #3963, opened 2026-09-07
bump zenoh to 1.10.1; use zenoh upstream: Fix multicast scouting mac

bogwi:
## Problem

fix multicast issue on mac, as reported by @jetsonearth and @TomCC7 

"zenoh's multicast scouting can't send on macOS (binds to 127.0.0.1 without IP_MULTICAST_IF, kernel rejects with EADDRNOTAVAIL), so local workers never discover each other and every run dies with a 120s set_transport timeout. "

## Solution

bump zenoh to 1.10.0  where this issue is solved

https://github.com/eclipse-zenoh/zenoh/pull/2671

Port scouting tests from upstream zenoh

## UPD
bump zenoh to 1.10.1, to mitigate this bug causing CI failures,
https://github.com/eclipse-zenoh/zenoh/pull/2759

## Additional

since we are bumping zenoh to 1.10.1 - ensure we are not breaking cause the bump may surface new issues

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -519,6 +519,8 @@ jobs:
       # redirects it to a per-process file under $RUNNER_TEMP/pytest-crash,
       # which is what makes a crash legible under xdist (one file per worker).
       PYTHONFAULTHANDLER: "1"
+      # Zenoh's own scouting decisions, for diagnosing workers that never link.
+      RUST_LOG: "zenoh::net::runtime::orchestrator=debug"
     permissions:
       contents: read  # For checkout
       id-token: write  # For codecov-action's OIDC upload
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -640,11 +640,10 @@ jobs:
                 - /var/cache/dimos-root-cache:/root/.cache
             markers: "self_hosted or skipif_no_ros"
             experimental: false
-          # macOS runner disabled for now — we don't want Mac tests.
-          # - os: macOS
-          #   container: null  # run on host — `container:` is Linux-only
-          #   markers: "self_hosted"
-          #   experimental: true
+          - os: macOS
+            container: null  # run on host — `container:` is Linux-only
+            markers: "self_hosted"
+            experimental: true
     runs-on:
       - self-hosted
       - ${{ matrix.os }}
--- a/dimos/protocol/service/test_zenoh_loopback_scouting.py
+++ b/dimos/protocol/service/test_zenoh_loopback_scouting.py
@@ -17,17 +17,11 @@
 Upstream: Fix multicast scouting on loopback#2671.
 """
 
+from collections.abc import Callable
 import json
-import socket
 import time
 
 import zenoh
-
-
-def _free_udp_port() -> int:
-    with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as sock:
-        sock.bind(("127.0.0.1", 0))
-        return int(sock.getsockname()[1])
 
 
 def _config(**settings: object) -> zenoh.Config:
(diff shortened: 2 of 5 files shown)


dimos/protocol/service/test_zenoh_loopback_scouting.py
```diff
@@ -0,0 +1,88 @@
...
+    address = f"224.0.0.224:{_free_udp_port()}"
+    responder = zenoh.open(
+        _config(
+            mode="router",
+            listen__endpoints=[],
+            scouting__multicast__autoconnect=[],
+            **_loopback_multicast(address),
+        )
+    )
+    scout = zenoh.scout(what="router", config=_config(**_loopback_multicast(address)))
```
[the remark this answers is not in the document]
bogwi: If zenoh.scout() raises here, - zenoh is broken and this whole file is red anyway

dimos/protocol/service/test_zenoh_scouting.py
```diff
@@ -208,3 +212,28 @@ def test_zero_timeout_disables_the_wait(zenoh_defaults: None) -> None:
         _await_elapsed(_FakeSession([]), connect=["tcp/192.0.2.199:7447"], connect_timeout=0.0)
         < 1.0
     )
+
+
+def _free_udp_port() -> int:
```
Dreamsorcerer: There's fixtures for this:
https://pytest-asyncio.readthedocs.io/en/stable/reference/fixtures/index.html#unused-udp-port-and-unused-udp-port-factory

From experience, it's safer to use unused_udp_port_factory if possible, as the unused_udp_port can cause race conditions with parallel tests (same is true with this implementation).

dimos/protocol/service/test_zenoh_scouting.py
```diff
@@ -154,7 +158,7 @@ def _await_elapsed(
     """Seconds _await_connect blocks against a session with the given links."""
     service = ZenohService(mode=mode, connect=connect, connect_timeout=connect_timeout)
     started = time.monotonic()
-    service._await_connect(session)
+    service._await_connect(cast("zenoh.Session", session))
```
Dreamsorcerer: As a general rule we should avoid casts. If we can't type this correctly now, an ignore comment is better than a cast().

A cast is simply a lie to the type checker. If something changes in future which breaks it in a different way, then the type checker won't report a new error (and it's less likely to report when the existing error is fixed incidentally too).

If you wanted to look at fixing the error, a better approach is probably using something like `mock.create_autospec(zenoh.Session, spec_set=True, instance=True, info=...)` for the fake session.

.github/workflows/ci.yml
```diff
@@ -640,11 +640,10 @@ jobs:
...
             experimental: false
-          # macOS runner disabled for now — we don't want Mac tests.
-          # - os: macOS
-          #   container: null  # run on host — `container:` is Linux-only
-          #   markers: "self_hosted"
-          #   experimental: true
+          - os: macOS
+            container: null  # run on host — `container:` is Linux-only
+            markers: "self_hosted"
+            experimental: true
```
Dreamsorcerer, line 648: If the tests all end up passing, then this needs to become mandatory.
```suggestion
            experimental: false
```

bogwi: mac test fails are unrelated to the bump, https://github.com/dimensionalOS/dimos/actions/runs/34141424106/job/101804529020?pr=3963

Dreamsorcerer: Seems there's some new failures, but the Zenoh parts are fixed.
Let's merge this with experimental true, then I'll take a look at fixing the nix failure tomorrow.

Dreamsorcerer (dismissed):

Dreamsorcerer (approved):

PR #3777, opened 2026-08-31
Yimo/feat/host daemon

poorwym:
## Summary

This PR adds a minimal DimOS Host daemon that allows a machine to register as a Host over Zenoh, be discovered remotely, and execute deployed Blueprints.

Highlights:

- Add a minimal `HostDaemon` with `describe/start/status/stop` operations
- Allow at most one active deployment per Host
- Identify each machine with a persistent `host_id`
- Use a file lock to enforce a single `dimos host serve` instance per machine
- Discover online Hosts through Zenoh liveliness
- Resolve a Host by ID or unique exact name
- Diagnose Host identity, Zenoh configuration, connectivity, and version
- Validate multicast and cross-router network topologies
- Validate remote Blueprint startup, RPC calls, error propagation, and cleanup

## Host CLI

```text
Usage: dimos host [OPTIONS] COMMAND [ARGS]...

Run and inspect DimOS Hosts

Commands:
  id        Show this machine's persistent Host ID.
  list      List Hosts currently visible through Zenoh liveliness.
  describe  Describe one online Host by ID or unique exact name.
  doctor    Check the local Host identity, Zenoh configuration,
            connection, and version.
  serve     Serve one Host over the configured Zenoh fabric.
```

The complete command set is:

| Command | Description | Options |
|---|---|---|
| `dimos host id` | Show or create this machine's persistent Host ID | — |
| `dimos host list` | List Hosts currently visible through Zenoh liveliness | `--json`, `--timeout` |
| `dimos host describe <host>` | Get the full descriptor by Host ID or unique exact name | `--json`, `--timeout` |
| `dimos host doctor` | Check the Host ID, Zenoh configuration, connection, and DimOS version | — |
| `dimos host serve` | Start the Host daemon and publish its liveliness token | `--name`, `--tag/-t` |

Examples:

```bash
dimos host id
dimos host list
dimos host list --json --timeout 5
dimos host describe <host-id>
dimos host describe <unique-name> --json
dimos host doctor
dimos host serve --name lab-host --tag gpu --tag arm
```

`dimos host serve` uses a file lock to enforce a single Host service per machine. Starting it again fails immediately:

```text
Error: Host service is already running on this machine
```

## Host lifecycle

A Host announces its availability with the following Zenoh liveliness key:

```text
dimos/hosts/<host_id>/live
```

Each Host exposes the following RPC endpoints:

```text
hosts/<host_id>/describe
hosts/<host_id>/start
hosts/<host_id>/status
[...]

PR #3692, opened 2026-08-26
Yimo/feat/hosted

poorwym:
This PR introduces the basic API design for hosted support. It is an initial implementation intended to establish a foundation and is open for public discussion and further iteration. @spomichter @leshy

PR #3967, opened 2026-09-08
Yimo/feat/host fragment

poorwym:
This PR adds the core infrastructure for deploying a DimOS Blueprint across multiple Hosts:

1. Use `.hosted()` to describe where a module should run;
2. Automatically place modules based on Host name, tags, state, load, and code revision;
3. Compile one Blueprint into multiple immutable `HostFragment` objects;
4. Route cross-Host streams through run-scoped Zenoh topics;
5. Support run/Host-scoped module RPC and coordinator RPC;
6. Start each Fragment in an isolated child process and manage its start, status, stop, and logs.

## `.hosted()` API

```python
app = autoconnect(
    Camera.blueprint().hosted(host="edge-a"),
    Detector.blueprint().hosted(tags={"gpu"}),
    ResultSink.blueprint(),  # No constraint: stays on the controller
)
```

```python
blueprint.hosted()                    # Any available remote Host
blueprint.hosted(host="edge-a")     # Exact Host ID or unique name
blueprint.hosted(tags={"gpu"})       # Host must contain all tags
blueprint.hosted(local=True)          # Force placement on the controller
```

`.hosted()` can be applied to a single module or a composed Blueprint. A composed Blueprint becomes one placement unit and is assigned to a single Host.

## Core Behavior

- Modules without placement constraints stay on the controller, preserving existing single-Host behavior;
- Cross-Host streams use `dimos/runs/<run-id>/streams/<stream>/<message-type>`;
- Each Fragment is bound to its `run_id`, `generation`, `host_id`, application revision, and SHA-256 digest;
- A Host validates the target Host, epoch, schema, digest, and code revision;
- Starting the same Fragment again is idempotent; stopping requires a matching run, generation, and digest;
- Automatic placement co-locates modules connected by module references. Explicit assignments can still be used to exercise cross-Host module RPC.

This PR does not include code distribution, failover, migration, replicas, or resource-aware scheduling. The standard `dimos run` command is also not yet connected to the full discovery → placement → dispatch workflow.

---

# Validation on `yimo/temp/host_e2e`

This branch adds a Host client, scenario runner, and Docker Compose environment on top of the feature PR.

## 1. Switch Branches

```bash
git switch yimo/temp/host_e2e
```

## 2. Run the Three Automated Scenarios

```bash
experimental/hosted/compose/run.sh all
```

| Scenario | Module distribution | What it validates |
| --- | --- | --- |
| `basic` | source → `edge-a`; transform → `compute-a`; sink → controller | Exact name, tags, local default, and cross-Host streams |
| `replay` | replay → `replay-a`; transform → `compute-a`; sink → controller | SQLite Replay across Hosts |
| `sim` | MuJoCo → `sim-a`; odom probe → `compute-a`; sink → controller | Real simulation and typed odometry across Hosts |

### Compose topology

```mermaid
flowchart LR
    C["verify-* / Controller"] -->|"discover + lifecycle RPC"| Z["Zenoh Router"]
    Z <--> E["edge-host\nedge-a / edge"]
    Z <--> G["compute-host\ncompute-a / compute,gpu"]
    Z <--> R["replay-host\nreplay-a / replay"]
    Z <--> S["sim-host\nsim-a / sim"]
    E <-->|"run-scoped streams"| G
    G <-->|"run-scoped streams"| C
    R <-->|"run-scoped streams"| G
    S <-->|"run-scoped streams"| G
```

The Compose file has three layers:

1. **Router:** one dedicated Zenoh router provides the shared discovery, RPC, and data fabric;
2. **Execution Hosts:** `edge-host`, `compute-host`, `replay-host`, and `sim-host` each run an independent `HostDaemon` with a stable name and tags;
3. **Scenario controller:** one profile-specific `verify-*` container owns the original Blueprint, discovers Hosts, resolves placement, compiles Fragments, and checks the final application result.

All services use the same locally built image, so every Host has the same code and application revision. They share the same runtime configuration: Zenoh transport, client mode, explicit connection to `tcp/router:7447`, two local workers, and no multicast discovery. `sim-host` and the simulation controllers also receive additional shared memory for MuJoCo.

### Control plane and data plane

The controller uses the router to discover Host liveliness tokens and call Host RPCs such as `describe`, `start`, and `stop`. This is the **control plane**.

After deployment, application messages do not pass through the controller as a proxy. The Fragment compiler pins every boundary stream to a run-scoped Zenoh key, and the modules exchange data directly through the router. This is the **data plane**.

```text
Controller Blueprint
  -> discover Host descriptors
  -> resolve .hosted() constraints
  -> compile one HostFragment per selected Host
  -> start consumer Fragments
  -> start producer Fragments
  -> wait for ValidationSink
  -> stop all Fragments
```

Starting consumers first prevents short-lived replay or source messages from being published before their remote subscribers are ready.

### Scenario module graphs

Each profile changes only the scenario controller; the router and four Host services keep the same topology.

```text
basic:
  BasicSource@edge-a
    -> BasicTransform@compute-a
    -> ValidationSink@controller

replay:
  ReplaySource@replay-a
    -> ReplayTransform@compute-a
    -> ValidationSink@controller

sim:
  MujocoSimModule@sim-a
    -> SimOdomProbe@compute-a
    -> ValidationSink@controller
```

`basic` exercises exact-name placement, tag placement, the local default, and two Host boundaries. `replay` replaces the synthetic source with a real SQLite-backed replay. `sim` replaces it with a real headless MuJoCo process and a typed `PoseStamped` odometry stream.

The `ValidationSink` is deliberately local and is the end-to-end assertion point. It writes a result only after all required values arrive. The controller then prints the resolved placement and application result, stops each Fragment using its run/generation/digest identity, and exits with the scenario status. The shell runner propagates that status through `--exit-code-from` and removes the Compose containers and network.

The scenarios can also be run individually:

```bash
experimental/hosted/compose/run.sh basic
experimental/hosted/compose/run.sh replay
experimental/hosted/compose/run.sh sim
```

## 3. Run the Visual Scenario

```bash
experimental/hosted/compose/run.sh visual
```

After `VISUAL:READY` appears, open:

```text
http://localhost:9878/?url=rerun%2Bhttp%3A%2F%2Flocalhost%3A9877%2Fproxy
```

The visual profile extends the simulation graph with a controller-local Rerun bridge:

```text
MovingMujocoSimModule@sim-a
  ├─ odom -> SimOdomProbe@compute-a -> ValidationSink@controller
  └─ odom -> RerunBridgeModule@controller -> gRPC :9877 -> Web Viewer :9878
```

MuJoCo updates the robot's real root pose and velocity on `sim-a`. The same run-scoped odometry stream is consumed by the probe for an automated assertion and by Rerun for visualization. Static ground and target geometry are logged by the bridge, while the robot body is attached to the live odometry transform.

Unlike the finite profiles, `visual` does not exit after validation. `VISUAL:READY` means Host discovery, placement, Fragment startup, cross-Host odometry, and the application-level assertion have already succeeded; the controller is kept alive so the browser can remain connected.

The blue robot should move continuously in a circle, with the yellow arrow rotating to show its direction of travel. The red target and ground remain fixed, making motion easy to verify.

image
image


experimental/docs/hosted/architecture.md
```diff
@@ -0,0 +1,755 @@
+# Distributed Host Architecture
```
paul-nechifor, line 1: This file exists in https://github.com/dimensionalOS/dimos/pull/3692/changes#diff-cf2dda3032e3e1fd5df6450def8d5e77d4086a3d89df2d0671d06e0eb51ebcc5

Shouldn't it be merged in that PR?

Also, it doesn't look like you implemented any of the changes discussed there.

paul-nechifor: And docs should be in `docs/` not `experimental/`

dimos/cli/commands/host.py
```diff
@@ -0,0 +1,367 @@
...
+DISCOVERY_KEY = "dimos/hosts/*/live"
+DEFAULT_DISCOVERY_TIMEOUT = 2.0
+
+
+def _load_host_id(path: Path) -> str:
+    try:
+        host_id = path.read_text().strip()
+    except FileNotFoundError:
+        path.parent.mkdir(parents=True, exist_ok=True)
+        host_id = uuid.uuid4().hex
```
paul-nechifor, line 50: Why create your own host id when you can use `/etc/machine-id`?

dimos/cli/commands/host.py
```diff
@@ -0,0 +1,367 @@
...
+
+
+@host_app.command()
+def serve(
+    name: str | None = typer.Option(None, "--name", help="Human-readable Host name"),
+    tags: list[str] = typer.Option([], "--tag", "-t", help="Placement tag; repeatable"),
+) -> None:
+    """Serve one Host over the configured Zenoh fabric."""
+    from dimos.hosted.daemon import (
+        FRAGMENT_SCHEMA_VERSION,
```
paul-nechifor: Tests and linting are breaking because of this.

dimos/cli/commands/host.py
```diff
@@ -0,0 +1,367 @@
...
+from __future__ import annotations
+
+from collections.abc import Callable, Iterator
+from contextlib import ExitStack, contextmanager
+import fcntl
+from importlib.metadata import version as package_version
+import json
+from pathlib import Path
+import threading
+from typing import TYPE_CHECKING, Any, Never, TextIO
```
paul-nechifor: Never doesn't exist in Python 3.10.

dimos/cli/commands/host.py
```diff
@@ -0,0 +1,367 @@
...
+        HOST_CONTROL_RPC_NAME,
+        HOST_LIVELINESS_KEY,
+        HOST_PROTOCOL_VERSION,
+        HostDaemon,
+    )
+    from dimos.protocol.rpc.zenohrpc import ZenohRPC
+    from dimos.protocol.service.zenohservice import ZenohSessionPool
+
+    try:
+        lock_file = _acquire_host_lock(HOST_LOCK_PATH)
```
paul-nechifor: Rather than creating your own file locking primitives, use `filelock`

dimos/core/coordination/blueprints.py
```diff
@@ -228,6 +239,40 @@ def global_config(self, **kwargs: Any) -> "Blueprint":
             global_config_overrides=MappingProxyType({**self.global_config_overrides, **kwargs}),
         )
 
+    def hosted(
+        self,
+        *,
+        host: str | None = None,
+        tags: Iterable[str] = (),
+        local: bool = False,
+    ) -> "Blueprint":
```
paul-nechifor: The problem with having both `host` and `local` is that you can say `.hosted(host="something", local=True)` in which case, what do you follow? Do you put it on host "something" or locally.

I think it would be clearer if you removed `local`. You can express it by `.hosted(host=None)`.

dimos/core/coordination/blueprints.py
```diff
@@ -196,6 +206,7 @@ class Blueprint:
 
     requirement_checks: tuple[Callable[[], str | None], ...] = field(default_factory=tuple)
     configurator_checks: "tuple[SystemConfigurator, ...]" = field(default_factory=tuple)
+    hosted_placements: tuple[HostedPlacement, ...] = field(default_factory=tuple)
```
paul-nechifor, line 222: This does not seem like the right structure because, among others, it allows for invalid states to be stored. For example, what if you have two HostedPlacements with the same host names?

module_names is a touple. What if you have the same module twice, or the same module in two different HostedPlacement?

This would be simpler (note that I've removed `local` because it's not needed as mentioned in the other commnet):

```python
class Placement
    host: str | None = None
    tags: frozenset[str] = frozenset()

class Blueprint:
    placements: Mapping[str, Placement] # instance name to Placement
```

If a module is not in placement, then it's local.

dimos/core/module.py
```diff
@@ -113,6 +113,9 @@ class ModuleConfig(BaseConfig):
     # once (see BlueprintAtom.instance_name). Changes the RPC topic prefix
     # from the class name to this name.
     instance_name: str | None = None
+    # Optional RPC-only address. Unlike instance_name, this does not change the
+    # module's Blueprint identity, configuration key, or stream namespace.
+    rpc_name: str | None = None
```
paul-nechifor, line 118: Correct me if I'm wrong, but the reason you're introducing this is because you need a key that's unique across the network so you need to include the machine id in this string, right?

If so, then this should not be called `rpc_name`, but something like `_machine_id` and the module should receive it from the coordinator.

And then introduce a `rpc_name` property that's computed from the instance name plus machine id.

dimos/hosted/daemon.py
```diff
@@ -0,0 +1,344 @@
...
+                host_id=self._host_id,
+                epoch=self._epoch,
+                name=self._name,
+                tags=self._tags,
+                versions=dict(self._versions),
+                state=self._host_state_locked(),
+                active_run_ids=tuple(sorted(self._deployments)),
+            )
+
+    @rpc
```
paul-nechifor: Why is this an rpc?

dimos/hosted/daemon.py
```diff
@@ -0,0 +1,344 @@
...
+from dimos.utils.logging_config import set_run_log_dir
+
+HostState = Literal["available", "starting", "running", "stopping", "failed"]
+
+HOST_PROTOCOL_VERSION = 2
+HOST_LIVELINESS_KEY = "dimos/hosts/{host_id}/live"
+HOST_CONTROL_RPC_NAME = "hosts/{host_id}"
+DEFAULT_STARTUP_TIMEOUT = 60.0
+DEFAULT_STOP_TIMEOUT = 5.0
+DEFAULT_LOG_ROOT = STATE_DIR / "hosted" / "runs"
```
paul-nechifor, line 49: There are other files which use `STATE_DIR / "hosted"`. You should probably make that a constant and reuse it.

dimos/hosted/daemon.py
```diff
@@ -0,0 +1,344 @@
...
+    try:
+        payload = fragment.load_payload()
+        remote_module_refs = {
+            (reference.consumer_name, reference.reference_name): (
+                reference.provider_type,
+                reference.rpc_name,
+            )
+            for reference in payload.remote_module_references
+        }
+        coordinator = ModuleCoordinator.build(
```
paul-nechifor, line 319: I think the order is wrong. You're splitting the original blueprint into fragments and then running the fragments with ModuleCoordinator, but ModuleCoordinator.build does things like `_verify_no_name_conflicts` (which checks that `ch: In[Str]` cannot connect to `ch: Out[Bool]`.

You have to change the order here to verify the blueprint first, then split it and allocate it to hosts.

If you do it like this, `ch: In[Str]` and `ch: Out[Bool]` can end up on different fragments and the conflict is never noticed... until you send data.

dimos/hosted/daemon.py
```diff
@@ -0,0 +1,344 @@
...
+@dataclass(slots=True)
+class _Deployment:
+    fragment: HostFragment
+    state: HostState
+    log_dir: Path
+    process: BaseProcess
+    error: str | None = None
+
+
+class HostDaemon:
```
paul-nechifor, line 82: We already have a daemon in `dimos`. This can be a bit confusing. I thought the idea was to use something else. You say:

> Each execution device runs a small, persistent **Host service**.

...and that it would be a standalone binary: https://github.com/dimensionalOS/dimos/pull/3692/changes#r3877532556

experimental/docs/hosted/architecture.md
poorwym: This is actually due to how the PRs are stacked. This PR is based on #3777, which still hasn’t been merged.

I’ve already made quite a bit of progress on the next part, so I opened this PR first mainly to show the current progress and get some early feedback. That’s also why some of the changes discussed there aren’t reflected in this PR yet.

dimos/cli/commands/host.py
poorwym: `/etc/machine-id` identifies the OS instance, not the DIMOS host instance. In particular, container replicas created from the same image may share the same machine ID, causing identity collisions. A persisted random UUID gives each host instance its own stable and independent identity.

I ran a small test in docker image with two replica:

```
-- replica-a --
container_id=18574ccf11efb4e60fab928ef94453d679042bf90a8cb3db84225ebd19b3deaf
hostname=replica-a
cgroup=0::/;
machine_id=0123456789abcdef0123456789abcdef
app_instance_id=55ee540c-ac5f-45be-99fd-6095d814e644
app_instance_id_state=generated

-- replica-b --
container_id=b007c582e49171e82be055600ea25ca10aa2f9655d438e9c58b624047400132c
hostname=replica-b
cgroup=0::/;
machine_id=0123456789abcdef0123456789abcdef
app_instance_id=6309109b-8584-4724-9df4-eb2e42311482
app_instance_id_state=generated
```

dimos/cli/commands/host.py
poorwym: Yes, so I fixed it in PR #3777. Sorry again for miscommunication.

dimos/hosted/daemon.py
poorwym: Yes you're right. I'll fix this.

poorwym: Thanks for pointing this out

dimos/core/module.py
poorwym: You’re right that the network-visible RPC identity should be owned by the coordinator rather than chosen by the module itself. That is exactly the concern I was trying to address here.
In the current implementation, the fragment compiler already has the complete deployment context, so it computes the canonical RPC address:
```
runs/{run_id}/hosts/{host_id}/modules/{module_name}
```
It then passes that final address to the module as `rpc_name`. The module only uses it to register its RPC handlers; it does not affect the module’s Blueprint identity, configuration key, or stream namespace.
The reason I pass the complete address instead of only a machine ID is that a machine ID plus instance_name is not always unique. For example, the same Host can run two separate runs containing a module with the same instance name. Both modules would then compute the same RPC name and collide unless the run_id were also included.
Computing the address inside Module would therefore require passing the full deployment context and would couple the base module class to the hosted RPC naming convention. I think keeping that policy in the coordinator/compiler and injecting the resulting address preserves the separation more cleanly.
That said, I can clarify the comment to make it more explicit that rpc_name is coordinator-assigned deployment metadata rather than a user-defined module identity.

dimos/hosted/daemon.py
poorwym: This is explained in https://github.com/dimensionalOS/dimos/pull/3777#discussion_r3954003411

dimos/core/coordination/blueprints.py
poorwym: You’re right. This is over-designed, and your suggestion is clearer. I’ll change it as you suggested.

dimos/hosted/daemon.py
paul-nechifor: But why not remove it?

poorwym: Oh, sorry I thought you just want to figure out the reason. I'll remove it.

dimos/hosted/daemon.py
poorwym: A standalone binary is definitely the end goal. However, achieving that now would require reimplementing the relevant type system, which feels out of scope for the MVP. The current Host daemon is already minimally coupled to DimOS and only depends on a few shared types. I’d be happy to make extracting it into a standalone binary a longer-term goal for hosted modules.

dimos/core/coordination/blueprints.py
poorwym: Good point about invalid states. I added validation to reject empty placement units, duplicate module names within a unit, and references to unknown module instances.

I kept the grouped placement representation because mapping each instance directly to a `Placement` would lose the distinction between one composed co-location unit and multiple independent units with identical constraints. A module may also appear in overlapping units through nested composition; those constraints are intentionally combined.

dimos/hosted/daemon.py
poorwym: Good idea!

PR #3973, opened 2026-09-08
Ivan readme codeowner

leshy:
Broken installer added as our main install path to README.MD so making sure this type of stuff doesn't happen with owning readme


paul-nechifor (approved):

leshy:
role change

PR #3974, opened 2026-09-08
fix(deps): pin rerun-sdk to the version dimos-viewer is built on

leshy:
`dimos-viewer==0.32.0a2` reports itself as `v0.32.0-alpha.1`; `rerun-sdk` was pinned to `0.32.0` final. `re_sdk`'s spawn check treats an alpha as incompatible with the release, so every spawn prints:

```
⚠ The version of the Rerun Viewer available on your PATH does not match the version of your Rerun SDK ⚠
> Rerun Viewer: v0.32.0-alpha.1 (executable: "dimos-viewer")
> Rerun SDK: v0.32.0
```

The SDK was moved from `0.32.0a1` to `0.32.0` in 74d6b9e6a (#2671) without moving the viewer with it. This moves it back so the two match.

`VideoCodec.VP8`/`VP9` only exist from 0.32.0 final (and the shipped viewer can't decode them anyway), so they're dropped from the `CompressedVideo.to_rerun` codec map — unsupported formats already raise a clear `ValueError`.

### DIFF
--- a/dimos/msgs/foxglove_msgs/CompressedVideo.py
+++ b/dimos/msgs/foxglove_msgs/CompressedVideo.py
@@ -92,8 +92,7 @@ class CompressedVideo(Timestamped):
             "h264": rr.VideoCodec.H264,
             "h265": rr.VideoCodec.H265,
             "av1": rr.VideoCodec.AV1,
-            "vp8": rr.VideoCodec.VP8,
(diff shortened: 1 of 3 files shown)


leshy:
closing as https://github.com/dimensionalOS/dimos/pull/2671 is more important but we need to update dimos-viewer

PR #3982, opened 2026-09-08
fix(manipulation): detect Viser execution completion and consume plans

TomCC7:
## Contribution path

Bug fix; no tracking issue.

## Problem

- Viser stays in `EXECUTING` after a trajectory finishes because nonblocking execution never refreshes its cached status. Planning again requires manually cancelling.
- The panel keeps the plan after dispatch.

## Solution

- Refresh active execution during status reads using the existing zero-timeout wait. This makes one synchronous coordinator RPC and allows planning again when execution completes.
- Clear the plan before dispatch, including on rejection or error. During motion, the plan is `NONE` and the module is `EXECUTING`.
- Remove the redundant `None` check for the execution manager, which is initialized at startup.

## How to Test

```bash
uv run dimos run dual-xarm6-planner-coordinator
```
drag the arms and execute the plan, see UI updates correctly when motion finishes.

## AI assistance

Codex (GPT-6) wrote the fix and regression tests, ran validation, and drafted this description.

## Checklist

### DIFF
--- a/dimos/manipulation/manipulation_module.py
+++ b/dimos/manipulation/manipulation_module.py
@@ -399,11 +399,7 @@ class ManipulationModule(Module):
             operation_status = OperationStatus[self._state.name]
             error = self._error_message or None
             has_pending_plan = self._last_plan is not None
-        execution_status = (
-            self._execution_manager.status
-            if hasattr(self, "_execution_manager")
-            else ExecutionStatus.IDLE
-        )
+        execution_status = self._execution_manager.status
         return ManipulationSnapshot(
             timestamp=time.time(),
             operation_status=operation_status,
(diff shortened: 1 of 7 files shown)


dimos/manipulation/manipulation_module.py
```diff
@@ -414,9 +415,19 @@ def get_state(self) -> ManipulationSnapshot:
 
     def get_operation_status(self) -> OperationStatus:
         """Return the current operation status without collecting telemetry."""
+        self._refresh_execution_status()
         with self._lock:
             return OperationStatus[self._state.name]
 
+    def _refresh_execution_status(self) -> None:
+        """Poll active nonblocking execution once, without waiting for motion."""
+        manager = getattr(self, "_execution_manager", None)
```
TomCC7: why would this be ever None? should not check this during runtime IMO

dimos/manipulation/visualization/viser/state.py
```diff
@@ -66,7 +66,6 @@ class PlanStatus(str, Enum):
     PLANNING = "planning"
     FRESH = "fresh"
     STALE = "stale"
-    EXECUTING = "executing"
```
TomCC7, line 69: why is this state removed? what's the state during execution then?

dimos/manipulation/manipulation_module.py
```diff
@@ -414,9 +415,19 @@ def get_state(self) -> ManipulationSnapshot:
...
+        if manager is not None and manager.status in {
+            ExecutionStatus.ACCEPTED,
+            ExecutionStatus.EXECUTING,
+        }:
+            self.wait_for_execution(timeout=0.0)
```
TomCC7, line 428: is this a blcoking call? why we need it here?

paul-nechifor (approved):

PR #3944, opened 2026-09-04
feat(manipulation): add client-only Arm SDK for dimos shell

TomCC7:
## Contribution path

- Discussion: https://github.com/dimensionalOS/dimos/discussions/3923
- Scope: client-only motion conveniences over the existing manipulation RPCs.
- Generic Spec discovery and RPC completion landed in #4009.
- Independent perception startup fix: #4008. Typed pick/place changes have been dropped, not moved to a follow-up.
- closes DIM-1472

## Problem

Basic Python motion scripts must select a planning group, construct messages and mappings, and check planning and execution results by hand. The standalone manipulation runner duplicates runtime and client lifecycle already handled by `dimos run` and `dimos shell`.

## Solution

`Arm` is a client-side API for controlling one arm at a time. It binds a planning group once, so each call can use joint positions or XYZ values directly. In `dimos shell`, use the existing connection:

```python
from dimos.manipulation.sdk import Arm

arm = Arm.from_app(app)
```

| API | What it does | Returns |
|---|---|---|
| `Arm.from_app(app)` | Select the unique pose-capable arm, with or without a gripper. Use `group=` and `instance_name=` when selection is ambiguous. | `Arm` |
| `arm.info` | Inspect the group ID, joint order, frames, and gripper capability. | `PlanningGroupInfo` |
| `arm.state()` | Read the arm's state, including available joint presets. | `PlanningGroupState` |
| `arm.joints()` | Read a fresh joint-position array in `arm.info.joint_names` order. | NumPy array |
| `arm.pose()` | Read the current end-effector pose. | `PoseStamped` |
| `arm.move_joints(positions)` | Plan and execute a move to joint positions supplied as a list, tuple, or NumPy array. | `ExecutionResult` |
| `arm.move_pose(position, orientation=None)` | Plan and execute a move to world-frame XYZ. Optional orientation is an XYZW quaternion; omitting it preserves the current orientation. | `ExecutionResult` |
| `arm.move_linear(dx=0, dy=0, dz=0)` | Execute a straight world-frame translation through the existing motion primitive. The SDK adds no path-planning call. | `MoveResult` |
| `arm.set_gripper_position(position)` | Set normalized gripper travel: `0.0` closed, `1.0` open. | `CommandResult` |
| `arm.open_gripper()` / `arm.close_gripper()` | Open or close the selected arm's gripper. | `CommandResult` |
[...]

PR #4009, opened 2026-09-08
feat(porcelain): add Spec discovery and RPC shell completion

TomCC7:
## Contribution path

Split from #3944. Related discussion: https://github.com/dimensionalOS/dimos/discussions/3923

## Problem

Clients should be able to find a module by its Spec without also knowing its implementation name, and discover its RPC methods through shell completion.

## Solution

`app.find_module_by_spec(MySpec)` finds the deployed module whose RPC signatures match the Spec, using the same checks as blueprint wiring.

Multiple matches require `instance_name=`. The module's class must be importable in the client. Name-based lookup still works.

`RPCClient.__dir__` includes RPC names so IPython can complete them. Method signatures, docstrings, and transport ownership are unchanged. This includes the completion change from #4020.

The Arm SDK in #3944 builds on this change.

## How to test

Use this branch's provisioned environment. Run only one blueprint on the transport bus.

In terminal one, start the mock dual-arm stack:

```bash
source .venv/bin/activate
dimos run dual-xarm6-planner-coordinator
```

Once startup finishes, open a shell in terminal two:

```bash
source .venv/bin/activate
dimos shell
```

Find the motion module by Spec:

```python
from dimos.manipulation.manipulation_spec import ManipulationSpec

motion = app.find_module_by_spec(ManipulationSpec)
motion.list_planning_groups()
```

You should see the left and right planning groups. This does not move either arm.

Check shell completion and inspection:

- Type `motion.list_` and press Tab. `list_planning_groups` should appear.
- Run `motion.list_planning_groups?`. Its signature and docstring should appear without a `self` parameter.
- Run `type(motion)`. It should still be `dimos.core.rpc_client.RPCClient`.

Check that name-based and explicit-instance lookup return the same proxy:

```python
app.get_module("ManipulationModule") is motion

app.find_module_by_spec(
    ManipulationSpec,
    instance_name=motion.remote_name,
) is motion
```

Both expressions should return `True`.

Check a missing instance:

```python
app.find_module_by_spec(ManipulationSpec, instance_name="missing")
```

This should raise `LookupError`.

Exit the shell, reopen it, and repeat the first lookup. The runtime should still be running. When finished, exit the shell and stop the test runtime with Ctrl-C in terminal one.

## AI assistance

Codex (GPT-5-based coding agent) implemented the change with user direction.

## Checklist


dimos/porcelain/dimos.py
```diff
@@ -132,10 +135,62 @@ def list_modules(self) -> list[ModuleInfo]:
 
         return [_describe_module(descriptor) for descriptor in descriptors]
 
-    def get_module(self, name: str) -> ModuleHandle:
-        """Return a module proxy by exact instance name or unique class name."""
+    @overload
+    def get_module(self, name: Callable[..., S], *, instance_name: str | None = None) -> S: ...
```
paul-nechifor: You say "Clients should be able to find a module by its Spec"

But then add `find_module_by_spec`, don't overload `get_module`.

paul-nechifor (approved):

PR #4019, opened 2026-09-09
feat(galaxea): r1pro kronknav blueprint fed by lidar and head depth

jeff-hykin:
r1pro nav blueprint using kronk nav. Lidar and head depth both feed the raytracing map, each with its own tf frame.


jeff-hykin:
Superseded: rebuilt this branch fresh off main so the diff no longer carries unrelated main commits (.github/, bin/). Old history preserved as `jeff/feat/r1_nav_old`. Continues in a new PR on the same branch.

PR #4021, opened 2026-09-09
fix(memory): store depth images losslessly instead of as JPEG

jeff-hykin:
Depth frames were being routed to the JPEG codec because `codec_for()` dispatches on payload type alone and `Image` covers both colour and depth. Float32 depth crashed out of turbojpeg so it could not be recorded at all, and uint16 depth silently came back as wrong metres.

### DIFF
--- a/dimos/memory/codecs/jpeg.py
+++ b/dimos/memory/codecs/jpeg.py
@@ -14,7 +14,9 @@
 
 from __future__ import annotations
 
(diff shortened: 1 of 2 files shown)


jeff-hykin:
Folded into #4146 — the R1 depth path needs this fix, and keeping it as a separate base branch made #4146's diff carry unrelated main commits.

PR #3873, opened 2026-09-01
refactor(xarm): one grasping stack, hardware by default

mustafab0:
Cleaning up the xArm grasping blueprints, and drawing the sim/hardware boundary in one place instead of writing the stack out once per target.

| Before | Now | What it does |
|---|---|---|
| `xarm-perception` | `xarm-grasp` | The real arm: coordinator, RealSense wrist camera, scene registration (moondream + edgetam), pick and place, top-down heuristic grasps, viser |
| `xarm-room-sim` | `xarm-grasp --simulation mujoco` | The same stack against `data/xarm_grasp_sim`: MuJoCo wrist camera, owlv2 + yolo, headless |
| `xarm-perception-agent` | `xarm-grasp-agent` | The above plus an MCP agent |
| `xarm-perception-sim` | unchanged | Plain planning and teleop sim on `data/xarm7`, not part of the grasping stack |
| `xarm-perception-sim-agent` | unchanged | The above plus an MCP agent |

PR #4035, opened 2026-09-09
fix(porcelain): apply config overrides before resolving the blueprint

mustafab0:
`Dimos.run()` resolved the target first and applied the instance's config overrides second. 

Resolving a blueprint name imports its module, and blueprints settle the arm-versus-sim split at import time, `SIMULATED = bool(global_config.simulation)` at module scope, so `Dimos(simulation="mujoco").run("xarm-grasp")` captured the default, composed the hardware stack, and then tried to reach an xArm and a RealSense that are not attached.

Found by greptile on #3873, which added the tenth such blueprint. The cause is here, not there.

### DIFF
--- a/dimos/porcelain/dimos.py
+++ b/dimos/porcelain/dimos.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 import atexit
 import importlib
 import inspect
+import sys
 import threading
 from typing import Any, TypeAlias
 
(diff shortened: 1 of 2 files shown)


mustafab0:
Solution is not great.

Issue tracked here to better solve this: https://github.com/dimensionalOS/dimos/issues/4100

PR #4043, opened 2026-09-10
fix/sdk source install

TomCC7:
**Superseded by the three-feature stack: SDK web/source packaging is in #4044; the application dependency-policy helper is in #4047.**

Current review order: #4042 → #4044 → #4047. Stack: https://github.com/dimensionalOS/dimos/stack/4064

---

Installing DimOS as an external Git dependency currently depends on web assets from the developer checkout. Build missing assets during wheel/sdist creation and define the SDK dependency-source policy consumed by app creation.

Validation: source-build and dependency-policy tests pass; a full external SDK Git install built its own assets successfully. The source archive passes the repository packaging check.

PRD: https://linear.app/dimensional/project/dimos-developer-onboarding-v1-3df889ba9a65

Review in order: #4042 → #4043 → #4044 → #4045 → #4046 → #4047. Replaces the implementation in #4027–#4030.

PR #4044, opened 2026-09-10
Make the SDK installable outside its repository

TomCC7:
Make dimOS installable and its native modules buildable outside the dimOS repository. Fresh Git and source-archive installs succeed without generated Cockpit/web-SDK assets or Deno. Python packaging never builds the frontend; already-built assets are still included when present. Wheels and source archives include the native inputs required by the installed SDK; builds use writable caches instead of modifying site-packages.

The native changes cover source discovery, installed-module build paths, and the reduced Point-LIO PCL dependency set needed for macOS builds. This combines the SDK packaging work from #4043 with the former native-build PR. The application dependency-policy helper is in #4047, alongside dimup init.

Validation: 46 installer/packaging/native-source tests, strict dimup typing, and pre-commit pass at this layer. Unit CI extends #4042 with SDK packaging and native-source tests. Direct-source and sdist-derived wheels built with no frontend assets and a failing Deno executable; both retained native inputs and the compiled extension. #4047 runs the three-platform application and native replay journeys.

PRD: https://linear.app/dimensional/project/dimos-developer-onboarding-v1-3df889ba9a65

Review order: #4042 → #4044 → #4047.


dimos/experimental/memory/rust_recorder.py
```diff
@@ -95,6 +95,7 @@ class RustRecorderConfig(NativeModuleConfig):
 
     executable: str = "result/bin/dimos-memory-recorder"
     build_command: str = "nix build -L .#dimos-memory-recorder"
+    bundled_sources: bool = True
```
TomCC7, line 98: what is this mod?

dimos/core/native_module.py
```diff
@@ -120,6 +122,7 @@ class NativeModuleConfig(ModuleConfig):
     executable: str
     build_command: str | None = None
     cwd: str | None = None
+    bundled_sources: bool = False
```
TomCC7, line 125: will this single change allow source build to be working from pypi installation?

PR #4050, opened 2026-09-10
fix_7_mac_tests: 2 fix - 5 skip

bogwi:
Fix 7 failing tests on mac.

2 - real fix, they should now pass on macOS

The 5 nix ones: not fixable as tests, they are skipped


TomCC7 (approved):
Manipulation changes lgtm

.github/workflows/ci.yml
mustafab0: Hesitant to bring this back.

We only have one Mac runner which causes huge bottleneck on CI. We have it as optional but the test failing removed PR from merge queue.

We should first confirm the merge queue is okay and then re enable this after

PR #4053, opened 2026-09-10
feat(manipulation): UFactory Lite 6 support

leshy:
The Lite 6 speaks the xArm SDK

New blueprints: `coordinator-lite6`, `lite6-planner-coordinator`, `keyboard-teleop-lite6`. Address comes from `--lite6-ip`


dimos/core/global_config.py
```diff
@@ -49,6 +49,7 @@ class GlobalConfig(BaseSettings):
     unitree_aes_128_key: str | None = None
     xarm7_ip: str | None = None
     xarm6_ip: str | None = None
+    lite6_ip: str | None = None
```
TomCC7, line 52: we'll handle the cleanup after control coordinator refactor

dimos/hardware/manipulators/xarm/adapter.py
```diff
@@ -431,6 +439,18 @@ def _write_gripper(self, position: float) -> bool:
         """Command the gripper in SDK units (0-850)."""
         if not self._arm:
             return False
+        if self._lite6:
+            opening = position > (XARM_GRIPPER_MIN + XARM_GRIPPER_MAX) / 2
```
TomCC7, line 443: so there's no api to control that gripper for a partial opening?

mustafab0: Appears to be a gpio control gripper. Probably just a Binary toggle

leshy: yeah it's an air pump based gripper without feedback. idk if you guys specify grippers independantly of arms etc since these are changable

mustafab0: Yes we do

Grippers can be controlled as a simple joint by itself. We haven't made some proper examples for it but should

mustafab0 (dismissed):

paul-nechifor (dismissed):

paul-nechifor (approved):

PR #3846, opened 2026-08-31
Add zenoh transport to the C++ native SDK

jeff-hykin:
add zenoh to C++

PR #4057, opened 2026-09-10
feat(native): zenoh transport for the C++ native module SDK

leshy:
The C++ native SDK had the `Transport` seam and one implementation behind it. This adds the second one, so `DIMOS_TRANSPORT=zenoh` runs a C++ module instead of throwing "LCM only"

### DIFF
--- a/native/cpp/include/dimos/native/zenoh_transport.hpp
+++ b/native/cpp/include/dimos/native/zenoh_transport.hpp
@@ -1,0 +1,312 @@
+// Copyright 2026 Dimensional Inc.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Zenoh implementation of the Transport seam over zenoh-c. The session settings
+// and per-channel publisher QoS come from the coordinator's launch line, the
+// same `session` and `qos` blocks the Rust native module reads.
+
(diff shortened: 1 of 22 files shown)


jeff-hykin:
Why not use the zenoh C++ api? (Instead of the C api)

native/cpp/include/dimos/native/zenoh_transport.hpp
```diff
@@ -0,0 +1,312 @@
...
+    ZenohTransport(const ZenohTransport&) = delete;
+    ZenohTransport& operator=(const ZenohTransport&) = delete;
+
+    void publish(const std::string& channel, std::vector<uint8_t> data) override {
+        const z_loaned_publisher_t* publisher = publisher_for(channel);
+        if (publisher == nullptr) {
+            return;
+        }
+        z_owned_bytes_t payload;
+        z_bytes_copy_from_buf(&payload, data.data(), data.size());
```
jeff-hykin, line 187: Move should be a lot more performant than copy here

leshy:
closing for https://github.com/dimensionalOS/dimos/pull/3846

PR #4065, opened 2026-09-10
fix(mem): skip streams whose payload type cannot be imported in mem rerun

spomichter:
## Problem

`dimos mem rerun <db>` crashed with `ModuleNotFoundError` when the store contained a stream whose payload class does not exist on the current branch, e.g. `april_tag` (pickled `TagInfo` from `dimos.mapping.recording`) or `pose_graph` (`dimos.navigation.jnav.msgs`).

Opening the stream resolves the stored payload class by import path, and that happened one line before the existing "no `to_rerun()`" skip.

## Fix

Treat an unresolvable payload type like a non-renderable stream: print a skip line and continue with the rest.

```
  skip april_tag: payload type unavailable (No module named 'dimos.mapping.recording')
  skip pose_graph: payload type unavailable (No module named 'dimos.navigation.jnav')
april_tags 100% [17/17] ...
color_image ...
```

Verified on a 2026-09-03 recording that renders everything else.

### DIFF
--- a/dimos/memory/cli/render.py
+++ b/dimos/memory/cli/render.py
@@ -77,10 +77,13 @@ def render_store(
     renderable = []
     t0: float | None = None
     for name in store.list_streams():
-        stream = store.streams[name]
         try:
+            stream = store.streams[name]
             first = stream.first()
         except LookupError:
+            continue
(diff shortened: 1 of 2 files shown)


spomichter:
@greptile review

PR #4066, opened 2026-09-10
chore: make manipulation extra a complete workflow bundle

TomCC7:
## Contribution path

## Problem

`uv sync --extra manipulation` installed planners and arm SDKs but omitted perception, agent/web, and simulation dependencies used by the shipped manipulation blueprints.

## Solution

- Make `manipulation` compose `planning`, `base`, `sim`, and `cpu`. Expose narrower `control` and `planning` extras.
- Move EdgeTAM and `timm` into `perception`. Keep embedding models, their runtime dependencies (`gdown` and TensorBoard), notebook tools, Google Maps, and Portal in `misc`. Keep xArm SDK ownership in `control`.
- Declare Socket.IO in `web`: `WebsocketVisModule` imports it directly and runs its ASGI server.
- Move the Go2 Quest blueprint into the Go2 tree so arm Quest imports do not require the Unitree SDK. CLI blueprint names stay the same; the registry is regenerated.
- Keep `uv sync --extra manipulation --inexact` as the recommended setup and error hint. The bundle supplies its dependencies; `--inexact` preserves additional installed packages.
- Remove the redundant Linux/macOS-only marker from `can-motor-control`; retain hardware/architecture-specific restrictions.

The existing CI workflow is unchanged; this PR adds no smoke-check script or CI step. `all` and the self-hosted dependency group retain `misc` coverage.

Navigation follow-up: leave an equivalent complete navigation workflow extra to the navigation team. This PR does not reorganize navigation dependencies.

## How to Test

```bash
uv run --no-sync dimos run xarm-perception-sim-agent
```

The agentic simulation requires an OpenAI API key, robot/model assets, and CUDA or MPS for the default EdgeTAM backend. `control` and `planning` remain available for smaller installations.

Validated locally after the review changes:

- A fresh installation with `--no-default-groups --extra manipulation` loaded all 51 registered manipulation and arm Quest blueprints without `misc`, pytest, or the Unitree SDK.
- A real Pink QP step, a timed RoboPlan Cartesian trajectory, Viser startup/shutdown, EdgeTAM encoder inference on CPU with untrained weights, and ten xArm MuJoCo steps passed. No physical hardware or paid API calls were used.
- An independently installed Portal package survived `uv sync --extra manipulation --inexact`.
- 83 targeted pytest cases passed, covering the registry, Quest, xArm, Pink install hints, and Viser lifecycle.
- All changed-file pre-commit checks, lockfile consistency, and built-wheel metadata checks passed. All 557 package versions remain unchanged.
- Metadata checks confirm `misc` remains included in `all` and the self-hosted dependency group, while `manipulation` does not depend on it.

The local wheel used `DIMOS_ALLOW_MISSING_COCKPIT=1` because this checkout lacks built cockpit assets. Full trained EdgeTAM inference and an agent-driven pick-and-place run were not validated locally; no usable GPU driver was available.

## AI assistance

Codex (GPT-6) implemented the changes and ran validation.

## Checklist


docs/capabilities/manipulation/agentic.md
```diff
@@ -11,10 +11,12 @@ the underlying planning and perception stack.
 Install the manipulation dependencies:
 
 ```bash
-uv sync --extra manipulation --inexact
+uv sync --extra manipulation
```
TomCC7: need to keep the inexact as user might want other dependencies

pyproject.toml
```diff
@@ -252,6 +229,13 @@ web = [
 ]
 
 perception = [
+    "timm>=1.0.15",  # runtime backbone for the EdgeTAM/sam2 segmentation config
+    "edgetam-dimos>=1.0.1",
+    "open_clip_torch==3.2.0",
+    "torchreid==0.2.5",
+    # torchreid imports these at runtime but omits them from its package metadata.
+    "gdown>=5.2.2",
```
TomCC7: why is this a perception dependency?

pyproject.toml
```diff
@@ -180,30 +180,6 @@ Issues = "https://github.com/dimensionalOS/dimos/issues"
 Changelog = "https://github.com/dimensionalOS/dimos/releases"
 
 [project.optional-dependencies]
-misc = [
```
TomCC7, line 183: maybe keep this group I have no idea what these are for...

pyproject.toml
```diff
@@ -242,6 +218,7 @@ agents = [
 
 web = [
     "fastapi>=0.115.6",
+    "python-socketio>=5.16.1",  # WebsocketVisModule, also imported by Quest visualization
```
TomCC7, line 232: should this be part of web?

pyproject.toml
```diff
@@ -280,32 +264,36 @@ unitree-dds = [
...
-    "drake>=1.40.0; sys_platform != 'darwin' and platform_machine != 'aarch64'",
-
-    # Hardware SDKs
+# Coordinators, arm hardware adapters, and keyboard/Cartesian control.
+control = [
     "piper-sdk",
     "xarm-python-sdk>=1.17.0",
     "a750-control; sys_platform == 'linux' and platform_machine == 'x86_64'",
-
     "can-motor-control>=0.0.8; sys_platform == 'linux' or sys_platform == 'darwin'",
```
TomCC7: if these are the only platforms we supported we don't need to add this filter?

.github/workflows/ci.yml
```diff
@@ -724,6 +723,18 @@ jobs:
           uv pip install --python .codecov-venv/bin/python codecov-cli==11.2.8
       - name: Install dependencies
         run: uv sync --group tests-self-hosted --frozen
+      - name: Check manipulation extras without development dependencies
```
TomCC7: don't think we need a ci for this...

dimos/manipulation/planning/kinematics/pink_solver.py
```diff
@@ -28,9 +28,7 @@
     import pinocchio
     import qpsolvers
 except ImportError as exc:
-    msg = (
-        "Pink IK dependencies not found; install them with: uv sync --extra manipulation --inexact."
-    )
+    msg = "Pink IK dependencies not found; install them with: uv sync --extra control."
```
TomCC7: why not keep manipulation as user typically just install this single extra

mustafab0 (approved):

paul-nechifor (approved):

PR #4074, opened 2026-09-11
feat(native): add JSON-RPC 2.0 support to NativeModule

omarespejel:
Refs dimensionalOS/dimos#3506 and dimensionalOS/dimos#3516.

#4152 adds the reusable Python RPC transport. This draft still needs to be updated to use it.

This draft gets Python calling a Rust `NativeModule` over Zenoh with JSON-RPC 2.0

My recommendation is to make JSON-RPC an `RPCSpec` implementation next so Python and native modules can use the same RPC interface, while existing calls stay on pickle and move later one by one

`dimos/rpc/v1` is our DimOS protocol version  
JSON-RPC 2.0 is the request and reply format

This draft keeps discovery in the existing method list. Choosing a version, running versions side by side, and adding other encodings are follow-up work

There’s also a small toy planner example. Startup checks method names and a launch token. Native calls don’t retry automatically.

This draft is scoped to small, JSON-compatible commands and replies. It does not yet add RPC to an existing Rust module or define how types like `Twist` cross the wire. Large data, such as point clouds, should stay binary. [Microduck](https://github.com/pollen-robotics/microduck/blob/6507d2e960417aaa4ecd38eccf59b2dcf586ecd2/docs/design/architecture.md#L172-L197) uses a similar split: JSON-RPC for control, with video and audio on a separate path. [`rmw_zenoh`](https://github.com/ros2/rmw_zenoh/blob/c833e326048d97b26ea58882b247d2c642710e4d/docs/design.md#L100-L107) serializes ROS messages as CDR bytes over Zenoh.

```
uv run pytest examples/native_rpc/test_native_rpc.py --no-cov
```

49 tests pass locally.

Used Codex.

### DIFF
--- a/examples/native_rpc/README.md
+++ b/examples/native_rpc/README.md
@@ -1,0 +1,18 @@
+# Native RPC toy
+
+Python starts a Rust `ToyPlanner` through `NativeModule` and calls `plan(start, goal)`.
+It returns the two points, not a collision-checked path. No robot needed.
+
+```sh
+uv run pytest examples/native_rpc/test_native_rpc.py --no-cov
+```
+
+`@native_rpc` declares Rust-owned methods. The stdin launch carries their names;
+`start()` waits for `_ready` to confirm this launch and its methods. Python RPC stays on pickle.
+
+JSON-RPC 2.0 request/reply over Zenoh only: named JSON parameters, no retries,
+notifications, batches or typed messages. The Rust runner handles one call at a time;
+streaming modules, concurrent handlers and bake are left for review. No authentication is added;
+use a trusted network. Tests use loopback with discovery off.
+
+Non-finite floats are rejected. Numbers outside serde_json's range return a parse error.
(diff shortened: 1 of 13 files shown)


examples/native_rpc/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+```sh
+uv run pytest examples/native_rpc/test_native_rpc.py --no-cov
+```
+
+`@native_rpc` declares Rust-owned methods. The stdin launch carries their names;
+`start()` waits for `_ready` to confirm this launch and its methods. Python RPC stays on pickle.
+
+JSON-RPC 2.0 request/reply over Zenoh only: named JSON parameters, no retries,
+notifications, batches or typed messages. The Rust runner handles one call at a time;
```
leshy, line 14: > The Rust runner handles one call at a time;

this seems like an arbitrary restriction that would bite us?

omarespejel: agreed, one at a time is only for this toy not something I want in the RPC contract

PR #4096, opened 2026-09-12
feat(manipulation): execute planar-base plans on a feedback base controller

mustafab0:
Prob: #3844 let the planner use an unbounded planar base, but execute() still rejected any plan that moved it. 

Solution: This adds the other half: a base_trajectory coordinator task that follows the plan's x, y and yaw commands on a twist base, sampling on the coordinator's tick clock and correcting against odometry, while the arm and torso commands still go to the joint trajectory task. 

The manipulation module splits a whole-body plan between the two and cancels one side whenever the other aborts, is preempted or is refused.

r1pro-planar-preview now runs on a mock chassis that integrates its own odometry, so this is testable in viser with `dimos run r1pro-planar-preview`.

### DIFF
--- a/dimos/control/tasks/base_trajectory_task/base_trajectory_task.py
+++ b/dimos/control/tasks/base_trajectory_task/base_trajectory_task.py
@@ -1,0 +1,335 @@
+# Copyright 2026 Dimensional Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Track a planner's timed base trajectory on a twist base."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+import math
+import threading
+from typing import Any
+
+from pydantic import ConfigDict, NonNegativeFloat, PositiveFloat
+
+from dimos.control.task import (
+    BaseControlTask,
+    ControlMode,
+    CoordinatorState,
+    JointCommandOutput,
+    ResourceClaim,
+)
+from dimos.control.tasks.trajectory_task.trajectory_task import (
+    TrajectoryExecutionResult,
+    TrajectoryExecutionStatus,
+)
+from dimos.msgs.trajectory_msgs.JointTrajectory import JointTrajectory
+from dimos.msgs.trajectory_msgs.TrajectoryStatus import TrajectoryState, TrajectoryStatus
+from dimos.protocol.service.spec import BaseConfig
+from dimos.utils.logging_config import setup_logger
+from dimos.utils.trigonometry import angle_diff
+
+logger = setup_logger()
+
+_FB_CLAMP_LINEAR = 0.15
+_FB_CLAMP_YAW = 0.4
+_SPEED_SLACK = 1e-3
+
+
+def _clamp(value: float, limit: float) -> float:
+    return max(-limit, min(limit, value))
+
+
+@dataclass
+class BaseTrajectoryTaskConfig:
+    joint_names: list[str]
+    priority: int = 10
+    kp: tuple[float, float, float] = (1.0, 1.0, 1.0)
+    max_linear: float = 1.0
+    max_angular: float = 2.0
+    start_tolerance: float = 0.1
+    goal_tolerance: float = 0.05
+    orientation_tolerance: float = 0.1
+    max_tracking_error: float = 0.5
+    max_yaw_tracking_error: float = 0.5
+    settle_timeout: float = 2.0
+    stop_hold_s: float = 0.5
+    stale_pose_timeout: float = 0.3
+
+
+class BaseTrajectoryTask(BaseControlTask):
+    """Follow an (x, y, yaw) trajectory in the odometry frame with feedforward and feedback.
+
+    The trajectory's three columns are x, y and yaw in that order; yaw may be
+    unwrapped. Commands are body-frame (vx, vy, wz) on the twist joints, which
+    report odometry through their position fields. A finished, failed, cancelled
+    or preempted run reports its final state at once and streams zero velocity
+    for ``stop_hold_s``.
+    """
+
+    def __init__(self, name: str, config: BaseTrajectoryTaskConfig) -> None:
+        if len(config.joint_names) != 3:
+            raise ValueError(
+                f"BaseTrajectoryTask '{name}' needs 3 joints (vx, vy, wz), "
+                f"got {len(config.joint_names)}"
+            )
+        self._name = name
+        self._config = config
+        self._joints = list(config.joint_names)
+        self._lock = threading.Lock()
+        self._state = TrajectoryState.IDLE
+        self._error = ""
+        self._trajectory: JointTrajectory | None = None
+        self._start_t: float | None = None
+        self._elapsed = 0.0
+        self._stop_until: float | None = None
+        self._stop_pending = False
+        self._armed_t: float | None = None
+        self._last_pose_t: float | None = None
+
+    @property
+    def name(self) -> str:
+        return self._name
+
+    def claim(self) -> ResourceClaim:
+        return ResourceClaim(
+            joints=frozenset(self._joints),
+            priority=self._config.priority,
+            mode=ControlMode.VELOCITY,
+        )
+
+    def is_active(self) -> bool:
+        return (
+            self._state == TrajectoryState.EXECUTING
+            or self._stop_pending
+            or self._stop_until is not None
+        )
+
+    def execute(self, trajectory: JointTrajectory) -> TrajectoryExecutionResult:
+        """Accept a trajectory; its start is checked against odometry on the first tick."""
+        problem = _trajectory_problem(trajectory, self._config)
+        if problem:
+            return TrajectoryExecutionResult(TrajectoryExecutionStatus.INVALID_TRAJECTORY, problem)
+        with self._lock:
+            if self._state == TrajectoryState.EXECUTING:
+                return TrajectoryExecutionResult(
+                    TrajectoryExecutionStatus.ALREADY_EXECUTING,
+                    f"Base trajectory task '{self._name}' is already executing",
+                )
+            self._trajectory = trajectory
+            self._start_t = None
+            self._elapsed = 0.0
+            self._stop_until = None
+            self._stop_pending = False
+            self._armed_t = None
+            self._error = ""
+            self._last_pose_t = None
+            self._state = TrajectoryState.EXECUTING
+        return TrajectoryExecutionResult(TrajectoryExecutionStatus.ACCEPTED)
+
+    def cancel(self) -> bool:
+        with self._lock:
+            if self._state != TrajectoryState.EXECUTING:
+                return False
+            self._stop(None, TrajectoryState.ABORTED, "cancelled")
+        return True
+
+    def get_status(self) -> TrajectoryStatus:
+        with self._lock:
+            duration = self._trajectory.duration if self._trajectory is not None else 0.0
+            progress = 1.0 if duration <= 0.0 else min(1.0, self._elapsed / duration)
+            return TrajectoryStatus(
+                state=self._state,
+                progress=progress,
+                time_elapsed=self._elapsed,
+                time_remaining=max(0.0, duration - self._elapsed),
+                error=self._error,
+            )
+
+    def compute(self, state: CoordinatorState) -> JointCommandOutput | None:
+        with self._lock:
+            if self._stop_pending:
+                self._stop_until = state.t_now + self._config.stop_hold_s
+                self._stop_pending = False
+            if self._stop_until is not None:
+                if state.t_now >= self._stop_until:
+                    self._stop_until = None
+                return self._command(0.0, 0.0, 0.0)
+            if self._state != TrajectoryState.EXECUTING or self._trajectory is None:
+                return None
+            if self._armed_t is None:
+                self._armed_t = state.t_now
+
+            pose = self._read_pose(state)
+            if pose is None:
+                last_seen = self._last_pose_t if self._last_pose_t is not None else self._armed_t
+                if state.t_now - last_seen > self._config.stale_pose_timeout:
+                    return self._fail(state.t_now, "base odometry is unavailable")
+                return self._command(0.0, 0.0, 0.0)
+
+            first_tick = self._start_t is None
+            if self._start_t is None:
+                self._start_t = state.t_now
+            self._elapsed = state.t_now - self._start_t
+            reference, reference_velocity = self._trajectory.sample(self._elapsed)
+            position_error = math.hypot(reference[0] - pose[0], reference[1] - pose[1])
+            yaw_error = abs(angle_diff(reference[2], pose[2]))
+
+            config = self._config
+            if first_tick and max(position_error, yaw_error) > config.start_tolerance:
+                return self._fail(
+                    state.t_now,
+                    f"trajectory starts {position_error:.3f} m / {yaw_error:.3f} rad from the "
+                    "base's odometry",
+                )
+            if (
+                position_error > config.max_tracking_error
+                or yaw_error > config.max_yaw_tracking_error
+            ):
+                return self._fail(
+                    state.t_now,
+                    f"base is {position_error:.3f} m / {yaw_error:.3f} rad off the trajectory",
+                )
+            if self._elapsed >= self._trajectory.duration:
+                if (
+                    position_error < config.goal_tolerance
+                    and yaw_error < config.orientation_tolerance
+                ):
+                    self._stop(state.t_now, TrajectoryState.COMPLETED, "")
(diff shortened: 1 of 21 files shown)


dimos/control/tasks/planar_base_trajectory_task/planar_base_trajectory_task.py
```diff
@@ -0,0 +1,335 @@
...
+    return max(-limit, min(limit, value))
+
+
+@dataclass
+class BaseTrajectoryTaskConfig:
+    joint_names: list[str]
+    priority: int = 10
+    kp: tuple[float, float, float] = (1.0, 1.0, 1.0)
+    max_linear: float = 1.0
+    max_angular: float = 2.0
```
TomCC7, line 61: I feel like these two should be robot-specific instead of task-specific

dimos/control/tasks/base_trajectory_task/base_trajectory_task.py
```diff
@@ -0,0 +1,335 @@
...
+@dataclass
+class BaseTrajectoryTaskConfig:
+    joint_names: list[str]
+    priority: int = 10
+    kp: tuple[float, float, float] = (1.0, 1.0, 1.0)
+    max_linear: float = 1.0
+    max_angular: float = 2.0
+    start_tolerance: float = 0.1
+    goal_tolerance: float = 0.05
+    orientation_tolerance: float = 0.1
```
TomCC7: should this be `goal_orientation_tolerance`?

dimos/control/tasks/base_trajectory_task/base_trajectory_task.py
```diff
@@ -0,0 +1,335 @@
...
+_FB_CLAMP_YAW = 0.4
+_SPEED_SLACK = 1e-3
+
+
+def _clamp(value: float, limit: float) -> float:
+    return max(-limit, min(limit, value))
+
+
+@dataclass
+class BaseTrajectoryTaskConfig:
```
TomCC7: maybe `PlanarBaseTrajectoryTask`?

dimos/control/tasks/planar_base_trajectory_task/planar_base_trajectory_task.py
```diff
@@ -0,0 +1,335 @@
...
+
+    The trajectory's three columns are x, y and yaw in that order; yaw may be
+    unwrapped. Commands are body-frame (vx, vy, wz) on the twist joints, which
+    report odometry through their position fields. A finished, failed, cancelled
+    or preempted run reports its final state at once and streams zero velocity
+    for ``stop_hold_s``.
+    """
+
+    def __init__(self, name: str, config: BaseTrajectoryTaskConfig) -> None:
+        if len(config.joint_names) != 3:
```
TomCC7, line 84: can do something like this to decouple init with config checking:
```python
from pydantic import BaseModel, ConfigDict, Field, NonNegativeFloat, PositiveFloat

class BaseTrajectoryTaskParams(BaseConfig):
    model_config = ConfigDict(allow_inf_nan=False)

    # exactly 3 joints: vx, vy, wz
    joint_names: list[str] = Field(min_length=3, max_length=3)

    kp: tuple[NonNegativeFloat, NonNegativeFloat, NonNegativeFloat] = (1.0, 1.0, 1.0)
    max_linear: PositiveFloat = 1.0
    max_angular: PositiveFloat = 2.0
    start_tolerance: PositiveFloat = 0.1
    goal_tolerance: PositiveFloat = 0.05
    orientation_tolerance: PositiveFloat = 0.1
    max_tracking_error: PositiveFloat = 0.5
    max_yaw_tracking_error: PositiveFloat = 0.5
    settle_timeout: NonNegativeFloat = 2.0
    stop_hold_s: NonNegativeFloat = 0.5
    stale_pose_timeout: PositiveFloat = 0.3
```

dimos/manipulation/execution_manager.py
TomCC7, line 1: This looks more like a temporary solution to me. The base trajectory task is very specific and we might add a lot more similar cases in the future (for example 3d base, or linear rail). There's two ways that make this design more extendable IMO:
1. decouple the logic to the controller level by adding a generic `trajectorytask` which dispatch to different handling logic separately
2. formalize the "execution dispatching" to a registry style (I think moveit does this) which binds specific joints to a task. So that we can actually swap tasks more flexibly. (tho I don't see this design in moveit to be very useful as most of time it just dispatch to JTC)

TomCC7 (changes requested):

TomCC7:
and would this current design be easy to transfer to real hardware? I'm not sure what interface the hardware provides (like twist action?)

dimos/control/tasks/planar_base_trajectory_task/planar_base_trajectory_task.py
mustafab0: We do have robot specific limits at the connection layer.

We should be able to configure clamps per task as well, when executing a certain task I may want to clamp the overall velocity without necessarily changing the overall robot config.

dimos/control/tasks/base_trajectory_task/base_trajectory_task.py
mustafab0: We need both position and orientation tolerance. I can rename it better though

dimos/control/tasks/planar_base_trajectory_task/planar_base_trajectory_task.py
mustafab0: This requires a few more changes here. I will link this with issue and track it to later fix

dimos/manipulation/execution_manager.py
mustafab0: This is true. It is a temporary solution in sim to start execution for now. 

> The base trajectory task is very specific and we might add a lot more similar cases in the future (for example 3d base, or linear rail)

Yes I agree. The Idea is to have a single `BaseTrajectoryTask` that works for all robots even drones that have 6dofs. Since the current planar base planner requires 3dof i stuck with that for now.

1. Yes makes sense. I already have something like a holonomic trajectory controller which is super generic. However that one is designed to be a trajectory follower without timing constraint. 
The `BaseTrajectoryTask` should replace the same  but with timing control
2. The Control coordinator is already managing this. In the current execution manager we are doing the same, maybe we want to just make it more generic?

mustafab0:
> and would this current design be easy to transfer to real hardware? I'm not sure what interface the hardware provides (like twist action?)

Yes should be very straightforward. I just need to supply odom from nav or directly the flowbase and we can translate this pretty easy.

dimos/manipulation/execution_manager.py
TomCC7: for 2, control coordinator handles the arbitration from a different path, where command goes in through module IO. however execution manager and these trajectory tasks uses RPC overriding the arbitration design largely, and you are currently handling trajectory task dispatching logic in manip module execution manager

dimos/control/tasks/planar_base_trajectory_task/planar_base_trajectory_task.py
TomCC7: no I guess fine since the verification layer is thin. but in general since we largely depend on pydantic already it makes sense to fully leverage it's feature (pydantic dataclass can be the dropin replacement of dataclass)

dimos/manipulation/execution_manager.py
mustafab0: > you are currently handling trajectory task dispatching logic in manip module execution manage

Yes, I believe this is also the design intention. Since the manipulation module is doing the trajectory timing for the whole motion. The `execution_manager` can then also take care of dispatching from here, and monitor. 

Also the RPC commands don't over ride the arbitration design. Arbitration happens after a task computes the next action, so this would not affect the execution logic as such.

dimos/control/tasks/planar_base_trajectory_task/planar_base_trajectory_task.py
```diff
@@ -0,0 +1,340 @@
...
+        return TrajectoryExecutionResult(TrajectoryExecutionStatus.ACCEPTED)
+
+    def cancel(self) -> bool:
+        with self._lock:
+            if self._state != TrajectoryState.EXECUTING:
+                return False
+            self._stop(None, TrajectoryState.ABORTED, "cancelled")
+        return True
+
+    def get_status(self) -> TrajectoryStatus:
```
TomCC7: this function is never invoked

dimos/control/tasks/planar_base_trajectory_task/planar_base_trajectory_task.py
```diff
@@ -0,0 +1,340 @@
...
+    def _feedback(
+        self, reference: list[float], pose: tuple[float, float, float]
+    ) -> tuple[float, float, float]:
+        x, y, yaw = pose
+        ex_world = reference[0] - x
+        ey_world = reference[1] - y
+        cos_yaw, sin_yaw = math.cos(yaw), math.sin(yaw)
+        ex_body = cos_yaw * ex_world + sin_yaw * ey_world
+        ey_body = -sin_yaw * ex_world + cos_yaw * ey_world
+        kp = self._config.kp
```
TomCC7, line 296: need to verify on hardware whether we need damping

dimos/manipulation/execution_manager.py
```diff
@@ -50,17 +61,27 @@ def __init__(
...
         self._operation_lock = threading.Lock()
         self._state_lock = threading.Lock()
+        self._poll_lock = threading.Lock()
         self._default_timeout = default_timeout
         self._poll_interval = poll_interval
+        self._base_task = base_task
+        self._base_joint_names = list(base_joint_names)
         self._active = False
         self._latest_result: ExecutionResult | None = None
+        self._legs: tuple[str, ...] = (JOINT_TRAJECTORY_TASK_NAME,)
```
TomCC7: how many legs does r1pro have? lol

dimos/manipulation/execution_manager.py
```diff
@@ -83,36 +104,72 @@ def execute(
...
+            with self._poll_lock:
+                with self._state_lock:
+                    self._run_id += 1
+                    run_id = self._run_id
+                    self._legs = tuple(legs)
+                    self._run_done = threading.Event()
+                    run_done = self._run_done
+                self._store(accepted, active=True)
+                if base_trajectory is not None:
+                    self._watchdog = threading.Thread(
```
TomCC7: shouldn't watchdog be a generic thing (if added)?

dimos/manipulation/execution_manager.py
```diff
@@ -208,27 +266,133 @@ def cancel(self, timeout: float = 1.0) -> ExecutionResult:
...
+                )
+            mapped = self._result_from_status(status).status
+            return ExecutionResult(mapped, f"{leg}: {status.error}", trajectory_status=primary)
+        if all(status.state is TrajectoryState.COMPLETED for status in statuses.values()):
+            return ExecutionResult(ExecutionStatus.COMPLETED, trajectory_status=primary)
+        return ExecutionResult(ExecutionStatus.EXECUTING, trajectory_status=primary)
+
+    def _watch(self, run_done: threading.Event, run_id: int) -> None:
+        # Non-blocking runs are otherwise only polled when someone reads status.
+        while not run_done.wait(self._poll_interval):
```
TomCC7, line 331: for jtt the current way is to poll everytime when visualizer request for status (as visualization is the only status consumer). I guess either align/unify status check.

dimos/manipulation/manipulation_module.py
```diff
@@ -166,6 +167,26 @@ class ManipulationModuleConfig(ModuleConfig):
     default_speed_scale: float = Field(default=1.0, gt=0.0, le=1.0)
     linear_speed_scale: float = Field(default=0.5, gt=0.0, le=1.0)
     execution_timeout: float = Field(default=60.0, gt=0.0)
+    # Coordinator joint name -> model joint name, so a twist base's odometry
+    # joints (chassis/vx, ...) can feed the model's planar base joints.
```
TomCC7, line 171: why would we need this? even if action interface is velocity we should be still able to have state interface to be velocity? overriding seems super messy

dimos/manipulation/execution_manager.py
TomCC7: > Yes, I believe this is also the design intention.

Yes this is the second direction I'm proposing which is the more generalized version of dispatching logic.

lmk if you want to address in separate pr but current implementation is very dirty IMO. the special handling exists across all layers

dimos/manipulation/execution_manager.py
mustafab0: Lol, `legs` in this context means a part of the full planned trajectory going to a specific task.

dimos/control/tasks/planar_base_trajectory_task/planar_base_trajectory_task.py
mustafab0: There is no damping in this case because we are just treating this planar base as a first order system, since we are solving for error in position.

The `Kp` here is actually a feedforward term, which we use when we tune the gain of the system. This controller currently only supports a mock so has everything tuned to `1` .

dimos/control/tasks/planar_base_trajectory_task/planar_base_trajectory_task.py
mustafab0: We are invoking this in the `execution_manager`, when we call the `poll` method

dimos/manipulation/execution_manager.py
mustafab0: I think better names should be used because legs can definitely be confusing

dimos/manipulation/execution_manager.py
mustafab0: Since we are discussing about how the manipulation_module becomes a module with no lifecycle. I think it would be nice to attach a watchdog thread to the whole module, just like how control coordinator has one to keep track of things.

That is sort of the concept behind the `state_monitors` in ManipulationModule.

Is this similar to what you are suggesting by "generic"

Maybe I shouldn't call this a `_watchdog` I can just call it `_execution_thread` and we can address a proper watchdog later.

lmk your thoughts.

TomCC7: no I mean from lifecycle perspective all trajectory tasks should be equal right? then we should treat them uniformly (both joint and base trajectory task). I think better design would be we have a trajectory task spec or something which aligns rpcs they expose

dimos/manipulation/manipulation_module.py
mustafab0: That is true.

This is because I am forcing the odometry as `joints`.

There is a cleaner fix here to expose odometry as a topic. But this requires a bigger refactor. I will create a issue link here and track it to resolve this in a follow up. 

I have sort of been waiting for the control coordinator refactor to happen to fix this properly.

mustafab0: https://github.com/dimensionalOS/dimos/issues/4104

TomCC7: lol yeah. seems like control coordinator refactor is blocking a lot here... I'll start working on that next week

TomCC7: can we add a TODO in code so we don't forget? even better we can just link linear issue number

dimos/manipulation/manipulation_module.py
```diff
@@ -166,6 +167,26 @@ class ManipulationModuleConfig(ModuleConfig):
...
+    joint_state_aliases: dict[str, str] = Field(default_factory=dict)
```
TomCC7, line 172:
```suggestion
    # TODO: #4104 
    joint_state_aliases: dict[str, str] = Field(default_factory=dict)
```

dimos/manipulation/execution_manager.py
mustafab0: True I am adding a ticket here for the same: https://github.com/dimensionalOS/dimos/issues/4107

dimos/manipulation/execution_manager.py
mustafab0: Ahh I see. 

I am hesitant to add any Task Spec. I think we are sort of converging on a shape for trajectory task but as we add more use cases we can make it more concrete.

I think this is alright for now.

TomCC7 (approved):

dimos/hardware/drive_trains/mock/test_adapter.py
```diff
@@ -0,0 +1,36 @@
...
+    """Facing +y, driving forward moves the base along world +y, and yaw wraps past pi."""
+    adapter = MockTwistBaseAdapter(dof=3, integrate_odometry=True, step_dt=0.01)
+    adapter.set_odometry([0.0, 0.0, math.pi / 2])
+
+    _drive(adapter, [1.0, 0.0, 0.0], seconds=2.0)
+    _drive(adapter, [0.0, 0.0, 1.0], seconds=2.0)
+
+    x, y, yaw = adapter.read_odometry()
+    assert abs(x) < 1e-6
+    assert abs(y - 2.0) < 1e-6
```
paul-nechifor, line 35: Nit: you can use `assert y == pytest.approx(2)`

dimos/control/tasks/planar_base_trajectory_task/planar_base_trajectory_task.py
```diff
@@ -0,0 +1,378 @@
...
+        self._trajectory: JointTrajectory | None = None
+        self._start_t: float | None = None
+        self._elapsed = 0.0
+        self._stop_until: float | None = None
+        self._stop_pending = False
+        self._armed_t: float | None = None
+        self._last_pose_t: float | None = None
+
+    @property
+    def name(self) -> str:
```
paul-nechifor, line 104: BaseControlTask already does this.

paul-nechifor (approved):

PR #4112, opened 2026-09-13
feat(evals): compare Pi Bash baseline with dimcode

spomichter:
We need to sandbox our control cases for evals such that theyre not cheating and using dimOS, so we get a clean comparison of no dimOS astra versus yes dimOS astra. The Pi adapter cannot run dimcode's production gateway or provide a Bash baseline without DimOS access. now we can compare the same model in Pi+Bash against dimcode+DimOS using the existing eval cases and runner.

Also added more agressive typing since some of the files were hard to read. Abstraced much into dataclasses. 

Added allowed tools to harnesses and excluded so we can eval Bash against No Bash, or any other tool eval. Previously we enforced allowed skills via the dimOS modules added but that only works for dimensional @skills. Needed it for built in Pi skills which live in the JS

## Problem

The Pi adapter cannot run dimcode's production gateway or provide a Bash baseline without DimOS access. We need to compare the same model in Pi+Bash against dimcode+DimOS using the existing eval cases and runner.

## Solution

- Add `DimcodeAdapter(PiAdapter)`. It adapts gateway startup, session configuration and events while inheriting provider setup, tracing, request limits, ATIF conversion and process cleanup.
- Add shared `AgentConfig.allowed_tools` and `dimos evals run --allow bash,grep`. Pi and dimcode share the native Pi extension that selects tools and blocks excluded calls. Unsupported adapters reject explicit allowlists.
- Compose recording isolation with `PiAdapter(sandbox=True, allowed_tools=("bash", "grep"))`. Linux bubblewrap excludes DimOS, host files, credentials and network access. Selected observations arrive as CSV/PNG/JSON; camera PNGs are attached to the initial message. There is no separate baseline agent subclass or unrestricted fallback.
- Configure both provider SDKs through Pi's public extension API; use typed paths, gateway commands, configs and incoming events. Config/cache locations follow `dimos.constants`, and native home/state/cache settings are inherited. Reuse Pi's model registry for both arms. Record model turns, tool calls, request attempts and adapter time; retain partial trajectories, count errors in the denominator and preserve unknown costs.

The starting comparison is **Astra + Bash (Pi)** versus **Astra in dimcode + DimOS**. Both use the Pi runtime; no native Codex/Claude Code adapters or new model loop are introduced. This measures the combined product/robotics-stack effect.

The baseline currently supports recordings. Live tasks need a bounded vendor-SDK interface, and publication runs still need grader/unrelated-data isolation for the product arm. The two observation delivery paths are documented.

## How to Test

With `OPENAI_API_KEY`, Pi 0.85.1, dimcode, Linux bubblewrap and the `go2_short` recording available:

```bash
dimos evals run dimos.evals.suites.examples --agent dimos.evals.agents.pi --allow bash,grep --set sandbox=true --set model=gpt-6-astra --set thinking=medium --set max_steps=12 --set max_output_tokens=4096
dimos evals run dimos.evals.suites.examples --agent dimos.evals.agents.dimcode --set model=gpt-6-astra --set thinking=medium --set max_steps=12 --set max_output_tokens=4096
```

These exercise the existing one-frame LiDAR point count and ten-frame person-presence cases. They are integration pilots, not a broad performance ranking.

Validation on the rebased branch:

- `python -m pytest dimos/evals -q`: **210 passed**, one default deselection.
- Mypy: **all 19 changed Python files passed**.
- Twenty-three native-runtime tests use shared OpenAI/Anthropic and Pi/dimcode/sandbox fixtures. They verify endpoints, model selection, output caps, tool execution/rejection, no-tool mode, missing/stale extension setup and sandboxed grep failures/quoting. Additional tests cover malformed events, disconnects and unknown costs. Both JavaScript extensions passed strict TypeScript checks. No paid model trials were run for this revision.
- All pre-commit checks passed. No dependency changes.

## Checklist


dimos/evals/agents/bash_only.py
```diff
@@ -0,0 +1,157 @@
...
+        "--setenv", "LANG", "C.UTF-8", "/bin/bash", "--noprofile", "--norc", "-c",
+    ]  # fmt: skip
+
+
+class BashOnlyConfig(PiAdapterConfig):
+    tools: tuple[str, ...] = ("bash",)
+    builtin_guidance: bool = False
+
+
+class BashOnly(PiAdapter):
```
spomichter: bad

dimos/evals/agents/pi.py
```diff
@@ -184,139 +210,190 @@ def run(
...
         return {
-            **passed,
-            "PI_CODING_AGENT_DIR": str(run_dir / ".pi-agent"),
-            "PI_SKIP_VERSION_CHECK": "1",
-            "PI_TELEMETRY": "0",
+            **{
+                k: v
+                for k, v in os.environ.items()
+                if not (self.config.sandbox and k.startswith("DIMOS_"))
+            },
```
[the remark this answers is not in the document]
spomichter: Fixed in 4d706282a4: _build_process_env passes an explicit allowlist again (PATH, HOME, LANG/LC_*, TERM, TMPDIR, XDG dirs, SSL cert vars) plus the provider key named by _key_env and DIMOS_* outside the sandbox; nothing else from the host. test_model_proxy_keeps_native_provider_and_no_secret now sets an unrelated AWS_SECRET_ACCESS_KEY and asserts it does not reach the process.

dimos/evals/agents/pi.py
```diff
@@ -184,139 +210,190 @@ def run(
...
+    def _build_process_env(self, paths: RunPaths) -> dict[str, str]:
         return {
-            **passed,
-            "PI_CODING_AGENT_DIR": str(run_dir / ".pi-agent"),
-            "PI_SKIP_VERSION_CHECK": "1",
-            "PI_TELEMETRY": "0",
+            **{
+                k: v
+                for k, v in os.environ.items()
+                if not (self.config.sandbox and k.startswith("DIMOS_"))
```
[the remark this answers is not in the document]
spomichter: Fixed in 4d706282a4, same change as the thread above: dimcode inherits the same allowlist through super()._build_process_env, so its Bash sees the provider key, DIMCODE_HOME and DIMOS_* only.

dimos/evals/runner.py
```diff
@@ -160,24 +164,44 @@ def run_case(self, case: EvalCase, agent: Agent) -> EvalResult:
                 )
                 started = time.monotonic()
                 trajectory = agent.run(case.inputs, env, case_dir, timeout_s=case.timeout_s)
-                case.environment.settle(max(0.0, case.timeout_s - (time.monotonic() - started)))
+                agent_duration_s = time.monotonic() - started
```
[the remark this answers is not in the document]
spomichter: Fixed in 4d706282a4 and df609edcf4: agent_duration_s is measured in a finally around agent.run and carried on the base EvalResult, so a run that raises keeps its elapsed time. test_failed_agent_run_keeps_its_duration covers it.

spomichter:
@greptileai review

dimos/evals/agents/lib/runtime.js
```diff
@@ -0,0 +1,67 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { readFileSync, writeFileSync } from "node:fs";
+
+/** @type {{provider: string, base_url: string, key_env: string, allowed_tools: string[] | null, max_output_tokens: number | null}} */
+const config = JSON.parse(
+  readFileSync(new URL("./runtime.json", import.meta.url), "utf8"),
```
spomichter, line 19: this wil be deleted in the stackedPR

dimos/evals/agents/pi.py
```diff
@@ -184,139 +216,196 @@ def run(
         )
         if self.config.max_steps == 0:
             return events.trajectory.build("max_steps")
-        upstream = os.environ.get("OPENAI_BASE_URL", "https://api.openai.com/v1")
+        paths = RunPaths.for_run(run_dir)
+        upstream = os.environ.get(
```
spomichter: horrible

dimos/evals/agents/pi.py
```diff
@@ -184,139 +216,196 @@ def run(
         )
         if self.config.max_steps == 0:
             return events.trajectory.build("max_steps")
-        upstream = os.environ.get("OPENAI_BASE_URL", "https://api.openai.com/v1")
+        paths = RunPaths.for_run(run_dir)
```
spomichter, line 220: why RunPaths seems likeuseless dataclass

PR #4136, opened 2026-09-14
fix(gps): wire websocket vis into demo

### LINKED ISSUE #4131: demo-gps-nav fails startup because required WebsocketVisSpec module reference cannot be resolved
### Description

# demo-gps-nav fails startup because required WebsocketVisSpec module reference cannot be resolved

## Summary

`demo-gps-nav` fails during startup because `GpsNavSkillContainer` has a required `WebsocketVisSpec` module reference, but the blueprint does not provide any module that satisfies that specification.

This reproduces consistently when running the blueprint startup smoke test in isolation.

## Reproduction

```bash
cd ~/dimensional/dimos

uv run pytest \
  'dimos/robot/test_blueprint_startup.py::test_blueprint_starts[demo-gps-nav]' \
  -m '' \
  -vv -s --tb=long
```

## Actual behavior

The blueprint exits during startup.

`GpsNavSkillContainer` requires the following module reference:

```text
GpsNavSkillContainer
  _vis: WebsocketVisSpec
  optional=False
```

No module in the blueprint satisfies the required `WebsocketVisSpec`, so the coordinator fails while resolving module references.

The startup smoke test reports:

```text
Blueprint 'demo-gps-nav' exited during startup with code 1.
```

## Expected behavior

`demo-gps-nav` should either:

1. Include a module satisfying the required `WebsocketVisSpec`, or
2. Make the visualization reference optional if visualization is not required for the GPS navigation demo.

A registered software-only blueprint should not fail startup because a required module reference cannot be resolved.

## Notes

This was reproduced independently after the full macOS test suite exposed the failure, so it does not appear to be caused by test ordering or state left behind by another test.

This is separate from #4130. The failure is caused by an unresolved required module reference rather than missing robot hardware, credentials, IP configuration, or an external asset.

During cleanup, the run also logged an `McpServer/stop` error involving `cannot join current thread`. That appears secondary to the blueprint construction failure and is not the primary issue reported here.

## Suggested labels

`bug`

---

## QA provenance clarification

This issue was discovered during a macOS release QA pass using an **additional local runtime startup harness** that launched registered blueprints and checked whether they remained alive through startup.

The local QA harness was:

```text
dimos/robot/test_blueprint_startup.py
```

That file was **not tracked on `main`** and was not an existing DimOS CI test.

The underlying failure documented in this issue was subsequently reproduced in isolation, so the bug itself remains valid. This clarification is only to correct the original issue wording where it may have implied that `test_blueprint_startup.py` was already part of the repository or CI.

Runtime startup coverage itself is being discussed separately in #4130.

christiefhyang:
## Contribution path

- Small, safe change that does not need a separate implementation issue beyond the bug report.

## Summary

Fixes #4131.

`GpsNavSkillContainer` requires a `WebsocketVisSpec` module reference and calls it when setting GPS travel goal points.

The `demo-gps-nav` blueprint did not include any module satisfying that spec, so startup failed while resolving required module references.

Add `WebsocketVisModule` to the demo blueprint and add a regression test verifying that the demo includes a websocket visualization provider.

## Testing

```bash
uv run pytest dimos/agents/skills/test_gps_nav_skills.py -q
# 3 passed

uv run pytest \
  'dimos/robot/test_all_blueprints.py::test_blueprint_is_valid[demo-gps-nav]' \
  -q
# 1 passed

uv run pytest \
  'dimos/robot/test_blueprint_startup.py::test_blueprint_starts[demo-gps-nav]' \
  -m '' \
  -q
# 1 passed
```

## Scope

This only changes the `demo-gps-nav` blueprint composition and adds regression coverage.

No changes are made to `GpsNavSkillContainer` behavior.


spomichter:
websocket vis is deprecated

christiefhyang:
closing this since websocket vis is deprecated and @ezen needs to port the gps blueprint to rerun

christiefhyang:
websocket vis is deprecated and @ezen needs to port the gps blueprint to rerun

PR #4144, opened 2026-09-14
fix(zenoh): wait for a session's links only once

KrishnaH96:
## Problem
In the Rerun WASD panel the first few key presses do nothing. The Go2 starts moving about 4 s after the first press.

## Root cause
Every zenoh service waits up to `zenoh_connect_timeout` (1 s) for a link to the dialled endpoints when it starts, and publishers start on their first message. The first teleop message makes `MovementManager` publish on four cold topics (goal cancel x3, then `cmd_vel`), so its handler blocks for 4 s. On a WebRTC Go2 the dialled `tcp/ROBOT_IP:7447` never links, so every wait runs to the timeout.

On current release/main branch try this to see the difference youself:

Run this normally, and reproduce the issue where intial rerun wasd clicks are not working.
```
OBSTACLE_AVOIDANCE=false uv run dimos run unitree-go2
```

Then when i run the same command with this flag: ZENOH_CONNECT_TIMEOUT=0, the bot start moving right away with first click on WASD rerun panel.

```
ZENOH_CONNECT_TIMEOUT=0 OBSTACLE_AVOIDANCE=false uv run dimos run unitree-go2
```

## Solution
Sessions are pooled per config. The pool now remembers a session whose link wait timed out, and later services on that session skip the wait. A session waits at most once.

## How to test
`uv run dimos run unitree-go2`, tap W once in the Rerun panel: the robot moves on the first tap, and "Zenoh endpoints not linked" appears only during startup.

`uv run pytest dimos/protocol/service/test_zenoh_connect_timeout.py`

## AI assistance

## Checklist


dimos/protocol/service/zenohservice.py
```diff
@@ -290,6 +291,15 @@ def close_all(self) -> None:
                 except zenoh.ZError as e:
                     logger.warning("Zenoh session close failed", session_key=key, error=str(e))
             self._sessions.clear()
+            self._unlinked.clear()
+
+    def gave_up_linking(self, config: ZenohConfig) -> bool:
```
paul-nechifor: I was looking at this and wondering why you named this function the same as the next one...

Maybe call this `did_give_up_linking` so it's easier to tell apart from the next one.

KrishnaH96: the ealier functions name is give_up_linking and here it is gave_up_linking (past tense)
I can also name it did_give_up_linking though

paul-nechifor (approved):

paul-nechifor (dismissed):

paul-nechifor (approved):

PR #4147, opened 2026-09-14
fix(config): keep opaque module config values when shipping to workers

KrishnaH96:
## Problem

`dimos --simulation mujoco run unitree-g1-groot-wbc` fails at startup with "Failed to deploy module: ValidationError ... visual_override.world/g1/joints Input should be callable". Every blueprint whose Rerun config uses the URDF renderers is affected: G1 GR00T in sim and on hardware, Alfred vis-nav.

## Root cause
Since #3305 the blueprint config parser ships each module config as `model.model_dump(mode="python")`. Pydantic serialises dataclass instances to dicts, so `UrdfRobotJointStateRerunFactory` and `UrdfRobotStaticRerunFactory` reach the worker as dicts and fail the callable check. Plain-function renderers, as used by the Go2 blueprints, are left alone, which is why those still work.

## Solution
The parser ships the explicitly set fields as validated, recursing through models, dicts and lists as `model_dump` did, and leaves any other object untouched. Regression test with a callable dataclass in a module config.

## How to test
`uv run dimos --simulation mujoco run unitree-g1-groot-wbc` deploys the Rerun bridge and draws the G1.
`uv run pytest dimos/core/coordination/blueprint_config/`

## AI assistance

## Checklist


KrishnaH96:
Superseded by #4141, which fixes the same model_dump conversion with a broader set of tests and already has its backport (#4169). Closing this one. The finding and the GR00T sim verification are recorded in DIM-1669.

PR #4151, opened 2026-09-14
remove tests which don't run in CI

paul-nechifor:
## Problem

Some tests don't run in CI and looks like they never will.

## Solution

Remove them.


hvent90 (approved):
10/10

aclauer (commented):
There are even more e2e tests that are skipped and need an openai key, should we remove those too maybe?

aclauer:
> There are some more e2e tests that are skipped and need an openai key, should we remove those too maybe?

idk if anyone is actually running those other tests locally ever either, I know I'm not haha

aclauer (approved):

paul-nechifor:
> There are even more e2e tests that are skipped and need an openai key, should we remove those too maybe?

Yeah, there are more things we need to delete. Some should probably be converted to tool_* tests though. I was trying to be conservative in this round. But we should delete more.

PR #4155, opened 2026-09-15
fix(porcelain): load OpenCV before PyAV so cv2 windows work from the …

KrishnaH96:
## Problem
In the wiki's Python API sequence, `cv2.imshow` after `Dimos.connect()` and `peek_stream("color_image")` hangs at 100% CPU and cannot be interrupted.

## Root cause
PyAV bundles a private libxcb. When PyAV loads before OpenCV, OpenCV's Qt window plugin links its xcb helpers against that copy and livelocks after the X extension handshake. Since #3617's WebRTC transport the Go2 stack imports PyAV (through aiortc), so a connected client loads it before the user's `import cv2`. On 0.0.13 no Go2 path loaded PyAV. Reproduced with only `import av; import cv2; cv2.imshow(...)`.

## Solution
The Python API module imports OpenCV first, so anything the API loads afterwards, including PyAV, cannot break a later `cv2.imshow`. One import with a comment.

## How to test
Blueprint running, then in a REPL: `from dimos import Dimos; app = Dimos.connect(); img = app.peek_stream("color_image", 1.0); import cv2, numpy as np; cv2.imshow("i", np.array(img.data)); cv2.waitKey(0)` shows the frame.

## AI assistance

## Checklist


KrishnaH96:
@paul-nechifor please review and approve.

paul-nechifor (approved):

PR #4163, opened 2026-09-15
Alfred Full stack integration

mustafab0:
Alfred integration with nav + manipulation


dimos/robot/diy/alfred/alfred_posedomain.json
mustafab0, line 1: This file is used as a tuning config to the holonomic pose controller for trajectory following (nav)

dimos/robot/diy/alfred/alfred_model.py
mustafab0, line 1: I am not happy about this file. I would much rather this be part of config.py 

But can simplify later.

PR #4168, opened 2026-09-15
hyperspace: a live query surface, and make it fast by default

jeff-hykin:
The whole of hyperspace against main. Supersedes #4059 jeff/feat/hype, which is the first 111 commits of this.

Two halves:

**A query surface.** `find` / `heatmap` / `area` as skills over one implementation, answers published as they arrive, and `dimos map live` writing a self-contained 3D page per query.

**Made to pick its own fast path.** One device chooser for the CLI and the modules; detector precision, image preparation and the ranking member resolved from the hardware; the patch index held on the accelerator as pieces; the text towers placed after the detector and index have taken their share.

Measured on sf_office, first answer: Mac 1.33s to 0.96s, CudaLaptop 1.93s to 1.37s. Worst query anywhere 3.25s to 1.52s. Place counts unchanged or better throughout.

Two things deliberately left off, each with the number it costs written beside it in the code: `gpu_preprocess` on Metal (twice the detector speed, four of eleven traffic lights) and a device-held index past Metal's slice limit.


jeff-hykin:
Superseded by #4170 jeff/feat/hype2, which starts from main. This branch had danvi/dim1486/generalize-dan-perception-stack merged into it, so 24 of its 69 files were Dan's still-open work.

PR #4186, opened 2026-09-16
build(nix): add the shared flakes for the native module SDKs

jeff-hykin:
Gives native/rust and native/cpp a flake each, with nothing consuming them yet, so the modules that follow can name ref=main instead of pinning a branch.


jeff-hykin:
Superseded by #4187, which now targets main directly and already contains these three commits.

PR #4187, opened 2026-09-16
Fix all the nix

jeff-hykin:
- CI should be 30% faster
- standardize native modules
- crate2nix
- fix nix store problems


.github/workflows/autofix.yml
```diff
@@ -6,7 +6,7 @@ permissions: {}
 
 jobs:
   lint:
-    timeout-minutes: 1
+    timeout-minutes: 15
```
jeff-hykin: ruff/uv was already 40sec, this adds just slightly more (~7 sec) but I don't want a 1min timeout to become flakey

.github/workflows/autofix.yml
```diff
@@ -16,6 +16,11 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v7
+        with:
+          # The shared-flake guard below diffs against the PR base, and relocking
+          # compares the pinned revision's tree with this one -- neither is answerable
+          # from a depth-1 checkout.
+          fetch-depth: 0
```
jeff-hykin: / ` fetch-depth: 0` needs explaining

.github/workflows/ci.yml
```diff
@@ -244,8 +253,16 @@ jobs:
           key: cachix-published-${{ runner.os }}-${{ runner.arch }}-${{ needs.cachix-build-check.outputs.inputs-hash }}
 
   lint:
-    timeout-minutes: 10
+    # Not 10 minutes any more: this job now runs the cargo-fmt and cargo-clippy hooks,
+    # and with no shared cargo workspace left they compile each module's dependency
+    # graph in turn. The cache below makes repeat runs cheap; a cold run is the
+    # expensive one.
+    timeout-minutes: 30
```
jeff-hykin: / doesnt seem right

.github/workflows/ci.yml
```diff
@@ -273,15 +290,73 @@ jobs:
           deno-version: ${{ steps.deno-version.outputs.version }}
           cache: true
 
+      # The cargo-fmt and cargo-clippy hooks enter each module's own devShell, which
```
jeff-hykin: / will be removed

.github/workflows/ci.yml
```diff
@@ -273,15 +290,73 @@ jobs:
...
+            ~/.cargo/git/db
+            target
+            !target/debug/incremental
+          key: cargo-lint-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('**/Cargo.lock') }}
 
   rust:
     timeout-minutes: 30
     runs-on: ubuntu-latest
+    env:
+      CARGO_INCREMENTAL: "0"
```
jeff-hykin: / why

.github/workflows/autofix.yml
```diff
@@ -6,7 +6,7 @@ permissions: {}
 
 jobs:
   lint:
-    timeout-minutes: 1
+    timeout-minutes: 1.5
```
jeff-hykin, line 9: 40sec of ruff/uv and ~7sec of nix. 

The .5 is effectively padding to make sure this isn't flakey

.github/workflows/ci.yml
```diff
@@ -292,18 +320,41 @@ jobs:
...
+            !target/debug/incremental
+          key: cargo-test-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            cargo-test-${{ runner.os }}-${{ runner.arch }}-
       - name: cargo test
-        run: cargo test --workspace --all-features --locked
+        run: |
+          set -euo pipefail
+          needs_system_libs=(
+            dimos/hardware/sensors/camera/realsense/rust  # librealsense
```
jeff-hykin: Will change

jeff-hykin:
Todo:
- make CI rust use nix
- figure out something for maturin / pyo3 needing UV (not nix) python at compile time

.github/workflows/autofix.yml
```diff
@@ -30,5 +30,25 @@ jobs:
         run: uv run ruff format
       - name: uv lock
         run: uv lock
+
+      - name: Did native/ change?
```
jeff-hykin, line 34: this is for bumping flake locks whenever the root (./native/rust or ./native/cpp) stuff changes

.github/workflows/ci.yml
```diff
@@ -155,8 +156,18 @@ jobs:
           authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
       - name: Build native modules
         env:
-          BUILD_WORKERS: "1"
+          BUILD_WORKERS: "4"
```
jeff-hykin, line 159: free performance. No Idea why it was set to 1

.github/workflows/ci.yml
```diff
@@ -155,8 +156,18 @@ jobs:
...
         run: python3 bin/build-native-modules
+      - name: Lint native modules
```
jeff-hykin, line 161: linting is now a nix thing

.github/workflows/ci.yml
```diff
@@ -293,15 +305,28 @@ jobs:
           deno-version: ${{ steps.deno-version.outputs.version }}
           cache: true
 
+      - name: Install Nix
+        env:
+          INPUT_EXTRA_NIX_CONFIG: |
+            http-connections = 50
```
jeff-hykin, line 311: make cache loading faster

.github/workflows/ci.yml
```diff
@@ -293,15 +305,28 @@ jobs:
...
+        uses: pre-commit/action@v3.0.1
+        with:
+          extra_args: --all-files --verbose
+
 
   rust:
     timeout-minutes: 30
     runs-on: ubuntu-latest
+    env:
+      CARGO_INCREMENTAL: "0"
```
jeff-hykin, line 329: time save: prevents creating an always-gets-busted cache

.github/workflows/ci.yml
```diff
@@ -397,16 +460,14 @@ jobs:
...
+      - name: List C++ native modules
         id: find-modules
         run: |
-          modules=$(find . -name flake.nix \
-            -not -path './.git/*' \
-            -not -path '*/build/*' \
-            -not -path '*/result/*' \
-            -printf '%h\n' \
+          modules=$(git ls-files '*/flake.nix' \
+            | sed 's|/flake.nix$||' \
```
jeff-hykin, line 467: native/cpp now has a flake.nix, need this find to not find that one

.github/workflows/autofix.yml
```diff
@@ -30,5 +30,25 @@ jobs:
...
+            echo "changed=true" >> "$GITHUB_OUTPUT"
+          fi
+      - name: Install Nix
+        if: steps.shared.outputs.changed == 'true'
+        run: bash docker/ros/install-nix.sh
+      - name: Relock the modules onto this revision
+        if: steps.shared.outputs.changed == 'true'
+        run: python3 bin/relock-shared-flakes
+
+      # Unguarded: it commits the ruff and uv fixes too, which every PR may need.
```
Dreamsorcerer, line 52:
```suggestion
```

.github/workflows/ci.yml
Dreamsorcerer: There's not enough memory on the runners to handle more than 1 simultaneous build, depending on which modules get built (it was 2 originally, and then had to reduce to 1 after more module changes).
```suggestion
          # Runners have limited memory and can OOM with parallel builds.
          BUILD_WORKERS: "1"
```

.github/workflows/ci.yml
Dreamsorcerer: Is lint not a trivial thing without notable dependencies?

.github/workflows/ci.yml
```diff
@@ -155,8 +156,18 @@ jobs:
           authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
       - name: Build native modules
         env:
-          BUILD_WORKERS: "1"
+          BUILD_WORKERS: "4"
         run: python3 bin/build-native-modules
+      - name: Lint native modules
+        env:
+          BUILD_WORKERS: "4"
```
Dreamsorcerer, line 163: Probably only need to specify this if we want to reduce the worker count.
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -155,8 +156,18 @@ jobs:
...
+        run: python3 bin/build-native-modules --lint
+      # Published so the `native` job substitutes them instead of compiling the
+      # test targets itself on every run.
+      - name: Test rust native modules
+        env:
+          BUILD_WORKERS: "4"
```
Dreamsorcerer, line 169:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -155,8 +156,18 @@ jobs:
...
+        run: python3 bin/build-native-modules --tests
```
Dreamsorcerer, line 170: Checking I'm understanding this. This is some kind of nix tests that verifies the built module in isolation, right? No Python tests or dependencies used?

.github/workflows/ci.yml
```diff
@@ -223,7 +235,7 @@ jobs:
           authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
       - name: Build native modules
         env:
-          BUILD_WORKERS: "1"
+          BUILD_WORKERS: "4"
```
Dreamsorcerer, line 238:
```suggestion
          # Runners have limited memory and can OOM with parallel builds.
          BUILD_WORKERS: "1"
```

.github/workflows/ci.yml
```diff
@@ -293,15 +305,28 @@ jobs:
           deno-version: ${{ steps.deno-version.outputs.version }}
           cache: true
 
+      - name: Install Nix
```
Dreamsorcerer, line 308: Why are we installing Nix here? Seems like pre-commit is skipping the native lint, so I'm a little confused why this would be needed.

.github/workflows/ci.yml
```diff
@@ -334,25 +359,62 @@ jobs:
...
-          python -c "import dimos_voxel_ray_tracing, dimos_mls_planner"
-          python -m pytest -c /dev/null --rootdir . --noconftest --import-mode=importlib -p no:cacheprovider -v \
-            dimos/mapping/ray_tracing/test_voxel_map.py \
-            dimos/mapping/ray_tracing/test_transformer.py \
-            dimos/navigation/nav_3d/mls_planner/test_mls_planner.py \
-            dimos/navigation/nav_3d/mls_planner/test_transformer.py
+          python bin/pyo3-extensions --build
+          python bin/pyo3-extensions --test
       - name: Bake e2e tests
         run: uv run pytest -m bake_e2e dimos/cli/bake/test_bake_e2e.py --no-cov
```
Dreamsorcerer, line 370: We should move these into the actual tests job, right? We should be building and testing against every Python version and OS.

Dreamsorcerer: Maybe leave to a followup PR.

.github/workflows/ci.yml
```diff
@@ -334,25 +359,62 @@ jobs:
...
+        with:
+          path: |
+            ~/.cargo/registry/index
+            ~/.cargo/registry/cache
+            ~/.cargo/git/db
+            target
+            !target/debug/incremental
+          key: cargo-test-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('**/Cargo.lock') }}
+
+  flake-locks-current:
```
Dreamsorcerer, line 384: Isn't autofix doing this already?

.github/workflows/ci.yml
```diff
@@ -334,25 +359,62 @@ jobs:
...
+          git ls-files '*flake.nix' | sed 's|/flake.nix$||;s|^flake.nix$|.|' | sort -u > /tmp/flakes
+          : > /tmp/stale
+          while read -r dir; do
+            echo "::group::$dir"
+            nix flake lock --no-update-lock-file "path:$dir" || echo "$dir" >> /tmp/stale
+            echo "::endgroup::"
+          done < /tmp/flakes
+          if [ -s /tmp/stale ]; then
+            while read -r dir; do
+              echo "::error::flake.lock is stale or incomplete in $dir -- run 'nix flake lock' there and commit it"
```
Dreamsorcerer, line 413: Can probably just put this in the first loop instead of iterating twice?

.github/workflows/ci.yml
```diff
@@ -457,6 +511,15 @@ jobs:
...
           done
+      - name: Install uv
+        uses: astral-sh/setup-uv@v10.0.1
+        with:
+          enable-cache: true
+          prune-cache: true
+      - name: Native module e2e tests
+        run: |
+          uv sync --group tests --frozen
+          uv run pytest -m native_e2e dimos/hardware/sensors/lidar/livox/test_e2e.py --no-cov
```
Dreamsorcerer, line 522: Again, think we should be restoring the modules in the tests jobs and running the Python tests there.

If all modules are built in Nix, is this native job still needed, or can it be replaced with build-cachix and restoring in tests?

.github/workflows/ci.yml
```diff
@@ -946,7 +1010,7 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
       - name: Build rust native modules
         if: contains(matrix.markers, 'self_hosted')
-        run: cargo build --release --locked -p dimos-livox -p dimos-virtual-mid360
+        run: python3 bin/build-native-modules
```
Dreamsorcerer, line 1013: Will this not get replaced with the 'Provision native modules from Cachix' step?

bin/build-native-modules
```diff
@@ -65,6 +82,24 @@ import urllib.request
...
+    """nix's name for this machine, e.g. aarch64-darwin.
+
+    Asked lazily, never at import: `--inputs-hash` runs on a runner that has no nix
+    yet -- that is the whole point of the gate -- and must stay a git-only operation.
+    """
+    proc = subprocess.run(
+        ["nix", "eval", "--raw", "--impure", "--expr", "builtins.currentSystem"],
+        capture_output=True,
+        text=True,
+        check=False,
```
Dreamsorcerer, line 98: Won't this miss a command failure (and then maybe end up returning "" or something)?
```suggestion
        check=true,
```

bin/build-native-modules
```diff
@@ -477,10 +517,98 @@ def verify_published(modules: tuple[DiscoveredModule, ...]) -> None:
...
+    proc = subprocess.run(
+        ["nix", "build", "-L", attr, "--no-link"],
+        cwd=str(REPO_ROOT / build_dir),
+        capture_output=True,
+        text=True,
+        check=False,
+    )
+    body = (proc.stdout or "") + (proc.stderr or "")
+    if proc.returncode != 0:
+        raise RuntimeError(f"{check} {build_dir} exited {proc.returncode}\n{body}")
```
Dreamsorcerer, line 601: Why not just use check?
```suggestion
        check=True,
    )
    body = (proc.stdout or "") + (proc.stderr or "")
```

bin/module-dir-guard
Dreamsorcerer, line 1: Would this not make more sense as just a regular test? I think I have some similar tests already, relating to things that can break the cache in build-cachix-check.

dimos/core/test_build_native_modules.py
```diff
@@ -207,37 +208,193 @@ def test_ast_extraction_matches_runtime() -> None:
...
+    for module in modules:
+        inputs = _SCRIPT._collect_input_paths(module)
+        assert inputs == {module.build_dir}, (
+            f"{module.qualname}: hashed inputs {sorted(inputs)} are not just "
+            f"{module.build_dir!r} — the flake reaches outside its own directory"
         )
 
 
-@pytest.mark.skipif(not _IN_GIT_CHECKOUT, reason="needs git HEAD for object hashes")
+def test_no_module_reaches_the_repository_root() -> None:
```
Dreamsorcerer, line 238: Looks similar to test_no_module_hashes_the_repo_root above, maybe needs to merged into that test.

PR #4200, opened 2026-09-17
Select hosted e2e and native_e2e on Mac self-hosted-tests.

bogwi:
Run the coordinator, datacollection, daemon, viewer websocket, relay, and relay-bridge e2e tests, plus the livox native loopback, on Mac CI as well as on Linux. Introduce dedicated test marks - `pytest.mark.macos_ci`, and `pytest.mark.skipif_macos_ci`


.github/workflows/ci.yml
```diff
@@ -796,7 +796,9 @@ jobs:
           - os: macOS
             container: null  # run on host — `container:` is Linux-only
             # web_browser: the cockpit browser e2e that Linux runs in the `web` job.
-            markers: "self_hosted or web_browser"
+            # native_e2e: livox live loopback that Linux runs in the rust job.
+            # macos_ci: hosted job tests also collected here; not in hosted not (...).
```
Dreamsorcerer, line 800: I'm still a little confused over this marker. It seems like it's meant to only run in CI, but I see nothing skipping it locally. It's called ci, but doesn't run in the normal CI job, only the self-hosted one.

Why not use the existing self_hosted marker with a normal OS skipif?

.github/workflows/ci.yml
```diff
@@ -796,7 +796,9 @@ jobs:
           - os: macOS
             container: null  # run on host — `container:` is Linux-only
             # web_browser: the cockpit browser e2e that Linux runs in the `web` job.
-            markers: "self_hosted or web_browser"
+            # native_e2e: livox live loopback that Linux runs in the rust job.
```
Dreamsorcerer, line 799: I'm also a little confused why we'd have a marker that is run in the rust job and one self_hosted job? If the tests work here, should it not be removed from the rust job, in which case the entire marker can be removed instead?

.github/workflows/ci.yml
bogwi: this PR puts pytestmark = pytest.mark.macos_ci on files that were already in hosted tests (daemon, coordinator, datacollection, viewer ws, relay). 

Hosted Linux tests still collects them because they are not self_hosted / web_browser / native_e2e. Mac -m now includes macos_ci so Mac collects the same files.

Remarking files marked macos_ci with self_hosted would drop them from hosted tests -> (not (self_hosted or …)). Linux self-hosted would then pick them up.

.github/workflows/ci.yml
bogwi: mixing a bit select markers (pytest -m) with skipif markers (pytest_collection_modifyitems). 

It does not mean ci-only or move off the rust job.
 
Mac -m adding native_e2e is the same pattern as web_browser: Linux keeps the specialist job and Mac self-hosted-tests also selects that mark.

---

If native_e2e is removed from the rust job and the mark is deleted, Linux CI stops running test_live_loopback_handshake_and_stream unless that function is marked self_hosted (then Linux self-hosted would run it, rust would not). 

Mac work is not a reason to drop the Linux rust step. We have the native_e2e mark so two jobs can name the same set without putting self_hosted on it which would also pull it into Linux self-hosted and out of hosted tests.

.github/workflows/ci.yml
Dreamsorcerer: But, don't we already have them running in #3681 then?

Dreamsorcerer: Ah, it looks like some of these are the ones I've been skipping due to the runner not sleeping appropriately. I think I'd be more inclined to switch them to self_hosted if we can't get them working on the basic runner, rather than introducing more markers (we have too many already...).

.github/workflows/ci.yml
Dreamsorcerer: > If native_e2e is removed from the rust job and the mark is deleted, Linux CI stops running test_live_loopback_handshake_and_stream unless that function is marked self_hosted (then Linux self-hosted would run it, rust would not).

Yeah, that's what I'm expecting. If the test can be run in self-hosted, why does it need to run in the rust job? I think the entire rust job needs to go.

My expectation going forward is that rust modules are built in one job (I'd assume they will all move into the cachix-build job in future), and then the built artifacts will be restored in our tests/self-hosted-tests jobs and used in the tests.

bogwi: Yes, I agree with you but this PR stack inherited this marker.

This PR' main purpose is to enable remaining tests for mac ci not messing with old. These enabled test well pass green locally and should be run on the mac runner as well. 

We aim for the full parity with linux ci. 


But let's refactor those rust jobs in the consecutive PRs.

PR #4203, opened 2026-09-17
Feat/ivan/local planner trajectory

leshy:
there are a few things here that can be split into separate PRs but nav team is the only consumer so keeping here unless we decide otherwise

### restructure of navigation/ much cleaner folder, new specs

```python

class GlobalPlanner(Protocol):
    tf: In[TFMessage]
    goal: In[PointStamped]

    path: Out[Path]


class BlindLocalPlanner(Protocol):
    """Shapes the global route without looking: no map."""

    planner_path: In[Path]
    tf: In[TFMessage]

    path: Out[Path]


class MapLocalPlanner(Protocol):
    """Routes around what the local map says is there."""

    planner_path: In[Path]
    local_map: In[PointCloud2]
    tf: IO[TFMessage]

    path: Out[Path]


class TrajectoryFollower(Protocol):
    tf: IO[TFMessage]
    path: In[Path]

    nav_cmd_vel: Out[Twist]


```

### bridge hacky selective sub feature
we need to be able to explicitly subscribe to just SOME topics not all, so that they are not sent over network or decoded by py

### module introspection fixes
module.io() hides default internal RPCs

### lidar module config standardized
removed global env reading, switched to local module config

### distributed zenoh architecture for nav

we run `go2-zenoh-motion-pointlio` on go2 jetson and `dimos --robot-ip go22 run go2-viewer` on the laptop

this means zenoh (network robust) subs go over network, and rerun rendering and RPC happens locally on the laptop

### TODO

- requires go2web on the actual go2 or jetson for zenoh control, this needs to be reimplemented and moved to dimos in follow up PR
- experiment with global planner actually running the local planner internally so we have a single planner


dimos/navigation/test_tf_pose.py
```diff
@@ -0,0 +1,103 @@
+# Copyright 2026 Dimensional Inc.
```
leshy, line 1: wat

native/rust/dimos-module/src/pointcloud.rs
```diff
@@ -0,0 +1,187 @@
...
+}
+
+impl std::error::Error for ExtractError {}
+
+fn count(v: i32, what: &'static str) -> Result<usize, ExtractError> {
+    usize::try_from(v).map_err(|_| ExtractError(what))
+}
+
+/// The finite xyz points of a little-endian float32 cloud, in order.
+pub fn extract_xyz(msg: &PointCloud2) -> Result<Vec<[f32; 3]>, ExtractError> {
```
aclauer, line 36: I think this should actually go in the lcm repository in the pointcloud struct, but we can do that later. I need to publish that crate on crates.io anyway.

dimos/navigation/local_planner/rust/src/planner.rs
```diff
@@ -0,0 +1,3168 @@
...
+    pub fan_yaw_per_m: f64,
+    pub fan_yaw_done: f64,
+    pub speed_lookahead: f64,
+}
+
+/// `embodiment/base.py::Embodiment`, field for field: the body a module is
+/// configured with, deserialised straight from its config.
+#[derive(Clone, Debug, serde::Deserialize, serde::Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct Emb {
```
aclauer, line 94: should this struct be declared somewhere outside of the local planner? it doesn't seem like it really belongs with a specific module

dimos/navigation/local_planner/rust/src/planner.rs
```diff
@@ -0,0 +1,3168 @@
...
+/// the planner's own time, because it sits on the per-sample path of the footprint scan (`fill_x` /
+/// `fill_y`) and on every distance-field lookup, which means every clearance
+/// evaluation and every `seg_free` step. The `as i64` that always follows it
+/// then costs a `cvttsd2si` plus the compare-and-cmov chain Rust needs to make
+/// the cast saturating.
+///
+/// Adding 1.5 * 2^52 forces the significand to shed its fractional bits under
+/// the ambient rounding mode -- which is round-to-nearest-ties-to-even, and
+/// which Rust never changes -- leaving the rounded integer in the low mantissa
+/// bits, biased by the constant. Subtracting the constant's own bit pattern
```
aclauer: almost 1000 lines of this file are comments. If they were shortened I think it would help both the AI and me.

native/rust/dimos-module/src/pointcloud.rs
leshy: yeah I agree. TBH I'm not sure if lcm repository should go to dimos

dimos/navigation/local_planner/rust/src/planner.rs
leshy: jesus christ

dimos/navigation/local_planner/rust/src/planner.rs
leshy: hm maybe for now it does but soon it might not? since only local planner is embodiment aware, everything else blindly deals with points

aclauer (approved):

aclauer (approved):

paul-nechifor (approved):

PR #4204, opened 2026-09-17
Remove skipif_macos_bug from tests that pass on Darwin.

bogwi:
1) Remove skipif_macos_bug from tests that pass on Darwin.

2) ([9bc0b60](https://github.com/dimensionalOS/dimos/pull/4204/commits/9bc0b6041727a66a23053b7fb6d4d5d1b5f675a5)) Removed the sqlite-vec Darwin skip and the matching `skipif_macos` / `skipif_aarch64` on `TestStandaloneComponents`, then marked `test_replay.py`, `test_store.py`, and `test_stream.py` with `macos_ci` so Mac self-hosted-tests now runs those 246 tests (they were hosted-only before).


dimos/memory/test_store.py
```diff
@@ -29,11 +28,11 @@
 from dimos.memory.blobstore.base import BlobStore
 from dimos.memory.vectorstore.base import VectorStore
 
-_SKIP_SQLITE_VEC = platform.machine() == "aarch64" or platform.system() == "Darwin"
-
 if TYPE_CHECKING:
     from dimos.memory.store.base import Store
 
+pytestmark = pytest.mark.macos_ci
```
Dreamsorcerer, line 34: Are these due to extensions not being loaded? If so, I've fixed this in my PR by using the uv version of Python, rather than the system one.

bogwi: no, these are not skips, these are adds with clear intent - add to mac ci

PR #4207, opened 2026-09-17
feat(evals): last-line parsers for prose-then-answer replies

spomichter:
## Problem

Anthropic models answer eval prompts with a short explanation and then the value. The suite parsers read the wrong thing: `first_number` takes the first list index of an enumerated explanation, `yes_no` requires the reply to open with the word, and `choice` matched by lowercase words so the article "a" counted as option A. In the apartment matrix this turned four correct Fable answers into failures.

## Solution

Three parsers in `dimos.evals.scorers`, each falling back to the existing behaviour when the last line is not a bare value:

- `last_number(text)`: a bare number on the last line, else `first_number`.
- `last_yes_no(text)`: a bare yes/no on the last line, else `yes_no`.
- `letter(options)`: the last standalone letter from `options`, case-sensitive.

Cases opt in by using them; nothing existing changes.

## How to Test

```sh
uv run pytest dimos/evals/test_evals.py -k "parsers or scorer" -q
```

## AI assistance

Implemented with Claude Code; tests run locally.

## Checklist


dimos/evals/scorers.py
```diff
@@ -75,6 +75,44 @@ def parse(text: str) -> str:
...
+
+def last_yes_no(text: str) -> str:
+    """A bare yes/no on the reply's last line, else ``yes_no`` on the whole reply."""
+    last = _last_line(text).lower()
+    return last if last in ("yes", "no") else yes_no(text)
+
+
+def _last_line(text: str) -> str:
+    lines = [line.strip().strip("*_`") for line in text.strip().splitlines() if line.strip()]
+    return lines[-1].rstrip(".!").strip() if lines else ""
```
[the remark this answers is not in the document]
spomichter: Fixed in 136d3001c9: _last_line strips emphasis and trailing punctuation in one pass, so **4**. and **yes**. read as 4 and yes; both cases added to test_last_line_parsers.

dimos/evals/scorers.py
```diff
@@ -75,6 +75,44 @@ def parse(text: str) -> str:
...
+    def parse(text: str) -> str:
+        found = pattern.findall(text)
+        if not found:
+            raise ValueError(f"no option letter {options} in reply: {text[:80]!r}")
+        return str(found[-1])
+
+    return parse
+
+
+def last_number(text: str) -> float:
```
hvent90: What scenarios would we use `last_number` vs `first_number`? Same for `last_yes_no` and `yes_no`

dimos/evals/scorers.py
```diff
@@ -75,6 +75,44 @@ def parse(text: str) -> str:
     return parse
 
 
+def letter(options: str) -> Callable[[str], str]:
```
hvent90: could we use `choice` in place of this?

dimos/evals/scorers.py
spomichter: yes fixed

dimos/evals/scorers.py
spomichter: Yes, with one flag. choice matches case-insensitively, so choice(["A","B","C","D"]) reads the article in "it is a kitchen" as option A. In 98727e3e38 letter is gone and choice takes case_sensitive=True for lettered options; test_parsers_take_the_answer_after_an_explanation covers the article case.

PR #4210, opened 2026-09-18
feat(agents): TypeSafe reactive agent: world state in, joystick cmd_vel out

spomichter:
## Problem

DimOS agents are LLM chat loops. There is no way to run a fast, calibrated, text-only decision model as a reactive controller: perception in, typed decisions out, at a fixed rate, with confidence the code can gate on.

## Solution

`TypeSafeAgent` (`dimos/agents/typesafe/`) drives a robot from TypeSafe System One (Jev) decisions:

- **WorldState in**: `odom`, `detections_3d` (or `detections_2d`), `lidar` and the latched goal text are folded into one JSON state per tick. New `to_json()` on `PoseStamped`, `Detection2DArray`, `Detection3DArray` and `PointCloud2` (8-sector clearance summary in the robot frame). Numbers are paired with word buckets (`bearing`, `distance`, sector `state`) because the model reads words better than floats.
- **One request per tick** (default 2 Hz): a 3-way Choice per joystick axis (`x`, `y`, `yaw`), a `stop` Noul, and a `target` Choice over the object labels. Choice is single-select, so simultaneous forward + strafe comes from independent per-axis questions.
- **cmd_vel out**: the model picks direction only. Code scales turn rate by bearing error and speed by distance to the chosen target, declares arrival under 0.5 m, ramps the Twist at 10 Hz from a separate thread, and zeroes on stop, error, or a deadman that outlasts one slow inference (`max(1 s, 2.5 / rate_hz)`).
- Holds (no model call) when odom or detections are missing; all-zero decisions for 5 s clear the goal.
- Keeps the existing agent contract (`human_input` / `agent` / `agent_idle`), so humancli, the cockpit chat and the evals adapter attach unchanged.

Blueprints: `unitree-go2-typesafe` and `unitree-go2-typesafe-demo` (adds `DemoObjects`, a fixed-object stand-in for a 3D detector, since no Go2 stack publishes `Detection3DArray` yet).

The HTTP client is ~60 lines over `requests`; `typesafe-sdk` is on PyPI but blocked by `exclude-newer` until 2026-09-22 and not needed for one endpoint.

Spec: `dimensionalOS/agents` `engineering/typesafe-agent/spec.md`.

## Testing

- 20 unit tests: message `to_json`, world-state buckets, drive decoding (per-axis gate, blend, stop, target), agent loop with a fake client (hold, stop word, inference failure, deadman, arrival).
[...]

PR #4212, opened 2026-09-18
feat(agents): TypeSafeAgent: world state in, joystick cmd_vel out

spomichter:
## Problem

DimOS agents are LLM chat loops. There is no way to run a fast, calibrated, text-only decision model as a reactive controller: perception in, typed decisions out, at a fixed rate, with confidence the code can gate on.

## Solution

`TypeSafeAgent` (`agent.py`, generic: `trigger()` / `state(trigger)` / `questions()` in, `choices` / `scores` / `nouls` out) and `TypeSafeNavigationAgent` (`navigation.py`: pose, detections and scan in; `cmd_vel` and a world-frame `goal` out; `world_state.py` builds the JSON, `drive.py` is the adapter) drives a robot from TypeSafe System One decisions. Rewrite of #4210 from scratch on `main`, applying the review there.

- **State**: `world_state.build_world_state()` folds odom (`PoseStamped` or `Odometry`), 3D or 2D detections and the scan into one JSON state. New typed `to_json()` on `PoseStamped`, `Detection2DArray`, `Detection3DArray`, `PointCloud2` (8-sector clearance in the robot frame). Numbers carry word buckets (`bearing`, `distance`, sector `state`).
- **Questions**: `drive.questions()` builds one 3-way Choice per joystick axis (`x`, `y`, `yaw`), a `stop` Noul and a `target` Choice over the visible labels, in the documented API shapes via `client.choice()` / `client.noul()`.
- **Request**: one `requests.post` in `TypeSafeAgent._post` to `POST /v1/systemone`, on every trigger update or the latest snapshot at `max_hz`; answers publish by type as soon as the response lands. Key from `global_config.typesafe_api_key`.
- **Drive**: `drive.decode()` selects each axis by the chosen option's probability; `agent._steer()` scales turn rate by bearing error and speed by distance to the chosen target, declares arrival under 0.5 m, ramps the Twist at 10 Hz from its own thread, zeroes on a deadman of `max(1 s, 2.5 / rate_hz)`. Holds (no request) without detections; a reply for a goal that changed in flight is dropped; gives up after 5 s without motion.
- Keeps the `human_input` / `agent` / `agent_idle` contract, so humancli and the cockpit chat attach unchanged.

Blueprints:
- `habitat-typesafe`: `habitat_teleop` + `HabitatConnection(publish_objects=True)`. HM3D annotations carry no object boxes, so the native renders the semantic sensor and boxes each visible instance from its depth pixels. Generated LCM constructors share nested defaults; `Detection3D` is built fresh at every level.
- `unitree-go2-typesafe`, with a DimSim e2e that publishes a chair and drives to it.

Spec: `dimensionalOS/agents` `engineering/typesafe-agent/spec.md`.

## Testing

- 68 unit tests (message `to_json`, world-state buckets, drive decoding, the generic agent with a fake `_post`, the navigation agent: hold, arrival, odometry input, failure, deadman, goal point, goal-change race; habitat server round trips). mypy clean on all changed files.
- Live habitat run on this branch (`habitat-typesafe`, headless, traced): chair 4.5 m ahead at spawn, `target=chair`, drove 3.8 m and stopped at the chair.
- DimSim e2e (`test_dimsim_typesafe_agent.py`) passed at c807291b0 (86 s); rerun on the review commit pending, the box is busy with evals.

Run: `TYPESAFE_API_KEY=... uv run dimos run habitat-typesafe`, then `uv run humancli --transport zenoh` and `go to the chair`.

Supersedes #4210.


dimos/agents/typesafe/agent.py
```diff
@@ -0,0 +1,334 @@
...
+        )
+        qs = questions(tuple(dict.fromkeys(o["label"] for o in state["objects"])))
+        answers = self._ask(state, qs)
+        self._trace(state, qs, answers)
+        drive = decode(
+            answers,
+            min_confidence=self.config.min_confidence,
+            stop_threshold=self.config.stop_threshold,
+        )
+        self._steer(state, drive)
```
omarespejel: hey @spomichter  , I found a goal-change race in a local test. I delayed the `_ask` reply and switched from chair to door. When the chair reply arrived, it still restored the chair position and published a forward command. That old position also showed up as `goal_point` in the next request.
I have a test that reproduces it. Are you already working on this? If not, I can put up a small fix.

dimos/agents/typesafe/agent.py
```diff
@@ -0,0 +1,334 @@
...
+from dimos.msgs.geometry_msgs.Twist import Twist
+from dimos.msgs.nav_msgs.Odometry import Odometry
+from dimos.msgs.sensor_msgs.PointCloud2 import PointCloud2
+from dimos.msgs.vision_msgs.Detection2DArray import Detection2DArray
+from dimos.msgs.vision_msgs.Detection3DArray import Detection3DArray
+from dimos.utils.logging_config import setup_logger
+
+logger = setup_logger()
+
+API_KEY_ENV = "TYPESAFE_API_KEY"
```
spomichter: bad -- NO CONSTANTS HIDDEN IN FILES

dimos/agents/typesafe/agent.py
```diff
@@ -0,0 +1,334 @@
...
+Vec3 = tuple[float, float, float]
+Ask = Callable[[WorldState, dict[str, Question]], Answers]
+ZERO: Vec3 = (0.0, 0.0, 0.0)
+PUBLISH_HZ = 10.0
+# ponytail: fixed steering gains; config them if a robot needs different ones.
+SLOW_WITHIN_M = 1.5
+TURN_FULL_AT_DEG = 45.0
+
+
+def typesafe_api_key() -> str | None:
```
spomichter, line 50: typesafe api key probably add to globalconfig or just check how we handle with openaiapi key

dimos/agents/typesafe/agent.py
```diff
@@ -0,0 +1,334 @@
...
+    if os.environ.get(API_KEY_ENV):
+        return None
+    return f"{API_KEY_ENV} is not set. Create a key at https://console.typesafe.ai/settings/keys"
+
+
+class _Latest(Generic[T]):
```
spomichter: No we already have helpers for this check transport operators

dimos/agents/typesafe/agent.py
```diff
@@ -0,0 +1,334 @@
...
+    model: str = "jev-latest"
+    rate_hz: float = 2.0
+    timeout_s: float = 5.0
+    stale_s: float = 2.0
+    deadman_s: float | None = None  # None: max(1, 2.5 / rate_hz), must outlast one slow inference
+    linear_speed: float = 0.5
+    angular_speed: float = 0.8
+    linear_accel: float = 0.8
+    angular_accel: float = 1.6
+    min_confidence: float = 0.5
```
spomichter: No these are navigation related configs, dont put here either put in some typesafe constants.py or in the drive.py or whatever navigation. this agent needs to be generic to all typesafe usage not just nav usage

dimos/agents/typesafe/agent.py
```diff
@@ -0,0 +1,334 @@
...
+    stop_threshold: float = 0.7
+    reached_m: float = 0.5
+    give_up_s: float = 5.0  # goal clears after this long without motion
```
spomichter: same with this its nav goal related

dimos/agents/typesafe/agent.py
```diff
@@ -0,0 +1,334 @@
...
+    config: TypeSafeAgentConfig
+
+    odom: In[PoseStamped]
+    odometry: In[Odometry]  # same pose, nav_msgs flavour (habitat)
+    detections_3d: In[Detection3DArray]
+    detections_2d: In[Detection2DArray]
+    lidar: In[PointCloud2]
+    human_input: In[str]
+
+    cmd_vel: Out[Twist]
```
spomichter: i would have typesafe agent have a generic interface for each typesafe output type (choices, score, noul) and then subclass with TypesafeNavigationAgent

dimos/agents/typesafe/agent.py
```diff
@@ -0,0 +1,334 @@
...
+    def _infer_loop(self) -> None:
+        period = 1.0 / self.config.rate_hz
+        while not self._stop_event.is_set():
+            t0 = time.monotonic()
+            try:
+                self._tick()
+            except Exception:
+                logger.exception("TypeSafeAgent tick failed")
+                self._set_target(ZERO)
+            self._stop_event.wait(max(0.0, period - (time.monotonic() - t0)))
```
spomichter: I think inference loop should be set by the input stream which in this case is WorldState or whatver. So maybe depending on the input stream we have some operator maybe reactive X that does input stream concatenantion and conversion to Json, and that frequency is the default freqnecy of the inference loop

then if we want we can also have a throttle feature that enforces say max 1hz and then at 1hz increments it takes the latest snapsho from the input and sends to JEv

dimos/agents/typesafe/agent.py
```diff
@@ -0,0 +1,334 @@
...
+            return
+        state = build_world_state(
+            goal,
+            pose,
+            self._motion,
+            detections_3d=det3d,
+            detections_2d=det2d,
+            lidar=self._lidar.get(stale),
+            goal_xy=self._goal_xy,
+            image_size=self.config.image_size,
```
spomichter: this is nav speciic and should be subclassed

dimos/agents/typesafe/agent.py
```diff
@@ -0,0 +1,334 @@
...
+        answers = self._ask(state, qs)
+        self._trace(state, qs, answers)
+        drive = decode(
+            answers,
+            min_confidence=self.config.min_confidence,
+            stop_threshold=self.config.stop_threshold,
+        )
+        self._steer(state, drive)
+
+    def _steer(self, state: WorldState, drive: Drive) -> None:
```
spomichter: also nav specific it looks like

dimos/agents/typesafe/agent.py
```diff
@@ -0,0 +1,334 @@
...
+        while not self._stop_event.is_set():
+            with self._lock:
+                stale = time.monotonic() - self._decided_at > deadman
+                target = ZERO if self._goal is None or stale else self._target
+                cur = self._current
+                nxt = tuple(
+                    c + max(-s, min(s, t - c)) for c, t, s in zip(cur, target, steps, strict=True)
+                )
+                self._current = (nxt[0], nxt[1], nxt[2])
+                publish = self._goal is not None or any(cur)
```
spomichter: publish loop should by defualt publish as soon as we get inference response from JEv. maybe if we want we acn have a param to throttle output or can enforce on a output by output basis. FOr example some choices can throttle other choices can take at full speed. Need  a proposal here but for now can merge and let the input throttling handle impicitly the output throttle 

Deadman is specifically needed for Nav i dont think we need a deadman on every Jev output topic so sublcass that.

dimos/agents/typesafe/agent.py
```diff
@@ -0,0 +1,334 @@
...
+
+    def _trace(self, state: WorldState, qs: dict[str, Question], answers: Answers) -> None:
+        if self.config.trace_dir is None:
+            return
+        d = Path(self.config.trace_dir)
+        d.mkdir(parents=True, exist_ok=True)
+        self._seq += 1
+        (d / f"{self._seq}-request.json").write_text(
+            json.dumps({"body": {"state": state, "questions": qs}})
+        )
```
spomichter: No dont make directorries like that internally we have a way we handle this i think by default i .local or .config check how its done for saving of eval outputs and memory outputs.

dimos/agents/typesafe/drive.py
```diff
@@ -0,0 +1,204 @@
...
+    return qs
+
+
+@dataclass(frozen=True)
+class Drive:
+    x: float
+    y: float
+    yaw: float
+    stop: bool
+    confidence: float
```
spomichter, line 127: we can have confidence but dont use it for selection - probabilty is used for selection of choices

dimos/agents/typesafe/world_state.py
```diff
@@ -0,0 +1,164 @@
...
+if TYPE_CHECKING:
+    from dimos.msgs.geometry_msgs.PoseStamped import PoseStamped
+    from dimos.msgs.sensor_msgs.PointCloud2 import PointCloud2
+    from dimos.msgs.vision_msgs.Detection2DArray import Detection2DArray
+    from dimos.msgs.vision_msgs.Detection3DArray import Detection3DArray
+
+MAX_OBJECTS = 20  # closest first; the model reads words, not a scene graph
+_DISTANCE = ((0.5, "touching"), (1.5, "near"), (4.0, "mid"), (math.inf, "far"))
+_BEARING_2D = ("far_left", "left", "center", "right", "far_right")
+_SIZE_2D = ((0.4, "filling_view"), (0.15, "large"), (0.03, "medium"), (-1.0, "small"))
```
spomichter: these should be tpypesave spsecific constants i assume and under a section called  perceptoin /world state

dimos/msgs/sensor_msgs/PointCloud2.py
```diff
@@ -30,10 +30,28 @@
...
+    "right",
+    "ahead_right",
+)
+
+
+class SectorJson(TypedDict):
+    clear_m: float
+    state: str
+
+
```
spomichter: megaslop no do not change our sensor messages for tyepsafe specific tstuff this is bad. put ti somwehere else under typesafe/ dir

dimos/msgs/sensor_msgs/PointCloud2.py
```diff
@@ -998,6 +1016,38 @@ def filter_by_height(
...
+            pts = pts - np.array([origin.x, origin.y, origin.z])
+            c, s = np.cos(-origin.yaw), np.sin(-origin.yaw)
+            pts = np.column_stack(
+                (c * pts[:, 0] - s * pts[:, 1], s * pts[:, 0] + c * pts[:, 1], pts[:, 2])
+            )
+        r = np.hypot(pts[:, 0], pts[:, 1])
+        keep = (pts[:, 2] > z_min) & (pts[:, 2] < z_max) & (r > 0.05) & (r < max_range)
+        pts, r = pts[keep], r[keep]
+        sector = np.round(np.arctan2(pts[:, 1], pts[:, 0]) / (np.pi / 4)).astype(int) % 8
+        out: dict[str, SectorJson] = {}
```
spomichter, line 1033: we can keep this here probably godo to have

dimos/msgs/vision_msgs/Detection2DArray.py
```diff
@@ -11,13 +11,28 @@
...
+    w: float
+    h: float
+
+
+class Detection2DJson(TypedDict):
+    label: str
+    score: float
+    bbox: BBoxJson
+
+
```
spomichter, line 35: fine to keep here

dimos/msgs/sensor_msgs/PointCloud2.py
spomichter: json stuff we can keep in here but not the naming stuff

dimos/msgs/sensor_msgs/PointCloud2.py
```diff
@@ -998,6 +1016,38 @@ def filter_by_height(
             timestamp=self.ts,
         )
 
+    def to_json(
```
spomichter, line 1007: fine to keep

dimos/simulation/habitat/server.py
```diff
@@ -206,20 +212,86 @@ def tf_msg(links: list[tuple[str, str, Any, Any]], ts: float) -> bytes:
...
+    """
+    pts, valid = _pixels(depth, k, trunc, stride)
+    ids = semantic[::stride, ::stride][valid]
+    pts = pts @ world[:3, :3].T + world[:3, 3]
+    out = []
+    for i in np.unique(ids):
+        label = labels[i] if 0 <= i < len(labels) else ""
+        sel = pts[ids == i]
+        if label in STRUCTURE or len(sel) < min_points:
+            continue
```
spomichter, line 276: clean all of this up need heavy typing and readable

dimos/simulation/habitat/server.py
```diff
@@ -206,20 +212,86 @@ def tf_msg(links: list[tuple[str, str, Any, Any]], ts: float) -> bytes:
...
+        _stamp(d.header, ts)
+        d.header.frame_id = "world"
+        d.results = [
+            ObjectHypothesisWithPose(hypothesis=ObjectHypothesis(class_id=label, score=1.0))
+        ]
+        d.results_length = 1
+        cx, cy, cz = (float(v) for v in center)
+        sx, sy, sz = (float(v) for v in size)
+        d.bbox = BoundingBox3D(
+            center=LCMPose(position=Vector3(x=cx, y=cy, z=cz), orientation=Quaternion(w=1.0)),
```
spomichter: IS this for continuous detection in a habitat? but habitats typically start from ground truth/god view? so why is this needed

dimos/agents/typesafe/agent.py
spomichter: Fixed in d03865f41: `set_goal` bumps a generation counter, `state()` records it, and `on_answers` drops a reply whose generation no longer matches, so a late chair answer neither moves the robot nor latches the chair as `goal_point`. Covered by `test_goal_change_during_inference_drops_the_stale_answer`. Thanks for the repro.

dimos/agents/typesafe/agent.py
spomichter: Moved to `typesafe/constants.py` (sectioned: API, perception/world state, navigation). The key itself is no longer a file constant, see next thread.

dimos/agents/typesafe/agent.py
spomichter: Now `global_config.typesafe_api_key`, read from `TYPESAFE_API_KEY` like `DIMOS_TRANSPORT`, and listed in `SECRET_CONFIG_FIELDS` so it redacts. The agent uses `self.config.g.typesafe_api_key`; the blueprint requirement checks the same field.

dimos/agents/typesafe/agent.py
spomichter: Replaced with the existing `getter_streaming(observable, timeout=0, nonblocking=True)`; staleness is judged by the message `ts`. The pose is not read through a getter at all: it is the trigger message handed to `state(trigger)`.

dimos/agents/typesafe/agent.py
spomichter: Moved: `TypeSafeAgentConfig` now has only `model`, `max_hz`, `timeout_s`, `trace`. Everything navigation-related lives in `TypeSafeNavigationConfig` (navigation.py) with defaults from the navigation section of `constants.py`.

dimos/agents/typesafe/agent.py
spomichter: Same, now in `TypeSafeNavigationConfig`.

dimos/agents/typesafe/agent.py
spomichter: Done: generic `TypeSafeAgent` publishes `choices`, `scores`, `nouls` (typed `Out` per answer type) and exposes `trigger()` / `state(trigger)` / `questions(state)` / `on_answers()`. `TypeSafeNavigationAgent` subclasses it and owns cmd_vel, goal, ramp and deadman. Blueprints use the subclass.

dimos/agents/typesafe/agent.py
spomichter: Done: the loop is gone. The base subscribes to `trigger()` (for nav: the odom/odometry stream) and infers on each update; with `max_hz` set it pipes through `ops.sample(1/max_hz)`, i.e. the latest snapshot at that rate. One request in flight; triggers that land during it are dropped. Nav defaults `max_hz=2`.

dimos/agents/typesafe/agent.py
spomichter: Moved into `TypeSafeNavigationAgent.state()`.

dimos/agents/typesafe/agent.py
spomichter: Moved into `TypeSafeNavigationAgent._steer`.

dimos/agents/typesafe/agent.py
spomichter: Base agent now publishes answers the moment the response lands (no output loop). The ramped cmd_vel publisher and the deadman are in the navigation subclass only. Per-output throttling left for the proposal you mention.

dimos/agents/typesafe/agent.py
spomichter: Now `trace: bool`, writing under `DIMOS_RUN_LOG_DIR/typesafe` (the per-run dir the CLI sets and temporal memory uses), falling back to `LOG_DIR`. No caller-chosen directories.

dimos/agents/typesafe/drive.py
spomichter: Done: `decode()` selects on `probabilities[choice] >= min_probability` for the axes and the target; `confidence` stays on `Drive` for the log line only.

dimos/agents/typesafe/world_state.py
spomichter: Moved to `constants.py` under `# --- Perception / world state`.

dimos/msgs/sensor_msgs/PointCloud2.py
spomichter: Removed from the sensor message. Sector names live in `typesafe/constants.py` and are passed in: `PointCloud2.to_json(origin, sectors=..., ...)`, which also makes the method N-way generic. Only the `SectorJson` return type sits next to the method.

dimos/msgs/sensor_msgs/PointCloud2.py
spomichter: Kept.

dimos/msgs/vision_msgs/Detection2DArray.py
spomichter: Kept.

dimos/msgs/sensor_msgs/PointCloud2.py
spomichter: Kept, with the sector names now a parameter.

dimos/simulation/habitat/server.py
spomichter: Rewritten: `VisibleObject(NamedTuple)` with typed numpy arrays, `FloatArray` alias, explicit `instance_id` / `hits` / `low, high` names, and `objects_msg` takes a `Sequence[VisibleObject]`.

dimos/simulation/habitat/server.py
spomichter: Habitat gives no god view here: HM3D's semantic annotations are per-face instance ids with no object boxes (every `semantic_scene.objects[i].aabb` is zero in this dataset). The only way to get a labelled 3D box is to render the semantic sensor and unproject its pixels with depth, which is what this does, per frame, for what is in view. If we want scene-wide ground truth we would precompute per-instance boxes from the mesh once at load; happy to add that as `publish_objects="scene"` in a follow-up.

omarespejel:
hey, I reproduced both Habitat CI failures locally. Adding `sim.semantic_scene.objects = []` after `sim = hs.Simulator.return_value` in `test_explicit_spawn_uses_real_constructor` fixes both

dimos/agents/typesafe/agent.py
omarespejel: thanks, the new check fixes the original case. I found one more window though. If the goal changes after the check but before `_steer` writes the target, the old reply can still latch the old position and send a forward command
happy to add the test if you want

PR #4223, opened 2026-09-18
fix(go2): stamp camera_info on each publish

KrishnaH96:
## Problem

`GO2Connection.publish_camera_info` republishes one `CameraInfo` whose `ts` is frozen at import. `--record` stamps messages with their own `ts`, so a recording stores every `camera_info` copy at one instant, before all other streams. In replay they fire once at the start, the image never gets its pinhole, and the 3D view shows "2D visualizers require a pinhole ancestor" with no camera frustum. Live runs hide it because a fresh copy arrives every second.

Seen on a real Go2 recording: 67 `camera_info` messages, one unique timestamp, 15 s before the first odom.

Same one-line change as in #3914, split out with a test so it can land and be backported on its own. #3914 keeps the mem rerun renderer fix.

## Solution

Publish `self.camera_info_static.with_ts(time.time())`. `with_ts` already exists and copies every field.

## How to Test

`uv run pytest dimos/robot/unitree/go2/test_connection.py`

On a real Go2: `dimos --record run unitree-go2`, then `dimos mem summary` shows `camera_info` at 1 Hz across the whole run and replay shows the frustum.

## AI assistance

Claude Code (Claude Fable 5.1) helped find the root cause from a recording and draft the fix and test. I reviewed and understand every line.

## Checklist


KrishnaH96:
Duplicate of the timestamp line already in #3914. Closing in favour of it. I moved the test from here onto #3914.

PR #4232, opened 2026-09-19
feat(sim): auto-disable mujoco shadows when rendering is slow

spomichter:
@bogwi Dan — heads up, this is a slop PR (Claude-authored from a debugging session on my Mac) — shared in case it's helpful. Take what's useful or close it.

## Problem

`dimos --simulation run unitree-go2` runs at ~0.4x realtime on Apple Silicon. Profiling showed physics is fine (0.8 ms/step, 6.3x realtime headroom with the office scene) — the cost is shadow-mapping the ~1220-mesh scene in the offscreen camera renders: **28.6 ms with shadows vs ~7 ms without** per 640x360 frame, out of a 50 ms video frame budget.

## Changes

- `mujoco_shadows: Literal["auto", "on", "off"] = "auto"` in `GlobalConfig`. `auto` benchmarks one shadowed offscreen render at sim startup (~150 ms) and zeroes `model.vis.quality.shadowsize` when it exceeds 30% of the video frame budget. This must happen **before** `launch_passive`/`Renderer` create their GL contexts — the shadow framebuffer is allocated at context creation, so toggling `mjRND_SHADOW` scene flags at runtime leaves the interactive viewer shadow-rendering the whole scene regardless (tried that first; didn't help).
- Skip `CoreMLExecutionProvider` for the locomotion policy ONNX session: CoreML partitions the 14-node graph (13/14 nodes) and pays a per-call copy overhead making it ~20x slower than plain CPU inference (0.335 ms vs 0.014 ms).

Machines where the shadowed render is cheap (discrete GPUs measure ~2-5 ms) keep shadows; nothing changes for them.

## Measured (M-series Mac, go2 + office1 scene)

| | before | after |
|---|---|---|
| worst-case loop iteration | 92 ms | 28 ms |
| realtime factor | 0.38x | 1.25x |

Verified end-to-end: startup logs `Shadowed render took 28.6 ms of the 50 ms frame budget; disabling shadows` and the sim runs at realtime.


dimos/simulation/mujoco/mujoco_process.py
bogwi, line 267:
```suggestion
            frame_sim_time = model.opt.timestep * config.mujoco_steps_per_frame
            time_until_next_step = frame_sim_time - (time.time() - step_start)
```
Drop shadows on a slow machine and drop core ml are fine fixes. Yet they surfaced another latent bug int the render loop.

Each iteration of the main loop steps physics mujoco_steps_per_frame times: 7 x 0.005 s -  so 35 ms of sim time. The sleep at the end pads the iteration to model.opt.timestep only, 5 ms of wall time - so the loop was never paced to realtime. Its speed was whatever the viewer sync and the offscreen renders happened to cost per machine.

With shadows on, on Apple macs, render was slow that one iteration took about 35 ms and the sim looked close to realtime. With shadows off the iteration takes about half of that, the same 35 ms of sim time is consumed per iteration and the go2 runs at about 2x. 

On my m4 Pro this is go2 bolt running the office. On your Air - the iteration is slower, so it probably looked fine. Linux with a discrete GPU most likely ran above realtime the whole time and nobody had a before and after to compare against.

Add these two lines and go2 will 1) move with correct speed and 2) movement + Shift correctly paces up too.

bogwi: This might go as a separate PR as well, yes, though the changes surfaced the bug, so let's not leave it for later

dimos/simulation/mujoco/mujoco_process.py
```diff
@@ -70,6 +70,39 @@ def stop(self) -> None:
...
+        renderer.close()
+
+    budget_ms = 1.0 / VIDEO_FPS * 1e3
+    is_slow = render_ms > budget_ms * 0.3
+    if is_slow:
+        logger.warning(
+            f"Shadowed render took {render_ms:.1f} ms of the {budget_ms:.0f} ms frame "
+            "budget; disabling shadows (force with mujoco_shadows=on)"
+        )
+    return is_slow
```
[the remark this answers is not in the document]
bogwi: mocked benchmark test is not wanted, so will not do, the behavior is exercised on every sim start and logs which branch it took

dimos/core/global_config.py
```diff
@@ -108,6 +108,10 @@ class GlobalConfig(BaseSettings):
     mujoco_global_map_from_pointcloud: str | None = None
     mujoco_start_pos: str = "-1.0, 1.0"
     mujoco_steps_per_frame: int = 7
+    # Shadow-mapping the office scene costs ~4x per offscreen render on
+    # integrated GPUs (e.g. Apple Silicon), dropping the sim below realtime.
+    # "auto" keeps shadows and turns them off if the sim falls behind realtime.
+    mujoco_shadows: Literal["auto", "on", "off"] = "auto"
```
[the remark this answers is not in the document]
bogwi: [aa140da](https://github.com/dimensionalOS/dimos/pull/4232/commits/aa140dac4702c5c45005259d626c5b3e7d188d27)

dimos/simulation/mujoco/mujoco_process.py
```diff
@@ -222,7 +264,8 @@ def _run_simulation(config: GlobalConfig, shm: ShmReader) -> None:
                 last_lidar_time = current_time
 
             # Control simulation speed
-            time_until_next_step = model.opt.timestep - (time.time() - step_start)
+            frame_sim_time = model.opt.timestep * config.mujoco_steps_per_frame
+            time_until_next_step = frame_sim_time - (time.time() - step_start)
```
paul-nechifor: This change makes it so the speed is no longer faster than real time, making the robot move quite slowly.

paul-nechifor (changes requested):
The speed of the robot is too low now.

dimos/simulation/mujoco/mujoco_process.py
bogwi: Entirely disagree. 

What the blueprint commands:

1) unitree-go2 (smart/unitree_go2.py:42-50) uses ReplanningAStarPlanner. Its local planner sets _speed = 0.55 m/s (replanning_a_star/local_planner.py:67), the PController scales that down with heading error and pins it at _min_linear_velocity = 0.2 (controllers.py:43, :73-74). 

2) So the planner commands at most 0.55 m/s forward. 

3) MujocoConnection passes twist unchanged into shared memory (mujoco_connection.py:412-414), MockController copies it as [forward, lateral, yaw] (mujoco_process.py:56-64), and the Go1 policy feeds it straight into its observation (policy.py:97). 

Policy's command units are m/s and rad/s in sim time. 

At 1x realtime, 0.55 m/s commanded is 0.55 m/s **observed**, **which is what the real go2 does under the same blueprint**. 

That is the speed you call slow. But this the speed the blueprint comnnads

---

Having "faster than realtime is a defect", not a feature

Why should it run on my mac m4pro or Linux RTX 5070 x N times faster and on slow machines slowe when the shadows are reduced?

---

I have just tested on **3 laptops**

After the change:
same identical speed, 0.5 m/s when walking and 1m/s when Shift is pressed for Apple m1pro, m4pro, and rtx 5070, meaning the changed ARE valid!

Before the change:
m1pro: runs visibly x3 faster than 0.5m/s for unitree-go2 blueprint
m4pro12: runs visibly x4 faster than 0.5 m/s for unitree-go2 blueprint
rxt 5070: runs visibly x N faster than 0.5m/s for unitree-go2 blueprint -> IMPOSSIBLE TO CONTROL THE ROBOT AT ALL.

So, switching off the shadows SHOULD NOT AFFECT THE RENDER SPEED

bogwi:
> The speed of the robot is too low now.

It is not, run this on your machine and you will see the defect in action,

23407d7c96a023a11b806d4f5ecb82b97c913ec4 -> without my changes

`git checkout 23407d7c96a023a11b806d4f5ecb82b97c913ec4 && uv run dimos --simulation run unitree-go2`

bogwi:
> The speed of the robot is too low now.

If the you want faster-than-realtime for iteration speed, that should be an explicit, deterministic knob (a realtime factor applied to frame_sim_time), not a side effect of GPU speed. That is a separate feature from this PR and by intend a) defeats the purpose of simulation and b) affects benchmarks we do by the factor of N, so entirely incorrect

paul-nechifor (approved):

PR #4235, opened 2026-09-19
go2: GO2DDS native module, the DDS controller out of go2web

leshy:
The Go2 DDS controller from go2web's `dimos-helper`, as a dimos rust native module. Runs on the Jetson, talks CycloneDDS to the robot, same port profile as `GO2Zenoh` so blueprints swap between them.


dimos/core/native_module.py
```diff
@@ -133,6 +137,8 @@ class NativeModuleConfig(ModuleConfig):
 
     cli_exclude: frozenset[str] = frozenset()
     cli_name_override: dict[str, str] = Field(default_factory=dict)
+    # Ports the python wrapper publishes itself; the native process never sees them.
+    python_ports: frozenset[str] = frozenset()
```
aclauer, line 141: When does python use ports that the native binary doesn't?

leshy: I had inherited normalized base.py for go2 which took over camerainfo loop, but I actually agree, this breaks our boundaries in a bad way and will spread if we allow, so will undo, and camerainfo loop will be taken over by the dds -> zenoh client

dimos/robot/unitree/go2/dds/rust/src/rtp.rs
```diff
@@ -0,0 +1,66 @@
...
+    }
+    let csrc_count = (buf[0] & 0x0F) as usize;
+    let has_extension = buf[0] & 0x10 != 0;
+    let padded = buf[0] & 0x20 != 0;
+    let mut offset = 12 + 4 * csrc_count;
+    if has_extension {
+        let words = u16::from_be_bytes([*buf.get(offset + 2)?, *buf.get(offset + 3)?]) as usize;
+        offset += 4 + 4 * words;
+    }
+    let end = buf.len() - if padded { *buf.last()? as usize } else { 0 };
```
omarespejel, line 37: This underflows when a 12-byte packet has padding set and ends in `0xff`. Debug and test builds panic, while release drops it. `checked_sub` would make both drop it, and this packet can be a regression test

dimos/robot/unitree/go2/dds/rust/src/module.rs
```diff
@@ -0,0 +1,1021 @@
...
+            lidar_imu: self.lidar_imu.clone(),
+            joint_state: self.joint_state.clone(),
+            imu: self.imu.clone(),
+            battery: self.battery.clone(),
+            joy: self.joy.clone(),
+            tf: self.tf.clone(),
+            stop: self.stop.clone(),
+            handle: handle.clone(),
+        };
+        self.threads.push(std::thread::spawn(move || dds.run(rx)));
```
omarespejel: Is there another path that sends this failure to the process supervisor? With a bad `iface` participant creation panics in this thread after 30s but `start()` has already returned so the parent stays alive
I reproduced it with the release binary. The worker panicked, but the parent was still running at 37s. Since DimOS watches the process and `stop()` drops the join result, the module looks alive even though its DDS loop is gone

dimos/robot/unitree/go2/base.py
```diff
@@ -0,0 +1,216 @@
...
+        super().start()
+        self.spawn(self._publish_camera_info())
+        # Deferred: a verb sent before the transport matches the robot side is dropped.
+        timer = threading.Timer(5.0, self._startup_pose)
+        timer.daemon = True
+        timer.start()
+        self.register_disposable(Disposable(timer.cancel))
+
+    def _startup_pose(self) -> None:
+        self.set_lidar(False)
```
omarespejel: With `lidar_on=True`, `GO2DDS` writes ON at startup but this timer sends `lidar off` five seconds later. I checked both sides: Python sent the command and the release binary read the same command as `Lidar(false)` and wrote `OFF` to `rt/utlidar/switch`.

The same timer runs in `go2_dds_basic` where `GO2DDS` is the only lidar source. Point-LIO already sets `lidar_on=False`. Was this shutdown meant only for Point-LIO or am I missing why it needs to live in `Go2Base`?

